PR #4, opened 2026-04-22
feat: xr_media_hub IPC layer

nvddr:
## Summary

Implements `xr_media_hub.ipc` — the shared-memory + ZMQ IPC layer that decouples the LiveKit connector from agent code. Two dependencies only: `pyzmq` and `msgpack`.

- **ConnectorEndpoint**: producer — writes CPU frames into a shm ring buffer, signals hub via ZMQ PUSH. Creates and owns its own ring buffer; registers with hub on startup.
- **HubEndpoint**: server — dispatches `on_frame`, `on_audio`, `on_data`, `on_participant` callbacks; resolves connector topology via registry; enforces participant isolation (no cross-participant routing).
- **ShmRingBuffer**: 64-byte slot headers, zero-copy memoryview reads.
- Multi-track: `FrameSignal` and `AudioChunk` carry `(participant_id, track_id)`; ZMQ topics namespaced as `audio.<pid>.<track>` for prefix-filter subscriptions.
- Bidirectional: `send_return_audio/data()` lets agents push back to XR clients.
- Dynamic join/leave via `notify_participant_joined/left()` — no static participant lists.

## Test plan
- [ ] ConnectorEndpoint can push frames and audio without deadlocking
- [ ] HubEndpoint dispatches all callbacks correctly
- [ ] Return path reaches the correct participant only
- [ ] Multi-connector topology is transparent to consumer callbacks

### DIFF
--- a/server-runtime/requirements.txt
+++ b/server-runtime/requirements.txt
@@ -1,0 +1,2 @@
+pyzmq>=26.0
+msgpack>=1.0
--- a/server-runtime/main.py
+++ b/server-runtime/main.py
@@ -1,0 +1,52 @@
+"""XR-Media-Hub entry point (stub)."""
+from __future__ import annotations
+
+import asyncio
+import logging
+
+from xr_media_hub.ipc import HubEndpoint, PixelFormat, SlotView, AudioChunk
+
+logging.basicConfig(level=logging.INFO)
+log = logging.getLogger(__name__)
+
+SHM_NAME        = "xr_hub_frames"
+NUM_SLOTS       = 10
+MAX_FRAME_BYTES = 12_441_600  # 4K NV12 worst-case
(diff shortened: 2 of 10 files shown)


server-runtime/requirements.txt
```diff
@@ -0,0 +1,2 @@
+pyzmq>=26.0
```
yanziz-nvidia, line 1: latest is 27.1.0 https://pypi.org/project/pyzmq/

server-runtime/requirements.txt
```diff
@@ -0,0 +1,2 @@
...
+msgpack>=1.0
```
yanziz-nvidia, line 2: latest is 1.1.2 https://pypi.org/project/msgpack/

yanziz-nvidia (approved):

PR #20, opened 2026-04-22
feat: video frame support + NVDEC/NVENC enforcement

nvddr:
## Summary

Adds on-demand video frame pixel delivery to `ProcessorEndpoint` and enforces NVIDIA hardware codec availability at startup.

**Video frame delivery**: processors receive `FrameSignal` metadata at full rate via `on_frame()` — no pixel copy per frame. Pixels only copied on explicit `await ep.request_frame(signal)`. Hub holds one SHM slot per `(participant, track)`, keeping it `_STATE_READY` until the next frame. Concurrent `request_frame()` calls for the same track are coalesced. Echo-agent demonstrates ~1 fps pixel sampling alongside full-rate metadata.

**Hardware codec enforcement**: connector refuses to start if `libnvcuvid.so` (NVDEC) or `libnvidia-encode.so` (NVENC) are absent — prevents silent use of the bundled OpenH264 software fallback, which is royalty-bearing. Platform coverage: Linux (`.so.1`), Windows (`.dll`); macOS blocked outright. `XR_AI_SKIP_HWCODEC_CHECK=1` escape hatch for development only.

**Housekeeping**: `uv.lock` files removed and gitignored.

## Test plan
- [ ] `request_frame()` returns pixel data without blocking the receive loop
- [ ] Startup fails clearly on a machine without NVDEC/NVENC
- [ ] `XR_AI_SKIP_HWCODEC_CHECK=1` bypasses the check with a visible warning


server-runtime/xr_media_hub/ipc/_hub.py
```diff
@@ -173,12 +179,43 @@ async def _dispatch(self, type_id: int, msg) -> None:
...
+            held = self._latest_slots.get(key)
+            if held is None:
+                log.debug("FRAME_REQUEST for %s/%s — no frame held", msg.participant_id, msg.track_id)
+                return
+            _, view = held
+            sig = view.signal
+            frame_data = FrameData(
+                seq=sig.seq, pts_us=sig.pts_us,
+                width=sig.width, height=sig.height, fmt=sig.fmt,
+                data=bytes(view.data[:sig.data_sz]),
```
yanziz-nvidia, line 214: cursor mentioned it could block event loop for ~2-5ms per 4K frame

yanziz-nvidia (approved):

PR #21, opened 2026-04-22
feat: VLM agent sample, multi-camera, HTTPS, agent status

nvddr:
## Summary

Adds the VLM agent sample (Cosmos-Reason1-7B), multi-camera selection, TLS support, and a live agent status indicator.

**VLM agent** (`agent-samples/vlm-agent/`): uses `nvidia/Cosmos-Reason1-7B` (commercial-OK). Accepts data channel messages as vision queries, replies on `vlm.response`. Model loaded at startup (warm on first query). Frames capped at ~1 MP before tokenization to prevent CUDA context corruption from oversized vision token sequences. Retriable soft errors reply to client; inference failures crash via `os._exit(1)` (CUDA context corruption is unrecoverable).

**ProcessorEndpoint fix**: user callbacks dispatched as `asyncio` tasks — fixes the deadlock causing "Frame data unavailable" on every VLM query.

**Agent status channel**: `set_status("processing"|"idle")` sends on reserved topic `_agent.status`; routed through return-data path to LiveKit room. Web client shows a pulsing orange dot while processing, green when idle.

**Multi-camera**: `CameraConfig` gains `deviceId`; camera selector dropdown appears when multiple cameras are detected.

**HTTPS**: auto-generates a self-signed cert in `~/.local/share/xr-ai/` when `web_server_tls=true`. Required for camera access from non-localhost devices.

## Test plan
- [ ] VLM agent responds to data channel queries with vision answers
- [ ] Agent status dot updates during inference
- [ ] Camera selector appears when multiple cameras are available
- [ ] HTTPS web server works; camera access works on a remote device


agent-samples/vlm-agent/vlm_agent/__main__.py
```diff
@@ -0,0 +1,337 @@
...
+            # Hub had no slot for this track yet — client can retry.
+            await self._reply(pid, "Frame data unavailable — please retry.", msg.pts_us)
+            return
+
+        image = _frame_to_pil(frame)   # raises on unrecognised format → crash
+        log.info("vlm  pid=%r  %dx%d  query=%r", pid, frame.width, frame.height, query[:60])
+
+        await self._ep.set_status("processing", pid)
+        loop   = asyncio.get_running_loop()
+        answer = await loop.run_in_executor(None, self._vlm.infer, image, query)
```
yanziz-nvidia, line 265: needs to handle multi client if that's the intended use case

yanziz-nvidia (approved):

PR #43, opened 2026-04-28
feat: add Piper TTS server, VLM streaming TTS, reorganize TTS services

nvddr:
- Add ai-services/tts/piper/ — Piper TTS server (ONNX, non-autoregressive, ~100 ms/sentence on CPU) as a drop-in alternative to Magpie on port 8105
- Add streaming TTS to vlm-agent-worker: sentences are synthesized in parallel as the VLM generates tokens; audio plays sentence-by-sentence via an ordered pipeline task, not after the full response completes
- Add SSE streaming to vlm-server: uses TextIteratorStreamer + background thread, strips <think> preamble before yielding tokens to callers
- Reorganize ai-services/tts/ — magpie and piper are now subfolders under tts/; rename magpie entry point to magpie_tts_server for consistency
- Serialize Magpie synthesis calls with the existing lock (model is not re-entrant)
- Update AGENTS.md and DEPENDENCIES.md to reflect new structure and servers


ai-services/vlm-server/vlm_server/__main__.py
```diff
@@ -145,6 +145,79 @@ def infer(self, images: list, text: str, max_new_tokens: int) -> str:
...
+                    past_think = True
+                    tail = preamble.split("</think>", 1)[1].lstrip()
+                    if tail:
+                        yield tail
+                elif not preamble.lstrip().startswith("<") and len(preamble) >= 8:
+                    # No think block — emit accumulated preamble and stream directly.
+                    past_think = True
+                    yield preamble
+        t.join()
+
```
yanziz-nvidia, line 220: if not past_think and preamble:
  yield preamble


just in case `elif not preamble.lstrip().startswith("<") and len(preamble) >= 8:` never hit

ai-services/tts/piper/piper_tts_server/__main__.py
```diff
@@ -0,0 +1,215 @@
...
+    import asyncio
+    import uvicorn
+
+    if not cfg.get("voice"):
+        print("[piper_tts_server] 'voice' is required in config", file=sys.stderr)
+        sys.exit(1)
+
+    model_cache = _resolve_model_cache(cfg, yaml_dir)
+    port = int(cfg.get("port", _DEFAULT_PORT))
+    host = cfg.get("host", "0.0.0.0")
```
yanziz-nvidia, line 179: We should all default to 127.0.0.1 for safety. 0.0.0.0 binding can be an opt-in flag

yanziz-nvidia (approved):

PR #49, opened 2026-04-30
feat(ios): fix iPhone window sizing, keyboard dismiss, and camera toggle

wenxind-nvidia:
- Add UILaunchScreen to Info.plist for native resolution on all iPhones
- Add Assets.xcassets with AppIcon and AccentColor placeholders
- Register asset catalog in project.pbxproj (file ref + resources build phase)
- Add keyboard dismissal via scroll, toolbar Done button, and onSubmit
- Add front/back camera position picker with live switching
- Disable camera picker on visionOS (ARKit passthrough only)

### DIFF
--- a/client-samples/ios-visionos/App/ContentView.swift
+++ b/client-samples/ios-visionos/App/ContentView.swift
@@ -27,6 +27,18 @@ struct ContentView: View {
             .navigationTitle("StreamKit Sample")
             #if os(iOS)
             .navigationBarTitleDisplayMode(.large)
+            .scrollDismissesKeyboard(.interactively)
+            .toolbar {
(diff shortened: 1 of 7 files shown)


client-samples/ios-visionos/App/ContentView.swift
```diff
@@ -150,6 +164,19 @@ struct ContentView: View {
     @ViewBuilder
     private var cameraRow: some View {
         if model.connectionState == .connected {
+            @Bindable var m = model
+
+            Picker("Camera Mode", selection: $m.cameraPosition) {
```
nvddr, line 169: Instead of manually enumerating front and back, it should be a loop going over the enum I think

nvddr (approved):

PR #62, opened 2026-05-01
ci: add SPDX pre-commit/CI + uv-lock CI; plus fix surface hub startup failures cleanly

yanziz-nvidia:


scripts/check_spdx_headers.py
nvddr, line 1: Can we move this into a separate CI-scripts thing? I don't think this should belong in a top-level scripts directory.

nvddr (approved):

PR #71, opened 2026-05-02
fix(transcript-mcp): surface path validation as local before sink

yanziz-nvidia:
In _resolve_existing, hoist self._check(...) into a jsonl local right before each return. For the slow path, validate the constructed path before .exists() rather than after, so the sanitizer sits on the data-flow path SonarQube traces into the sink.


yanziz-nvidia:
hahaha claude code literally just changed wrappers. Not sure why SonarQube returns this to be recommended.

wenxind-nvidia (approved):

PR #104, opened 2026-05-06
Use single LLM + prompt tuning for XR render, docker workflow fix

nvddr:


yanziz-nvidia (approved):

nvddr:
Triage of the Copilot review comments (this PR is already approved, so tracking rather than churning the branch):

1. **`container_on_port()` vs `--network host` — valid, real.** `build_run_argv` runs containers with `--network host`, so `docker ps --filter publish=<port>` never matches and the stop helper falls through to `pid_on_port`, which can't cleanly stop a docker-mode container. Fix direction: containers are started with a deterministic `--name`, and we already have `container_running(name)`/`container_exists(name)` — the stop path should resolve the container by name rather than by published port. Tracking as a follow-up since it needs a docker host to verify.
2. **`clear_scene()` clears in-memory state even if `disp.forward("scene.remove", …)` fails** — valid; should gate the state clear on a successful (`ok: True`) forward. Follow-up.
3. **worker `clear_scene` result ignored** — same root cause as (2); the await result should be checked so the log line isn't misleading. Follow-up.

Filing these together; none block the approved scope.

yanziz-nvidia (commented):

## Summary

Consolidates the two-LLM setup (Llama-Nemotron 8B for quick-ack + Nemotron-3-Nano 30B for reasoning) into a single Nemotron-3-Nano server at port 8107, with thinking toggled per-request via `enable_thinking`. Also adds a `clear_scene` MCP tool called before `start_xr` to prevent stale objects, and tightens the system/quick-ack/validate prompt wording. PR has been open since 2026-05-06 and is currently **conflicting with main** (`mergeStateStatus: DIRTY`) — needs a rebase before merge.

|     | Finding |
|-----|---------|
| 🚫 | None |
| 💡 | [`agent-samples/xr-render-demo/worker/agent.py:101`](https://github.com/NVIDIA/xr-ai/blob/bc44ca2d31e952401b5fb63bfe631fac6a80be9a/agent-samples/xr-render-demo/worker/agent.py#L101) — `clear_scene` failure is silently ignored: `await self._call_render("clear_scene", {})` has no result check, while the immediately following `start_xr` call does. A failed clear leaves stale objects in LOVR while the code proceeds as if the slate is clean. At minimum log a warning on `None` return, or gate `start_xr` on the clear succeeding. |
| 🔍 | [`agent-samples/xr-render-demo/worker/processors.py:769`](https://github.com/NVIDIA/xr-ai/blob/bc44ca2d31e952401b5fb63bfe631fac6a80be9a/agent-samples/xr-render-demo/worker/processors.py#L769) — `max_tokens` for the thinking path dropped from 2048 → 1536 alongside the `thinking_budget` reduction from 1024 → 512. The comment says "512-token budget covers 3-axis arithmetic" — worth a quick note in the YAML or changelog if the thinking budget was tuned empirically, so future reviewers know why 512 was chosen. |

- `agent-samples/xr-render-demo/worker/agent.py:101` — After `await self._call_render("clear_scene", {})`, check the returned dict for `ok == False` (or `None`) and log a warning; optionally skip `start_xr` if the clear failed to avoid a session with leftover scene state.

PR #113, opened 2026-05-07
test(utils): add unit tests for launcher, vllm, and logging leaf modules

nvddr:
## Summary

- Adds 89 offline unit tests for the three stdlib/loguru-only leaf packages
  (`xr-ai-launcher`, `xr-ai-vllm`, `xr-ai-logging`) that previously had zero
  test coverage.
- Updates `tests/pyproject.toml` to add editable-install sources for the three
  new packages; updates `DEPENDENCIES.md` in the same commit per project rules.

**Files touched:**
- `tests/test_launcher_cloudxr_env.py` — env-file parser (quoting, comments, blank lines, error paths)
- `tests/test_launcher_credentials.py` — read/write/load/ensure helpers, file permissions, env priority
- `tests/test_launcher_gpu.py` — nvidia-smi output → GPU profile detection (Ada, Blackwell, Spark, fallback)
- `tests/test_launcher_stack.py` — `Process`/`Parallel` dataclass defaults, immutability, `launch_mode` values
- `tests/test_vllm_docker.py` — `build_run_argv` argv shape, `_registry_for`, `_already_logged_in`, container helpers
- `tests/test_vllm_lifecycle.py` — `health_url`, `health_ok`, `wait_until_healthy` polling semantics
- `tests/test_logging_setup.py` — namespace/timestamp env stamping, log file path, intercept handler, verbose flag
- `tests/pyproject.toml` — add editable deps for xr-ai-launcher, xr-ai-logging, xr-ai-vllm
- `DEPENDENCIES.md` — updated to reflect new editable deps in xr-ai-tests

## Test plan

- [x] All 89 new unit tests pass locally: `uv run pytest test_launcher_*.py test_vllm_*.py test_logging_setup.py -v` → **89 passed**
- [x] Existing 38 integration tests unaffected: `uv run pytest` (excluding new files) → **38 passed**
- [x] No GPU, Docker, ZMQ, or network access required — all tests are offline
- [x] `tests/pyproject.toml` change also updates `DEPENDENCIES.md` in the same commit

### DIFF
--- a/tests/test_launcher_credentials.py
+++ b/tests/test_launcher_credentials.py
@@ -6,7 +6,6 @@ from __future__ import annotations
 
 import json
 import os
-from pathlib import Path
 from unittest.mock import patch
 
 import pytest
--- a/tests/test_launcher_credentials.py
+++ b/tests/test_launcher_credentials.py
@@ -52,7 +52,7 @@ class TestReadWrite:
         assert json.loads(creds_file.read_text()) == {"HF_TOKEN": "abc"}
         assert oct(creds_file.stat().st_mode & 0o777) == oct(0o600)
 
-    def test_write_filters_non_string_values(self, fake_creds_dir):
+    def test_read_filters_non_string_values(self, fake_creds_dir):
         """_read skips non-string values written by other tools."""
         creds_file, _ = fake_creds_dir
         creds_file.write_text(json.dumps({"HF_TOKEN": "tok", "BAD": 123}))
--- a/tests/test_launcher_credentials.py
+++ b/tests/test_launcher_credentials.py
@@ -1,0 +1,143 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Unit tests for xr_ai_launcher._credentials (read/write/load/ensure helpers)."""
+from __future__ import annotations
+
+import json
+import os
+from pathlib import Path
+from unittest.mock import patch
+
+import pytest
+
+# Import the private helpers directly to exercise them in isolation.
+import xr_ai_launcher._credentials as _creds
+
+
+@pytest.fixture(autouse=True)
+def _clean_env(monkeypatch):
+    monkeypatch.delenv("HF_TOKEN",    raising=False)
+    monkeypatch.delenv("NGC_API_KEY", raising=False)
+
+
+@pytest.fixture()
+def fake_creds_dir(tmp_path, monkeypatch):
+    """Point _CREDS_FILE and _HF_TOKEN_FILE to tmp_path so tests never
+    touch ~/.config or ~/.cache."""
+    creds_file   = tmp_path / "credentials.json"
+    hf_token_file = tmp_path / "hf_token"
+    monkeypatch.setattr(_creds, "_CREDS_FILE",    creds_file)
+    monkeypatch.setattr(_creds, "_HF_TOKEN_FILE", hf_token_file)
+    return creds_file, hf_token_file
+
+
+class TestReadWrite:
+    def test_read_missing_file_returns_empty(self, fake_creds_dir):
+        assert _creds._read() == {}
+
+    def test_read_valid_json(self, fake_creds_dir):
+        creds_file, _ = fake_creds_dir
+        creds_file.write_text(json.dumps({"HF_TOKEN": "tok123"}))
+        assert _creds._read() == {"HF_TOKEN": "tok123"}
+
+    def test_read_corrupt_json_returns_empty(self, fake_creds_dir):
+        creds_file, _ = fake_creds_dir
+        creds_file.write_text("not json{{{")
+        assert _creds._read() == {}
+
+    def test_write_creates_file_with_restricted_permissions(self, fake_creds_dir):
+        _creds._write({"HF_TOKEN": "abc"})
+        creds_file, _ = fake_creds_dir
+        assert creds_file.exists()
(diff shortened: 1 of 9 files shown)


yanziz-nvidia (commented):
A few non-blocking notes inline — overall solid coverage. Approving once the inline items are looked at.

tests/test_logging_setup.py
```diff
@@ -0,0 +1,92 @@
...
+
+from xr_ai_logging import setup_logging
+
+
+@pytest.fixture(autouse=True)
+def _clean_log_env(monkeypatch, tmp_path):
+    """Redirect log output to tmp_path and clear timestamp/namespace stamps."""
+    monkeypatch.delenv("XR_AI_LOG_NAMESPACE", raising=False)
+    monkeypatch.delenv("XR_AI_LOG_TIMESTAMP", raising=False)
+    monkeypatch.setenv("XR_AI_LOG_ROOT", str(tmp_path))
```
yanziz-nvidia, line 28: `setup_logging` calls `logger.remove()`, which is a global Loguru reset. Harmless today, but if any other module ever adds its own loguru sink, those tests will be silently broken. Consider adding `logger.remove()` to the fixture teardown via `yield`:

```python
    monkeypatch.setenv("XR_AI_LOG_ROOT", str(tmp_path))
    yield
    logger.remove()
```

tests/test_logging_setup.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+class TestSetupLoggingReturnValue:
+    def test_returns_path_object(self, tmp_path):
+        result = setup_logging("test-proc")
+        assert isinstance(result, Path)
+
+    def test_log_file_created(self, tmp_path):
+        log_file = setup_logging("test-proc")
+        assert log_file.exists()
```
yanziz-nvidia, line 43: This relies on Loguru creating the file synchronously when `add()` is called even with `enqueue=True`. True today, but it's an undocumented impl detail — a one-line comment would help future readers know why the assertion is safe before any record is written.

tests/test_launcher_credentials.py
```diff
@@ -0,0 +1,142 @@
...
+        assert _creds._read() == {}
+
+    def test_write_creates_file_with_restricted_permissions(self, fake_creds_dir):
+        _creds._write({"HF_TOKEN": "abc"})
+        creds_file, _ = fake_creds_dir
+        assert creds_file.exists()
+        assert json.loads(creds_file.read_text()) == {"HF_TOKEN": "abc"}
+        assert oct(creds_file.stat().st_mode & 0o777) == oct(0o600)
+
+    def test_write_filters_non_string_values(self, fake_creds_dir):
```
yanziz-nvidia: Misleading name — this tests `_read()` filtering of non-string values, not `_write()`. Suggest renaming to `test_read_filters_non_string_values` (the docstring already says "_read skips...").

tests/test_vllm_docker.py
```diff
@@ -0,0 +1,161 @@
...
+    def test_unqualified_name_no_registry(self):
+        # A bare name with no slash and no dot/colon in the first component.
+        assert _registry_for("myimage") is None
+
+    def test_registry_with_port(self):
+        # host:port/image has a colon in the first segment
+        assert _registry_for("localhost:5000/myimage:latest") == "localhost:5000"
+
+    def test_plain_image_name(self):
+        assert _registry_for("myimage") is None
```
yanziz-nvidia: This is a duplicate of `test_unqualified_name_no_registry` above (same input `"myimage"`, same expected `None`). Either drop one, or change the input to something meaningfully different (e.g. `"myimage:latest"` — also no slash, no registry, but exercises the tag-stripping path).

tests/test_vllm_docker.py
```diff
@@ -0,0 +1,161 @@
...
+
+
+class TestAlreadyLoggedIn:
+    def test_no_docker_config(self, tmp_path, monkeypatch):
+        monkeypatch.setattr("xr_ai_vllm._docker._DOCKER_CONFIG", tmp_path / "config.json")
+        assert not _already_logged_in("nvcr.io")
+
+    def test_registry_in_auths(self, tmp_path, monkeypatch):
+        import json
```
yanziz-nvidia: `import json` inline here (and again at line 49) — move to the top-level imports for consistency with the rest of the test files.

nvddr:
Thanks — addressed all five in c4234a2:

- **test_logging_setup.py:21** — added `yield + logger.remove()` teardown to the autouse fixture so tests can't leak loguru sinks into one another.
- **test_logging_setup.py:31** — added a one-line comment explaining why the file path exists immediately under `add(enqueue=True)`.
- **test_launcher_credentials.py:55** — renamed to `test_read_filters_non_string_values`.
- **test_vllm_docker.py:33** — dropped the duplicate. The suggested `"myimage:latest"` variant is a better test, but it actually surfaces a real bug in `_registry_for` (the bare-tag path returns the input string instead of `None` because `:` matches the registry heuristic). Fixing that is a behavioral change and feels out of scope for a tests-only PR — I'll open a separate PR for it.
- **test_vllm_docker.py:42** — hoisted `import json` to the top of the module.

tests/test_vllm_docker.py
nvddr: Resolved by dropping the duplicate (your option A) in b8412e0. 

For option B (the suggested `"myimage:latest"` replacement): I tried that and it surfaced an actual bug in `_registry_for` — the bare-tag form returns `"myimage:latest"` instead of `None`, because the function uses `:` in the first segment-before-slash as a host:port marker, and a no-slash image makes the entire string the first segment. That is a behavioral fix beyond the scope of this tests-only PR, so it is split out into #117.

yanziz-nvidia (approved):

PR #114, opened 2026-05-07
test(agent-sdk): add encode/decode round-trip tests for xr_ai_agent codec

nvddr:
## Summary

- Adds 33 unit tests for `xr_ai_agent._codec` (encode/decode round-trips for all 12 registered message types).
- Tests cover: type_id byte preservation, all field values, binary data fidelity, `PixelFormat` enum round-trips, wire-format invariants (first byte = type_id), and unknown type_id error handling.
- No new dependencies — `xr-ai-agent` is already a dep of `xr-ai-tests`.
- File touched: `tests/test_agent_codec.py` only.

## Test plan

- [x] All 33 new tests pass: `uv run pytest test_agent_codec.py -v` → **33 passed in 0.04s**
- [x] No GPU, Docker, network, or ZMQ required — pure msgpack encode/decode
- [x] No `pyproject.toml` or `DEPENDENCIES.md` changes (xr-ai-agent already in tests deps)
- [x] File-disjoint from all other PRs in this batch

### DIFF
--- a/tests/test_agent_codec.py
+++ b/tests/test_agent_codec.py
@@ -1,0 +1,285 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Unit tests for xr_ai_agent._codec (encode/decode round-trips) and _types."""
+from __future__ import annotations
+
+import pytest
+
+from xr_ai_agent._codec import decode, encode
+from xr_ai_agent._types import (
+    AudioChunk,
+    ConnectorRegistration,
+    ControlMessage,
+    DataMessage,
+    FrameData,
+    FrameRequest,
+    FrameSignal,
+    MsgType,
+    ParticipantEvent,
+    PixelFormat,
+    ReturnAudioFlush,
+    RosterRequest,
+)
+
+
+# ── helpers ────────────────────────────────────────────────────────────────────
+
+def rt(msg_type: int, obj) -> object:
+    """Encode *obj* as *msg_type*, then decode the wire bytes. Returns the decoded object."""
+    return decode(encode(msg_type, obj))[1]
+
+
+def rt_type_id(msg_type: int, obj) -> int:
+    """Return the decoded type_id from the wire bytes."""
+    return decode(encode(msg_type, obj))[0]
+
+
+# ── type_id round-trips ────────────────────────────────────────────────────────
+
+class TestTypeIdPreservation:
+    """The type byte written by encode must come back from decode unchanged."""
+
+    def test_frame_signal_type_id(self):
+        msg = FrameSignal(0, 1, 1000, 640, 480, PixelFormat.NV12, 460800)
+        assert rt_type_id(MsgType.FRAME_SIGNAL, msg) == MsgType.FRAME_SIGNAL
+
+    def test_audio_chunk_type_id(self):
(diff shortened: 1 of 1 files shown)


yanziz-nvidia (commented):
Two non-blocking robustness suggestions inline. Codec coverage looks thorough.

tests/test_agent_codec.py
```diff
@@ -0,0 +1,285 @@
...
+        """Even the smallest message (RosterRequest, empty payload) must have
+        at least 1 byte for the type header."""
+        wire = encode(MsgType.ROSTER_REQUEST, RosterRequest())
+        assert len(wire) >= 1
+
+    def test_unknown_type_id_raises_on_decode(self):
+        """A byte sequence with an unregistered type_id must raise KeyError."""
+        import msgpack
+        # Build a fake wire message with type_id=255 (not registered).
+        fake = bytes([255]) + msgpack.packb([], use_bin_type=True)
```
yanziz-nvidia: Hard-coding type_id=255 means this test will silently pass if 255 ever gets registered later — the `KeyError` path stops being exercised. Consider picking the sentinel dynamically (scan `_DECODERS` for the first unused id) and asserting on the missing key:

```python
free_id = next(i for i in range(256) if i not in _DECODERS)
fake = bytes([free_id]) + msgpack.packb([], use_bin_type=True)
with pytest.raises(KeyError, match=str(free_id)):
    decode(fake)
```

tests/test_agent_codec.py
```diff
@@ -0,0 +1,285 @@
...
+        assert out.pts_us         == orig.pts_us
+        assert out.width          == orig.width
+        assert out.height         == orig.height
+        assert out.fmt            == orig.fmt
+        assert out.data_sz        == orig.data_sz
+        assert out.participant_id == orig.participant_id
+        assert out.track_id       == orig.track_id
+
+    def test_pixel_format_preserved(self):
+        for fmt in PixelFormat:
```
yanziz-nvidia: If a new `PixelFormat` is added that the codec doesn't support, this loop produces an opaque failure. `pytest.mark.parametrize("fmt", list(PixelFormat))` would give a distinct test ID per format and make failures immediately actionable.

tests/test_agent_codec.py
```diff
@@ -0,0 +1,285 @@
...
+
+
+# ── wire format sanity ──────────────────────────────────────────────────────────
+
+class TestWireFormat:
+    def test_first_byte_is_type_id(self):
+        """The wire format starts with the type byte (u8)."""
+        msg = FrameRequest("pid", "track")
+        wire = encode(MsgType.FRAME_REQUEST, msg)
+        assert wire[0] == int(MsgType.FRAME_REQUEST)
```
yanziz-nvidia, line 269: `wire[0] == int(MsgType.FRAME_REQUEST)` implicitly asserts the enum value fits in a single byte. That wire contract is load-bearing but unstated — a one-line comment (`# type byte is u8; enum values must stay in 0..255`) would help future readers.

nvddr:
Thanks — addressed all three in f026955:

- **line 114** — `test_pixel_format_preserved` is now `@pytest.mark.parametrize("fmt", list(PixelFormat))` so each format gets its own test id.
- **line 267** — added a one-line comment on the wire byte assertion explaining the u8 contract that registered `MsgType` values must respect.
- **line 283** — sentinel is now picked dynamically from `_decoders` (`next(i for i in range(256) if i not in _decoders)`), and the `pytest.raises(KeyError, match=str(free_id))` asserts the missing key.

yanziz-nvidia (approved):

PR #116, opened 2026-05-07
fix(logging): demote lovr log to files, add banner at task completion

wenxind-nvidia:
* utils/xr-ai-launcher/xr_ai_launcher/_stack.py: print a one-shot "All processes ready" banner from run_stack once every process has signaled ready, mirroring xr_ai_logging._print_log_dir_banner so the run begins and ends with matching brackets in the user's terminal. Add Process.quiet_native_output: bool = False; when set, _forward routes captured lines that don't carry a loguru HH:MM:SS.SSS prefix through stdlib log.debug instead of printing.

* utils/xr-ai-launcher/xr_ai_launcher/_processes.py: ManagedProcess routes captured subprocess stdout/stderr to <log_dir>/<name>.log when the orchestrator has stamped per-run env vars (XR_AI_LOG_*); falls back to today's prefixed prints for standalone use. Parses LOVR_LOG\\t<level>\\t<tag>\\t<msg> markers (emitted by the custom lovr.log callback in render-mcp's main.lua) for authoritative severity routing — warn/error mirror to the terminal, everything else is file-only. Native bytes that bypass lovr.log fall back to a case-insensitive \\b(error|warn|warning|fatal|panic)\\b regex so genuine C-level failures still surface in the terminal.

* utils/xr-ai-logging/xr_ai_logging/__init__.py: add print_task_done_banner(label, *, status, detail, duration_s) — a stdlib-only stderr banner with green/yellow/red status colors (done/interrupted/error), exported in __all__.

* agent-samples/simple-vlm-example/worker/agent.py: wrap _handle_query in try/except CancelledError/except Exception/finally with outcome tracking; emit print_task_done_banner per task end.

* agent-samples/xr-render-demo/worker/processors.py: same banner wiring in RenderSceneProcessor._drain. Demote per-turn CTX dump ("pre-fetched context for turn" + _trace_log "CTX") and per-tool "tool call" + _trace_log "TOOL" records from logger.info to .debug.

* agent-samples/xr-render-demo/worker/xr_render_demo_worker.py: lower the dedicated trace-file sink from level=INFO to level=DEBUG so the now-DEBUG CTX/TOOL records still reach /tmp/xr-agent-trace.log; USER/ACK/RESP records stay at INFO and remain visible in the terminal.

* agent-mcp-servers/render-mcp/xr_app/main.lua: install a custom lovr.log callback that emits a tab-separated LOVR_LOG marker on stdout for the launcher to parse. Migrate all 20 bare print() call sites to lovr.log(message, "info", "render-mcp-scene") so every Lua-emitted line carries an authoritative LOVR level via the override.

* agent-samples/xr-render-demo/main.py: enable quiet_native_output=True on the oxr-mcp Process so the OpenXR loader's C-level chatter (no Python timestamps, bypasses loguru) routes to orchestrator.log at DEBUG instead of polluting the terminal; oxr-mcp's own Python loguru lines (timestamped) keep flowing through.

* agent-mcp-servers/oxr-mcp/oxr_mcp_server/__main__.py: pass log_config=None to uvicorn.Config so its uvicorn / uvicorn.error / uvicorn.access loggers fall back to root and get intercepted by the loguru bridge installed in setup_logging. log_level= "warning" still applies (via setLevel on the three uvicorn loggers), so only WARNING+ uvicorn records reach loguru and the terminal — but they reach it as proper loguru-formatted lines
  (timestamped) instead of uvicorn's "ERROR:    message" raw stderr
  format.

### DIFF
--- a/agent-mcp-servers/oxr-mcp/oxr_mcp_server/__main__.py
+++ b/agent-mcp-servers/oxr-mcp/oxr_mcp_server/__main__.py
@@ -358,7 +358,13 @@ async def _serve(cfg: Config, ready_file: Path | None = None) -> None:
     logger.info("oxr-mcp: ready to serve get_head_pose (session opens lazily on first request)")
     try:
         app = build_mcp(source).http_app(path="/mcp")
-        uv_cfg = uvicorn.Config(app, host=cfg.host, port=cfg.port, log_level="warning")
+        # log_config=None skips uvicorn's own dictConfig install so its
+        # uvicorn / uvicorn.error / uvicorn.access loggers fall back to root
+        # and get intercepted by the loguru bridge installed in setup_logging.
+        # log_level="warning" still applies (set independently via setLevel
+        # on each uvicorn logger), so only WARNING+ records reach loguru.
+        uv_cfg = uvicorn.Config(app, host=cfg.host, port=cfg.port,
+                                log_level="warning", log_config=None)
         server = uvicorn.Server(uv_cfg)
         logger.info("oxr-mcp  mcp=/mcp  port={}", cfg.port)
         if ready_file:
--- a/agent-mcp-servers/render-mcp/xr_app/main.lua
+++ b/agent-mcp-servers/render-mcp/xr_app/main.lua
@@ -9,11 +9,23 @@
 --   { op="scene.remove", value={ id } }
 --   { op="scene.scale",  value={ id, scale } }    -- high-frequency audio path
 
-print("[render-mcp-scene] main.lua: top of file")
+-- ── lovr.log override ────────────────────────────────────────────────────────
+-- Emit a tab-separated marker so render-mcp's Python side (the
+-- xr_ai_launcher._processes._forward forwarder) can route by authoritative
+-- LOVR severity. Levels follow https://lovr.org/docs/lovr.log: debug, info,
+-- warn, error. Anything calling lovr.log(...) below (or any LOVR engine
+-- message routed through this callback) flows through here.
+function lovr.log(message, level, tag)
+    if message:sub(-1) == "\n" then message = message:sub(1, -2) end
+    io.stdout:write(string.format("LOVR_LOG\t%s\t%s\t%s\n", level, tag or "", message))
+    io.stdout:flush()
+end
+
+lovr.log("main.lua: top of file", "info", "render-mcp-scene")
 
 local zmq = require("lib.zmq")
 local mp  = require("lib.msgpack")
-print("[render-mcp-scene] lib.zmq + lib.msgpack loaded")
+lovr.log("lib.zmq + lib.msgpack loaded", "info", "render-mcp-scene")
 
 -- ── Scene state ───────────────────────────────────────────────────────────────
 
@@ -35,7 +47,7 @@ local ok, err = pcall(function()
 end)
 if not ok then
     recv_err = tostring(err)
-    print("[render-mcp-scene] ZMQ error: " .. recv_err)
+    lovr.log("ZMQ error: " .. recv_err, "info", "render-mcp-scene")
 end
 
 -- ── Helpers ───────────────────────────────────────────────────────────────────
@@ -71,7 +83,7 @@ end
 local function handle_scene_add(v)
     local id = v.id
     if not id then
-        print("[render-mcp-scene] scene.add: missing id — dropping")
+        lovr.log("scene.add: missing id — dropping", "info", "render-mcp-scene")
         return
     end
     local ptype      = v.type or "sphere"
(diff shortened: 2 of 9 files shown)


nvddr (commented):
## Summary

Three pieces of plumbing land here:
1. `ManagedProcess` gains a per-name file sink under `<log_dir>/<name>.log` plus a `LOVR_LOG\t<level>\t<tag>\t<msg>` parser that mirrors only `warn`/`error` to the terminal.
2. `_stack._forward` gains an opt-in `quiet_native_output` flag that demotes non-loguru-formatted lines to stdlib `log.debug` (file-only).
3. `xr_ai_logging` exports `print_task_done_banner` and `_stack` prints a matching "All processes ready" banner.

The change set is coherent and best kept as one PR — splitting the launcher utils from the consumer wiring would force ordered merges with no review benefit. CI is green (lock-check, tests 3.11/3.12, SPDX, CodeQL).

The design respects the launcher's stdlib-only contract (no new imports outside stdlib) and the `xr_ai_logging` loguru-only contract. PR description claims (banner shows, lovr → file, no terminal regression) check out against the diff.

## Blocking issues

None — the change is structurally sound and consistent with existing conventions.

## Suggestions

**1. Add unit tests for the new pure logic (`utils/xr-ai-launcher/`, `utils/xr-ai-logging/`).**
The PR introduces several easy-to-cover pure functions and parsers, into modules that already have a test suite (`tests/test_launcher_stack.py`, `tests/test_logging_setup.py`). Concretely:
- `_LOVR_LOG_PREFIX` parsing — match cases, malformed-marker fallthrough, empty-tag handling.
- `_LOVR_TERMINAL_RE` — case-insensitive match, no false-positives on identifier substrings (the "no false-match inside `…_error_…`" claim in the comment is a testable invariant).
- `_open_log_file` — env-vars-set vs unset paths, `OSError` fallthrough returning `None`.
- `_LOGURU_TIME_RE` in `_stack.py` — round-trip a real loguru-formatted line through `setup_logging`'s format string and assert the regex matches; this also pins the coupling described in the next item.
- `print_task_done_banner` — no-TTY plain-text variant (current tests don't have to assert ANSI), unknown-status fallback to "done" color.
- `Process(quiet_native_output=…)` — extend `test_launcher_stack.py::test_defaults` to assert the new field default.

**2. The `_LOGURU_TIME_RE` regex in `_stack.py` is implicitly pinned to `xr_ai_logging.setup_logging`'s format string.**
```python
_LOGURU_TIME_RE = re.compile(r"^\d{2}:\d{2}:\d{2}\.\d{3}\s")
```
The comment above it documents the coupling, but there is no runtime assertion or test linking the two. If anyone ever drops `.SSS` or prepends a character to the loguru stderr format in `setup_logging`, every line from a `quiet_native_output=True` process silently routes to DEBUG (terminal-invisible) instead of being printed. A test that emits a real loguru line through the configured format and asserts the regex matches would lock this in.

**3. `_open_log_file` reconstructs the `log_<ns>_<ts>` directory name by hand.**
The same path shape is built by `xr_ai_logging.setup_logging` (`log_dir = _resolve_log_root() / f"log_{ns}_{stamp}"`). It would be cheap to expose a tiny helper from `xr_ai_logging` (e.g. `current_log_dir() -> Path | None`) and have the launcher call it — except the launcher cannot import `xr_ai_logging` (loguru dep). So duplication is the right call here; just mention the coupling explicitly in `_open_log_file`'s docstring with a pointer to `xr_ai_logging.setup_logging` so the next person to touch either side knows to keep them in lockstep.

**4. Banner duration in `xr-render-demo` `_drain` includes the quick-ack phase.**
`t0 = time.monotonic()` is captured before `_quick_ack`; the banner reports the agentic-loop outcome but the elapsed time covers ack + still-task setup + agentic. That's defensible as "wall-clock for the user's turn," but the docstring/comment should make the choice explicit so a future reader doesn't try to "fix" it back to wrapping only `_agentic_task`.

## Nits

- `nit:` `_processes._forward`'s `level in ("warn", "error")` is exact-match. LOVR-spec levels are exactly `debug|info|warn|error` so this is correct against the spec, but a developer typo of `lovr.log("...", "warning", ...)` in Lua would silently go file-only. Either add `"warning"` to the tuple or assert against a frozen set with a comment pointing at the LOVR docs URL already in the module-level comment.
- `nit:` In `print_task_done_banner`, the docstring says unknown statuses "fall back to the `done` color" — consider explicitly listing `("done", "interrupted", "error")` as the accepted set in the docstring, since the colors dict is module-private.
- `nit:` The `Process` dataclass docstring's parameter alignment shifted from the prior 12-char column to 21 chars to fit `quiet_native_output`. Minor cosmetic but consider whether the multi-paragraph description for `quiet_native_output` (5 lines vs 1 line for every other field) belongs in the field comment or moved into the body of the docstring under a "Notes" heading.
- `nit:` `_processes.py`'s file-write fallback prints to `stdout` on `OSError|ValueError` ("File closed mid-shutdown — fall back so we don't lose the line."). Consider routing the fallback through stdlib `log.debug` rather than `print`, so it threads through the orchestrator's intercept handler like every other launcher-internal line.

## Verdict

**Comment-only** — no blockers. The change is well-scoped and ships fine as-is. Strongly recommend addressing **Suggestion 1** (tests) before merge, since the new pure functions are exactly the kind of regression surface this codebase otherwise covers well, and the format-coupling assertions in suggestion 2 are most cheaply expressed as tests.

— nvddr-reviewer

nvddr (approved):

PR #117, opened 2026-05-07
fix(xr-ai-vllm): _registry_for must require a slash, not just a colon

nvddr:
## Summary
- `_registry_for("myimage:latest")` used to return `"myimage:latest"` because the first segment-before-slash (the whole string, when no slash exists) contains `:` from the tag, which the registry heuristic mistakes for a `host:port` marker.
- Add the missing `"/" in image` precondition so bare tagged images correctly parse as having no registry.

Surfaced via review feedback on #113 — the bot's suggested replacement for a duplicate test (`"myimage:latest"` instead of `"myimage"`) immediately failed and revealed the bug. Kept that PR tests-only and split the behavioral change here.

## Manual verification
Ran every case through the patched `_registry_for`:

| Input | Old result | New result |
|---|---|---|
| `myimage` | `None` ✓ | `None` ✓ |
| `myimage:latest` | `"myimage:latest"` 🐛 | `None` ✓ |
| `library/myimage` | `None` ✓ | `None` ✓ |
| `library/myimage:latest` | `None` ✓ | `None` ✓ |
| `nvcr.io/nvidia/vllm:26.04-py3` | `"nvcr.io"` ✓ | `"nvcr.io"` ✓ |
| `localhost:5000/myimage:latest` | `"localhost:5000"` ✓ | `"localhost:5000"` ✓ |
| `docker.io/library/myimage` | `"docker.io"` ✓ | `"docker.io"` ✓ |

## Test plan
- [ ] Once #113 lands (which adds `xr-ai-vllm` to `tests/pyproject.toml` and the `_registry_for` test class), follow up by adding a regression case to `tests/test_vllm_docker.py::TestRegistryFor` for the `myimage:latest` input. Held back from this PR to keep it file-disjoint with #113 and avoid touching `tests/pyproject.toml` / `DEPENDENCIES.md`.

### DIFF
--- a/tests/test_vllm_docker.py
+++ b/tests/test_vllm_docker.py
@@ -30,6 +30,11 @@ class TestRegistryFor:
         # host:port/image has a colon in the first segment
         assert _registry_for("localhost:5000/myimage:latest") == "localhost:5000"
 
+    def test_tagged_unqualified_name_no_registry(self):
+        # A bare image with a tag must not be misread as a registry just
+        # because the tag's `:` looks like a host:port marker.
+        assert _registry_for("myimage:latest") is None
+
 
 class TestAlreadyLoggedIn:
     def test_no_docker_config(self, tmp_path, monkeypatch):
--- a/tests/test_vllm_docker.py
+++ b/tests/test_vllm_docker.py
@@ -35,6 +35,14 @@ class TestRegistryFor:
         # because the tag's `:` looks like a host:port marker.
         assert _registry_for("myimage:latest") is None
 
+    def test_namespace_no_registry(self):
+        # slash present but first segment has no dot or colon — Docker Hub library namespace
+        assert _registry_for("library/myimage") is None
+
+    def test_tagged_namespace_no_registry(self):
+        # same as above with an explicit tag — still not a registry reference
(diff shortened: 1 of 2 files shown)


yanziz-nvidia (commented):
Approved by yanziz

## Summary (for human)

Fixes `_registry_for` so a bare tagged image like `"myimage:latest"` is no longer returned as its own registry — the prior `":" in head` heuristic matched the tag's colon when no `/` separator was present. The patch adds a `"/" in image` guard before the host-vs-port check. No `pyproject.toml` changes, so no `DEPENDENCIES.md` update needed. CI green (pytest 3.11/3.12, SPDX, lock-check, CodeQL).

|     | Finding |
|-----|---------|
| 🚫 | None |
| 💡 | The PR body's manual-verification table covers `library/myimage` and `library/myimage:latest` (slash, no dot/colon in head → expect `None`), but neither is in the automated suite. Adding them locks in the fix against future regressions for cheap. |
| 🔍 | None |

- In `tests/test_vllm_docker.py`, add two cases to `TestRegistryFor`: `_registry_for("library/myimage")` and `_registry_for("library/myimage:latest")` should both return `None`.

yanziz-nvidia (approved):

PR #118, opened 2026-05-07
feat(ios-client): replace hard-coded camera picker with dynamic enumerate in case of external camera

wenxind-nvidia:
…ration

Use AVCaptureDevice.DiscoverySession to enumerate available cameras (built-in wide-angle + external) instead of a fixed front/back enum. Pass the selected AVCaptureDevice directly via LiveKit's CameraCaptureOptions(device:) API. Camera names come from AVCaptureDevice.localizedName — no hard-coding.

Default selection: back camera if available, otherwise index 0. Selection persisted by uniqueID across launches. visionOS unchanged (ARKit passthrough only, picker hidden).


nvddr (commented):
**Review by nvddr-reviewer**

## Summary
PR replaces `CameraConfig.Position { front, back }` with a free-form `AVCaptureDevice?` so `AVCaptureDevice.DiscoverySession` (built-in wide-angle + `.external`) can drive an iOS camera picker. The motivation — surfacing external/UVC cameras — is sound, and the App-side changes are well-organized (gated on `#if os(iOS)`, persisted by `uniqueID`, sensible default-selection fallback). Below are the concerns from a Round-1 architectural pass.

## Blocking issues

**1. Undeclared public SDK break.**
The PR is titled `feat(ios-client)` and the description frames this as an iOS picker change, but it removes/changes public surface in `StreamKit`:
- Removes `CameraConfig.Position` enum (`front`, `back`).
- Removes `CameraConfig.init(position:)`.
- Removes `CameraConfig.rear` preset.
- Weakens `Sendable` to `@unchecked Sendable`.

Per the sample's `README.md`, `StreamKit` is presented as a reusable SDK with a documented "Adding a custom backend" extension point — third-party consumers conforming to `StreamingBackend` will fail to compile. Either:
- Make the change additive: keep `position:` (and `.rear`) as a convenience that, on iOS, resolves internally to a discovery-session lookup; add a new `device:` initializer for the picker case. Or
- Explicitly call out the breaking SDK change in the PR description (and ideally in `docs/changelog.md` per `AGENTS.md`'s docs rule), and confirm with maintainers that StreamKit's surface is allowed to break.

I'd lean toward additive — `position` is a strictly less expressive subset of the new API and is cheap to keep.

**2. Hot-plug gap contradicts the PR's stated motivation.**
`availableCameras` is enumerated exactly once, at `AppModel` instantiation, in the property's default-value expression (`App/AppModel.swift:57`). External UVC cameras are precisely the class of device most likely to be plugged/unplugged at runtime. Concretely:
- Plug a UVC camera *after* the app launches → it never appears in the picker until the app restarts.
- Unplug the currently selected camera mid-session → `selectedCamera` resolves to a now-stale `AVCaptureDevice`, and the next `startCamera` will fail in a way the user can't recover from in-app.

Minimum bar to match the PR's intent: observe `AVCaptureDevice.wasConnectedNotification` / `wasDisconnectedNotification` (or KVO `discoverySession.devices`), refresh `availableCameras`, and re-resolve `selectedCameraID` with the same fallback that `resolveInitialCameraID()` already implements. Disconnect-mid-session also needs at least a user-visible error path (the existing `lastError` is fine).

## Suggestions

**3. Verify `CameraCaptureOptions(device: nil)` behavior in livekit-client-sdk-swift 2.13.**
On a device where `enumerateCameras()` returns `[]` (uncommon, but possible — e.g., iPad with no UVC + restricted permissions, or future hardware where neither device type matches), `selectedCamera?.device` is `nil` and the iOS path now calls `CameraCaptureOptions(device: nil)` (`StreamKit/.../LiveKitBackend.swift:283`). I haven't been able to confirm the LiveKit SDK's behavior for nil here from a primary source — please verify it has a sane fallback (e.g., default device) rather than throwing/crashing on the publish path, and add a guard in `startCamera` if not.

**4. `@unchecked Sendable` justification.**
`CameraConfig` was `Sendable`, now `@unchecked Sendable` solely so it can carry `AVCaptureDevice?`. `AVCaptureDevice` is documented thread-safe, so `@unchecked` is defensible — but Swift-6 strict concurrency mode is on (`Package.swift:34: .swiftLanguageMode(.v6)`), and the diff carries no comment explaining the escape hatch. Two options:
- One-line `// AVCaptureDevice is documented thread-safe; CameraConfig is otherwise immutable.` next to the `@unchecked` attribution. Or
- Keep `CameraConfig` properly `Sendable` by storing `uniqueID: String` (the picker already persists this), and resolve to `AVCaptureDevice` at the LiveKit boundary in `makeIOSTrack(config:)`. This also keeps `CameraConfig` cheaply value-comparable without the hand-rolled `==`.

The second option is also nicer for testability — `CameraConfig` stays a plain value, no AVFoundation dependency leaking up to consumers who only want to express "this camera by ID".

**5. README still describes only front/back.**
`README.md:24` was updated to "Camera device selection (iOS only)" — good — but §7 ("Build and run") still says `iOS / iPadOS device | AVCaptureSession front/back camera`. Per `AGENTS.md`'s "documentation rule" ("update in the same task as the code change"), that row should mention external/UVC cameras now too.

## Nits

- `nit:` `App/AppModel.swift:64` — `cameraOnDemand` block has a multi-sentence comment that includes a parenthetical about `bool(forKey:)` semantics duplicated from line 35. Per `AGENTS.md` "Comments" rules, a single sentence on the contract is enough; the duplicate isn't load-bearing. Out of scope to clean up since you're not changing those lines, but the next pass through that file could trim it.
- `nit:` `App/CameraEnumeration.swift:8` — `@unchecked Sendable` on `CameraInfo`: same comment as #4. Since every stored property except `device: AVCaptureDevice` is already `Sendable`, the same uniqueID-based design would let this type be plain `Sendable` too.
- `nit:` `App/AppModel.swift:247` — the deletion of `// `error.localizedDescription` strips domain/code/underlying cause.` removes a comment that does answer "why isn't the obvious version correct?" per the AGENTS comment rules. The DEBUG block immediately below it makes a lot more sense with that one-liner kept; consider re-adding.
- `nit:` `App/ContentView.swift:179` — `Picker("Camera", selection: $m.selectedCameraID)` binds to `String?` but no `ForEach` row carries a `nil` tag. If `selectedCameraID` is ever `nil` (initial enumeration returned empty, then refreshed) the picker has no matching row — SwiftUI will silently show whichever row it picks. A `.tag(String?.none)` placeholder row, or guarding the Picker behind `if model.selectedCameraID != nil`, makes the contract explicit.

## Verdict
Request changes. Items 1 and 2 are blocking on architectural grounds — the SDK break needs to be either reverted to additive or explicitly declared, and the hot-plug gap directly undermines the PR's stated purpose. Items 3–5 are addressable in the same revision; nits are non-blocking.

yanziz-nvidia (commented):

## Summary

Adds an `isCameraStarting` in-flight guard to `startCamera()` to prevent concurrent re-entry during the `await` suspension window, and refactors the iOS camera `Picker` from hard-coded `.front`/`.back` tags to a `ForEach` over `CameraConfig.Position.allCases` with a `CaseIterable` + `displayName` extension. The PR title and body describe a larger external-camera enumeration that was present in earlier revisions; the squashed head contains only these two targeted fixes. All CI checks pass (CodeQL, pytest, SPDX, uv lock).

|     | Finding |
|-----|---------|
| 🚫 | None |
| 💡 | [`App/AppModel.swift:277-285`](https://github.com/NVIDIA/xr-ai/blob/b36bc22b29c62e3a0e52b85a54efa92542fefa06/client-samples/ios-visionos/App/AppModel.swift#L277-L285) — **`switchCamera(to:)` calls `session?.startCamera(...)` without the `isCameraStarting` guard.** If a `startCamera()` task is in-flight when a camera switch arrives, two concurrent camera-start requests can race at the LiveKit layer. Applying the same guard pattern (or funneling through `startCamera()` after stopping) would be consistent with the fix this PR introduces. |
| 🔍 | [`App/ContentView.swift:369`](https://github.com/NVIDIA/xr-ai/blob/b36bc22b29c62e3a0e52b85a54efa92542fefa06/client-samples/ios-visionos/App/ContentView.swift#L369) — **`allCases` is `[.back, .front]` while `CameraConfig.init` defaults to `.front`.** The picker renders "Back" first; swapping to `[.front, .back]` would match the type's documented default and the typical iPhone UX ordering. |

- `client-samples/ios-visionos/App/AppModel.swift:277-285`: In `switchCamera(to:)`, check `isCameraStarting` before calling `session?.startCamera(...)`, or stop the current camera first and then call `startCamera()` — prevents a concurrent start race when a switch arrives while `startCamera()` is suspended at its `await`.
- `client-samples/ios-visionos/App/ContentView.swift:369`: Change `allCases` from `[.back, .front]` to `[.front, .back]` to match `CameraConfig.init`'s default of `.front` and typical iOS picker ordering (front listed first).

yanziz-nvidia (approved):

PR #120, opened 2026-05-10
fix(livekit-tls): make TLS the default, fix /rtc/v1 proxy, drop client-side toggle

nvddr:
## Summary

- `web_server_tls` now defaults to **true**. The hub's same-origin `wss://<host>:<web_server_port>/rtc` proxy is the only client-facing signaling path; LiveKit's native 7880 stays on loopback.
- Proxy widened to `/rtc/{tail:path}` (livekit-client v2.x appends `/rtc/v1`) and now forwards `Authorization` + end-to-end headers (the Swift SDK sends the JWT in the header, not the query string).
- Self-signed cert is generated as a CA (`BasicConstraints CA:TRUE` + KeyUsage/EKU/SubjectKeyIdentifier) so iOS exposes the Full Trust toggle; SAN is populated from routable IPv4 addresses via the UDP-connect trick so the cert covers the LAN IP that clients actually type. Stale caches (non-CA or missing-IP) auto-regenerate with a log banner directing the user to reinstall the profile.
- `TrustingSessionDelegate` for the iOS `/token` fetch is no longer gated on `#if DEBUG` — Release builds work.
- New `GET /cert` serves the cert with `application/x-x509-ca-cert` MIME for one-tap iOS profile install (cert bytes cached at startup).
- Mobile samples lose the `secure`/HTTPS-token toggle; default port `7880` → `8080`. Web non-XR client auto-detects scheme + port from `window.location.protocol`.
- Refactor pass: extracted `_lk_proxy.mount_rtc_proxy()` so `_web_server.py` and `_token_server.py` no longer duplicate the four `/rtc` routes; stdlib `wsgiref.util.is_hop_by_hop` for the proxy header filter; cert parsed once and shared between `_is_ca_cert` + `_cert_ipv4_san`; comments trimmed per `AGENTS.md` (decision history lives in `docs/changelog.md`).
- Docs updated: top-level + Android + iOS READMEs, `docs/networking.md`, `architecture.md`, `troubleshooting.md`, `xr-render-demo.md`, `changelog.md`.

## Test plan

- [x] `tests/`: 166/166 pass
- [x] TestClient smoke: `/token` returns `wss://<host>:<port>`; `/cert` returns 200 with `application/x-x509-ca-cert` + Content-Disposition; `/cert` returns 404 when `web_server_tls` is off; both `/rtc/validate` and `/rtc/v1/validate` resolve to the proxy; header filter keeps `Authorization` and drops hop-by-hop
- [x] Cert generated end-to-end: `openssl x509 -text` shows `CA:TRUE`, KeyUsage `digital_signature + key_encipherment + key_cert_sign`, EKU `SERVER_AUTH`, SAN with `127.0.0.1` plus detected local IPv4 addresses
- [x] Regen path: written a CA-marked cert with only `127.0.0.1` in SAN; `ensure_self_signed_cert` detects missing IPs and regenerates with the banner
- [x] End-to-end against a live LiveKit container on the AWS box that hit the original Slack issue
- [x] iOS sample: rebuild, install `/cert` profile, enable Full Trust, connect with host = remote IP, port = 8080 — wss handshake completes
- [x] Android sample: rebuild, connect with host = remote IP, port = 8080 — wss handshake completes (no profile install needed; `TrustAllCerts.kt` handles it)


yanziz-nvidia (commented):

## Summary

Makes TLS the default, fixes the LiveKit `/rtc/v1` proxy path for livekit-client v2.x, and removes the per-client `secure` toggle in favor of protocol auto-detection. Also auto-regenerates self-signed certs when SAN/CA bits are stale, adds a `/cert` endpoint, and consolidates the duplicated rtc-proxy routing into `mount_rtc_proxy()`.

|     | Finding |
|-----|---------|
| 🚫 | None |
| 💡 | [`client-samples/android/app/src/main/java/com/nvidia/xrai/streamkitsample/streamkit/backends/livekit/LiveKitBackend.kt:226`](https://github.com/NVIDIA/xr-ai/blob/8ff5091920b223ee064685a4ca3888b7e15557e9/client-samples/android/app/src/main/java/com/nvidia/xrai/streamkitsample/streamkit/backends/livekit/LiveKitBackend.kt#L226) — **Surface an actionable error when the token fetch fails before the cert is installed.** Old code used `TrustAllCerts.socketFactory` here; the new code doesn't. First-time users who tap Connect before Install CA Cert hit an opaque `SSLHandshakeException`. Catch it and show a snackbar pointing at the Install button, or scope a bootstrap trust-all the way `CertInstall.kt` already does for the cert download. |
| 💡 | [`server-runtime/xr_media_hub.yaml:32`](https://github.com/NVIDIA/xr-ai/blob/8ff5091920b223ee064685a4ca3888b7e15557e9/server-runtime/xr_media_hub.yaml#L32) — **Consider dropping the `web_server_tls=False` path entirely.** It's now half-supported: the web client auto-detects, but native clients hardcode `https://`/`wss://` with the `secure` toggle gone. Given auto-regen makes localhost TLS zero-friction, the flag is mostly a footgun. One TLS story is simpler to debug than two. |
| 💡 | [`client-samples/ios-visionos/StreamKit/Sources/StreamKit/Backends/LiveKit/LiveKitBackend.swift:586`](https://github.com/NVIDIA/xr-ai/blob/8ff5091920b223ee064685a4ca3888b7e15557e9/client-samples/ios-visionos/StreamKit/Sources/StreamKit/Backends/LiveKit/LiveKitBackend.swift#L586) — **Document why `TrustingSessionDelegate` is no longer `#if DEBUG`-guarded.** A 2–3 line comment stating that the trust-all is deliberate for the bootstrap token fetch (before the CA profile is installed) and that the wss handshake still uses system trust with the installed profile would prevent this from reading as a regression. |
| 🔍 | [`server-runtime/xr_media_hub/transport/livekit/_token_server.py:42`](https://github.com/NVIDIA/xr-ai/blob/8ff5091920b223ee064685a4ca3888b7e15557e9/server-runtime/xr_media_hub/transport/livekit/_token_server.py#L42), [`_web_server.py:51`](https://github.com/NVIDIA/xr-ai/blob/8ff5091920b223ee064685a4ca3888b7e15557e9/server-runtime/xr_media_hub/transport/livekit/_web_server.py#L51) — `@app.on_event("shutdown")` is deprecated since FastAPI 0.95 (project pins ≥0.111). Migrate to `lifespan` async context manager. |
| 🔍 | [`client-samples/android/app/src/main/java/com/nvidia/xrai/streamkitsample/CertInstall.kt:51`](https://github.com/NVIDIA/xr-ai/blob/8ff5091920b223ee064685a4ca3888b7e15557e9/client-samples/android/app/src/main/java/com/nvidia/xrai/streamkitsample/CertInstall.kt#L51) — `catch (_: Exception)` swallows failures silently; the Install button then does nothing with no user feedback. Log at minimum, ideally toast. |

- `client-samples/android/.../LiveKitBackend.kt` `fetchToken()` (L226): catch `javax.net.ssl.SSLHandshakeException` and surface a user-facing message pointing at the Install CA Cert button; or apply `TrustAllCerts.socketFactory` to the bootstrap token fetch the way `CertInstall.kt` already does.
- `server-runtime/xr_media_hub.yaml`, `transport/livekit/config.py`, `_web_server.py`: remove the `web_server_tls` flag and the `if cfg.web_server_tls` conditionals. TLS becomes mandatory. Update AGENTS.md / READMEs to drop references to the disabled path.
- `client-samples/ios-visionos/.../LiveKitBackend.swift` L586: prepend the `TrustingSessionDelegate` class with a 2–3 line comment stating that the previously `#if DEBUG`-guarded trust-all is intentional for the bootstrap token fetch only, and that the wss handshake still uses system trust with the installed CA profile.
- `server-runtime/.../_token_server.py` L42 and `_web_server.py` L51: replace `@app.on_event("shutdown")` with a `@asynccontextmanager` lifespan passed to `FastAPI(..., lifespan=...)`.
- `client-samples/android/.../CertInstall.kt` L51: replace `catch (_: Exception)` with `catch (e: Exception) { Log.w("CertInstall", "...", e); null }`; in the caller (`MainActivity`) show a toast when the returned intent is null.

server-runtime/xr_media_hub/transport/livekit/config.py
```diff
@@ -52,10 +52,13 @@ class LiveKitConnectorConfig:
...
-    # ~/.local/share/xr-ai/ on first run. Supply cert_file/key_file to use
-    # your own certificate instead.
-    web_server_tls:    bool = False
+    # HTTPS is on by default — required for camera access from any device that
+    # isn't localhost, and required so the same-origin /rtc proxy can carry
+    # LiveKit signaling as wss:// without browser mixed-content blocks.
+    # A self-signed cert is auto-generated in ~/.local/share/xr-ai/ on first
+    # run; supply cert_file/key_file to use your own.
+    # Set to False only for localhost-only dev where the cert warning is noise.
+    web_server_tls:    bool = True
```
yanziz-nvidia, line 65: I wonder what is the case to set it to false now that we want secure websocket.

yanziz-nvidia (approved):

server-runtime/xr_media_hub/transport/livekit/config.py
nvddr: Fair callout — the old comment ("only for localhost-only dev where the cert warning is noise") is stale now that the install button is in. Two cases still genuinely want the hub to *not* terminate TLS itself, and I've rewritten the comment to spell both out (19c2ecf):

1. A TLS-terminating reverse proxy in front (nginx, Caddy, Cloudflare Tunnel) — the hub speaks plain `http://` + `ws://` to the proxy on the loopback and the proxy serves wss/https to clients. Common in any production deployment that wants a real public CA cert.
2. `localhost`-only dev — browsers grant camera/mic on `http://localhost` without TLS, so for fast iteration on the same machine the cert install/regen dance is friction with no payoff.

Happy to drop the field entirely if you'd rather, but my read is (1) is load-bearing for anyone running this past the LAN demo and (2) is a small but real ergonomics win during dev. The default is `True` either way.

PR #122, opened 2026-05-12
test(vlm-server): GPU smoke + fix device-order, docstring, orphan-vLLM teardown

nvddr:
## Summary

GPU smoke test for `ai-services/vlm-server` (Cosmos-Reason-7B) plus two adjacent fixes.

**New test** (`tests/test_gpu_vlm_server.py`, `@pytest.mark.gpu`): spawns the real `vlm_server` subprocess, sends a tiny 32×32 PNG via `/v1/chat/completions`, asserts a non-empty content string. Cold-start budget 1800 s so the test is self-sufficient on a fresh box (vLLM downloads + reuses from `~/.cache/huggingface`).

**Fixes folded in:**
- `ai-services/vlm-server/vlm_server/__main__.py` — moved HF cache env setup *before* the CUDA device probe so values set via env are respected (#6); replaced "OpenAI vision API format" docstring with "vLLM's OpenAI-compatible /v1/chat/completions endpoint" (#9).
- `tests/_helpers.py` — added `kill_orphan_vllm(port)` (probe `/health` → `pid_on_port` → SIGTERM → SIGKILL); the vlm wrapper Popens `vllm serve` with `start_new_session=True` (prod design — survives stack restart) so the wrapper's SIGTERM doesn't reach vLLM, and ~35 GB of VRAM would stay pinned across tests without explicit cleanup.

## Stacked relationship

This PR includes `feat/tests-gpu-marker`'s commit as a parent. Once #<scaffold-pr> lands, this diff shrinks to just the vlm-server changes.

## Test plan

- [x] `pytest -m gpu test_gpu_vlm_server.py -v` passes on L40S (~65 s warm, ~30 min cold)
- [x] `pytest -m "not gpu" -q` → 166 passed (non-regression)
- [x] No leaked VRAM after teardown (verified via `nvidia-smi --query-compute-apps`)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -52,4 +52,4 @@ jobs:
 
       - name: Run tests
         working-directory: tests
-        run: uv run pytest -v --tb=short --color=yes
+        run: uv run pytest -v --tb=short --color=yes -m "not gpu"
--- a/ai-services/vlm-server/vlm_server/__main__.py
+++ b/ai-services/vlm-server/vlm_server/__main__.py
@@ -8,8 +8,8 @@ Reads config, builds vLLM serve flags, and dispatches through ``xr_ai_vllm.serve
 to either the pip-installed ``vllm`` CLI or the NGC ``nvcr.io/nvidia/vllm``
 docker container — picked per-YAML via ``vllm_backend: pip|docker``.
 
-Images are passed as base64 data URLs in the ``image_url`` content block,
-same as the OpenAI vision API format.
+Serves vLLM's OpenAI-compatible /v1/chat/completions endpoint; images are
+passed as base64 data URLs in the ``image_url`` content block.
 
 Accepts ``--config <path>.yaml`` (auto-passed by xr-ai-launcher).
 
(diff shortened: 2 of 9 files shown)


yanziz-nvidia (commented):

## Summary

Adds a GPU smoke test (`@pytest.mark.gpu`) for the vlm-server that spawns a real vLLM subprocess, sends a 32x32 PNG, and asserts a non-empty response; also fixes CUDA_VISIBLE_DEVICES ordering in `__main__.py` and adds `kill_orphan_vllm` to prevent VRAM leaks across tests. All CI checks pass.

|     | Finding |
|-----|---------|
| 🚫 | None |
| 💡 | [`tests/test_gpu_vlm_server.py:101`](https://github.com/NVIDIA/xr-ai/blob/617043167fb380de79d474c7c3931f5eee0284f4/tests/test_gpu_vlm_server.py#L101) — `_pick_free_port` has a TOCTOU race: the port is released before `vllm serve` binds it, so another process can steal it between the two syscalls. Flakiness risk on loaded CI nodes. |
| 💡 | [`tests/_helpers.py:56-72`](https://github.com/NVIDIA/xr-ai/blob/617043167fb380de79d474c7c3931f5eee0284f4/tests/_helpers.py#L56-L72) — `kill_orphan_vllm` silently swallows all exceptions from `pid_on_port`, including `ImportError` for `xr_ai_vllm` not being installed. If the import fails, teardown returns without killing anything but the caller has no signal. |
| 🔍 | [`tests/test_gpu_vlm_server.py:29`](https://github.com/NVIDIA/xr-ai/blob/617043167fb380de79d474c7c3931f5eee0284f4/tests/test_gpu_vlm_server.py#L29) — `_weights_available()` is defined but never called to gate the test; the test only skips on missing `uv` or `httpx`, so on a box without weights it will hang for up to 1800 s before failing. |

> Fix if it makes sense in context — these are agent-generated suggestions, not human-vetted obligations. Skip anything wrong, already addressed, or not worth the churn.

- `tests/test_gpu_vlm_server.py:101`: replace `_pick_free_port()` with `SO_REUSEPORT` or pass `port=0` to vllm and parse its stdout, to eliminate the TOCTOU race.
- `tests/_helpers.py:56-72`: add `logging.warning(...)` (or re-raise) inside the broad `except Exception` so teardown silences are visible in test output.
- `tests/test_gpu_vlm_server.py:29-30`: call `if not _weights_available(): pytest.skip("Cosmos-Reason1-7B weights not found")` at the top of the test to avoid a 30-min cold hang.

yanziz-nvidia (approved):

PR #126, opened 2026-05-12
test(video): add NVENC recorder GPU tests + surface encoder-recreate failures

nvddr:
## Summary

GPU tests for `server-runtime/xr_media_hub/video/_recorder.py` (NVENC video recording) plus a silent-failure fix.

**New tests** (`tests/test_gpu_video_recorder.py`, `@pytest.mark.gpu`):
- `test_record_synthetic_frames`: feeds 30 synthetic NV12 frames to `VideoRecorder`, asserts at least one `.h264` chunk + matching `.json` sidecar on disk; checks H.264 NAL start code prefix.
- `test_resolution_change_surfaces_error`: feeds 10 frames at 640×480 then 10 at 1280×720; asserts either both resolutions appear in chunks or the new sticky `failed` flag is set — no silent frame drops.

**Fix folded in (bug #7):**
- `server-runtime/xr_media_hub/video/_recorder.py` — added a sticky `_TrackEncoder.failed` flag set when `_create_encoder()` raises during a resolution change. `_encode_frame` short-circuits when set, so we no longer dereference `encoder=None` and swallow `AttributeError` as a generic per-frame warning. The create-failure log is promoted from `warning` to `error` and now includes old→new resolution + pid.

## Test plan

- [x] `pytest -m gpu test_gpu_video_recorder.py -v` — both tests pass on L40S
- [x] `pytest -m "not gpu" -q` → no regressions
- [x] Skip path: tests skip cleanly when `PyNvVideoCodec` import fails or NVENC unreachable

### DIFF
--- a/server-runtime/xr_media_hub/video/_recorder.py
+++ b/server-runtime/xr_media_hub/video/_recorder.py
@@ -82,6 +82,9 @@ class _TrackEncoder:
     height:         int            = 0
     out_dir:        Path           = field(default_factory=Path)
     last_ts:        float          = 0.0
+    # Sticky once NVENC (re)creation fails — _encode_frame short-circuits so
+    # we don't dereference encoder=None and turn every frame into a silent drop.
+    failed:         bool           = False
 
 
 class VideoRecorder:
@@ -135,6 +138,8 @@ class VideoRecorder:
             enc = self._encoders[key]
 
         with enc.lock:
+            if enc.failed:
+                return
             if now - enc.last_ts < self._min_interval:
                 return
             enc.last_ts = now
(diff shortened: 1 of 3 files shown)


yanziz-nvidia (commented):

## Summary

Adds GPU integration tests for `VideoRecorder`'s NVENC path and folds in a one-line fix that promotes encoder-recreate failures to a sticky `failed` flag, preventing silent per-frame `AttributeError` drops on resolution changes. All CI checks pass.

|     | Finding |
|-----|---------|
| 🚫 | None |
| 💡 | [`tests/test_gpu_video_recorder.py:24`](https://github.com/NVIDIA/xr-ai/blob/3e4716237a9482523603ef122290ddf0ede5fd1c/tests/test_gpu_video_recorder.py#L24) — `pynvc` is bound but never used; the bare-except `try/except` block exists only to skip the module, so the alias is dead. Drop `as pynvc` (or alias to `_`) to silence the linter. |
| 🔍 | [`tests/test_gpu_video_recorder.py:83-85`](https://github.com/NVIDIA/xr-ai/blob/3e4716237a9482523603ef122290ddf0ede5fd1c/tests/test_gpu_video_recorder.py#L83) — the inner `except Exception: pass` around `probe.EndEncode()` is bare and uncommented; add `except Exception:  # best-effort teardown, ignore` to satisfy static analysis. |

> Fix if it makes sense in context — these are agent-generated suggestions, not human-vetted obligations. Skip anything wrong, already addressed, or not worth the churn.

- `tests/test_gpu_video_recorder.py:24`: drop the unused `pynvc` alias.
- `tests/test_gpu_video_recorder.py:83-85`: annotate the bare `except Exception: pass` around `EndEncode()` so static analysis stops flagging it.

yanziz-nvidia (approved):

PR #132, opened 2026-05-13
fix(simple-vlm-example): require fresh FrameSignal in _wait_for_camera_frame

wenxind-nvidia:
A stale FrameSignal from a track that has since stopped persists in self._latest and was being returned by the camera-start wait. The subsequent request_frame then served an 8x8 placeholder because the underlying track was gone, leaving the VLM with no real image. Require _is_fresh on both the immediate-shortcut path and the post-event path.

### DIFF
--- a/agent-samples/simple-vlm-example/worker/agent.py
+++ b/agent-samples/simple-vlm-example/worker/agent.py
@@ -378,11 +378,12 @@ class SimpleVlmAgent:
     async def _wait_for_camera_frame(
         self, pid: str, timeout: float,
     ) -> FrameSignal | None:
-        """Wait up to ``timeout`` seconds for any new FrameSignal for ``pid``.
+        """Wait up to ``timeout`` seconds for a fresh FrameSignal for ``pid``.
 
-        During the camera-start wait, age doesn't matter — any arriving frame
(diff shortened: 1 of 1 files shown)


nvddr (approved):
LGTM — pre-approving.

Correct, minimal fix. `_latest_signal` picks max(`pts_us`) across all `(pid, track_id)` keys, and a stopped track is only evicted on participant leave (`_on_participant`) — so a stopped-but-still-present track can dominate `max(pts_us)` until the participant departs. Gating both the immediate-shortcut and post-event branches on `_is_fresh` is the right place to fix it; the wait loop already re-clears the event and re-iterates, so a fresh frame from a newly-started track will wake it (or the deadline fires). Solid PR description with a concrete failure mode.

nit: the docstring "why" paragraph is on the long side relative to AGENTS.md's comments guidance (one to two sentences, lead with the contract). Something like:

> Wait up to ``timeout`` seconds for a fresh FrameSignal for ``pid``. Stale signals are rejected because a stopped track lingers in ``self._latest`` until the participant leaves, and returning one yields an 8x8 placeholder from ``request_frame``.

would tighten it. Non-blocking.

— nvddr-reviewer

PR #135, opened 2026-05-14
feat(xr-ai-models): unified LLM/VLM/STT/TTS SDK with OpenAI-compat clients

nvddr:
## Summary

Unit 1 of the modular AI-models refactor — introduces `agent-sdk/xr-ai-models/`, a new SDK that absorbs every hand-rolled `httpx` wrapper around `/v1/chat/completions`, `/v1/audio/transcriptions`, and `/v1/audio/speech` behind four service protocols (`LLMService`, `VLMService`, `STTService`, `TTSService`) with `OpenAICompat*` implementations covering all seven in-tree backends.

- Per-model quirks (reasoning-field aliasing for `nano_v3` vs `nemotron_v3`; `chat_template_kwargs` plumbing for `enable_thinking` / `thinking_budget`) live behind the SDK seam — callers do not branch on them.
- Built-in presets for `cosmos_vlm`, `llama_nemotron`, `nemotron3_nano`, `nemotron_omni`, `parakeet_stt`, `piper_tts`, `magpie_tts` — a sample's `yaml/models.yaml` entry becomes `kind: preset:<name>` + `base_url:`.
- `AGENTS.md` gains the new hard rule "All HTTP calls to AI services go through `agent-sdk/xr-ai-models`" and `DEPENDENCIES.md` / `docs/changelog.md` are updated in the same commit per the docs rule.
- No callers change in this PR — subsequent units migrate `vlm-mcp`, `simple-vlm-example`, `xr-render-demo`, and `xr-ai-pipecat` to depend on this SDK.

## Test plan

- [x] 46 new unit tests under `tests/test_models_*.py` against an `httpx.MockTransport` stub
- [x] Full non-GPU test suite (220 tests) regression-clean on Python 3.12
- [x] SPDX header check passes (305 files)
- [x] `uv lock` resolves cleanly for the new package
- [ ] Reviewer: `cd tests && uv sync && uv run pytest -m "not gpu" -k models`

### DIFF
--- a/agent-sdk/xr-ai-models/README.md
+++ b/agent-sdk/xr-ai-models/README.md
@@ -62,7 +62,7 @@ Built-in presets — see `xr_ai_models/presets/`:
 
 | Preset | Service it targets | Notes |
 |---|---|---|
-| `cosmos_vlm`     | vlm-server               | `enable_thinking=false` by default |
+| `cosmos_vlm`     | vlm-server               | image + video; `enable_thinking=false` by default. Video requires vlm-server's `max_videos_per_prompt >= 1` |
 | `llama_nemotron` | llama-nemotron-llm-server | OpenAI tool calling via llama3_json (server-side) |
 | `nemotron3_nano` | nemotron3-nano-llm-server | reasoning field: `reasoning` |
 | `nemotron_omni`  | nemotron-omni-llm-server  | reasoning field: `reasoning_content`, vision + video |
@@ -101,6 +101,9 @@ class LLMService(Protocol):
 class VLMService(Protocol):
     capabilities: Capabilities
     async def ask_image(self, image, question, *, system_prompt="",
+                        max_tokens=None, temperature=None,
+                        timeout=None) -> ChatResponse: ...
+    async def ask_video(self, video, question, *, system_prompt="",
                         max_tokens=None, temperature=None,
                         timeout=None) -> ChatResponse: ...
 
--- a/agent-sdk/xr-ai-models/README.md
+++ b/agent-sdk/xr-ai-models/README.md
@@ -1,0 +1,131 @@
+<!--
+  SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+  SPDX-License-Identifier: Apache-2.0
+-->
+
+# xr-ai-models
+
+Unified service protocols and OpenAI-compatible HTTP clients for the xr-ai
+model layer.  Worker code depends on the four protocols
+(`LLMService`, `VLMService`, `STTService`, `TTSService`) and constructs
+concrete clients from a `models.yaml` config — no hand-rolled httpx calls
+in callers, no model quirks leaking out of this package.
+
+## Why
+
+Before this package, every consumer (vlm-mcp, simple-vlm-example worker,
+xr-render-demo worker, xr-ai-pipecat) rolled its own httpx wrappers and
+hard-coded model quirks (`chat_template_kwargs`, served-model-name strings,
+the `reasoning` vs `reasoning_content` field difference between
+nano_v3 and nemotron_v3 parsers).  Swapping an LLM meant editing N files.
+
+After: one `models.yaml` per sample names the logical models the worker
+needs; `make_llm(config, "agent_llm")` returns something that satisfies
+`LLMService` regardless of which backend or quirks are involved.
+
+## Quickstart
+
+```python
+from xr_ai_models import load_models_config, make_llm, ChatMessage
+
+config = load_models_config("yaml/models.yaml")
+async with make_llm(config, "agent_llm") as llm:
+    resp = await llm.chat(
+        [ChatMessage(role="user", content="hello")],
+        max_tokens=128,
+        enable_thinking=True,
+    )
+    print(resp.content, resp.reasoning)
+```
+
+`models.yaml`:
+
+```yaml
+agent_llm:
+  kind:     preset:nemotron3_nano
+  base_url: http://localhost:8107
+
+vlm:
+  kind:     preset:cosmos_vlm
+  base_url: http://localhost:8100
+
+stt:
+  kind:     preset:parakeet_stt
+  base_url: http://localhost:8103
+
+tts:
+  kind:     preset:piper_tts
+  base_url: http://localhost:8105
+```
+
+Built-in presets — see `xr_ai_models/presets/`:
+
+| Preset | Service it targets | Notes |
+|---|---|---|
+| `cosmos_vlm`     | vlm-server               | `enable_thinking=false` by default |
+| `llama_nemotron` | llama-nemotron-llm-server | OpenAI tool calling via llama3_json (server-side) |
+| `nemotron3_nano` | nemotron3-nano-llm-server | reasoning field: `reasoning` |
+| `nemotron_omni`  | nemotron-omni-llm-server  | reasoning field: `reasoning_content`, vision + video |
+| `parakeet_stt`   | stt-server               | |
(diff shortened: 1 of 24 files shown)


yanziz-nvidia (commented):

## Summary

Introduces `agent-sdk/xr-ai-models/` — a new SDK package that consolidates seven hand-rolled `httpx` wrappers for LLM/VLM/STT/TTS services behind four typed protocols and one `OpenAICompat*` implementation each. `AGENTS.md`, `DEPENDENCIES.md`, and `docs/changelog.md` are all updated in the same commit per project convention. CI is green across CodeQL, pytest 3.11/3.12, SPDX check, and `uv lock`.

|     | Finding |
|-----|---------|
| 🚫 | None |
| 💡 | [`agent-sdk/xr-ai-models/xr_ai_models/openai_compat.py:893`](https://github.com/NVIDIA/xr-ai/blob/63145b8fa26ed0b9959a7ed4ca8cc72b909064e6/agent-sdk/xr-ai-models/xr_ai_models/openai_compat.py#L893) — `logger.error` fires before `raise_for_status()` so the log line is emitted even for errors the caller silently retries or catches; callers end up with spurious ERROR log noise. Consider logging at WARNING, or only logging when the caller does not raise (i.e. move the log inside an except block). Affects all four client classes at the same pattern. |
| 🔍 | [`agent-sdk/xr-ai-models/xr_ai_models/config.py:566`](https://github.com/NVIDIA/xr-ai/blob/63145b8fa26ed0b9959a7ed4ca8cc72b909064e6/agent-sdk/xr-ai-models/xr_ai_models/config.py#L566) — `_require_str` rejects an empty string (`not val`) but a `base_url` of `"  "` (whitespace only) passes that guard then gets silently stripped by `.rstrip("/")` later, leaving a nonsensical URL. `val.strip()` before the emptiness check would close the gap. |

- In `openai_compat.py` `chat`, `stream`, `transcribe`, `synthesize`: demote `logger.error` to `logger.warning` (or log inside an `except httpx.HTTPStatusError` block after `raise_for_status`) so callers that catch and retry don't get spurious ERROR log entries — the exception itself is the signal.
- In `config.py` `_require_str`: change `if not isinstance(val, str) or not val` to `if not isinstance(val, str) or not val.strip()` to reject whitespace-only strings before they reach URL construction.

agent-sdk/xr-ai-models/README.md
```diff
@@ -0,0 +1,131 @@
...
+  base_url: http://localhost:8103
+
+tts:
+  kind:     preset:piper_tts
+  base_url: http://localhost:8105
+```
+
+Built-in presets — see `xr_ai_models/presets/`:
+
+| Preset | Service it targets | Notes |
```
yanziz-nvidia, line 63: do we plan for a custom one? or someone needs to update code repo by sending a PR to add a new type and model support?

agent-sdk/xr-ai-models/xr_ai_models/config.py
```diff
@@ -0,0 +1,182 @@
...
+from pathlib import Path
+from typing import Any, Literal, TypeVar
+
+import yaml
+
+from . import presets as _presets
+from ._utils import merge_dicts
+
+
+Category = Literal["llm", "vlm", "stt", "tts"]
```
yanziz-nvidia, line 23: Wonder how would it scale when multi model comes out. Seems capabilities more important than category?

nvddr:
Pushed `a05c730` adding video support — and answering your "capabilities vs category" question with a concrete case.

## What's in the new commit

Cosmos-Reason1-7B is a Qwen2.5-VL fine-tune **primarily designed for video reasoning**, but the SDK as I'd introduced it exposed only `ask_image`. So the preset was undersell­ing the model and there was no typed path to send video. The new commit adds:

- `VideoPart(url, type="video_url")` content type + `VideoInput` alias mirroring `ImagePart`/`ImageInput`.
- `VLMService.ask_video(video, question)` — same call shape as `ask_image`, builds a `[VideoPart, TextPart]` user message.
- `_normalize_video` / `_sniff_video_mime` handle mp4 + webm + path-suffix mime, matching the image helpers.
- `cosmos_vlm` preset now declares `capabilities.video=True` with a docstring noting the server-side opt-in (vlm-server's `max_videos_per_prompt` defaults to 0; vLLM reserves tens of GiB of activation memory for video at startup).
- `ask_video` raises `ValueError` if `capabilities.video` is False — typed error at the seam instead of a vLLM 500 once a deployment hasn't opted in.

4 new wire-trace tests (mp4 bytes-sniff, webm path-suffix, http URL passthrough, capability-off guard) — full test count is now 50 in `-k models`, 224 in the non-GPU suite, all green. SPDX clean.

## Re: your two design questions

**Re `README.md:63` ("do we plan for a custom one?"):**
Presets are intentionally in-tree only. The whole reason `presets/<name>.py` exists is so that the per-model quirks (reasoning-field name, `chat_template_kwargs` defaults, declared capabilities) live in one reviewable place rather than scattered across every sample's YAML. Adding a new in-tree model is a ~10-line dict in a new `presets/<name>.py` plus a registry entry in `presets/__init__.py` — same pattern as the seven existing ones. There's no out-of-tree "register-your-own preset" hook by design. Out-of-tree users (when we have any) can always inline the full spec without a preset reference (the `category: …` field documents that path in the README).

**Re `config.py:23` ("Seems capabilities more important than category"):**
They're orthogonal:
- `category` pins the **typed contract** — which protocol you get and which methods that protocol exposes (`chat` vs `ask_image` vs `ask_video` vs `transcribe` vs `synthesize`).
- `capabilities` describes **runtime features** — what the protocol's content carries, what knobs the model supports (`vision`, `video`, `tool_calls`, `reasoning`, `streaming`).

This commit is a concrete example: Cosmos was already `category: vlm` (its contract is "answer questions about visual media"), but its capabilities needed to grow (`vision` → `vision + video`) and the protocol's surface needed to grow alongside (`ask_image` → `ask_image + ask_video`). When future models extend modalities further, we grow `capabilities` and (if it needs a typed method) grow the protocol; we rarely move `category`. Nemotron-Omni is the other side of the coin — it's primarily a chat model with vision/video capabilities, so `category: llm` and callers reach video via `LLMService.chat()` with multi-part content.

I won't auto-resolve the threads — feel free to push back if you want the design split differently.

agent-sdk/xr-ai-models/README.md
nvddr: Short answer: presets are in-tree only by design — new model = ~10-line dict in `presets/<name>.py` plus a registry entry in `presets/__init__.py`, same as the existing seven. Out-of-tree users can always inline the full spec without a preset reference.

Long answer + rationale in the PR-level reply: https://github.com/NVIDIA/xr-ai/pull/135#issuecomment-4482795505

agent-sdk/xr-ai-models/xr_ai_models/config.py
nvddr: They're orthogonal: `category` pins the typed protocol contract (`ask_image` vs `chat` vs `transcribe` vs `synthesize`); `capabilities` describes which content parts / runtime features the protocol carries. When a model spans modalities we grow capabilities and (if it deserves a typed helper) grow the protocol — `a05c730` is a concrete example: Cosmos got `video: true` and `VLMService` got `ask_video`. We rarely move `category`.

Full rationale + the Nemotron-Omni / Cosmos comparison: https://github.com/NVIDIA/xr-ai/pull/135#issuecomment-4482795505

yanziz-nvidia: I see. It'd be good to add some comments

yanziz-nvidia (approved):

PR #137, opened 2026-05-14
refactor(xr-ai-pipecat): rewire SttClient/TtsClient on xr-ai-models SDK

nvddr:
## Summary

Unit 5 of the modular AI-models refactor — rewires pipecat's
`SttClient` and `TtsClient` to be thin wrappers around the SDK's
`OpenAICompatSTT` / `OpenAICompatTTS`.

- Public constructor signatures `(base_url, timeout=30.0)` preserved.
- Public methods `transcribe(audio, sample_rate, channels)` / `synthesize(text)` preserved.
- PCM→WAV conversion now lives inside the SDK — pipecat's local `_pcm_to_wav` is gone.
- `http_probe`, `mcp_probe`, `wait_for_services` are unchanged (unrelated to model HTTP).
- `DEPENDENCIES.md` updated in the same commit per the docs rule.
- New `tests/test_pipecat_services_wire.py` drives both clients against `_stub_openai` and asserts the wire shape (multipart for STT, JSON for TTS) matches the pre-migration goldens — 12 tests, all pass.

Stacked on #135 — Unit 5 of the modular AI-models refactor. When #135 merges, GitHub auto-retargets this PR to ``main``.

## Test plan
- [x] 12 new wire-trace tests pass.
- [x] Full non-GPU suite (\`uv run pytest -m \"not gpu\"\`) regression-clean (58 passed in the pipecat+models slice; full suite green).
- [x] SPDX header check: \`OK: 306 file(s) carry a valid SPDX header\`.
- [ ] GPU smoke skipped (no GPU available in sandbox).

### DIFF
--- a/DEPENDENCIES.md
+++ b/DEPENDENCIES.md
@@ -46,6 +46,7 @@ xr-ai-agent  (agent-sdk/)
 xr-ai-pipecat  (agent-sdk/xr-ai-pipecat/)
     └── xr-ai-agent   [editable: ..]
     └── xr-ai-logging [editable: ../../utils/xr-ai-logging]
+    └── xr-ai-models  [editable: ../xr-ai-models]
     └── pipecat-ai >=0.0.46
     └── numpy >=1.24
     └── scipy >=1.11
@@ -54,6 +55,9 @@ xr-ai-pipecat  (agent-sdk/xr-ai-pipecat/)
     Optional Pipecat transport bridge: connects ProcessorEndpoint (ZMQ IPC)
     to a Pipecat frame pipeline. Resamples hub float32 audio → 16 kHz int16
     for STT; converts TTS int16 PCM back to float32 AudioChunks for return.
+    SttClient / TtsClient are thin wrappers around xr-ai-models'
+    OpenAICompatSTT / OpenAICompatTTS — PCM→WAV conversion is handled by
+    the SDK. httpx is retained for http_probe() readiness checks.
     Not a dep of xr-ai-agent itself — import only in workers that use Pipecat.
 
 xr-ai-models  (agent-sdk/xr-ai-models/)
@@ -150,6 +154,7 @@ oxr-mcp-server  (agent-mcp-servers/oxr-mcp/)
 xr-ai-tests  (tests/)
     └── xr-ai-agent             [editable: ../agent-sdk]
     └── xr-ai-models            [editable: ../agent-sdk/xr-ai-models]
+    └── xr-ai-pipecat           [editable: ../agent-sdk/xr-ai-pipecat]
     └── xr-media-hub            [editable: ../server-runtime]    (pulls in livekit, livekit-api for the wss /rtc proxy + room-client tests)
     └── xr-ai-launcher          [editable: ../utils/xr-ai-launcher]
     └── xr-ai-logging           [editable: ../utils/xr-ai-logging]
--- a/agent-sdk/xr-ai-models/xr_ai_models/openai_compat.py
+++ b/agent-sdk/xr-ai-models/xr_ai_models/openai_compat.py
@@ -505,7 +505,7 @@ class OpenAICompatSTT:
         channels: int = 1,
         timeout: float | None = None,
     ) -> str:
-        wav_bytes = _pcm_to_wav(audio, sample_rate, channels) if sample_rate else audio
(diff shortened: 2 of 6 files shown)


yanziz-nvidia (commented):

## Summary

Rewires `xr_ai_pipecat.SttClient` and `TtsClient` to be thin wrappers around `OpenAICompatSTT` / `OpenAICompatTTS` from the new `xr-ai-models` SDK (introduced in #135, now merged). Public constructor signatures and `transcribe` / `synthesize` method signatures are preserved; PCM→WAV conversion moves into the SDK. AGENTS.md, DEPENDENCIES.md, and `docs/changelog.md` are updated in-commit per convention. CI green across all checks (pytest 3.11+3.12, CodeQL, SPDX, lock).

|     | Finding |
|-----|---------|
| 🚫 | None |
| 💡 | [`agent-sdk/xr-ai-pipecat/xr_ai_pipecat/services.py:29`](https://github.com/NVIDIA/xr-ai/blob/d3838a91e98198e9a5ef09dfee15c1748f3e0e20/agent-sdk/xr-ai-pipecat/xr_ai_pipecat/services.py#L29) — `SttClient` / `TtsClient` no longer expose `health_url` as a public attribute. The `xr-render-demo` worker never accessed it on these classes (confirmed), but the attribute existed on the pre-migration versions and was part of the informal interface. If the next migration unit (simple-vlm-example) ever switches from its own local `services.py` to importing from `xr_ai_pipecat`, that worker's `stt.health_url` / `tts.health_url` calls (lines 74–76 of `simple_vlm_example_worker.py`) would break silently. Delegating to `self._stt.health_url` / `self._tts.health_url` (which already exist on `OpenAICompatSTT` / `OpenAICompatTTS`) would preserve the attribute at zero cost. |
| 💡 | [`tests/test_pipecat_services_wire.py:62`](https://github.com/NVIDIA/xr-ai/blob/d3838a91e98198e9a5ef09dfee15c1748f3e0e20/tests/test_pipecat_services_wire.py#L62-L70) — `_wire_stt_client` / `_wire_tts_client` reach into `__new__` + private `_stt` / `_tts` to inject the stub transport. A cleaner alternative would be a factory helper that threads `client=` through the `SttClient` constructor — saves coupling the tests to the private field name across future refactors. |
| 🔍 | [`agent-sdk/xr-ai-models/xr_ai_models/openai_compat.py:1175`](https://github.com/NVIDIA/xr-ai/blob/d3838a91e98198e9a5ef09dfee15c1748f3e0e20/agent-sdk/xr-ai-models/xr_ai_models/openai_compat.py#L1175) — `OpenAICompatSTT.transcribe` uses `if sample_rate:` to gate PCM→WAV conversion, but `sample_rate: int | None = None` means a caller passing `sample_rate=0` would silently forward raw PCM. `if sample_rate is not None` is the standard pattern for optional numeric params. Not a real-world risk (0 Hz PCM is nonsense). |

- In `agent-sdk/xr-ai-pipecat/xr_ai_pipecat/services.py`, add `health_url` property to `SttClient` and `TtsClient` delegating to `self._stt.health_url` / `self._tts.health_url` to preserve the pre-migration public surface.
- In `agent-sdk/xr-ai-models/xr_ai_models/openai_compat.py` line 1175, change `if sample_rate:` to `if sample_rate is not None:` to guard against the zero-value edge case.
- In `tests/test_pipecat_services_wire.py`, consider replacing the `__new__` + private-field injection with a `client=`-accepting constructor overload or test-only factory, to avoid coupling tests to internal `_stt` / `_tts` field names.

yanziz-nvidia (approved):

nvddr:
All three items from the review addressed in 89163e1:

- **💡 `health_url` property**: added `@property health_url` to `SttClient` and `TtsClient` in `services.py`, delegating to `self._stt.health_url` / `self._tts.health_url`.
- **💡 Test injection via `client=` kwarg**: extended both wrapper constructors with an optional `client: httpx.AsyncClient | None = None` kwarg forwarded to the underlying SDK client; updated `_wire_stt_client` / `_wire_tts_client` in the test file to use that public path — `__new__` + private-field injection is gone.
- **🔍 `if sample_rate is not None:`**: changed the `if sample_rate:` guard in `OpenAICompatSTT.transcribe` to `if sample_rate is not None:` as suggested.

Branch rebased onto current `main` (the pre-squash SDK commits are gone), 236 non-GPU tests pass, SPDX clean.

PR #145, opened 2026-05-19
Add offline eval harness for xr-render-demo agent

blongs-nv:
This adds eval harness to the xr demo prompts along with a watcher that allows an agent to iterate on prompt changes.


yanziz-nvidia (commented):

## Summary

Adds an offline eval harness (`eval/eval.py`, `eval/eval_watch.sh`, `eval/README.md`) to `agent-samples/xr-render-demo/` for scoring the agent LLM's tool-calling against ~60 synthetic cases, plus a new AGENTS.md rule requiring eval cases whenever a prompt rule changes. CI is all green; no `pyproject.toml` changes so `DEPENDENCIES.md` needs no update.

|     | Finding |
|-----|---------|
| 🚫 | None |
| 💡 | [`agent-samples/xr-render-demo/eval/eval.py:812`](https://github.com/NVIDIA/xr-ai/blob/70de9b3acb8783e6e03d747064371146a8ea59e6/agent-samples/xr-render-demo/eval/eval.py#L812) — `shape_change_sphere_to_cube` comment says "a cube exists at the end" but the `result: []` path in `_check()` only verifies ≥1 mutating call; a bare `remove_primitive(sphere-0)` passes with no cube in the scene. Same pattern applies to `shape_and_color_change`. Either add a predicate that checks a `box`-typed primitive is present after execution, or update the comment to match what's actually verified. |
| 💡 | [`docs/xr-render-demo.md`](https://github.com/NVIDIA/xr-ai/blob/70de9b3acb8783e6e03d747064371146a8ea59e6/docs/xr-render-demo.md) — AGENTS.md documentation rule: new entry points in an existing sample should be reflected in the sub-repo docs in the same commit. `docs/xr-render-demo.md` is the canonical developer reference for this sample and has no mention of `eval/eval.py` / `eval_watch.sh`. A short "Eval harness" section pointing at `eval/README.md` would satisfy the rule. |
| 🔍 | [`agent-samples/xr-render-demo/eval/eval_watch.sh:67-76`](https://github.com/NVIDIA/xr-ai/blob/70de9b3acb8783e6e03d747064371146a8ea59e6/agent-samples/xr-render-demo/eval/eval_watch.sh#L67-L76) — `/proc`-based PID detection is Linux-only; `eval/README.md` doesn't call this out. Since the whole stack is Linux-bound (CUDA, LOVR, CloudXR) it's not a correctness gap, but a one-line note would prevent confusion on macOS dev boxes. |

- In `eval.py`, add a predicate for `shape_change_sphere_to_cube` (and `shape_and_color_change`) that verifies at least one `add_primitive(prim_type=box)` or `update_primitive(prim_type=box)` appears in the mutations, so the test matches the comment's stated intent.
- Add a short "Eval harness" section to `docs/xr-render-demo.md` (e.g. "Run `agent-samples/xr-render-demo/eval/eval.py` — see `eval/README.md`.") to satisfy the AGENTS.md documentation rule.
- Add a line to `eval/README.md` noting that `eval_watch.sh` requires Linux (`/proc` filesystem) and will not function on macOS.

yanziz-nvidia (approved):

nvddr:
Thanks for landing the eval harness — case taxonomy is broad (single-turn, multi-turn `history`, undo `recent_moves`, pose-override, gravity-aligned), range-based arg matching is the right call against LLM float jitter, and `_check_prompt_eval_overlap` is the kind of guard rail most eval harnesses skip. Watcher's sha1-not-mtime + `setsid` + exact-argv PID match all show thought.

A handful of follow-ups, ranked by severity. One blocker; the rest are suggestions you can land in this PR or a fast follow-up.

### 🚫 Blocker

- **`docs/xr-render-demo.md` needs an `## Eval harness` section.** This PR adds `eval.py` and `eval_watch.sh` as new entry points for the sample, and the AGENTS.md rule the PR *itself adds* says new entry points in an existing sample must be reflected in the sub-repo docs in the same commit. `eval/README.md` is great as the deep dive, but `docs/xr-render-demo.md` is where developers land first — a short section pointing at the README would satisfy the rule the PR is creating. (Echoing bot finding #2.)

### 💡 Suggestions

- **Tighten `result: []` semantics on shape-change cases.** `shape_change_sphere_to_cube` and `shape_and_color_change` use `result: []`, which the harness's `_check` interprets as "≥1 mutating call passed." A bare `remove_primitive(sphere-0)` clears the bar but leaves no cube in the scene — comment says one exists, the check doesn't verify it. `_check` already supports `predicate` callbacks (line 1709); a one-liner `predicate=lambda muts: any(_post_state_has(muts, type="box"))` closes the gap. (Echoing bot finding #1.)

- **Split `eval.py` (1879 LOC).** The file already has clean section banners — `CASES` data, `_exec_tool` + scene/pose mutators, `_check` + `_match_call`, `_run_one` + tool discovery, CLI/orchestration. Splitting these into `cases.py` / `exec.py` / `check.py` / `runner.py` / `eval.py` is mechanical (no logic change) and reviewers eat the 1879-LOC review once instead of every case-add reviewer paying it again. Strong suggestion.

- **Expand PR body.** One sentence for +2117 lines makes reviewers reconstruct intent from the diff. A few extra paragraphs covering: motivation (why an offline eval at all), case taxonomy (the variants listed above), how to extend, what's deliberately NOT covered (live worker pipeline, real render-mcp effects, real visual queries) — most of this is already in `eval/README.md`; lifting it into the body would help.

- **Add `set -eo pipefail` to `eval_watch.sh`.** Today `set -u` only catches unset vars; `sha1sum` failing transiently (file briefly absent during editor atomic-save, FS hiccup) is swallowed by `2>/dev/null` and the pipe to `awk`. Empty hash → spurious "change detected" → wasted restart. Self-healing but burns work. Also worth bailing out of the iteration explicitly when `file_hash` returns empty.

- **`--strict-overlap` flag for `_check_prompt_eval_overlap`.** Today the audit warns and continues, which is right for interactive iteration but means a future `system.txt` edit that introduces a worked-example overlap can silently degrade the eval. A `--strict-overlap` flag (default on in any CI-ish invocation, default off interactively) hard-fails on overlap and protects the contract long-term.

- **Resolve the `_max_steps_for` dead seam.** `_max_steps_for(case)` always returns `_MAX_STEPS=10`. Either parameterize per-case (some of the multi-turn / distractor cases genuinely benefit from more turns — `between_red_and_blue_with_distractors` can take 3–4 turns of scene reasoning) or delete the wrapper. Dead seams rot quietly.

- **Hardcoded relative paths.** `_HERE / "../worker/prompts/system.txt"` and `"../yaml/xr_render_demo_worker.yaml"` work today but break if `eval/` ever moves. A `--worker-root` arg (falling back to `git rev-parse --show-toplevel`-relative) is one line and future-proofs.

- **Surface tool-discovery degradation.** README correctly notes `vlm-mcp` / `video-mcp` may be down and the harness fails open — but the code is silent about it. A `logger.warning(...)` at startup when those tool sets are missing would prevent the failure mode where someone iterating prompts spends an evening blaming their wording for a tool the LLM no longer sees.

### 🔍 Nit

- **Document watcher's Linux-only PID detection.** `eval_watch.sh:31-37` uses `/proc/$pid/cmdline` which is Linux-only. Stack is Linux-bound (CUDA, LOVR, CloudXR) so not a correctness gap, but a one-line note in `eval/README.md` would prevent confusion on a macOS dev box. (Echoing bot finding #3.)

### Recap

Bot already 🟢 approved this; my notes are mostly hygiene + the one AGENTS.md self-violation. Happy to discuss any of these — particularly the eval.py split, since that one's the most invasive.

PR #146, opened 2026-05-20
fix(xr-media-hub): release held ring slots on participant leave (#143)

### LINKED ISSUE #143: Hub ring-buffer slot leak: new participants' video frames dropped 100% after enough tracks have ended
# Hub ring-buffer slot leak: new participants' video frames dropped 100% after enough tracks have ended

## Summary

`xr_media_hub/ipc/_hub.py` holds the latest SHM ring slot per `(participant_id, track_id)` and only releases it when the next frame for the same key arrives. When a track **ends**, its last slot is never released. The ring has 10 slots by default (`_DEFAULT_NUM_SLOTS = 10`). After 10+ unique track lifetimes the ring is permanently full and every subsequent frame from any participant is dropped with:

```
WARNING xr_media_hub.transport.livekit._room_client:_stream_video:374
  Ring buffer full — dropped frame from '<pid>'/'<track_id>'
```

The worker process then reports `tracks_seen=0` for the affected participant and the VLM responds with `"camera unavailable, please try again"` because `_wait_for_camera_frame` times out.

## Reproduction

1. Start `agent-samples/simple-vlm-example`.
2. Connect from any client, publish video, ask a vision question. Confirm it works.
3. Disconnect. Reconnect. Publish video again. Repeat 10+ times (each connect → publish → disconnect cycle creates a new `(pid, track_id)`). Trackable in the hub log via `"Video stream started"` / `"Video stream ended"` lines.
4. Connect a fresh participant (different identity), publish video, ask a vision question.
5. **Expected:** video frames flow to the worker, VLM answers.
6. **Actual:** every frame is dropped at the hub. Worker log shows `tracks_seen=0`. Agent replies `"camera unavailable, please try again"`.

## Diagnostic data

Real-world session: hub running for several hours, multiple `ios-client` and `web-client` test sessions, then a new participant `innoactive-spatial` joined.

```
$ grep "Ring buffer full" hub.log | awk -F\' '{print $2}' | sort | uniq -c
   3539 ios-client            # earlier tests
  38140 innoactive-spatial    # new participant — 100% drop rate

$ grep "first frame signal" worker.log | awk -F\' '{print $2}' | sort | uniq -c
      8 ios-client
      2 web-client
                              # zero for innoactive-spatial
```

The new participant produced 38,140 dropped frames and **zero** "first frame signal" deliveries to the worker. Earlier participants delivered first frames in 1 ms (`age_ms=1` in worker log).

## Root cause

`xr_media_hub/ipc/_hub.py`:

```python
elif type_id == MsgType.FRAME_SIGNAL:
    ...
    key = (msg.participant_id, msg.track_id)
    prev = self._latest_slots.pop(key, None)
    if prev:
        prev[0].release_slot(prev[1].signal.slot)
    view = ring.read_slot(msg)
    self._latest_slots[key] = (ring, view)
```

The slot is only released when a new frame for the same `(pid, track_id)` arrives. When a track ends (LiveKit `track_unsubscribed` or `participant_disconnected`), no further frames come, so `_latest_slots[(pid, track_id)]` holds the slot forever.

`xr_media_hub/transport/livekit/_room_client.py` handles `track_unsubscribed`:

```python
@self._room.on("track_unsubscribed")
def _on_track_end(track, _pub, _participant) -> None:
    self._cancel_track_task(track.sid)
```

— but no signal flows from here to the hub to release the corresponding slot.

## Suggested fix

When a track ends or a participant disconnects, release the held slot. Two options:

**Option A — release at the room_client (cleanest):**

Add a new `MsgType.TRACK_END` (or piggyback on `PARTICIPANT_EVENT` with `joined=False`) and have the hub release every slot keyed by that participant.

**Option B — bound the leak with a TTL:**

In the hub's `_dispatch` FRAME_SIGNAL handler, periodically scan `_latest_slots` and release entries older than N seconds. Less precise but resilient to other event sources.

Either fix should also bump `_DEFAULT_NUM_SLOTS` slightly (e.g. 16) since a single ill-timed reconnect can otherwise still hit the ceiling for a brief window.

## Workaround for operators

Restart `simple-vlm-example` (or just the hub). The new process has empty `_latest_slots`. The leak then re-accumulates over time.

## Environment

- `xr-ai` HEAD as of 2026-05-17
- Hub running on Ubuntu 24.04 / DGX Spark
- Clients: `web-client`, `ios-client` (`StreamKitSample`), and an Innoactive Spatial integration based on the same `StreamKit` package.

## References

- `server-runtime/xr_media_hub/ipc/_hub.py` — `_dispatch` FRAME_SIGNAL branch + `_latest_slots`
- `server-runtime/xr_media_hub/ipc/_connector.py` — `_DEFAULT_NUM_SLOTS = 10`
- `server-runtime/xr_media_hub/transport/livekit/_room_client.py` — `track_unsubscribed` handler

nvddr:
## Summary

Fixes #143 — the hub-side ring-buffer slot leak that drops 100% of frames from new participants after enough tracks have ended.

`HubEndpoint` holds the latest SHM ring slot per `(participant_id, track_id)` so processors can fetch pixels on demand without an eager copy. The slot was only released when the *next* `FRAME_SIGNAL` for the same key arrived — when a track ended, its slot was never released. After ~`_DEFAULT_NUM_SLOTS` connect/publish/disconnect cycles, the connector's ring filled and every subsequent frame was dropped with `Ring buffer full — dropped frame`.

## Fix

- **`server-runtime/xr_media_hub/ipc/_hub.py`** — On `PARTICIPANT_EVENT(joined=False)`, release every slot keyed by that participant_id. `notify_participant_left` already fires on both LiveKit `track_unsubscribed` and `participant_disconnected` (called from `_room_client._handle_left`), so this single hook covers the issue's repro without adding a new message type. Reuses the established "release_slot without `view.data.release()`" pattern from the `FRAME_SIGNAL` branch — the connector's ring is still live, so no explicit memoryview release is needed.
- **`server-runtime/xr_media_hub/ipc/_connector.py`** — Bump `_DEFAULT_NUM_SLOTS` 10 → 16 as defence-in-depth against the brief window between last frame and disconnect event. Adds ~75 MiB to worst-case 4K-NV12 shm footprint per connector.
- **`tests/test_participant_events.py`** — New `test_participant_leave_releases_held_slots` runs 6 cycles of join → push_frame → leave against the standard `num_slots=4` connector fixture (>num_slots so the leak would saturate without the fix), then asserts `hub._latest_slots == {}` and that a fresh participant can publish frames after the churn.
- **`docs/changelog.md`** — Decision entry documenting the fix and the explicit out-of-scope item (connector crash/OOM still leaks; no `joined=False` is emitted on unclean exit).

## Out of scope

A connector that crashes (OOM, SIGKILL) never emits `notify_participant_left`, so its slots still leak. A heartbeat or TTL scan would close that gap but is deferred until there's evidence it matters in practice. The issue's reproduction is fully covered by participant disconnect.

No `MsgType.TRACK_END` (Option A's first variant) was added — the same-pid track-end-without-disconnect case isn't in the repro and adding a new message type for a hypothetical would be scope creep.

## Test plan

- [ ] `tests` workflow passes (CI runs `pytest` from `tests/` on `ubuntu-latest`; the new test is laptop-friendly IPC and doesn't need GPU/Docker/NVENC)
- [ ] Manual: reproduce the issue's steps against `agent-samples/simple-vlm-example` — connect/publish/disconnect 16+ times, then a fresh participant joins and asks a vision question; verify frames flow and the VLM answers (instead of `"camera unavailable, please try again"`)
- [ ] Confirm `hub.log` no longer shows runaway `Ring buffer full` counts attributed to a single fresh participant

Note: this PR was prepared in an environment without `uv` / Python 3.11+, so the new pytest case has not been executed locally. The patch is small (one branch in `_dispatch`, a constant, and a focused regression test); CI is the authoritative signal.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -9,6 +9,40 @@ Significant decisions, in reverse-chronological order. Update this whenever a
 non-trivial architectural or design decision is made so the rationale is
 preserved and not re-litigated.
 
+### 2026-05-20 — Hub releases held ring-buffer slots on participant leave (#143)
+
+`HubEndpoint` holds the latest SHM ring slot per `(participant_id,
+track_id)` so processors can fetch pixels on demand without an eager
+copy. The slot was only released when the *next* FRAME_SIGNAL for the
+same key arrived — so when a track ended (LiveKit `track_unsubscribed`
+or `participant_disconnected`), its last slot stayed held forever.
+After enough connect/publish/disconnect cycles the ring filled with
+abandoned slots and every subsequent frame from any participant was
+dropped at the connector with `Ring buffer full — dropped frame`.
+A new participant could publish video and the worker would log
+`tracks_seen=0` until the hub was restarted.
+
+**Fix.** The hub now releases every slot keyed by a participant when
+that participant's `PARTICIPANT_EVENT(joined=False)` arrives. The
+`notify_participant_left` path already fires on both `track_unsubscribed`
++ `participant_disconnected` (it is called from `_room_client._handle_left`),
+so no new message type is required. Reuses the established
+"release_slot without `view.data.release()`" pattern from the FRAME_SIGNAL
+branch — the connector's ring is still live at this point, so the
+memoryview does not need an explicit release.
+
+Also bumped `_DEFAULT_NUM_SLOTS` from 10 → 16 so a single ill-timed
+reconnect within the in-flight window between last frame and disconnect
+event can't still hit the ceiling. The slot is 12.4 MiB at the 4K NV12
+ceiling, so six extra slots adds ~75 MiB to the worst-case per-connector
+shm footprint — cheap insurance.
+
+**Out of scope.** Connector crash / OOM still leaks slots — `joined=False`
+is only emitted by the live `notify_participant_left` path, not by an
+unclean exit. A heartbeat or TTL scan would close that gap; deferred
+until there's evidence it matters in practice. The issue's reproduction
+is wholly covered by participant disconnect.
+
 ### 2026-05-18 — `nightly XR AI test` workflow on self-hosted `gpu` runner
 
 The `gpu`-marked pytest suite (`tests/test_gpu_*.py`,
--- a/server-runtime/xr_media_hub/ipc/_connector.py
+++ b/server-runtime/xr_media_hub/ipc/_connector.py
@@ -37,7 +37,7 @@ ReturnAudioCallback      = Callable[[AudioChunk],        Awaitable[None]]
 ReturnDataCallback       = Callable[[DataMessage],       Awaitable[None]]
 ReturnAudioFlushCallback = Callable[[ReturnAudioFlush],  Awaitable[None]]
 
-_DEFAULT_NUM_SLOTS       = 10
+_DEFAULT_NUM_SLOTS       = 16
 _DEFAULT_MAX_FRAME_BYTES = 12_441_600  # 4K NV12
 
 
@@ -81,7 +81,7 @@ class ConnectorEndpoint:
         sub_addr        : Hub's PUB address  — connector subscribes for return traffic.
         connector_id    : Unique ID for this connector. Defaults to a UUID.
         shm_name        : Shared-memory segment name. Defaults to xr_conn_<connector_id>.
-        num_slots       : Ring buffer slot count (default 10).
+        num_slots       : Ring buffer slot count (default 16).
         max_frame_bytes : Max bytes per slot (default 4K NV12 = 12 441 600).
         """
         self._connector_id = connector_id or uuid.uuid4().hex
--- a/server-runtime/xr_media_hub/ipc/_hub.py
+++ b/server-runtime/xr_media_hub/ipc/_hub.py
@@ -102,7 +102,9 @@ class HubEndpoint:
         self._participant_connector: dict[str, str] = {}
         # (participant_id, track_id) → (ring, SlotView) of the latest frame.
         # The slot is held open (not released) until the next frame for the same
-        # track arrives, so pixels can be copied on demand without eager allocation.
+        # track arrives, the participant disconnects, or the hub shuts down — so
+        # pixels can be copied on demand without eager allocation while still
+        # bounding ring occupancy across participant churn.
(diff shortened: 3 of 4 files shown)


yanziz-nvidia (commented):

## Summary

Fixes the ring-buffer slot leak (#143) by releasing all `_latest_slots` entries for a participant when their `PARTICIPANT_EVENT(joined=False)` arrives, and bumps `_DEFAULT_NUM_SLOTS` 10→16 as defence-in-depth. Changelog entry added. No `pyproject.toml` touched; DEPENDENCIES.md is unaffected. CI is fully green (pytest 3.11 + 3.12, CodeQL, SPDX, uv lock all pass).

|     | Finding |
|-----|---------|
| 🚫 | None |
| 💡 | [`server-runtime/xr_media_hub/ipc/_hub.py:283-284`](https://github.com/NVIDIA/xr-ai/blob/5301417d15fd3e21abe174e7086a245647e576fc/server-runtime/xr_media_hub/ipc/_hub.py#L283-L284) — Disconnect-path `release_slot` skips `view.data.release()`, unlike `close()` which calls it with the comment "must release before ring.close() so mmap has no exported pointers." The connector ring stays alive across a disconnect, so the omission is consistent with `close()`'s comment — but the asymmetry is undocumented and easy to misread as a bug. A one-line inline comment explaining why `view.data.release()` is not needed here would match the precedent. |
| 🔍 | [`tests/test_participant_events.py:106-108`](https://github.com/NVIDIA/xr-ai/blob/5301417d15fd3e21abe174e7086a245647e576fc/tests/test_participant_events.py#L106-L108) — The final `assert ("fresh", "cam") in hub._latest_slots` verifies the key is present, not that a fresh frame can be read through it. Also leaves the ring holding one slot for `"fresh"` after the test — fixture teardown releases it rather than the test itself. Slightly inconsistent with the empty-check at line 101. |

- In `server-runtime/xr_media_hub/ipc/_hub.py` around line 283, add a short comment explaining why `view.data.release()` is intentionally omitted on the disconnect path — e.g. `# ring is still live; memoryview export does not block the mmap` — matching the precedent set in `close()`.

yanziz-nvidia (approved):

PR #131, opened 2026-05-12
feat(client-sdk): native LiveKit backend + zero-copy frame overload

gongxinheng:
## Summary

- Replaces the stub `LiveKitBackend.cpp` in `client-samples/native/StreamKit/` with a working implementation against the upstream LiveKit C++ SDK (`livekit::Room`, `LocalParticipant`, `AudioSource`, `VideoSource`).
- Adds a zero-copy `FrameSink::InjectVideoFrame(std::vector<std::uint8_t>&&, …)` overload. Hardware-validated 14× per-frame speedup on an embedded 32-bit ARMv7-A target (~70 ms → ~5 ms) — the span overload's forced alloc + memcpy collapsed an effective 720p30 camera path to ~6 fps; the move overload restores the in-house baseline.
- Includes 5 unit tests behind `STREAMKIT_BUILD_TESTS=ON` (off by default) plus a `STREAMKIT_FEEDBACK.md` capturing 12 cross-SDK API findings from the integration.

## Highlights

- **The move overload is strictly additive.** Existing `FrameSink` implementations keep working unchanged — the default impl forwards to the span overload, so backends that don't override pay the same cost as before. `LiveKitBackend` overrides both; the span entrypoint copies once and delegates to the move overload.
- **Native C++ is structurally the only path that hits this** — iOS / Android / Web use platform-managed capture (`CMSampleBuffer` / `ImageReader` / `MediaStream`) that's already reference-counted or shareable. Every native C++ caller benefits: CloudXR, native game-engine plugins, embedded camera SDKs. Embedded is just where the cost crossed from "wasteful" to "unusable" and made it measurable.
- **State-change dedupe.** `FireStateChanged` ensures no spurious initial `kDisconnected` on first `Connect()` and no doubled `kConnected` after a successful blocking `Room::Connect`. Regression test included.

## Test plan

- [x] Stub-mode build: `cmake -S . -B build && cmake --build build` (no `LIVEKIT_SDK_ROOT`) — lib + sample app build clean
- [x] Real-mode build: `cmake -S . -B build -DLIVEKIT_SDK_ROOT=/path/to/sdk -DLIVEKIT_LIB_DIR=...` + `cmake --build build --target streamkit` — static lib builds clean against an ABI-matching SDK install (sample-app link skipped on a host where the SDK binaries are cross-ABI)
- [x] Unit tests: `cmake -S . -B build -DSTREAMKIT_BUILD_TESTS=ON && cmake --build build && ctest --test-dir build` → 5/5 passing
[...]

PR #147, opened 2026-05-20
feat(client-sdk): AudioSink mixin + CameraConfig contract (#134)

### LINKED ISSUE #134: Native C++ StreamKit — integration findings from a partner port
Cross-posting from PR #131 — keeps the PR scoped to the API change and lets these findings be tracked independently.

> **Updated 2026-05-13.** Two items below want your design input; the rest are follow-up tasks I'll own as the native client maintainer and ship as separate small PRs. Listed here for visibility, no decision needed on those.

---

## Design questions for NVIDIA

The two items below sit on your side because they affect the cross-platform API surface — a new public interface and a docstring carve-out that the spec should be explicit about. Worth converging on a direction before anyone implements.

### A. Audio: no `AudioSink` companion to `FrameSink`

`FrameSink::InjectVideoFrame` lets a host push externally-captured video frames into the SDK — that's how an embedded camera HAL feeds LiveKit in the integration that motivated PR #131. There's no parallel mechanism for audio. Embedded hosts that capture mic frames from a platform HAL have to either subclass `LiveKitBackend` to reach the private `audio_source_` member, or invent something out of band. The subclass-and-reach-into-privates pattern is fragile and couples backends to implementation details.

**Suggested fix:** add a sibling `AudioSink` interface with `InjectAudioFrame(pcm_bytes, sample_rate, channels, samples_per_channel, timestamp_us)`. `LiveKitBackend` implements it analogously to `FrameSink`. iOS already has an `AudioSink` equivalent (`AudioInjectable`) so this mirrors an existing platform.

Probably the most actionable item — clean new interface, no breaking changes, unblocks every native consumer that wants mic input.

**Decision you'd make:** is `AudioSink` (paired with `FrameSink`) the right shape, or would you prefer a unified `MediaSink` covering both audio and video? Either works; converging on the design before implementation starts.

### B. Camera: `CameraConfig::Facing` is inert in pure C++

`CameraConfig` exposes `Facing::kFront` / `Facing::kBack` and an optional `device_id`. On iOS / Android / Web, the SDK reads those and asks the OS to open the matching camera. **In pure C++ there's no portable camera-open API** — the C++ SDK can't act on `Facing` or `device_id`. So they're inert: the backend has to defer to the host, which captures externally and pushes frames via `FrameSink`.

This isn't a bug as much as an unstated API contract. A reader of `CameraConfig.h` would assume `Facing` works everywhere; in C++ it silently doesn't.

**Suggested fix:** drop `CameraConfig::Facing` / `device_id` from the C++ surface, or split `CameraConfig` so platform-specific fields only exist on backends that can act on them. Either way, the spec should be explicit that `Facing` only takes effect when the backend ships a built-in camera-open implementation.

**Decision you'd make:** should the C++ surface continue pretending it has a built-in camera (which would require shipping platform-specific open code for Linux / macOS / Windows — substantial work), or accept that C++ is structurally different and consumers always drive via `FrameSink`?

---

## Follow-up tasks I'll own (FYI)

Implementation work I'll send as separate small PRs — no decision needed on your side. Listed here so you have visibility into what's queued up.

### C. Audio: `AudioConfig::MicrophoneMode` unmapped in C++

`AudioConfig::MicrophoneMode` exposes `kVoiceProcessing`, `kSoftwareProcessing`, `kRaw`, `kDisabled` — the first two are documented as enabling AEC / AGC / noise suppression. The iOS / Android / Web backends honour them. The C++ SDK's `AudioSource` ctor takes `(sample_rate, num_channels, queue_size_ms)` with no DSP knobs — those filters live in `AudioProcessingModule` with a different lifecycle.

**Plan:** docstring update in `AudioConfig.h` noting that `MicrophoneMode` is platform-managed and that the built-in C++ `LiveKitBackend` does not currently apply AEC/AGC/NS. If we later need software DSP for a specific consumer, integrating `AudioProcessingModule` is a separate larger piece of work I'll scope at that point.

### D. Camera: docstring drift on first-frame lazy publish

`FrameSink.h` says "the first `InjectVideoFrame()` call creates and publishes the track because LiveKit requires at least one captured frame to resolve dimensions" — that matches WebRTC's `BufferCapturer` model on iOS / JS. The C++ SDK's `livekit::VideoSource` constructor instead requires `width` and `height` up front. PR #131 already implements lazy-publish on first frame; the docstring just hasn't caught up.

**Plan:** small docstring tweak in `FrameSink.h` clarifying that "first-frame triggers publish" is the contract every backend honours regardless of underlying SDK shape.

### E. `LiveKitConfig::token_url` HTTP fetch portability

The iOS / Swift / Kotlin backends fetch JWTs via platform HTTP clients. There's no portable C++ equivalent, and forcing libcurl on every consumer is heavy — especially embedded, where TLS root stores and DNS resolvers may not be present. PR #131 keeps `FetchToken` virtual and throws by default; the embedded path passes inline tokens via `LiveKitConfig::token` and doesn't need the URL fetch.

**Plan:** keep the virtual + subclass-and-override pattern. If a desktop consumer needs built-in HTTP later, ship that as an opt-in CMake option (`STREAMKIT_ENABLE_TOKEN_FETCH`) so the embedded path stays libcurl-free.

### F. SDK convention notes

Two small "C++ SDK has conventions StreamKit's API should document":

- `livekit::ConnectionState` only has `Disconnected / Connected / Reconnecting` — backends must synthesise `kConnecting` by firing the callback manually before calling `Room::Connect`. PR #131's `FireStateChanged` helper handles this and dedupes the duplicate-fire that would otherwise come from the explicit pre-fire plus the SDK's own state event.
- `livekit::initialize()` is a process-global one-shot that must be called before any other LiveKit SDK call. PR #131 calls it lazily on first backend ctor.

**Plan:** add one paragraph each to `StreamingBackend.h`'s docstring covering both — so a contributor implementing a non-LiveKit backend (mock, custom WebRTC, etc.) knows what's expected.

nvddr:
## Summary

Resolves the two design questions raised by the partner on issue #134
against PR #131's native C++ StreamKit integration.

- **`AudioSink` interface** (`streamkit/AudioSink.h`) mirroring `FrameSink`. Public `InjectAudioFrame(pcm, rate, channels, samples_per_channel, ts)` entry point. `LiveKitBackend` now inherits from it — embedded / native hosts no longer have to subclass to reach a private `audio_source_`.
- **`CameraConfig::facing` and `device_id`** stay on the cross-platform struct but are now documented in `CameraConfig.h` and `StreamingBackend.h` as inert on backends without a built-in camera-open path. The C++ backend expects frames via `FrameSink`; the host chooses the camera.
- Partner-owned follow-ups in issue #134 (items C / D / E / F) intentionally left to the partner per their stated plan.

## Decision rationale (full notes in `docs/changelog.md`)

- **`AudioSink` over `MediaSink`** — different validation rules, different real-time characteristics, different zero-copy stories between audio and video. Collapsing them produces a leaky abstraction at no readability gain. Also matches Swift's `AudioInjectable`.
- **No `&&` overload on `AudioSink`** — 10 ms @ 48 kHz mono PCM is ~960 bytes; the asymmetry FrameSink carries for 1.4 MB-per-frame buffers is not justified here.
- **Documented `CameraConfig::Facing` contract** rather than splitting the struct per-platform — would have forked the cross-platform shape every other client depends on.

## Test plan

- [x] Stub-mode build compiles clean (`LiveKitBackend.cpp` + `StreamSession.cpp` under `STREAMKIT_HAVE_LIVEKIT=0`)
- [x] All 5 pre-existing tests pass: `streamkit_mapping_tests`, `streamkit_agent_status_tests`, `streamkit_frame_sink_tests`, `streamkit_session_tests`, `streamkit_livekit_backend_tests`
- [x] New test `streamkit_audio_sink_tests` passes: covers parameter-verbatim delivery and `AudioSink&` dispatch
- [ ] Real-mode build against an ABI-matching LiveKit SDK install — relies on `livekit::AudioFrame(rate, channels, samples_per_channel, int16_t*)` and `AudioSource::captureFrame(AudioFrame, timestamp_us)`. Stub-mode CI cannot verify; the upstream SDK signatures match the existing `VideoFrame` pattern in `LiveKitBackend.cpp` so I expect no surprises, but a reviewer with an SDK install should sanity-link

Closes #134.


yanziz-nvidia (commented):
Approved by yanziz

## Summary

Adds the `AudioSink` mixin to the native C++ StreamKit, letting hosts push PCM frames via `InjectAudioFrame` without subclassing `LiveKitBackend`, and documents `CameraConfig::facing`/`device_id` as inert on backends without a built-in camera-open path. No `pyproject.toml` changes; `DEPENDENCIES.md` is unaffected. README and `docs/changelog.md` are updated in the same commit.

|     | Finding |
|-----|---------|
| 🚫 | [`AudioSink.h:47-48`](https://github.com/NVIDIA/xr-ai/blob/500a7953283c252b188b1937b518812101b758b1/client-samples/native/StreamKit/include/streamkit/AudioSink.h#L47-L48) — **`\param pcm` says "Length must equal `channels * samples_per_channel * 2` bytes" but `std::span<const std::int16_t>::size()` returns an element count, not a byte count.** Body text at [lines 57–59](https://github.com/NVIDIA/xr-ai/blob/500a7953283c252b188b1937b518812101b758b1/client-samples/native/StreamKit/include/streamkit/AudioSink.h#L57-L59) repeats the same error ("Backends validate the byte count"). The actual validation in [`LiveKitBackend.cpp:392-395`](https://github.com/NVIDIA/xr-ai/blob/500a7953283c252b188b1937b518812101b758b1/client-samples/native/StreamKit/src/Backends/LiveKit/LiveKitBackend.cpp#L392-L395) correctly checks `pcm.size() \!= channels * samples_per_channel` (elements). A caller following the doc literally would size their span to 2× the elements and get `invalid_argument`. |
| 💡 | [`LiveKitBackend.cpp:259-268`](https://github.com/NVIDIA/xr-ai/blob/500a7953283c252b188b1937b518812101b758b1/client-samples/native/StreamKit/src/Backends/LiveKit/LiveKitBackend.cpp#L259-L268) — `StopAudio()` `#else` branch (stub mode) has no body, so `audio_armed_` is never cleared there. `StartAudio()` sets it in both branches (lines 252 and 255), but `StopAudio()` only clears it under `#if STREAMKIT_HAVE_LIVEKIT`. In stub-mode builds, once armed the sink never disarms — the silent-drop contract documented in the header ("silently drops when audio capture is not armed…after `StopAudio()`") does not hold in the test binary. |
| 🔍 | [`AudioSink.h:14-17`](https://github.com/NVIDIA/xr-ai/blob/500a7953283c252b188b1937b518812101b758b1/client-samples/native/StreamKit/include/streamkit/AudioSink.h#L14-L17) — File-level block comment explains the "no `&&` overload" design rationale. Per AGENTS.md, architectural rationale belongs in `docs/changelog.md` (where it already appears), not source comments. |

- `client-samples/native/StreamKit/include/streamkit/AudioSink.h:47-48` — change `\param pcm` doc from "Length must equal `channels * samples_per_channel * 2` bytes" to "Length must equal `channels * samples_per_channel` elements". Also fix lines 57–59: "Backends validate the **element count** against `channels * samples_per_channel` and throw `std::invalid_argument` on mismatch." Both sites use "byte count" which contradicts `std::span::size()` semantics and the actual validation.
- `client-samples/native/StreamKit/src/Backends/LiveKit/LiveKitBackend.cpp:268` — in `StopAudio()`, add an `#else` branch that sets `audio_armed_.store(false);` so stub-mode teardown matches real-mode. Pattern: mirror what `StartAudio()` already does in its `#else` at line 255.
- `client-samples/native/StreamKit/include/streamkit/AudioSink.h:14-17` — trim or remove the "intentionally has a single pure-virtual entry point" rationale block; the changelog already has it.

yanziz-nvidia (approved):
[AudioSink.h](https://github.com/NVIDIA/xr-ai/blob/500a7953283c252b188b1937b518812101b758b1/client-samples/native/StreamKit/include/streamkit/AudioSink.h#L47-L48) please check.

nvddr:
Addressed the COMMENTED review in 825e778 (on top of a fresh rebase from main):

| Sev | Item | Status |
|---|---|---|
| 🚫 | `AudioSink.h:47-48` + `:57-59` — span size is element count, not bytes | **Fixed**. Doc now specifies element count (`pcm.size() == channels * samples_per_channel`) with the byte-length equivalent in a parenthetical for memory clarity. Body text at 57-59 updated to "validate `pcm.size() == channels * samples_per_channel`". |
| 💡 | `LiveKitBackend.cpp:259-268` — stub-mode `StopAudio` never clears `audio_armed_` | **Fixed**. Moved `audio_armed_.store(false)` outside the `#if STREAMKIT_HAVE_LIVEKIT` so it disarms in both real and stub builds, matching `StartAudio`'s symmetry. |
| 🔍 | `AudioSink.h:14-17` — file-level rationale block duplicates changelog | **Not applicable.** The actual content at lines 14-17 is `// Sibling of FrameSink. Mirror of Swift AudioInjectable.` + the `## Workflow` heading — contract context, not the "single pure-virtual entry point" / `&&` overload rationale the reviewer-bot cited. The rationale block it referenced isn't in the header (and would have been removed in an earlier pass if it ever was). If a stale comment block actually exists somewhere and I'm missing it, point me at it. |

**Final state on `825e778`**:
- `reviewDecision=APPROVED`
- `mergeable=MERGEABLE`, `mergeStateStatus=CLEAN`
- 9/9 CI checks COMPLETED+SUCCESS (CodeQL, pytest 3.11/3.12, uv lock, SPDX, Analyze py + js-ts ×2)

Ready to merge when you're ready.

PR #148, opened 2026-05-21
feat(mcp-servers): XR_RUN_DIR-based defaults for transcripts + frame queries

nvddr:
Both `transcript-mcp` and `video-mcp` defaulted to fixed `/tmp/...` paths
for their on-disk state (`transcripts_dir`, `out_dir`). Multi-sample dev
on one box collides on those paths and orchestrators that already create
a per-run directory (`XR_RUN_DIR`) have no way to point the MCP servers
into it without overriding config files.

Both servers now consult `XR_RUN_DIR` first:

- `transcript-mcp`: defaults to `$XR_RUN_DIR/transcripts` when set,
  otherwise `/tmp/xr_transcripts`. Explicit `transcripts_dir:` in config
  still wins.
- `video-mcp`: defaults `out_dir` to `$XR_RUN_DIR/frames` when set,
  otherwise `/tmp/xr_video_queries`. Explicit `out_dir:` in config still
  wins. (`recordings_dir` is sample-specific, unchanged.)

`transcript-mcp`'s `run()` is also refactored from a sync
`uvicorn.run(...)` with a startup-event ready-file callback to an
explicit `asyncio.run(_serve())`, matching the pattern in `video-mcp`
and the other async MCP servers — the ready-file touch now happens
right before `server.serve()` rather than via a side-effect callback.

Behavior preserved when `XR_RUN_DIR` is unset.

### DIFF
--- a/agent-mcp-servers/transcript-mcp/transcript_mcp_server/__main__.py
+++ b/agent-mcp-servers/transcript-mcp/transcript_mcp_server/__main__.py
@@ -54,7 +54,9 @@ Config (transcript_mcp_server.yaml)
 from __future__ import annotations
 
 import argparse
+import asyncio
 import json
+import os
 import pathlib
 
 import uvicorn
@@ -304,19 +306,26 @@ def run() -> None:
 
     setup_logging("transcript-mcp")
 
-    transcripts_dir = cfg.get("transcripts_dir", "/tmp/xr_transcripts")
+    _run_dir = os.environ.get("XR_RUN_DIR")
+    _default_transcripts = (
+        str(pathlib.Path(_run_dir) / "transcripts") if _run_dir else "/tmp/xr_transcripts"
+    )
+    transcripts_dir = cfg.get("transcripts_dir") or _default_transcripts
     host            = cfg.get("host", "0.0.0.0")
     port            = int(cfg.get("port", 8200))
 
     store = TranscriptStore(transcripts_dir)
     app   = build_app(store)
 
-    if ns.ready_file:
-        ready_path = ns.ready_file
-        app.add_event_handler("startup", lambda: ready_path.touch())
+    async def _serve() -> None:
+        config = uvicorn.Config(app, host=host, port=port, log_level="warning")
+        server = uvicorn.Server(config)
+        logger.info("transcript-mcp-server  mcp=/mcp  port={}  dir={}", port, transcripts_dir)
+        if ns.ready_file:
+            ns.ready_file.touch()
+        await server.serve()
 
-    logger.info("transcript-mcp-server  mcp=/mcp  port={}  dir={}", port, transcripts_dir)
-    uvicorn.run(app, host=host, port=port, log_level="warning")
+    asyncio.run(_serve())
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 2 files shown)


wenxind-nvidia (approved):

nvddr:

### Overview
Small, focused change: both `transcript-mcp` and `video-mcp` now consult `XR_RUN_DIR` for default on-disk paths, and `transcript-mcp`'s `run()` is brought in line with the `asyncio.run(_serve())` pattern used by `vlm-mcp`, `oxr-mcp`, and `video-mcp`. Scope, naming, and structure all look consistent with the rest of `agent-mcp-servers/`. Nice cleanup.

### Concerns Worth Discussing
- **[low] Ready-file timing semantics changed in `transcript-mcp`.** Previously the touch happened via `app.add_event_handler("startup", ...)`, which fires during ASGI lifespan startup — i.e. *after* uvicorn binds the listening socket. The new code touches the ready-file just before `await server.serve()`, so the file may exist before the port is actually accepting connections. This matches what `vlm-mcp`, `oxr-mcp`, and `video-mcp` already do (so it's a consistency win), and `docs/process-model.md` only requires "ready to serve requests" — but it is a strictly weaker guarantee than the old behavior. Worth confirming no downstream process race-connects to port 8200 the instant the ready-file appears.

### Nits
- `cfg.get("transcripts_dir") or _default_transcripts` (and the analogous `out_dir` line in `video-mcp`) is a *truthiness* fallback, not a *missing-key* fallback. The pre-PR code used `cfg.get("...", "/tmp/...")`, so an explicit `transcripts_dir: ""` or `transcripts_dir: null` in YAML would previously be honored as-is and now silently routes to the `XR_RUN_DIR`-derived default. Almost certainly an improvement, but it's a tiny behavior delta vs. the PR body's "explicit config still wins" claim — empty string is no longer considered explicit. Optional to address.
- The shipped default configs (`transcript_mcp_server.yaml`, `video_mcp_server.yaml`, and `agent-samples/xr-render-demo/yaml/video_mcp_server.yaml`) all still hard-code the `/tmp/...` paths, so orchestrators that set `XR_RUN_DIR` while passing one of these default YAMLs will see no redirection. By design per the PR body, but a one-line README/changelog note for orchestrator authors ("drop the key to opt into `XR_RUN_DIR`") would save a confused 30 minutes.

- `agent-mcp-servers/transcript-mcp/transcript_mcp_server/__main__.py:309-328` — the new `_default_transcripts` resolution, the `or`-fallback semantics, and the `_serve()` closure / ready-file ordering. This is the heart of the change.
- `agent-mcp-servers/video-mcp/video_mcp_server/__main__.py:649-660` — env-var resolution and the existing `out_dir.mkdir(parents=True, exist_ok=True)` which handles the `$XR_RUN_DIR/frames` subdir creation. Looks fine; just worth a sanity glance.

- Matches `AGENTS.md` "every process accepts `--ready-file` and must `Path(path).touch()` when ready to serve" — still satisfied.
- `transcript-mcp` now aligns with the `async def _serve()` + `asyncio.run(...)` pattern used by `vlm-mcp`, `oxr-mcp`, and `video-mcp`. One small structural diff vs. the siblings: `transcript-mcp` keeps `_serve` as a closure inside `run()` rather than a module-level `async def _serve(cfg, ready_file=None)`. Not wrong — just slightly less consistent. Up to the author whether to lift it.
- `docs/changelog.md` is the documented home for significant decisions; a short entry noting the new `XR_RUN_DIR` precedence rule would fit the "documentation rule" in `AGENTS.md`.

### Deferred to Experts
- None — this change is small and self-contained. yanziz-nvidia / blongs-nv reviews look sufficient.

---

PR #149, opened 2026-05-21
fix(stt-server): swallow NeMo errors on short audio; cleaner persistent idle

nvddr:
Three small fixes on the stt-server wrapper around NeMo Parakeet:

1. **Catch NeMo exceptions on very short / noisy audio.** `backend.transcribe`
   can raise on inputs that are too short or below the model's confidence
   threshold; that exception bubbled up to FastAPI and surfaced as a 500.
   The agent's STT path expects an empty transcription in those cases, not
   a hard error. Wrapped in try/except — log at WARNING, return `""`.

2. **`_idle_until_stopped` no longer needs an event loop.** The persistent-
   mode wait path used `loop.run_in_executor(None, _health_ok, ...)`, which
   conflated the asyncio loop and the polling concern. Rewrote it as a
   plain blocking function taking a health URL; called via
   `run_in_executor` from `_run` only when actually needed. Less
   coupling, easier to reuse from a wrapper subprocess.

3. **`cuda_visible_devices` config passthrough.** STT's GPU selection lived
   only in the GPU-init path. For multi-GPU hosts where the user wants the
   STT server on a specific device without polluting the wrapper's parent
   env, accept `cuda_visible_devices:` in the YAML and export it before
   any CUDA init.

Also adds an internal `--_serve` argparse flag (suppressed from help) to
support the subprocess-spawn pattern used by the persistent-mode reuse
path. Not user-visible.

### DIFF
--- a/ai-services/stt-server/stt_server/__main__.py
+++ b/ai-services/stt-server/stt_server/__main__.py
@@ -311,6 +311,9 @@ def run() -> None:
                 if r.status == 200:
                     break
         except Exception:
+            # Server still coming up — connection refused / timeout is the
+            # expected case for the first ~20s while NeMo loads. Retry
+            # silently; the for/else below handles the hard timeout.
             pass
     else:
         sys.exit("[stt_server] timed out waiting for server to become healthy")
--- a/ai-services/stt-server/stt_server/__main__.py
+++ b/ai-services/stt-server/stt_server/__main__.py
@@ -24,9 +24,12 @@ Config keys
 import argparse
 import asyncio
 import os
+import subprocess
 import sys
 import tempfile
 import threading
+import time
+import urllib.request
 import warnings
 from pathlib import Path
 
@@ -150,6 +153,12 @@ def _build_app(cfg: dict, model_cache: Path):
                     tmp.write(audio_bytes)
                     tmp_path = tmp.name
                 return backend.transcribe(tmp_path)
+            except Exception as exc:
+                # NeMo can throw on very short / noisy audio — return empty
+                # rather than letting the exception become a 500.
+                import logging as _log
+                _log.getLogger("stt_server").warning("transcribe failed: %s", exc)
+                return ""
             finally:
                 if tmp_path:
                     os.unlink(tmp_path)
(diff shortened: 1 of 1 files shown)


wenxind-nvidia (approved):

nvddr:

### Overview
Nice, focused hardening pass on the stt-server. Returning `""` on NeMo transcription failures matches the agent-side contract, and decoupling `_idle_until_stopped` from the running loop reads cleaner. One real bug worth fixing and a couple of smaller things below.

### Concerns Worth Discussing
- **`CUDA_VISIBLE_DEVICES` is now set twice.** The new walrus block at the top of `_run` (`if cuda_devices := cfg.get("cuda_visible_devices"):`) duplicates the pre-existing block ~3 lines below (`cuda_vis = cfg.get("cuda_visible_devices")` … `os.environ["CUDA_VISIBLE_DEVICES"] = str(cuda_vis)`). Same key, same env var, same effective ordering — looks like the intent was to move the assignment earlier (before `_resolve_model_cache`) but the original wasn't deleted. Suggest dropping one of the two blocks.
- **Scope vs. PR description.** The body frames `--_serve` as a small "internal argparse flag (not user-visible)", but the diff actually adds ~50 lines of `subprocess.Popen` + health-poll + reuse logic in `run()` — that's a wrapper-mode refactor, not a flag. A sentence in the description covering the spawn/reuse split would help future readers (and reviewers) understand why `run()` grew.

- Are there any plans for a small test (even just a unit test that mocks `backend.transcribe` to raise and asserts the endpoint returns `{"text": ""}` instead of 500)? The repo has a `tests/` tree but nothing was added here.
- The new `except Exception` in `_run`/transcribe silently turns *every* NeMo failure into an empty transcription — short audio, OOM, weight corruption, etc. all look identical to the caller. Was the intent to swallow only the short-audio class of error, or genuinely everything? If everything, a one-line note in the warning log (input size, exception type) would make post-hoc debugging less painful.

- `ai-services/stt-server/stt_server/__main__.py` around the two `CUDA_VISIBLE_DEVICES` writes (≈ lines 195-203) — confirm only one should remain and which ordering is intended relative to `_resolve_model_cache`.
- The startup polling loop in `run()` (≈ lines 309-314, `except Exception: pass`) — this is what `github-code-quality` flagged as an empty-except. Worth deciding whether to adopt the suggested `continue` + explanatory comment or leave it; either way, resolving that review thread is likely what's keeping `mergeStateStatus=BLOCKED`.
- The new `subprocess.Popen(..., start_new_session=True)` spawn path — worth a quick eyeball on lifecycle: if the wrapper's `_idle_until_stopped` exits because the child stops responding to `/health`, is anything responsible for reaping the child, or is it expected to have already exited on its own?

Aligns with the repo's `utils/xr-ai-logging` setup pattern and the stdlib-only launcher style called out in `AGENTS.md`. No `pyproject.toml` or dependency changes, so no `DEPENDENCIES.md` update needed. SPDX headers already present.

### Deferred to Experts
- Multi-GPU / CUDA selection ordering (the duplicated env-var write) is best confirmed by someone who's actually exercised the stt-server on a multi-GPU host — the existing GPU-init path may have subtle ordering assumptions that aren't obvious from the diff.

---

PR #150, opened 2026-05-21
fix(xr-ai-vllm): docker rm after stop so config changes apply on restart

nvddr:
`stop_persistent_servers` previously ran `docker stop <name>` and left
the container around so the next `start_xr_ai_vllm` could `docker start`
it back. That kept the container's baked-in argv — including stale
`--limit-mm-per-prompt`, `--max-images-per-prompt`, etc. — so YAML
config edits silently failed to take effect until the operator
manually removed the container.

Now: after a successful `docker stop`, run `docker rm` so the next
launch goes through the full `build_run_argv` → `docker run` path with
the latest YAML. Failure to remove is non-fatal (some hosts forbid
`docker rm` even when stop succeeds) — log the manual `docker rm`
command and continue.

User-visible change: the message changes from "stopped" to "stopped and
removed" on the happy path.

### DIFF
--- a/utils/xr-ai-vllm/xr_ai_vllm/__init__.py
+++ b/utils/xr-ai-vllm/xr_ai_vllm/__init__.py
@@ -177,7 +177,19 @@ def stop_persistent_servers(
         if container_name:
             print(f"  [{label}] stopping container {container_name}…", flush=True)
             if _docker.stop_container(container_name):
-                print(f"  [{label}] stopped", flush=True)
+                # Remove the container so a restart picks up any config changes.
+                # Without removal, `docker start` reuses the old container with
+                # its baked-in argv (including stale --limit-mm-per-prompt etc).
+                try:
+                    import subprocess as _sp
+                    _sp.run(["docker", "rm", container_name],
+                            check=True,
+                            stdout=_sp.DEVNULL, stderr=_sp.DEVNULL)
+                    print(f"  [{label}] stopped and removed", flush=True)
+                except Exception:
+                    print(f"  [{label}] stopped (rm failed — "
+                          f"run `docker rm {container_name}` to apply config changes)",
+                          flush=True)
             else:
                 print(f"  [{label}] docker stop failed — check `docker ps -a`",
                       flush=True)
--- a/utils/xr-ai-vllm/xr_ai_vllm/__init__.py
+++ b/utils/xr-ai-vllm/xr_ai_vllm/__init__.py
@@ -177,16 +177,21 @@ def stop_persistent_servers(
         if container_name:
             print(f"  [{label}] stopping container {container_name}…", flush=True)
             if _docker.stop_container(container_name):
-                # Remove the container so a restart picks up any config changes.
-                # Without removal, `docker start` reuses the old container with
-                # its baked-in argv (including stale --limit-mm-per-prompt etc).
-                try:
-                    import subprocess as _sp
-                    _sp.run(["docker", "rm", container_name],
-                            check=True,
-                            stdout=_sp.DEVNULL, stderr=_sp.DEVNULL)
+                # Remove the container so the next launch goes through a full
+                # `docker run` and picks up YAML config changes (without removal,
+                # `docker start` reuses the old container with its baked-in
+                # argv — stale --limit-mm-per-prompt, extra_pip, etc.).
+                #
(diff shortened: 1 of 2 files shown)


wenxind-nvidia (approved):

nvddr:

### Overview
Small, well-motivated fix: after `docker stop`, also `docker rm` so the next `start_xr_ai_vllm` re-runs `build_run_argv` against the current YAML instead of resurrecting a container with stale baked-in argv (`--limit-mm-per-prompt` etc.). PR body explains the root cause clearly and the failure mode is non-fatal with a clean manual-recovery hint. Approved by @wenxind-nvidia already.

### Concerns Worth Discussing
- **Interaction with the `docker start -a` reuse branch in `_docker.run`.** That branch (currently `if container_exists(name) and not container_running(name): docker start -a ...`) exists specifically to skip reinstalling `hf_transfer` on restart. With this PR, the normal stop/start cycle will never hit it — every restart goes through the fresh `docker run` + `pip install -q hf_transfer` (+ any `extra_pip`) path. Net effect on Nemotron-Omni-style configs with `extra_pip=[mamba-ssm, causal-conv1d]` could be a meaningful added startup cost. Worth confirming this is acceptable, or worth documenting that the reuse branch is now only reachable after launcher crashes (not graceful stops).
- **`docker logs` history is lost on `rm`.** Mitigated in practice because `_start_log_streamer` mirrors container stdout/stderr to `xr-ai-logging/<container>.log` while running, plus `_append_post_mortem` runs on failure. So the bind-mount log file survives `rm`. Probably fine — flagging just so a human confirms no one was relying on `docker logs <name>` post-stop for debugging.
- **Bind mount only, no named volumes** (`-v {model_cache}:{model_cache}` is a host path), so `docker rm` does not delete the HF cache. Good.

### Nits
- Style consistency: rest of `_docker.py` uses `check=False` + inspect `returncode` (`_maybe_ngc_login`) or `check=True` + targeted `except CalledProcessError` (`stop_container`). The new block uses `check=True` + bare `except Exception`. Could narrow to `(subprocess.CalledProcessError, FileNotFoundError)` to match neighboring style.
- The inline `import subprocess as _sp` is unusual — `__init__.py` already imports `urllib.request` etc. at module top. Could move the import up, or (cleaner) put the `docker rm` behind a small `_docker.remove_container(name)` helper next to `_docker.stop_container` and keep `__init__.py` free of `subprocess`. Optional — not a blocker.

- `utils/xr-ai-vllm/xr_ai_vllm/__init__.py:L177-L195` — the new try/except block; check exception scope and message wording.
- `utils/xr-ai-vllm/xr_ai_vllm/_docker.py` `run()` — the `container_exists(...) and not container_running(...)` reuse branch; confirm the perf trade-off is acceptable now that graceful stops always force a cold start.
- Any caller / docs that referenced \"the container stays around after stop\" as a debugging affordance (`docker logs`, `docker inspect`).

Aligns with `AGENTS.md`-style stdlib-only contract of `xr-ai-vllm` (no new deps). No `DEPENDENCIES.md` impact (no `pyproject.toml` change). Matches the module's existing pattern of \"best-effort docker ops with print-style fallback messages.\"

---

nvddr:
Addressed the review findings in 58033b8 (on top of the rebase to current main):

|---|---|---|
| 💡 `docker start -a` reuse-path optimization bypassed (extra pip install on every restart) | **Documented in code** — comment block in `__init__.py:stop_persistent_servers` explains the tradeoff: persistent-servers exists to skip MODEL reloads (multi-minute vLLM weight load), not pip reinstalls (few seconds on warm cache). Forcing `docker rm` is the correct behavior because config edits must take effect on next launch. | `utils/xr-ai-vllm/xr_ai_vllm/__init__.py` |
| 🔍 Inline `import subprocess as _sp` + bare `except Exception` inconsistent with `_docker.py`'s narrow catch style | **Fixed** — extracted `_docker.remove_container(name)` helper using narrow `(FileNotFoundError, subprocess.CalledProcessError)` catches. Caller is now `if _docker.remove_container(container_name)`. | `utils/xr-ai-vllm/xr_ai_vllm/_docker.py`, `utils/xr-ai-vllm/xr_ai_vllm/__init__.py` |

CI: 9/9 green on `58033b8`. State: MERGEABLE + CLEAN + APPROVED.

PR #152, opened 2026-05-21
feat(web): camera preview + agent response display in the web sample

nvddr:
Two user-facing additions to the web sample's main page, useful for
debugging any voice+vision agent (today's `simple-vlm-example`,
tomorrow's `glasses-agent-nat`):

1. **Live camera preview card** — 16:9 black panel at the top of the
   page that mirrors the user's selected camera while the LiveKit
   track is publishing. Uses its own `navigator.mediaDevices.getUserMedia`
   stream so the preview keeps working even if the publish track
   restarts. Front-cameras (`facingMode: "user"`) are mirrored
   horizontally to match user expectation; back-cameras keep the
   sensor orientation the server sees. A "LIVE" badge floats over
   the top-left while active.

2. **Agent response card** — text panel below the preview that
   displays whatever was last sent on the `agent.response` data-channel
   topic. Wired via a new `onDataReceived(topic, data) -> boolean`
   hook in `App/core.js`'s `connect()` flow; returning `true`
   suppresses the message from the generic received-messages list.

Cleanup paths: `releasePreviewStream()` stops the local tracks on
camera stop, disconnect, or page unload (`pagehide`).

Removed the old small "Agent status dot" card from `index.html` since
the response panel covers the same intent more usefully. `core.js`
now guards the agent-status DOM lookup with `if (agentDot && agentLabel)`
so clients that don't render the status card (this one, and others) no
longer crash on render.

Also: `core.js::ConnectionState.DISCONNECTED` handler now invokes
`_sc?.()` (stop-camera callback) when the camera was active at
disconnect time — without this, the local camera track stays live
across reconnects.

### DIFF
--- a/client-samples/web/App/app.js
+++ b/client-samples/web/App/app.js
@@ -33,7 +33,23 @@ import {
 // Model state  (mirrors AppModel.swift field-for-field)
 // ─────────────────────────────────────────────────────────────────────────────
 
-const model = { ...createBaseModel() };
+const model = {
+  ...createBaseModel(),
+  /** @type {string|null} Most recent agent.response message text. */
+  agentResponse: null,
+};
+
+// Local camera preview stream (separate from the LiveKit publish stream).
+let _previewStream = null;
+
+function releasePreviewStream() {
+  if (!_previewStream) return;
+  _previewStream.getTracks().forEach(t => t.stop());
+  _previewStream = null;
+  const videoEl = $('camera-preview');
+  videoEl.srcObject = null;
+  videoEl.style.transform = '';
+}
 
 // ─────────────────────────────────────────────────────────────────────────────
 // Error toast
@@ -57,18 +73,91 @@ function showError(message) {
 // Render + bound actions
 // ─────────────────────────────────────────────────────────────────────────────
 
-function render()           { renderBase(model); }
+function render() {
+  renderBase(model);
+
+  // Camera preview elements.
+  const video       = $('camera-preview');
+  const placeholder = $('preview-placeholder');
+  const liveBadge   = $('preview-live-badge');
+  if (model.isCameraActive) {
+    video.classList.add('active');
+    placeholder.style.display = 'none';
+    liveBadge.classList.add('active');
+  } else {
+    video.classList.remove('active');
+    placeholder.style.display = '';
+    liveBadge.classList.remove('active');
+  }
+
+  // Agent response.
+  const responseEl = $('agent-response-text');
+  if (model.agentResponse) {
+    responseEl.textContent = model.agentResponse;
+    responseEl.classList.remove('empty');
+  } else {
+    responseEl.textContent = 'Waiting for agent…';
+    responseEl.classList.add('empty');
+  }
+}
 
 function enumerateCameras() { return _enumerateCameras(model, render); }
-function stopCamera()       { return _stopCamera(model, render, showError); }
-function startCamera()      { return _startCamera(model, { render, showError, enumerateCameras }); }
+
+async function stopCamera() {
+  try {
+    await _stopCamera(model, render, showError);
+  } finally {
+    releasePreviewStream();
+  }
+}
+
+async function startCamera() {
+  await _startCamera(model, { render, showError, enumerateCameras });
+  if (model.isCameraActive) {
+    try {
+      releasePreviewStream();
+      const constraints = model.selectedCameraId
+        ? { video: { deviceId: { exact: model.selectedCameraId } } }
+        : { video: true };
+      _previewStream = await navigator.mediaDevices.getUserMedia(constraints);
+      const videoEl = $('camera-preview');
+      videoEl.srcObject = _previewStream;
+
+      // Mirror the preview only for front-facing cameras — back cameras should
+      // show the same orientation the server receives (no transform).
+      const track = _previewStream.getVideoTracks()[0];
+      const facingMode = track?.getSettings?.()?.facingMode ?? '';
+      videoEl.style.transform = facingMode === 'user' ? 'scaleX(-1)' : '';
+    } catch { /* preview failure is non-fatal */ }
+  }
+  render();
+}
+
 function startAudio()       { return _startAudio(model, render, showError); }
 function stopAudio()        { return _stopAudio(model, render, showError); }
-function disconnect()       { return _disconnect(model, render); }
+async function disconnect() {
+  releasePreviewStream();
+  try {
+    await _disconnect(model, render);
+  } finally {
+    releasePreviewStream();
+    render();
+  }
+}
 function sendPing()         { return _sendPing(model, startCamera); }
 function sendCustom(text)   { return _sendCustom(model, text, showError); }
 function connect()          {
-  return _connect(model, { render, showError, enumerateCameras, stopCamera });
+  return _connect(model, {
+    render, showError, enumerateCameras, stopCamera,
+    onDataReceived(topic, data) {
+      if (topic === 'agent.response') {
+        model.agentResponse = new TextDecoder().decode(data);
+        render();
+        return true; // suppress from the received messages list
+      }
+      return false;
+    },
+  });
 }
 
 // ─────────────────────────────────────────────────────────────────────────────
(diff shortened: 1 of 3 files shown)


wenxind-nvidia (approved):

nvddr:
**Automated Summary Review** _(bot reviewer — please verify with a human; wenxind-nvidia has already approved)_

### Overview
Nice, self-contained UX addition — a 16:9 preview card and an agent-response panel make the sample noticeably more useful for debugging voice+vision agents. The cleanup paths (`releasePreviewStream()` on stop / disconnect / `pagehide`, plus the new `_sc?.()` call on `DISCONNECTED`) look thoughtful, and gating the old agent-status DOM lookup with `if (agentDot && agentLabel)` is the right move now that other clients can drop that widget.

### Concerns Worth Discussing

- **Dual `getUserMedia` for the same camera.** The preview opens its own stream in parallel with LiveKit's publish track. In practice browsers do share the underlying device across two `getUserMedia` calls from the same origin, but there are known sharp edges worth a manual smoke test:
  - When `model.selectedCameraId` is **null** on first start, `_startCamera` uses `CameraConfig.default` while the preview asks for `{ video: true }` — these can resolve to *different* physical devices, so the preview won't match what the server sees. Consider deriving the preview's `deviceId` from the LiveKit track's `getSettings().deviceId` after `_startCamera` resolves, rather than from `model.selectedCameraId`.
  - On iOS Safari and some Android Chrome builds, the second `getUserMedia` can briefly renegotiate the shared source and cause the publish track to drop a frame or downgrade resolution. Worth verifying on a real iPhone / Android device before merging.
  - A cheaper alternative that sidesteps all of this: clone the LiveKit local video track's `MediaStreamTrack` (`track.mediaStreamTrack.clone()` or just bind `srcObject` to a `MediaStream` wrapping it) instead of opening a second device handle. Same pixels, one device consumer.

- **`onDataReceived(topic, data) -> boolean` is fully permissive.** Any caller can intercept *any* topic and swallow it from the received-messages list — including `clientControl`, which has special handling in core. For this sample it's fine, but as a pattern propagated to other clients (web-xr, glasses-agent-nat) it'd be more robust to either (a) accept a `{ [topic]: handler }` map at the core layer so interception is per-topic and explicit, or (b) document that `clientControl` is reserved and core handles it before delegating. Right now a caller returning `true` for `clientControl` would silently break camera-on-demand.

- **Accessibility on the response panel.** `<p id=\"agent-response-text\">` is updated via `textContent` but has no live-region role, so screen readers won't announce new agent responses. Consider `role=\"status\" aria-live=\"polite\" aria-atomic=\"true\"` on either the `<p>` or the surrounding `.agent-response-card`. The \"LIVE\" badge could similarly benefit from `aria-label=\"Camera active\"` since it currently renders as the literal text \"LIVE\" with a decorative dot.

- **Mirroring heuristic misses the common laptop case.** `facingMode === 'user'` is only populated reliably on mobile — on desktops and laptops with built-in/external webcams the value is typically empty string, so the preview will **not** mirror in what is arguably the most common selfie-style setup. Options: (a) treat empty `facingMode` as user-facing when on a non-mobile UA, (b) expose a manual \"mirror preview\" toggle, or (c) mirror by default and un-mirror only when `facingMode === 'environment'`. Option (c) is closest to what users expect from FaceTime / Zoom / Meet.

### Nits _(all optional)_

- The empty `catch { /* preview failure is non-fatal */ }` around the preview `getUserMedia` swallows useful diagnostics (e.g., NotAllowedError on a fresh profile, OverconstrainedError when `deviceId: exact` no longer matches a removed device). A `console.warn` would help future debugging without surfacing a toast.
- `disconnect()` calls `releasePreviewStream()` twice (before *and* after `_disconnect`). The second call is a no-op on the cleared `_previewStream`, but the intent reads a bit confusingly — a comment or a single trailing call would be clearer.
- The `pagehide` handler calls `model.session?.disconnect()` directly rather than the app's `disconnect()` wrapper, so it skips the `render()` and the wrapper's `releasePreviewStream()` choreography. Minor since the page is going away, but worth a one-line comment noting the intent.
- `connect()` is now `async`-ish via its callback but the function itself isn't marked `async`. Cosmetic.

- Is the `agent.response` topic name finalized across client+server, or still in flux? If the latter, worth a brief `// TODO` near the intercept so it's findable.
- Should the response panel show *only the latest* message, or accumulate a short transcript? Today's behavior is last-wins; a multi-turn debugging session loses prior responses. (Probably fine for the sample — just confirming the design intent.)

- `client-samples/web/App/app.js` — the new `startCamera` body (the dual-`getUserMedia` + facingMode mirror logic) is the highest-risk block; please eyeball on at least one mobile target (iOS Safari and Android Chrome) and one laptop.
- `client-samples/web/App/app.js` — the `connect()` `onDataReceived` closure: confirm `clientControl` interception is not a footgun for downstream clients.
- `client-samples/web/App/core.js` — the `wasCameraActive` snapshot + `_sc?.()` call inside the `DISCONNECTED` branch: please confirm that calling `_sc` (which awaits `session.stopCamera()` on a session that's already disconnecting) is safe and won't reject. The implementation swallows StreamError, but if `model.session` is already nulled out elsewhere this could no-op silently.
- `client-samples/web/index.html` — the new preview / response cards sit above the existing connection section, which shifts the page layout meaningfully. Confirm this is the intended visual hierarchy on both desktop and the narrow `max-width: 540px` mobile layout.

Per the root `CLAUDE.md`, `AGENTS.md` is authoritative for conventions — I didn't surface any conflicts with the existing core/app split or naming. The change correctly keeps app-specific extension in `App/app.js` and only adds a generic `onDataReceived` extension point + an optional-DOM guard in `App/core.js`, which is consistent with the documented \"core stays generic, apps extend\" pattern.

### Deferred to Experts

- **Web media / device-handling specialist**: the dual-`getUserMedia` interaction with LiveKit on iOS Safari and Android Chrome warrants someone who's hit those corners before.
- **Accessibility reviewer**: live-region semantics on the agent-response panel.

---

nvddr:
Addressed the review findings in b6337e8:

|---|---|---|
| 💡 Dual-`getUserMedia` collision risk on iOS Safari / Android Chrome; deviceId divergence when `selectedCameraId === null` | **Partial fix + escalated** — fixed the deviceId divergence by pinning the preview's `getUserMedia` to `facingMode: 'user'` when no `deviceId` is selected (matches `LiveKitBackend.js:285`'s default). The dual-open itself stays, since the clean-fix (clone the LiveKit publish track) requires reaching `LiveKitBackend.#videoTrack` which is a JS private field with no public accessor — that would expand the change beyond this PR's scope. Noted as a StreamKit-side follow-up. | `client-samples/web/App/app.js` |
| 💡 `onDataReceived` topic permissiveness (any topic interceptable) | **Pushed back / no change** — by design for this hook. App-level `onDataReceived` returns `true` only for `agent.response`; other topics fall through to the default. Documenting reserved-topics is a separate API contract decision. | (no-op) |
| 💡 Missing `role="status" aria-live="polite" aria-atomic="true"` on agent response panel | **Fixed** — added on `<p id="agent-response-text">`. | `client-samples/web/index.html` |
| 🔍 Mirroring heuristic backwards for desktop webcams (`facingMode` is empty on most desktop cameras → no mirror on selfie cams) | **Fixed** — inverted: mirror by default, skip only when `facingMode === 'environment'`. | `client-samples/web/App/app.js` |

CI: green on `b6337e8`. State: MERGEABLE + CLEAN + APPROVED.

nvddr:
Tightened the mirroring logic in a6ae909 after a board review of the primary test case.

Primary case is XR / glasses / mobile-back-camera capture — those need **left = left, right = right** on the preview (matches the world the camera sees). The previous "mirror by default, skip on environment" heuristic worked for environment cams but had two false-positive paths:

- Desktop / laptop webcams typically don't report any `facingMode` → got mirrored.
- External USB or device-back cameras that fail to report `facingMode` → got mirrored.

| Concern | Disposition | Where |
|---|---|---|
| Mirror logic should be watertight for environment cameras (primary test case) | **Fixed** — inverted to: default no-mirror; flip ONLY when `facingMode === 'user'` (explicit front mobile cam). Environment, desktop-with-no-facingMode, and external/back-USB all pass through un-mirrored now. | `client-samples/web/App/app.js` |
| Default `getUserMedia` constraint forced `facingMode: 'user'` even when no deviceId chosen | **Fixed** — dropped to plain `{ video: true }` in the no-deviceId case so the preview and the LiveKit publish track converge on the same browser-default camera. Explicit deviceId selection (dropdown) is unchanged — both pin to that exact device. | `client-samples/web/App/app.js` |

Tradeoff acknowledged: desktop selfie-webcam users no longer get the FaceTime-style auto-mirror UX. The sample's audience is XR-test-driven, not webcam-call-driven; a manual mirror toggle can land later if that gap matters.

CI: 9/9 SUCCESS on `a6ae909`. State: MERGEABLE + CLEAN + APPROVED.

PR #153, opened 2026-05-21
fix(web): exception-safe teardown in LiveKitBackend; skip unpublish when disconnected

nvddr:
Two robustness fixes in `client-samples/web/StreamKit/Backends/LiveKit/LiveKitBackend.js`:

1. **`stopCamera`** no longer calls `room.localParticipant.unpublishTrack`
   when the room isn't connected (`room.state !== 'connected'`). LiveKit's
   client SDK throws if you try, leaking a hot camera track because the
   surrounding code never reached `track.stop()`. Now: clear the field
   first, then conditionally unpublish in a try/finally so `track.stop()`
   always runs.

2. **`#tearDown`** previously ran `room.disconnect()` first, then cleaned
   up audio elements + tracks. A thrown promise from `disconnect()` left
   audio `<audio>` nodes attached to the DOM and the camera/mic tracks
   live. Restructured to clear `this.#room` up front and wrap the
   destructive ops in try/finally so the audio elements, video track,
   audio track, and `sessionConfig` are always released — even if the
   network disconnect raises.

No behaviour change on the happy path.

### DIFF
--- a/client-samples/web/StreamKit/Backends/LiveKit/LiveKitBackend.js
+++ b/client-samples/web/StreamKit/Backends/LiveKit/LiveKitBackend.js
@@ -304,13 +304,17 @@ export class LiveKitBackend {
    * @returns {Promise<void>}
    */
   async stopCamera() {
-    if (!this.#videoTrack) return;
-
-    if (this.#room) {
-      await this.#room.localParticipant.unpublishTrack(this.#videoTrack);
-    }
-    this.#videoTrack.stop();
+    const track = this.#videoTrack;
+    if (!track) return;
     this.#videoTrack = null;
+
+    try {
+      if (this.#room && this.#room.state === 'connected') {
+        await this.#room.localParticipant.unpublishTrack(track);
+      }
+    } finally {
+      track.stop();
+    }
   }
 
   /**
@@ -362,27 +366,30 @@ export class LiveKitBackend {
    * @returns {Promise<void>}
    */
   async #tearDown() {
-    if (this.#room) {
-      // Remove all listeners before disconnecting to prevent stale callbacks.
-      this.#room.removeAllListeners();
-      await this.#room.disconnect();
-      this.#room = null;
+    const room = this.#room;
+    this.#room = null;
+    try {
+      if (room) {
+        // Remove all listeners before disconnecting to prevent stale callbacks.
+        room.removeAllListeners();
+        await room.disconnect();
+      }
+    } finally {
+      for (const el of this.#audioElements.values()) el.remove();
+      this.#audioElements.clear();
+
+      if (this.#videoTrack) {
+        this.#videoTrack.stop();
+        this.#videoTrack = null;
+      }
+
+      if (this.#audioTrack) {
+        this.#audioTrack.stop();
+        this.#audioTrack = null;
+      }
+
+      this.#sessionConfig = null;
     }
-
-    for (const el of this.#audioElements.values()) el.remove();
-    this.#audioElements.clear();
-
-    if (this.#videoTrack) {
-      this.#videoTrack.stop();
-      this.#videoTrack = null;
-    }
-
-    if (this.#audioTrack) {
-      this.#audioTrack.stop();
-      this.#audioTrack = null;
-    }
-
-    this.#sessionConfig = null;
   }
 
   /**
--- a/client-samples/web/StreamKit/Backends/LiveKit/LiveKitBackend.js
+++ b/client-samples/web/StreamKit/Backends/LiveKit/LiveKitBackend.js
@@ -242,12 +242,17 @@ export class LiveKitBackend {
    * @returns {Promise<void>}
    */
   async stopAudio() {
-    if (!this.#audioTrack) return;
-    if (this.#room) {
-      await this.#room.localParticipant.unpublishTrack(this.#audioTrack);
-    }
-    this.#audioTrack.stop();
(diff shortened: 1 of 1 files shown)


wenxind-nvidia (approved):

nvddr:

### Overview
Nice, focused robustness fix. Both changes follow the same exception-safe shape (snapshot field, null it out, do destructive work inside `try/finally`), and the PR description clearly explains the failure mode each one addresses. Happy-path behaviour is preserved, and the diff is small and easy to reason about.

### Concerns Worth Discussing
- **Silent loss of the disconnect/unpublish error.** Neither `try/finally` block swallows the inner exception (finally doesn't catch), so it still propagates — but that means `connect()` (line 144) and `disconnect()` (line 261) can now throw a `disconnect()` error *after* `#room`, audio elements, tracks, and `sessionConfig` are all gone. The previous code would throw before any of that cleanup. Functionally an improvement, but worth a one-line `console.warn` inside `finally` (or a swallow + warn) so the disconnect failure isn't invisible to anyone debugging from logs. Same applies to `stopCamera`'s unpublish path.
- **Parity gap with `stopAudio` (LiveKitBackend.js:244-251).** It still has the exact pre-fix shape: no `room.state === 'connected'` guard, no `try/finally`, so an `unpublishTrack` throw on a non-connected room will skip `audioTrack.stop()` and leak the mic. If the LiveKit SDK behaviour that motivated the camera fix applies symmetrically, `stopAudio` likely wants the same treatment in this PR (or an explicit follow-up note).

### Nits
- The comment on the new `#tearDown` line `// Remove all listeners before disconnecting to prevent stale callbacks.` is still accurate, but since `this.#room = null` now happens *before* `removeAllListeners()`, you could optionally tighten the comment to note that nulling-first is what makes re-entrant `#tearDown` calls (e.g. from `connect()` racing with `disconnect()`) idempotent.

- Does the LiveKit v2 client ever expose `'disconnecting'` as a `room.state` value? The header comment at LiveKitBackend.js:35-39 only enumerates `connected | connecting | disconnected | reconnecting`. `state !== 'connected'` is the safe direction either way (skips unpublish in any transitional state), so this is fine — just confirming the assumption is documented.
- Was the `unpublishTrack` throw observed for `stopCamera` specifically, or for any unpublish-while-not-connected? (Drives whether `stopAudio` needs the same fix — see above.)

- `client-samples/web/StreamKit/Backends/LiveKit/LiveKitBackend.js:306-318` — confirm that clearing `#videoTrack` before the `await` is intentional re: re-entrancy (a concurrent `startCamera` → `stopCamera` toggle would now see `#videoTrack === null` and early-return, while the previous in-flight `unpublishTrack` is still resolving. Likely fine but worth a thought.)
- `client-samples/web/StreamKit/Backends/LiveKit/LiveKitBackend.js:364-386` — verify that nothing inside the `finally` block depends on `#room`/`#videoTrack`/`#audioTrack` still being live. Skim says no, but a second pair of eyes on the ordering of releases is cheap.
- `client-samples/web/StreamKit/Backends/LiveKit/LiveKitBackend.js:244-251` — `stopAudio`, untouched by this PR but structurally identical to the bug being fixed.
- `client-samples/web/App/core.js:423-430, 489-497` — callers already wrap in try/catch (`stopCamera`) or don't (`disconnect`). Worth checking whether `disconnect(model, render)` should now expect a thrown error from `#tearDown` and surface it, since the new code preserves the throw after cleanup.

Diff is local to one file, matches existing style (private fields, JSDoc, the `room.state === 'connected'` guard pattern already used in `startAudio`/`startCamera`/`send`). No `DEPENDENCIES.md` or `AGENTS.md` touchpoints. Looks consistent with module conventions.

### Deferred to Experts
- LiveKit SDK behaviour assumptions (exact throw conditions for `unpublishTrack` on non-`connected` rooms, completeness of the `room.state` value set) would benefit from a quick sanity check by someone with hands-on LiveKit JS v2 experience — pinging @wenxind-nvidia in case worth a second look given their approval.

---

nvddr:
Addressed the review findings in 2c6d231:

|---|---|---|
| 💡 `stopAudio` parity gap (same unfixed shape as old `stopCamera` — clears field after await, no `room.state` guard, no try/finally) | **Fixed** — mirrored the new `stopCamera` pattern: local capture, null the field first, `room.state === 'connected'` guard before `unpublishTrack`, `track.stop()` in `finally`. | `client-samples/web/StreamKit/Backends/LiveKit/LiveKitBackend.js` |
| 💡 `#tearDown` try/finally silently swallows the inner `room.disconnect()` exception | **Fixed** — added a `catch (err)` between the existing try and finally that emits `console.warn('LiveKitBackend: room.disconnect() failed', err)`. | `client-samples/web/StreamKit/Backends/LiveKit/LiveKitBackend.js` |
| 🔍 Caller-side `disconnect` in `App/core.js` has no try/catch around `#tearDown` (which can now throw after partial cleanup) | **Pushed back / out of scope** — `App/core.js` was a separate PR (#152). Caller hardening is a follow-up; this PR scoped to the LiveKit backend boundary. | (no-op) |
| Question on `this.#room = null` before await | **Reasoning posted** — clearing the field up front is defensively correct: concurrent `stopCamera` / `#tearDown` calls now early-return safely, and `removeAllListeners()` ahead of the await prevents the "callback sees null `#room`" failure mode. | (reply) |

CI: green on `2c6d231`. State: MERGEABLE + CLEAN + APPROVED.

PR #140, opened 2026-05-14
feat(xr-render-demo): migrate worker to xr-ai-models SDK

nvddr:
## Summary

Stacked on #135 — Unit 4 of the modular AI-models refactor.

- Replaces the four `httpx.post(/v1/chat/completions)` call sites in `RenderSceneProcessor` with `LLMService.chat()` calls via the `xr-ai-models` SDK:
  - **quick-ack** (`llm`, max_tokens=40, temperature=0.0)
  - **still-working** (`llm`, max_tokens=24, temperature=0.9)
  - **validation** (`llm`, max_tokens=60, temperature=0.0)
  - **agentic-loop** (`agent_llm`, tools, enable_thinking, thinking_budget=1024)
- `WorkerConfig` drops the five model-server URL fields (`stt_server`, `tts_server`, `llm_server`, `agent_llm_server`, `vlm_server`); adds `models_yaml` path resolved relative to the YAML file.
- New `yaml/models.yaml` with five preset-backed entries (llama_nemotron, nemotron3_nano, parakeet_stt, piper_tts, cosmos_vlm) and their default port base URLs.
- `xr_render_demo_worker.py` loads `models.yaml` via `load_models_config` and constructs SDK services; startup health probes converted from `http_probe()` URLs to `<service>.health` callables; MCP probes unchanged.
- `_tools_openai: list[dict]` → `_tools: list[ToolDef]`; `ToolDef.to_openai()` reproduces the identical OpenAI wire shape so the upstream server sees no change.
- Agentic-loop response parsing uses `ChatResponse.reasoning` (SDK-normalized via `nemotron3_nano` preset's `reasoning_field: "reasoning"`) — no more manual `reasoning_content` vs `reasoning` branch.
- `processors.py`: `import uuid` and `import os` moved to top-level; unused `ChatResponse` import removed; `close()` now closes both LLM clients.

## Test plan

- [x] 9 new wire-trace golden tests in `tests/test_xr_render_demo_wire.py` covering all four LLM call sites, the agentic-loop `chat_template_kwargs` shape, reasoning-field normalization, tool-call parsing, and `ToolDef.to_openai()` round-trip
- [x] Full non-GPU suite: `229 passed, 32 deselected` — all baseline tests green
- [x] SPDX header check: 307 files OK
- [ ] GPU smoke skipped — no GPU available; stub-server tests cover the wire path

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #154, opened 2026-05-21
feat(xr-ai-vad): shared Silero VAD util; migrate simple-vlm-example

nvddr:
## Summary

- Adds **`utils/xr-ai-vad/`** — a small shared utility package that wraps Silero VAD (ONNX, no torch/GPU required) with an adaptive energy fallback. API is one `VadDetector` class with two async callbacks (`on_utterance`, optional `on_speech_start`); raw float32 PCM bytes in, int16 PCM utterance bytes out — same byte format `xr_ai_agent.AudioChunk.data` carries.
- Migrates **`agent-samples/simple-vlm-example/worker`** off its inline RMS energy gate to `VadDetector`. The previous speculative camera-on (fired the moment `speech_s` crossed `min_speech`) is preserved via the new `on_speech_start` hook.
- Worker YAML gains `silero_threshold` + `vad_noise_mult`; `silence_threshold` is now scoped to the energy fallback only (documented inline). Defaults match the glasses-agent-nat reference.
- `xr-render-demo` is **intentionally untouched** — see Follow-ups.
- `DEPENDENCIES.md` / `docs/changelog.md` / `docs/adding-a-sample.md` updated. `tests/test_xr_ai_vad.py` exercises the state machine via the energy fallback (no ONNX model needed in CI).

## Structural decision — why `utils/xr-ai-vad/`

Considered four shapes (per-sample copies / `utils/`-shared / `agent-sdk/`-shared / `agent-samples/_shared/`). Chose `utils/` for the same reasons `utils/xr-ai-logging/` lives there:

- **Not `agent-sdk/`** — that tree is the HTTP-client seam for AI services (LLM/VLM/STT/TTS). Local DSP on raw PCM doesn't fit.
- **Not per-sample copies** — silero VAD already exists in `xr-render-demo` (pipecat-coupled, on main) and `glasses-agent-nat` (clean async, on a separate branch in active dev). simple-vlm-example was about to become the third copy.
- **Not `agent-samples/_shared/`** — no such precedent exists; a `utils/` package applies equally to non-sample callers (tests, future tools).

**On AGENTS.md's "two concrete use-cases" rule**: on `main` today there is exactly one. The second (`xr-render-demo`) exists inline today and is the obvious next migration; the third (`glasses-agent-nat`) is already implemented on a sibling branch and was the reference for this PR. The API surface is small and stable enough to live behind one class.

**Why raw bytes in/out (no `xr-ai-agent` dep)**: keeps the `utils/` → `agent-sdk/` direction unbroken, mirrors `utils/xr-ai-logging/`'s focused dep footprint, and leaves the door open for the `xr-render-demo` migration (which has int16 PCM inputs) without an `AudioChunk` edge.

## Files changed

- `utils/xr-ai-vad/{pyproject.toml,xr_ai_vad/__init__.py,xr_ai_vad/detector.py}` — new package
- `agent-samples/simple-vlm-example/worker/{agent.py,audio.py,voice.py,pyproject.toml,simple_vlm_example_worker.py}` — migrate to `xr-ai-vad`; drop `rms` + `chunks_to_wav`; slim `VoiceState` to per-pid bookkeeping
- `agent-samples/simple-vlm-example/yaml/simple_vlm_example_worker.yaml` — new VAD keys with comments scoping silence_threshold to fallback
- `tests/{pyproject.toml,test_xr_ai_vad.py}` — register new editable dep + state-machine tests
- `DEPENDENCIES.md`, `docs/changelog.md`, `docs/adding-a-sample.md` — keep docs in sync per the same-commit rule

## Test plan

- [x] Syntax-check all touched Python (`ast.parse`) — clean
- [x] `pytest tests/test_xr_ai_vad.py` — 4/4 passing locally (energy-path state machine; on_speech_start one-shot per utterance; sub-min_speech utterances not emitted; `reset()` drops in-progress)
- [ ] CI: lock-check resolves `xr-ai-vad` editable across simple-vlm-example/worker + tests
- [ ] CI: pytest across Python 3.11 / 3.12
- [ ] Manual verification of `simple-vlm-example` end-to-end with a real mic + STT — recommend reviewers smoke-test that an utterance still ends correctly and the camera still warms up speculatively

## Notes for reviewers

- The silero call inside `feed()` is sync CPU work on the event loop — matches the existing pattern in both `xr-render-demo` and `glasses-agent-nat`. Windows are 512 samples / ~32 ms; per-call cost is sub-ms in practice. `run_in_executor` is available as a follow-up if profiling shows it.
- `silence_threshold` keeps its name (config compatibility) but is now only honored in the energy fallback path. Documented in both the YAML comments and the entry-point docstring.

## Follow-ups

- **Migrate `xr-render-demo` to `xr-ai-vad`** — deferred. Its silero is wedged into a Pipecat `FrameProcessor` (int16-PCM input, inline STT, downstream `TranscriptionFrame` push, filler-phrase filtering, trace logging). Extraction needs either a pipecat wrapper around the new detector or pulling the core detection loop out and re-threading the FrameProcessor around it. Worth its own PR.

### DIFF
--- a/DEPENDENCIES.md
+++ b/DEPENDENCIES.md
@@ -90,6 +90,18 @@ xr-ai-vllm  (utils/xr-ai-vllm/)
     container.  Imported by the four vllm wrappers and by the orchestrator
     `--stop` flow.
 
+xr-ai-vad  (utils/xr-ai-vad/)
+    └── numpy >=1.24
+    └── silero-vad >=5.1
+    Shared per-participant Silero VAD utterance detector for agent workers
+    that ingest microphone audio.  ONNX backend — no torch / GPU required
+    at runtime.  Falls back to an adaptive energy gate when silero-vad
+    fails to load.  Consumes raw float32 PCM bytes (the format
+    ``AudioChunk.data`` uses) and emits int16 PCM utterance bytes via an
+    async ``on_utterance`` callback; an optional ``on_speech_start`` hook
+    fires when speech first crosses ``min_speech`` for speculative
+    downstream warmup (e.g. start the camera before STT completes).
+
 xr-media-hub  (server-runtime/)
     └── xr-ai-agent  [editable: ../agent-sdk]
     └── pyzmq >=26.0
@@ -170,6 +182,7 @@ xr-ai-tests  (tests/)
     └── xr-media-hub            [editable: ../server-runtime]    (pulls in livekit, livekit-api for the wss /rtc proxy + room-client tests)
     └── xr-ai-launcher          [editable: ../utils/xr-ai-launcher]
     └── xr-ai-logging           [editable: ../utils/xr-ai-logging]
+    └── xr-ai-vad               [editable: ../utils/xr-ai-vad]
     └── xr-ai-vllm              [editable: ../utils/xr-ai-vllm]
     └── transcript-mcp-server   [editable: ../agent-mcp-servers/transcript-mcp]
     └── vlm-mcp-server          [editable: ../agent-mcp-servers/vlm-mcp]
@@ -346,7 +359,7 @@ the latest video frame via streaming VLM and replies with both
 | Sub-project | Package | Internal deps | External deps |
 |---|---|---|---|
 | Orchestrator | `simple-vlm-example` | `xr-ai-launcher` | — |
-| Worker | `simple-vlm-example-worker` | `xr-ai-agent`, `xr-ai-models [editable]` | numpy >=1.24, Pillow >=10.0, pyyaml >=6.0 |
+| Worker | `simple-vlm-example-worker` | `xr-ai-agent`, `xr-ai-models [editable]`, `xr-ai-vad [editable]` | numpy >=1.24, Pillow >=10.0, pyyaml >=6.0 (silero-vad pulled in via xr-ai-vad) |
 
 Worker calls stt-server (8103), vlm-server (8100), and piper-tts-server
 (8105) over HTTP via `xr-ai-models` SDK — no model weights loaded
--- a/DEPENDENCIES.md
+++ b/DEPENDENCIES.md
@@ -92,15 +92,14 @@ xr-ai-vllm  (utils/xr-ai-vllm/)
 
 xr-ai-vad  (utils/xr-ai-vad/)
     └── numpy >=1.24
-    └── silero-vad >=5.1
+    └── silero-vad >=5.1  (pulls torch + onnxruntime transitively)
     Shared per-participant Silero VAD utterance detector for agent workers
-    that ingest microphone audio.  ONNX backend — no torch / GPU required
-    at runtime.  Falls back to an adaptive energy gate when silero-vad
-    fails to load.  Consumes raw float32 PCM bytes (the format
-    ``AudioChunk.data`` uses) and emits int16 PCM utterance bytes via an
-    async ``on_utterance`` callback; an optional ``on_speech_start`` hook
-    fires when speech first crosses ``min_speech`` for speculative
-    downstream warmup (e.g. start the camera before STT completes).
+    that ingest microphone audio.  Uses the ONNX backend (no GPU required
+    at runtime).  Consumes raw int16 PCM bytes and emits int16 PCM utterance
+    bytes via an async ``on_utterance`` callback; an optional
+    ``on_speech_start`` hook fires when speech first crosses ``min_speech``
+    for speculative downstream warmup (e.g. start the camera before STT
+    completes).
 
 xr-media-hub  (server-runtime/)
     └── xr-ai-agent  [editable: ../agent-sdk]
@@ -392,7 +391,7 @@ forwarding.
 | Sub-project | Package | Internal deps | External deps |
 |---|---|---|---|
 | Orchestrator | `xr-render-demo` | `xr-ai-launcher`, `xr-ai-logging` | — |
-| Worker | `xr-render-demo-worker` | `xr-ai-agent`, `xr-ai-models` [editable], `xr-ai-pipecat` [editable], `xr-ai-logging` [editable] | numpy >=1.24, httpx >=0.27, fastmcp >=0.4, pyyaml >=6.0, silero-vad >=5.1 |
+| Worker | `xr-render-demo-worker` | `xr-ai-agent`, `xr-ai-models` [editable], `xr-ai-pipecat` [editable], `xr-ai-logging` [editable], `xr-ai-vad` [editable] | numpy >=1.24, httpx >=0.27, fastmcp >=0.4, pyyaml >=6.0 (silero-vad pulled in via xr-ai-vad) |
 
 Model endpoints (llm, agent_llm, stt, tts, vlm) are declared in
 `yaml/models.yaml` and loaded via `xr-ai-models` `load_models_config` /
--- a/agent-samples/simple-vlm-example/worker/agent.py
+++ b/agent-samples/simple-vlm-example/worker/agent.py
@@ -51,8 +51,9 @@ from xr_ai_agent import (AudioChunk, DataMessage, FrameSignal,
                           ParticipantEvent, ProcessorEndpoint)
 from xr_ai_logging import print_task_done_banner
 from xr_ai_models import STTService, TTSService, VLMService
+from xr_ai_vad import VadDetector
 
-from audio import chunks_to_wav, now_us, rms, wav_to_chunks
+from audio import int16_pcm_to_wav, now_us, wav_to_chunks
 from pixels import encode_image, frame_to_pil
 from voice import VoiceState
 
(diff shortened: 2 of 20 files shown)


wenxind-nvidia (approved):

nvddr:
## Scope correction disposition (head: `7ea0de0`)

| # | Finding | Status | What changed |
|---|---|---|---|
| 1 | Drop adaptive-energy fallback from `xr-ai-vad` | ✅ done | `silence_threshold`, `vad_noise_mult`, `_noise_floor`, `_energy_vad` removed from `utils/xr-ai-vad/xr_ai_vad/detector.py`. Silero load is no longer wrapped in `try/except` — if silero or torch can't import, construction raises. Canonical input flipped from `float32_bytes + n_samples` to int16 LE PCM bytes; `simple-vlm-example` does the `float32→int16` conversion at the call site (`worker/agent.py::_on_audio`). YAML config (`yaml/simple_vlm_example_worker.yaml`) and entry-point docstring (`simple_vlm_example_worker.py`) lose the fallback knobs. `tests/test_xr_ai_vad.py` swaps in a deterministic stub for the silero session and exercises the same four state-machine scenarios (no fallback path under test). |
| 2 | Migrate `xr-render-demo` to `xr_ai_vad.VadDetector` | ✅ done | `agent-samples/xr-render-demo/worker/processors.py::SttProcessor` no longer carries an inline silero session, pre-roll buffer, `_finalize`, `_speech_s`/`_silent_s`/`_stt_busy` bookkeeping, or the `_TORCH_AVAILABLE` shim. It now constructs a `VadDetector` and runs STT + filler filtering + `TranscriptionFrame` push from the `on_utterance` callback. Pipecat's int16 PCM is passed through directly (no conversion). `worker/pyproject.toml` adds `xr-ai-vad` (+ `tool.uv.sources` editable path) and drops the direct `silero-vad` dep. `worker/config.py` and `yaml/xr_render_demo_worker.yaml` drop the fallback-only knobs (`silence_threshold`, `vad_noise_mult`). `DEPENDENCIES.md` and `docs/xr-render-demo.md` updated to reflect the new dep + that VAD is Silero-only via the shared util. |

### CI
All 9 checks green on `7ea0de0`: CodeQL, pytest (3.11 + 3.12), uv lock, SPDX headers, CodeQL Analyze python/js-ts.

nvddr:
Runtime bug caught + fixed in e18007b.

**Bug**: first mic input in `simple-vlm-example/worker` crashed with `ModuleNotFoundError: No module named 'onnxruntime'` at `VadDetector.__init__` → `load_silero_vad(onnx=True)`.

**Root cause**: `silero-vad>=5.1` lists `onnxruntime` as an OPTIONAL extra, not a hard dep. `xr-ai-vad` always uses the ONNX backend, so consumers always need `onnxruntime` — but `uv sync` skipped it because nothing in the dep tree required it.

|---|---|---|
| `onnxruntime` missing in consumer venv | **Fixed** — added `onnxruntime>=1.17` explicitly to `xr-ai-vad`'s `[project] dependencies`. Removed the misleading comment that claimed silero-vad pulled it transitively. | `utils/xr-ai-vad/pyproject.toml` |
| Existing tests didn't catch the missing dep (real silero load runs at construction, then `_StubSilero` overwrites) | **Fixed** — added `test_real_silero_model_loads_in_consumer_venv` that constructs `VadDetector` with NO stub and feeds one real audio chunk. Any future missing-runtime regression (onnxruntime, torch, etc.) fails this test instead of first-mic-input in a consumer. | `tests/test_xr_ai_vad.py` |

The existing tests passed in CI because the workspace test venv happened to have `onnxruntime` installed transitively from another sibling package, so construction never raised in that environment. Per-sample worker venvs (`simple-vlm-example/worker/.venv`) don't share that transitive luck.

CI: 9/9 SUCCESS on `e18007b`. Rebased onto main after merge-state went BEHIND. State: MERGEABLE + CLEAN + APPROVED.

nvddr:
Second runtime bug caught + fixed in f62ddc7.

**Bug**: first mic input crashed with `ValueError: Input audio chunk is too short` from `silero_vad/utils_vad.py:_validate_input`.

**Root cause**: Silero v5 only accepts 16 kHz or 8 kHz and requires a fixed 512-sample window at 16 kHz (32 ms). LiveKit publishes WebRTC audio at 48 kHz by default, so `chunk.sample_rate` arriving at the detector is 48000. The detector hardcoded `_SILERO_WINDOW = 512` (right for 16 kHz only) and forwarded the raw `sample_rate` straight to silero. At 48 kHz, silero expected ~1536 samples for a 32 ms window, saw 512, raised.

|---|---|---|
| Detector passes raw `sample_rate` to silero; silero only accepts 16 k / 8 k | **Fixed** — `_classify` now resamples input to 16 kHz via `np.interp` (cheap linear, adequate for VAD), buffers at 16 k, calls silero with `sample_rate=16000`. The `on_utterance` callback still receives the original-rate PCM for full-fidelity STT. | `utils/xr-ai-vad/xr_ai_vad/detector.py` |
| Previous regression test fed at 16 kHz, didn't exercise the sample-rate-mismatch path | **Fixed** — added `test_real_silero_accepts_48khz_via_internal_resample` driving the detector at 48 kHz (the rate the hub actually sends) and asserting silero classify doesn't raise. | `tests/test_xr_ai_vad.py` |

CI: 9/9 SUCCESS on `f62ddc7`. State: MERGEABLE + CLEAN + APPROVED.

To pick up the fix locally: `uv sync` in `agent-samples/simple-vlm-example/worker/`, then restart the worker.

nvddr:
Third runtime bug — fixed in deb4f77.

**Bug**: xr-render-demo worker crashed at startup:

    FileNotFoundError: '.../xr-render-demo/yaml/yaml/models.yaml'

**Root cause**: `config.load_config` resolves `models_yaml` relative to the worker YAML's parent dir (`yaml/`). The YAML had `models_yaml: yaml/models.yaml`, which the resolver joined to `yaml/yaml/models.yaml` — one segment too many. Same applied to the in-code fallback default.

|---|---|---|
| `models_yaml` path doubled by config-dir-relative resolver | **Fixed** — value in YAML and in-code default both changed to bare `models.yaml`. The resolver still joins on `path.parent`, landing on the correct `agent-samples/xr-render-demo/yaml/models.yaml`. | `agent-samples/xr-render-demo/yaml/xr_render_demo_worker.yaml`, `agent-samples/xr-render-demo/worker/config.py` |

Note: simple-vlm-example uses a different resolver (CWD-relative, not config-dir-relative) — its `models_yaml: yaml/models.yaml` was correct under that convention. Unifying the two resolvers is out of scope for this PR.

CI: 9/9 SUCCESS on `deb4f77`. State: MERGEABLE + CLEAN + APPROVED.

nvddr:
Resolver convention triage + unification in 86993fd.

# Triage

| Question | Answer |
|---|---|
| When did xr-render-demo's `models_yaml` path-doubling enter main? | Commit `84a423c` = PR #140 \"feat(xr-render-demo): migrate worker to xr-ai-models SDK\", merged 2026-05-21T00:02Z. The doubled `yaml/yaml/models.yaml` shipped at #140 — has been on main ~6 hours; nobody ran the worker locally to hit it. |
| Why did simple-vlm and xr-render-demo resolve differently? | Migrated in separate PRs (#138 vs #140) with no shared resolver. `simple-vlm-example` (#138) used CWD-relative — `cfg.get(\"models_yaml\") → load_models_config(raw)` — relies on the launcher cd'ing into the sample root. `xr-render-demo` (#140) added a config-file-relative `path.parent / models_yaml_raw` for CWD-independence, but introduced the doubling. |

# Fix

`deb4f77` corrected xr-render-demo by switching its YAML + default to bare `models.yaml`. `86993fd` now unifies `simple-vlm-example` onto the **same config-file-relative convention** so both samples behave the same regardless of CWD:

| Item | Disposition | Where |
|---|---|---|
| simple-vlm resolver was CWD-relative, diverging from xr-render-demo | **Fixed** — `simple_vlm_example_worker.py:main()` now takes `config_path` and joins `models_yaml` to its parent dir when relative. `run()` threads `ns.config` through. | `agent-samples/simple-vlm-example/worker/simple_vlm_example_worker.py` |
| Default value `models_yaml: yaml/models.yaml` would double under the new resolver | **Fixed** — bare `models.yaml` in YAML + bare-basename in-code default. | `agent-samples/simple-vlm-example/yaml/simple_vlm_example_worker.yaml` |
| `models.omni.yaml` header-comment example still showed old path | **Fixed** — `models_yaml: models.omni.yaml`. | `agent-samples/simple-vlm-example/yaml/models.omni.yaml` |

Behavior preserved when the launcher passes absolute `--config` (production path). Bare `uv run` without `--config` now falls back to `models.yaml` relative to CWD instead of `yaml/models.yaml` — small surface-only change.

CI: 9/9 SUCCESS on `86993fd`. State: MERGEABLE + CLEAN + APPROVED.

PR #155, opened 2026-05-22
add wake word for simple-vlm-example. configure wake word in simple_vlm_example_worker.yaml

wenxind-nvidia:
Check wake word in STT text, ignore speech query if it does not start with wake word.


yanziz-nvidia (commented):

## Summary

Adds an optional wake-word gate to `simple-vlm-example`: STT transcripts not prefixed with the configured phrase are silently dropped; the prefix is stripped before the VLM sees the query. The bypass for stop-like phrases (`_STOP_RE`) is correctly scoped to the audio VAD path only. All CI checks pass; no `pyproject.toml` files were touched so no `DEPENDENCIES.md` update is required.

|     | Finding |
|-----|---------|
| 🚫 | None |
| 💡 | [`agent-samples/simple-vlm-example/yaml/simple_vlm_example_worker.yaml:45`](https://github.com/NVIDIA/xr-ai/blob/c1685d068af7fac708be8cde27f78d9f40b73768/agent-samples/simple-vlm-example/yaml/simple_vlm_example_worker.yaml#L45) — **The shipped YAML defaults `wake_word` to `"agent"`, which is a breaking change for anyone running the current sample.** Prior behavior was "no filter"; any utterance was dispatched. Users upgrading who don't read the YAML diff will find all their voice queries silently dropped. `AGENTS.md` names `simple-vlm-example` as the reference implementation (used as a template), so the default here sets a precedent. Consider defaulting to `""` (disabled) and showing `"agent"` only as a comment example, or call out the behavior change in a `docs/changelog.md` entry. |
| 💡 | [`agent-samples/simple-vlm-example/worker/agent.py:67-71`](https://github.com/NVIDIA/xr-ai/blob/c1685d068af7fac708be8cde27f78d9f40b73768/agent-samples/simple-vlm-example/worker/agent.py#L67-L71) — **The `_STOP_RE` module-level comment is misleading about when this path is reachable.** The comment says "dispatched even when a wake word is configured and absent", but when `wake_word` is empty (the code default), `_strip_wake_word` returns `text` (not `None`), so the `_STOP_RE` branch is never reached — it only fires when a wake word is actively configured. The comment as written implies the regex also guards the no-wake-word path, which is not true. Per `AGENTS.md` comment rules, the comment should state the invariant precisely: something like "only reached when `_wake_word` is set and the transcript failed the wake-word gate." |
| 🔍 | [`agent-samples/simple-vlm-example/worker/agent.py:218-219`](https://github.com/NVIDIA/xr-ai/blob/c1685d068af7fac708be8cde27f78d9f40b73768/agent-samples/simple-vlm-example/worker/agent.py#L218-L219) — **`_strip_wake_word` builds the pattern with the `(?i)` inline flag while `_STOP_RE` uses the `re.IGNORECASE` flag argument.** Both work, but the inconsistency makes it harder to scan. Prefer `re.IGNORECASE` passed to `re.match(pattern, text, re.IGNORECASE)` to match the style used elsewhere in the file. |

- In `simple_vlm_example_worker.yaml`, change `wake_word: "agent"` to `wake_word: ""` and add a comment showing `"agent"` as an example value, so the out-of-the-box experience is unchanged from before this PR.
- In `agent.py:67`, replace the `_STOP_RE` comment with: `# Only reached when _wake_word is set and the transcript failed the gate — lets "stop" interrupt regardless.`
- In `agent.py:218-219`, replace the inline `(?i)` flag with a `re.IGNORECASE` argument to `re.match`: `re.match(pattern, text, re.IGNORECASE)` (and remove `(?i)` from the pattern string).

agent-samples/simple-vlm-example/worker/agent.py
```diff
@@ -7,8 +7,12 @@
 Inputs
 ------
 * Audio chunks (mic):  VAD detects an utterance, STT turns it into text,
-                       which is then dispatched as a query.
+                       which is then dispatched as a query.  If ``wake_word``
+                       is configured, the transcript must start with it
```
yanziz-nvidia: is that how wake word supposed to work? I feel the word detection approach is a very weak way enable whether an agent should ignore or not. Have we tested robustness?

nvddr:
## Review disposition for `216d089`

Mapping each finding from the prior commit (`c1685d0`) to its disposition on this commit. Includes changes that went beyond what reviewers asked for — those came from a session-level reframing of the feature from "wake word" (phonetic detection) to "magic phrase" (deliberate opt-in framing).

### @yanziz-nvidia — design question (the load-bearing one)

> "is that how wake word supposed to work? I feel the word detection approach is a very weak way enable whether an agent should ignore or not. Have we tested robustness?"
> — https://github.com/NVIDIA/xr-ai/pull/155#discussion_r3292064888

You're right that text-prefix matching on STT output is not a robust wake-word engine, and this commit does **not** try to turn it into one. Instead it reframes what the feature is:

- **Renamed `wake_word` → `magic_phrase` everywhere.** The new name is meant to signal "a phrase the user deliberately says to address the agent," not "phonetic wake-word detection." A real wake-word engine (Porcupine, openWakeWord, etc.) is a separate dependency and a separate problem; this PR doesn't pretend to solve it.
- **Robustness scope, stated honestly.** Detection is post-STT string matching. STT errors, homophones ("agent" vs "a gent"), and accented speech will leak through or get rejected. We have not run a robustness study, and we shouldn't claim the gate is reliable against an adversarial speaker — it's a usability gate for cooperating users in noisy rooms, not a security boundary.
- **Why ship this as the gate for the reference sample.** The reference sample's previous behavior (no gate, every utterance dispatched) made it unusable in any room with ambient speech. The opt-in phrase is the smallest change that makes the reference sample demoable in a real environment without taking on a wake-word-engine dependency. If/when robustness matters for a production worker, the right move is a pluggable detector behind the same `magic_phrase` config slot, not deeper matching here.
- **Strict-prefix tightening (this commit).** The prior commit allowed up to 3 leading filler words before the phrase ("hey there agent …"). That undercuts the whole point of the gate — ambient speech containing the phrase mid-sentence would trigger the agent. The new matcher requires the phrase to be the literal prefix.
- **Connection greeting (this commit).** The agent now speaks a one-shot greeting on participant join that names the configured phrase: *"To talk to me, start your question with 'agent'."* Without this, a gated agent is indistinguishable from a broken one.

Concrete asks for you on this thread: (a) does the reframing address your concern enough to unblock this PR as the reference-sample default, with the understanding that production workers needing real robustness will route through a future pluggable detector? Or (b) do you want the gate held out of `simple-vlm-example` entirely until a pluggable detector exists? I'll route based on your call.

### Finding-by-finding table

| Source | Finding | Status at `216d089` | Where |
|---|---|---|---|
| Copilot #1 | `cfg.get("wake_word", "")` returns `None` for empty YAML value; passed into `.strip()` crashes | **Fixed.** Worker now uses `cfg.get("magic_phrase") or ""` and constructor uses `(magic_phrase or "").strip().lower()`. | `simple_vlm_example_worker.py` (cfg.get line); `agent.py:134` |
| Copilot #2 | `wake_word.strip().lower()` in constructor assumes `str`, crashes on `None` | **Fixed.** Defensive `(magic_phrase or "").strip().lower()`. | `agent.py:134` |
| Copilot #3 | Early return on missing wake word skips `_handle_query`, leaves camera on indefinitely | **Fixed.** `_schedule_camera_off(pid)` called before the missing-phrase early return. | `agent.py:206-210` |
| Copilot #4 | Same camera-off issue on "wake word only, no query" early return | **Fixed.** `_schedule_camera_off(pid)` called before that early return too. | `agent.py:213` |
| Copilot #5 | Module docstring says "must start with it" but matcher allowed leading fillers — doc/code mismatch | **Fixed by tightening the code, not the doc.** Matcher is now strict-prefix (no leading fillers); docstring is accurate as written. | `agent.py:11`, matcher at `agent.py:222-244` |
| Copilot #6 | Worker docstring says "ping" is not affected, but spoken "ping" is gated | **Fixed.** Docstring now clarifies that the data-channel "ping" is unaffected; spoken "ping" is gated when a phrase is configured. | `simple_vlm_example_worker.py` module docstring |
| Copilot #7 | YAML comment says "must start with this phrase" but matcher allowed fillers; "ping" claim misleading | **Fixed by tightening the code + rewriting the comment.** YAML comment now says strict prefix, no leading filler, and explicitly notes spoken "ping" is gated while data-channel "ping" is not. | `simple_vlm_example_worker.yaml` (magic_phrase block) |
| yanziz-bot 💡 #1 | YAML defaults `wake_word: "agent"` — breaking change vs. prior "no filter" behavior | **Declined, intentionally.** Reframing this gate as opt-in (`magic_phrase`) is precisely the design choice that says "the sample should not default to always-on." A `""` default would mean the reference sample ships as an always-on agent — the behavior the rename pushes back against. The greeting added in this commit also makes the gate self-explanatory on connect, which mitigates the "users find their queries silently dropped" failure mode that motivated the bot's suggestion. Open to changing this if @yanziz-nvidia disagrees with the framing. |
| yanziz-bot 💡 #2 | `_STOP_RE` module comment misleadingly implies it guards the no-wake-word path | **Fixed.** Comment at `agent.py:73-75` is rewritten to: *"Transcripts matching this pattern bypass the magic-phrase gate so the user can interrupt a response mid-flight without having to start with the configured phrase."* No longer claims it guards the empty-phrase path. |
| yanziz-bot 🔍 #3 | `_strip_wake_word` uses `(?i)` inline flag while `_STOP_RE` uses `re.IGNORECASE` arg — style inconsistency | **Not fixed — deferred as a nit.** The matcher at `agent.py:237-242` still builds its pattern with the inline `(?i)` flag. Functionally equivalent; happy to roll this into a follow-up. Flagging it explicitly here so it doesn't get lost. |

### Beyond reviewer asks (session-directed)

These changes weren't requested by reviewers — they came from a session-level decision about what the feature should be:

- **Rename `wake_word` → `magic_phrase` everywhere** (worker, YAML, agent constructor, module docstrings, log lines). Rationale above.
- **Strict-prefix matcher.** Dropped the 3-leading-filler allowance.
- **Connection-time TTS greeting.** Speaks the configured phrase on participant join, with an example utterance. Falls back to "Ask me anything about what you see" when the phrase is empty.

### Other notes for the board

- **No DCO sign-off on `216d089`.** `CONTRIBUTING.md` requires it; CI does not enforce. Not amending the commit here because a force-push would invalidate "Viewing changes since you last reviewed" for the prior reviewers. Flagging for the board's call on whether to fix-up before merge.

nvddr (approved):

agent-samples/simple-vlm-example/worker/agent.py
```diff
@@ -480,6 +661,88 @@ async def _say(self, pid: str, text: str, pts_us: int) -> None:
...
+        is acted on immediately."""
+        vs = self._voice.get(pid)
+        if vs is not None:
+            async with vs.dispatch_lock:
+                old = vs.current_task
+                if old is not None and not old.done():
+                    old.cancel()
+                    try:
+                        await old
+                    except (asyncio.CancelledError, Exception):
```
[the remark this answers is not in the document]
nvddr: Addressed in d2835de — split the bare except into `except asyncio.CancelledError: pass` (expected, the cancel we just issued) and `except Exception as exc: logger.warning(...)` (unexpected, logged but non-actionable since we want the task gone either way).

agent-samples/simple-vlm-example/worker/agent.py
nvddr: modified a bit, did some testing with wenxin and it's pretty good. THis is definitely needed to avoid noisy environments

PR #156, opened 2026-05-22
docs(readme): mark Docker + NVIDIA Container Toolkit as required prereqs

nvddr:
A fresh-box install of `model_servers` failed for a dev with:

    Error response from daemon: failed to discover GPU vendor from CDI:
    no known GPU vendor found

…because the NVIDIA Container Toolkit (and/or the CDI spec) wasn't
installed. Until now the README only mentioned the toolkit under
"Blackwell-profile prerequisites", which made it look like a
profile-specific extra. In reality every GPU profile (Spark, dual_48G_ada,
96G_blackwell) runs vLLM in `nvcr.io/nvidia/vllm` containers, so the
toolkit is universally required.

Promoted Docker + NVIDIA Container Toolkit into the main "Software"
requirements table, with the install/configure/CDI/smoke-test sequence
that Rohit needed to run inline. The Blackwell prereq line now only
mentions CUDA NVCC, since the toolkit is covered above.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,8 @@ endpoint and no local GPU is required for the agent or hub.
 | Python | 3.11 or 3.12 | 3.10 and 3.13 are not supported |
 | [uv](https://docs.astral.sh/uv/) | latest | dependency manager used by all samples |
 | NVIDIA driver | 570+ | required for local model inference |
+| Docker | 24+ | required: all vLLM-backed services (LLM, VLM) run in `nvcr.io/nvidia/vllm` containers |
+| NVIDIA Container Toolkit | latest | required: gives Docker access to the GPU.  Without it, `model_servers` fails with `failed to discover GPU vendor from CDI: no known GPU vendor found` |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -72,10 +74,34 @@ or virtual-environment setup needed.  If you do not have it:
 curl -LsSf https://astral.sh/uv/install.sh | sh
 ```
 
+The NVIDIA Container Toolkit install is one-time per host:
```
yanziz-nvidia: should point to https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/install-guide.html instead in case of outdated instructions

yanziz-nvidia (approved):

README.md
nvddr: Done — replaced the inline install commands with a pointer to https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/install-guide.html and kept only the one-line GPU smoke-test. `2e20532`.

PR #157, opened 2026-05-22
fix(xr-ai-vad): fix VAD end-of-speech detection and short-command registration

nvddr:
## Summary

- **Remove \`_busy\` gate in \`VadDetector\`** — the flag blocked utterance finalization while STT was in-flight, causing speech + trailing silence to pile up in the buffer until STT returned, appearing as "VAD doesn't detect when I stop speaking." The per-participant \`vs.transcribing\` guard in the agent already handles overlapping STT.
- **Call \`silero.reset_states()\` after each finalization/reset** — Silero's RNN hidden state carries over between utterances and biases silence probability in the next one.
- **Lower \`min_speech\` 0.3 s → 0.1 s in simple-vlm-example** — short words like "stop" (~0.2 s) never crossed the 0.3 s threshold. The VAD kept the utterance buffered but never fired; repeating it caused accumulated \`speech_s\` to eventually cross, which is why multiple "stops" worked. At 0.1 s, a single "stop" registers on the first attempt.

## Test plan

- [ ] All existing VAD unit tests pass (6/6)
- [x] Single "stop" during active TTS response cancels the response
- [x] Longer utterances with natural pauses still transcribe correctly (silence_duration=0.8 s unchanged)
- [x] Second utterance spoken during STT is detected at the correct time, not accumulated until STT returns

### DIFF
--- a/tests/test_xr_ai_vad.py
+++ b/tests/test_xr_ai_vad.py
@@ -42,6 +42,9 @@ class _StubSilero:
         rms = float(np.sqrt(np.mean(arr.astype(np.float32) ** 2))) if arr.size else 0.0
         return 0.9 if rms >= self._threshold else 0.0
 
+    def reset_states(self) -> None:
+        pass
+
 
 def _install_stub(vad: VadDetector) -> None:
     """Replace the loaded silero model with a deterministic stub."""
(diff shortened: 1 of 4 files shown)


wenxind-nvidia (approved):

tests/test_xr_ai_vad.py
```diff
@@ -42,6 +42,9 @@ def __call__(self, tensor, sample_rate: int) -> float:
         rms = float(np.sqrt(np.mean(arr.astype(np.float32) ** 2))) if arr.size else 0.0
         return 0.9 if rms >= self._threshold else 0.0
 
+    def reset_states(self) -> None:
+        pass
```
wenxind-nvidia, line 46: Is this a placeholder?

wenxind-nvidia (approved):

tests/test_xr_ai_vad.py
nvddr: Intentional no-op. The real `OnnxWrapper.reset_states()` zeros out its LSTM hidden state (`_state`), context buffer (`_context`), and last-sample-rate tracking. The stub has none of that — it's stateless — so `pass` is the correct implementation, not a placeholder.

nvddr: No intentionally pass here.

PR #158, opened 2026-05-22
docs(client-sdk): StreamKit native follow-ups C/D/E/F (#134)

### LINKED ISSUE #134: Native C++ StreamKit — integration findings from a partner port

gongxinheng:
## Summary

Ships the four follow-up items I committed to in issue #134 as the native client maintainer. These are the partner-owned tasks from that issue — the design-question items A (AudioSink) and B (CameraConfig contract) were already resolved by NVIDIA in PR #147.

| Issue #134 item | Status | This PR |
|---|---|---|
| A. `AudioSink` companion to `FrameSink` | ✅ Resolved in #147 | — |
| B. `CameraConfig::Facing` contract in C++ | ✅ Resolved in #147 | — |
| **C. `MicrophoneMode` unmapped in C++** | This PR | `AudioConfig.h` docstring |
| **D. FrameSink first-frame lazy publish** | This PR | `FrameSink.h` docstring |
| **E. `token_url` HTTP fetch portability** | This PR | `BackendConfiguration.h` + `LiveKitBackend.h` |
| **F. SDK convention notes** | This PR | `StreamingBackend.h` docstring |

Four small commits — one per item — so each maps cleanly to its issue thread.

### C — `AudioConfig::MicrophoneMode` inert on C++ `LiveKitBackend`

Pure docstring. The iOS / Android / Web backends honour `MicrophoneMode`, mapping it to platform hardware AEC/AGC/NS or WebRTC software DSP. The built-in C++ `LiveKitBackend` does neither — `livekit::AudioSource`'s ctor takes only `(sample_rate, channels, queue_size_ms)`, and the software-DSP path lives in `AudioProcessingModule` with a lifecycle that isn't wired up. Make the contract explicit at the struct level so embedded hosts know to rely on platform/HAL DSP upstream of `AudioSink::InjectAudioFrame`. Same shape as #147's documented `CameraConfig::facing` carve-out.

### D — `FrameSink` first-frame publish contract

Pure docstring. Existing text described first-frame publish as if it were intrinsic to LiveKit's `BufferCapturer` (the iOS shape). The C++ `livekit::VideoSource` ctor requires explicit width/height up front, which aren't known until the host pushes the first frame — so it lands on the same observable rule from the opposite direction. Make "no track exists until the first `InjectVideoFrame()` call" the documented cross-backend contract.

### E — `token_url` HTTP fetch portability

One tiny code change + docstrings. The C++ SDK has no portable HTTP client and forcing one (libcurl, cpp-httplib, Poco::Net) on every consumer is heavy for embedded targets — TLS root stores and DNS resolvers may not be present. The embedded path supplies `LiveKitConfig::token` inline (computed server-side).

  - **Code:** `LiveKitBackend::FetchToken` moves from `private` (non-virtual) to `protected` (virtual). The throw-by-default body stays. Subclass + override is now the documented extension point.
  - **Docs:** `LiveKitConfig::token_url` and `LiveKitBackend::FetchToken` docstrings spell out the contract.

Future opt-in built-in HTTP support (e.g. via a `STREAMKIT_ENABLE_TOKEN_FETCH` CMake option) stays a separate piece of work and out of scope here.

### F — Backend implementation conventions

Pure docstring. Two cross-backend conventions that PR #131 figured out empirically while wrapping LiveKit and that a contributor implementing a non-LiveKit backend (mock, custom WebRTC) needs to know up front:

  - **`kConnecting` synthesis.** `streamkit::ConnectionState` has `kConnecting`, but `livekit::ConnectionState` only exposes `Disconnected / Connected / Reconnecting`. Backends wrapping such an SDK must fire `kConnecting` manually before initiating the network operation, then dedupe the SDK's own `kConnected` against the explicit one.
  - **Lazy process-global init.** SDKs requiring a one-shot global init (`livekit::initialize()`) should perform it lazily on first instance construction behind a `std::once_flag` — not from a static initialiser.

The inline `MyBackend` example is also updated to show the `kConnecting → kConnected` sequence.

## Test plan

- [x] Pure docstring changes for C / D / F — no test impact.
- [x] E moves `FetchToken` to `protected virtual`. Throw-by-default body unchanged; no existing test or call site touches it.
- [x] Verified locally on macOS (AppleClang 17.0) in stub mode (`STREAMKIT_HAVE_LIVEKIT=0`): full build clean, all 6 test binaries pass — `streamkit_mapping_tests`, `streamkit_agent_status_tests`, `streamkit_frame_sink_tests`, `streamkit_audio_sink_tests`, `streamkit_session_tests`, **`streamkit_livekit_backend_tests`** (the one exercising the vtable shape change).

Closes #134.

### DIFF
--- a/client-samples/native/StreamKit/include/streamkit/Backends/LiveKit/LiveKitBackend.h
+++ b/client-samples/native/StreamKit/include/streamkit/Backends/LiveKit/LiveKitBackend.h
@@ -117,6 +117,20 @@ public:
                           int samples_per_channel,
                           int64_t timestamp_us) override;
 
+protected:
+    /// Fetches a LiveKit JWT from `config_.token_url`.
+    ///
+    /// The default implementation throws `TokenFetchFailedError` — the C++
+    /// SDK doesn't ship a portable HTTP client and the embedded path
+    /// supplies an inline `LiveKitConfig::token` instead. Subclass and
+    /// override with whichever HTTP client your target already links
+    /// against (libcurl, cpp-httplib, Poco::Net) when a token endpoint is
+    /// actually required. Expected response format: plain JWT string or
+    /// `{"token":"eyJ…"}`; the SDK appends `?identity=<identity>` to the
+    /// configured URL.
+    virtual std::string FetchToken(const std::string& url,
+                                   const std::string& identity);
+
 private:
     // Forward-declared in the .cpp; subclasses livekit::RoomDelegate and
     // bridges its event callbacks into this backend's on_* event hooks.
@@ -124,10 +138,6 @@ private:
 
     /// Disconnects the room, stops all tracks, fires kDisconnected.
     void TearDown();
-
-    /// Fetches a LiveKit JWT from `config_.token_url`.
-    /// GET <url>?identity=<identity> → plain string or {"token":"eyJ…"}.
-    std::string FetchToken(const std::string& url, const std::string& identity);
 
     /// Maps livekit::ConnectionState → StreamKit's ConnectionState and fires
     /// on_connection_state_changed. Called by the Delegate's event hooks.
--- a/client-samples/native/StreamKit/include/streamkit/Backends/StreamingBackend.h
+++ b/client-samples/native/StreamKit/include/streamkit/Backends/StreamingBackend.h
@@ -59,6 +59,8 @@ namespace streamkit {
 /// class MyBackend : public streamkit::StreamingBackend {
 /// public:
 ///     void Connect(const SessionConfig& config) override {
+///         on_connection_state_changed(ConnectionState::kConnecting);
+///         // … begin async transport setup …
 ///         on_connection_state_changed(ConnectionState::kConnected);
 ///     }
 ///     void Disconnect() override {}
@@ -71,6 +73,25 @@ namespace streamkit {
 ///
 /// auto session = StreamSession(std::make_unique<MyBackend>());
 /// ```
+///
+/// ## Conventions for backend implementations
+///
+/// **Synthesise `kConnecting` if the underlying SDK skips it.** The
+/// `ConnectionState` enum exposes `kDisconnected / kConnecting / kConnected /
+/// kReconnecting / kFailed`. Some SDKs only report two or three of these —
+/// `livekit::ConnectionState` for example has only `Disconnected / Connected
+/// / Reconnecting`. Backends wrapping such an SDK MUST fire
+/// `on_connection_state_changed(kConnecting)` themselves before initiating
+/// the network operation, and deduplicate the eventual `kConnected` if the
+/// SDK fires it after their own explicit fire. Consumers depend on seeing
+/// the connecting → connected edge to drive UI affordances.
+///
+/// **Own your SDK's process-global init lifecycle.** Some SDKs require a
+/// one-shot global initializer before any other API call (LiveKit's
+/// `livekit::initialize()` is the canonical example). Backends should invoke
+/// that lazily on first instance construction — not from a static initialiser,
+/// which produces order-of-init hazards across translation units. A simple
+/// `static std::once_flag` guarded call inside the ctor is sufficient.
 class StreamingBackend {
 public:
     virtual ~StreamingBackend() = default;
--- a/client-samples/native/StreamKit/include/streamkit/Config/AudioConfig.h
+++ b/client-samples/native/StreamKit/include/streamkit/Config/AudioConfig.h
@@ -7,11 +7,30 @@ namespace streamkit {
 
 /// Configures microphone capture passed to StreamSession::StartAudio().
 ///
+/// ## Platform contract for `MicrophoneMode`
+///
+/// `MicrophoneMode` is platform-managed. The iOS, Android, and Web backends
+/// honour it: `kVoiceProcessing` engages platform hardware AEC/AGC/NS;
+/// `kSoftwareProcessing` engages WebRTC's software DSP. The built-in C++
+/// `LiveKitBackend` **ignores `MicrophoneMode`** — its `livekit::AudioSource`
+/// constructor takes only `(sample_rate, channels, queue_size_ms)` with no
+/// DSP knobs, and the SDK's software DSP lives in `AudioProcessingModule`
+/// with a separate lifecycle that isn't wired up. Embedded hosts that need
+/// AEC/AGC/NS should rely on the platform/HAL DSP upstream of
+/// `AudioSink::InjectAudioFrame`.
+///
+/// If a desktop consumer needs in-process software DSP, integrating
+/// `AudioProcessingModule` into the C++ `LiveKitBackend` is a separate piece
+/// of work.
+///
 /// Mirror of Swift `AudioConfig` and Kotlin `AudioConfig`.
 struct AudioConfig {
 
     /// Controls which audio processing pipeline is applied before the PCM
     /// audio is handed to WebRTC.
+    ///
+    /// See the struct-level "Platform contract" note: ignored by the
+    /// built-in C++ `LiveKitBackend`.
     enum class MicrophoneMode {
         /// Platform hardware voice processing (e.g. platform AEC, AGC, NR).
         /// Disable WebRTC's own DSP to avoid double-processing.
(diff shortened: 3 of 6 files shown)


nvddr:

### Overview
Nice clean wrap-up of the four partner-owned follow-ups from #134. The split into one commit per item makes each map cleanly back to its issue thread, the rationale in the PR description is genuinely helpful, and 4-of-5 files are docstring-only with the single code change kept tightly scoped. Overall this reads as low-risk and well-thought-through.

### 🤔 Concerns Worth Discussing
- **Possible `docs/changelog.md` entry for the `FetchToken` visibility change.** `AGENTS.md` directs that "architectural rationale and historical context belong in `docs/changelog.md`, not in source comments" and asks contributors to "record significant new decisions in `docs/changelog.md` (reverse chronological)." Moving `FetchToken` from `private` → `protected virtual` makes it a documented extension point and effectively part of the consumer-facing surface — the same shape PR #147 captured with a changelog entry. Would a one-line entry here feel proportional, or do you read this as small enough to skip? Happy either way — flagging so the human reviewer can make the call.

### 👀 Suggested Areas for Human Reviewer Focus
- `client-samples/native/StreamKit/include/streamkit/Backends/LiveKit/LiveKitBackend.h` (around the new `protected:` block, ~L120-131) — the only behavioural change in the PR. Worth a quick check that (a) no `LiveKitBackend.cpp` internal caller still expects the old private non-virtual signature, and (b) the throw-by-default body remains the right default for overrides to compose against.
- `streamkit_livekit_backend_tests` (named in the PR description as the test that exercises the vtable change) — worth confirming it covers an override path, not just the throw-by-default path, so the new extension contract has a regression anchor.

### 🧭 Consistency with Repository
- The "platform contract" callout pattern in `AudioConfig.h` and `BackendConfiguration.h` mirrors the `CameraConfig::Facing` carve-out PR #147 added — consistent house style for "cross-platform field, inert on one backend."
- Docstring tone leads with the contract (not the design conversation), which lines up with `AGENTS.md`'s comment guidance. The forward-looking sentences ("Two ways forward in C++…", "If a desktop consumer needs in-process software DSP…") read as contract for current readers rather than design narration, so they sit on the right side of the line.
- No `pyproject.toml` touched, so the `DEPENDENCIES.md` requirement doesn't kick in.

---

nvddr (approved):

gongxinheng:
Merged latest `main` in (now includes #159).

On `FetchToken` I kept item E's design — `protected virtual std::string`, so embedded hosts can subclass and override it to return a real token — which intentionally supersedes #159's `[[noreturn]] void` change for this one function. The default body still throws `TokenFetchFailedError`; since the call is now a virtual dispatch it's no longer provably-always-throwing, so the original Sonar finding should clear without the `[[noreturn]]` form.

Verified locally in stub mode (`STREAMKIT_HAVE_LIVEKIT=0`, AppleClang 17, C++20): clean build + all 6 StreamKit test binaries pass.

@yanziz-nvidia — this is the reconciliation you flagged. Happy to add a `NOSONAR` with rationale if Sonar re-raises anything on the virtual default.

PR #159, opened 2026-05-24
fix: resolve Sonar findings in native samples

yanziz-nvidia:
## Summary

- resolve valid Sonar findings across StreamKit native code/tests and the xr-render-demo eval watcher
- make LiveKitBackend destruction non-throwing and clean up byte/string boundaries in native tests and samples
- replace non-secret random participant labels with process-local unique IDs
- modernize C++20 style findings with std::format, std::ranges, using enum, explicit lambda captures, and std::source_location

## Sonar notes

- Leaves `cpp:S5414` on `StreamSession` accepted rather than changing the public callback-field API.
- Leaves `cpp:S5817` on `LiveKitBackend::HandleConnectionStateChange` as a Sonar false positive: the function mutates connection state in real LiveKit builds; the analyzer appears to report the stub-build path.

## Testing

- `bash -n agent-samples/xr-render-demo/eval/eval_watch.sh`
- `clang++ -std=c++20 -Wall -Wextra` compile of touched StreamKit translation units
- manual compile and run of all six native StreamKit test executables
- manual compile of native sample app
- `git diff --check`

`cmake` was not available in the local shell, so I could not run the normal CMake/CTest path.

### DIFF
--- a/client-samples/native/StreamKit/include/streamkit/Backends/LiveKit/LiveKitBackend.h
+++ b/client-samples/native/StreamKit/include/streamkit/Backends/LiveKit/LiveKitBackend.h
@@ -62,7 +62,7 @@ class LiveKitBackend : public StreamingBackend,
                        public AudioSink {
 public:
     explicit LiveKitBackend(const LiveKitConfig& config);
-    ~LiveKitBackend() override;
+    ~LiveKitBackend() noexcept override;
 
     LiveKitBackend(const LiveKitBackend&)            = delete;
     LiveKitBackend& operator=(const LiveKitBackend&) = delete;
@@ -127,7 +127,8 @@ private:
 
     /// Fetches a LiveKit JWT from `config_.token_url`.
     /// GET <url>?identity=<identity> → plain string or {"token":"eyJ…"}.
-    std::string FetchToken(const std::string& url, const std::string& identity);
+    [[noreturn]] void FetchToken(const std::string& url,
+                                 const std::string& identity) const;
 
     /// Maps livekit::ConnectionState → StreamKit's ConnectionState and fires
     /// on_connection_state_changed. Called by the Delegate's event hooks.
@@ -144,7 +145,7 @@ private:
     /// Routes incoming data packets: intercepts "_agent.status",
     /// fires on_data_received for everything else.
     void HandleDataReceived(std::string_view topic,
-                            std::span<const std::byte> payload);
+                            std::span<const std::byte> payload) const;
 
     LiveKitConfig config_;
     SessionConfig session_config_;
--- a/client-samples/native/StreamKit/include/streamkit/Backends/LiveKit/LiveKitBackend.h
+++ b/client-samples/native/StreamKit/include/streamkit/Backends/LiveKit/LiveKitBackend.h
@@ -125,8 +125,8 @@ private:
     /// Disconnects the room, stops all tracks, fires kDisconnected.
     void TearDown();
 
-    /// Fetches a LiveKit JWT from `config_.token_url`.
-    /// GET <url>?identity=<identity> → plain string or {"token":"eyJ…"}.
+    /// Always throws `TokenFetchFailedError` — supply `config_.token`
+    /// instead. This backend ships no HTTP client.
     [[noreturn]] void FetchToken(const std::string& url,
                                  const std::string& identity) const;
 
--- a/agent-samples/xr-render-demo/eval/eval_watch.sh
+++ b/agent-samples/xr-render-demo/eval/eval_watch.sh
@@ -20,6 +20,7 @@ EVAL="$HERE/eval.py"
 WORKER="$HERE/../worker"
 SELF="$(readlink -f "$0")"
 DEBOUNCE_SECS=10
+TIME_FMT='+%H:%M:%S'
 
 last_hash=""
 pending=0
@@ -55,7 +56,8 @@ echo $$ > "$LOCK"
 # editors / language servers / git tools re-save the file without changing
 # bytes (focus changes, refresh-on-blur, etc.).
 file_hash() {
-    sha1sum "$PROMPT" 2>/dev/null | awk '{print $1}'
+    sha1sum "$PROMPT" 2>/dev/null | awk '{print $1}' || true
+    return 0
 }
 
 kill_running() {
@@ -66,17 +68,18 @@ kill_running() {
         kill -TERM -- "-$running_pid" 2>/dev/null || true
         sleep 0.3
         kill -KILL -- "-$running_pid" 2>/dev/null || true
-        echo "  ── aborted at $(date '+%H:%M:%S') ──" >> "$LOG"
+        echo "  ── aborted at $(date "$TIME_FMT") ──" >> "$LOG"
     fi
     wait "$running_pid" 2>/dev/null || true
     running_pid=""
+    return 0
 }
 
 trigger() {
     {
         echo
         echo "═══════════════════════════════════════════════════════════════"
-        echo "  $(date '+%H:%M:%S')  prompt=$PROMPT"
+        echo "  $(date "$TIME_FMT")  prompt=$PROMPT"
         echo "═══════════════════════════════════════════════════════════════"
     } >> "$LOG"
     # Reuse the worker's already-resolved venv so we don't go through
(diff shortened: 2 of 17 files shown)


yanziz-nvidia (commented):

## Summary

Second pass on the Sonar cleanup: drops the file-scope `using` declarations from `test_assert.h` (now scoped inside each test's `main()`), tightens `FetchToken` / `HandleDataReceived` qualifiers, and adds a sequence-counter participant ID. The `return 0` redundancy in `eval_watch.sh` is unresolved; the `FetchToken` docstring still describes a real fetch despite the `[[noreturn]] void` signature.

|     | Finding |
|-----|---------|
| 🚫 | None |
| 💡 | [`LiveKitBackend.h:128-131`](https://github.com/NVIDIA/xr-ai/blob/8f222ab9e1d414265881a88f73032d5b09417d46/client-samples/native/StreamKit/include/streamkit/Backends/LiveKit/LiveKitBackend.h#L128-L131) — **`FetchToken` docstring contradicts the `[[noreturn]] void` signature.** Doc says "GET → plain string or `{"token":"eyJ…"}`"; the implementation always throws and the call site at [`LiveKitBackend.cpp:191`](https://github.com/NVIDIA/xr-ai/blob/8f222ab9e1d414265881a88f73032d5b09417d46/client-samples/native/StreamKit/src/Backends/LiveKit/LiveKitBackend.cpp#L191) discards no value. Reword the docstring to describe the throw-always contract (e.g. "Stub: throws `MissingTokenError`; HTTP fetch not implemented"). |
| 🔍 | [`eval_watch.sh:60,75,92,98`](https://github.com/NVIDIA/xr-ai/blob/8f222ab9e1d414265881a88f73032d5b09417d46/agent-samples/xr-render-demo/eval/eval_watch.sh#L58-L100) — **Trailing `return 0` in `file_hash`/`kill_running`/`trigger`/`cleanup` still present.** Bash returns the last command's exit status automatically. |

- `client-samples/native/StreamKit/include/streamkit/Backends/LiveKit/LiveKitBackend.h:128-131` — Replace the "Fetches a LiveKit JWT… → plain string or `{"token":"eyJ…"}`" docstring with one that matches `[[noreturn]] void` (e.g. "Stub: token-URL fetch is not implemented; always throws `MissingTokenError`").
- `agent-samples/xr-render-demo/eval/eval_watch.sh:60,75,92,98` — Remove the trailing `return 0` lines from `file_hash`, `kill_running`, `trigger`, and `cleanup`.

nvddr:

### Overview
Nice focused Sonar pass — the changes are mechanical and consistent: typed test helpers replacing macros, `[[noreturn]]` on the stub `FetchToken`, `noexcept` dtor, `std::format` / `using enum` / `std::ranges` modernization, and the random→steady-clock+atomic participant ID. Scope is well-contained to StreamKit native + the eval watcher, and the PR description does a great job documenting which Sonar findings are deliberately deferred and why (`cpp:S5414`, `cpp:S5817`). CI is green on `8f222ab`.

### Concerns Worth Discussing
- **`FetchToken` is now an extension point that contradicts its own documentation.** The header docstring still describes a GET-and-parse contract (already flagged by yanziz-reviewer-bot), but the body comment in `LiveKitBackend.cpp` also says "desktop hosts that need a token endpoint can subclass LiveKitBackend and override FetchToken with their preferred HTTP client." With the new `[[noreturn]] void` signature, any override is also forced to be `[[noreturn]]` — the subclass-and-return-a-real-token path described in that comment no longer compiles. Either the signature should be `virtual std::string FetchToken(...)` (keeping the extension story) or that paragraph should be reworded to say "subclasses can populate `LiveKitConfig::token` before construction" or similar. Worth deciding which story is the intended one.

- The PR description notes `cmake` wasn't available locally, so the CMake/CTest path didn't run. CI on this PR shows pytest, CodeQL, SPDX, and `uv lock` jobs — I don't see a C++ build/test check covering StreamKit. Is StreamKit covered by a job I'm missing, or is this PR's compile/test coverage entirely the author's manual `clang++ -std=c++20` runs?

- `client-samples/native/StreamKit/src/Backends/LiveKit/LiveKitBackend.cpp:187-194` — the call site of `FetchToken` after the signature change: `token` is declared but only assigned in the inline-token branch; the `token_url` branch now relies on `FetchToken` always throwing for correctness. Fine today, but couples two files together implicitly. Worth a glance.
- `client-samples/native/StreamKit/src/Backends/LiveKit/LiveKitBackend.cpp:533-538` — the "subclass and override FetchToken" comment described above; please confirm the intended extension story matches the new signature.
- `client-samples/native/StreamKit/src/StreamSession.cpp:18-23` — `MakeBackend` is now a templated-lambda `std::visit` with an `if constexpr` and no terminal `else return`. Fine if `BackendConfiguration` has a single alternative, but worth verifying that's still true (otherwise it's a falling-off-the-end UB hazard for a new variant arm).
- `client-samples/native/StreamKit/include/streamkit/FrameSink.h:115-122` and `LiveKitBackendTests.cpp:62-66` — the new pattern of `auto owned = std::move(data); span_of(owned)` (instead of spanning the rvalue ref directly) is the right answer to Sonar's "rvalue ref param never moved from," but a quick check that the lifetime works as intended for the span is worth a beat.
- `client-samples/native/StreamKit/Tests/StreamKitTests/test_assert.h` — the `SK_EXPECT*` macros are now inline functions in `streamkit::test` with `std::source_location` defaults. Each test now needs `using streamkit::test::Expect;` / `ExpectEq;` inside `main()` — please sanity-check that pattern is intentional vs. exposing them via the header. Also worth a peek that `std::source_location` is available on every toolchain StreamKit needs to support.
- `client-samples/native/StreamKit/src/Backends/LiveKit/AgentStatusParser.h:21` — the `NOSONAR` comment on the `reinterpret_cast` is the only one in the diff; please confirm the suppression rationale ("payload bytes are JSON text") is the right justification.
- The two deferred Sonar findings called out in the PR description (`cpp:S5414` on the callback-field API, `cpp:S5817` claimed-false-positive on `HandleConnectionStateChange`) — both are author judgment calls worth a human sanity-check rather than rubber-stamping.

### Nits
- yanziz-reviewer-bot's two existing nits (FetchToken docstring vs `[[noreturn]] void`; the redundant `return 0` lines in `eval_watch.sh`'s `file_hash` / `kill_running` / `trigger` / `cleanup`) remain open. Neither is blocking — flagging only so they're not silently dropped.

The changes align with `AGENTS.md`'s comment philosophy ("answer 'what non-obvious thing' or 'why isn't the obvious version correct'") — the surviving comments do that work. No `pyproject.toml` / `DEPENDENCIES.md` impact; no SPDX header concerns; no docs-rule impact (StreamKit READMEs aren't touched).

### Deferred to Experts
- LiveKit transport semantics (`HandleConnectionStateChange` Sonar false-positive claim, the stub-vs-real connection-state mutation path) is best validated by someone who has run the real LiveKit build path recently.

---

nvddr:
@yanziz-nvidia Look for conflicts with PR #158 for token change

nvddr:
@gongxinheng Please take a look at these changes and let us know if they are all good. Thanks!

client-samples/native/StreamKit/include/streamkit/Backends/LiveKit/LiveKitBackend.h
```diff
@@ -127,7 +127,8 @@ class LiveKitBackend : public StreamingBackend,
 
     /// Fetches a LiveKit JWT from `config_.token_url`.
     /// GET <url>?identity=<identity> → plain string or {"token":"eyJ…"}.
-    std::string FetchToken(const std::string& url, const std::string& identity);
+    [[noreturn]] void FetchToken(const std::string& url,
```
gongxinheng: nit: Update the doc for the signature change as well

gongxinheng (approved):
LGTM, thanks!

nvddr (approved):

client-samples/native/StreamKit/include/streamkit/Backends/LiveKit/LiveKitBackend.h
yanziz-nvidia: done

yanziz-nvidia:
@gongxinheng please help reconcile PR https://github.com/NVIDIA/xr-ai/pull/158 and see if you'd like to undo the Sonar fix. We can seek other ways if needed. Let me verify Sonar fix first.

PR #160, opened 2026-05-26
XR demo spatial prompt improvements

blongs-nv:
Changes to prompts and new MCP functions for moving objects around in XR demo.

Updates eval to test more things and remove ambiguous/redundant tests.

Previous score: 40/64

With this change: 58/66

### DIFF
--- a/agent-mcp-servers/oxr-mcp/oxr_mcp_server/__main__.py
+++ b/agent-mcp-servers/oxr-mcp/oxr_mcp_server/__main__.py
@@ -367,30 +367,10 @@ def build_mcp(source: PoseSource) -> FastMCP:
         if not pose.get("is_valid"):
             return {"error": "pose unavailable"}
         p = pose["position"]
-        f = pose["forward"]
-        r = pose["right"]
         ox = p["x"] if origin_x is None else origin_x
         oy = p["y"] if origin_y is None else origin_y
         oz = p["z"] if origin_z is None else origin_z
-
-        # Project forward onto the y=0 plane and renormalise so head
-        # pitch/roll don't bleed into horizontal moves. Recover right as
-        # forward × world_up so the basis stays orthonormal.
-        fx, fz = f["x"], f["z"]
-        mag = math.sqrt(fx*fx + fz*fz)
-        if mag < 1e-6:
-            # Looking straight up/down — fall back to right's projection.
-            rx0, rz0 = r["x"], r["z"]
-            mag2 = math.sqrt(rx0*rx0 + rz0*rz0)
-            if mag2 < 1e-6:
-                fx, fz = 0.0, -1.0
-            else:
-                rx0, rz0 = rx0 / mag2, rz0 / mag2
-                fx, fz = rz0, -rx0
-        else:
-            fx, fz = fx / mag, fz / mag
-        rx, rz = -fz, fx  # forward × (0,1,0) gives right in OpenXR Y-up.
-
+        (fx, fz), (rx, rz) = _ground_basis(pose)
         return {
             "x": round(ox + fx*forward + rx*right,        3),
             "y": round(oy + up,                            3),
@@ -399,7 +379,7 @@ def build_mcp(source: PoseSource) -> FastMCP:
 
     def _ground_basis(pose: dict) -> tuple[tuple[float, float], tuple[float, float]]:
         """Return ((fx, fz), (rx, rz)) — pose.forward / pose.right projected onto
-        the y=0 plane and renormalised. Used by place_user_relative."""
+        the y=0 plane and renormalised."""
         f, r = pose["forward"], pose["right"]
         fx, fz = f["x"], f["z"]
         mag = math.sqrt(fx * fx + fz * fz)
--- a/agent-mcp-servers/oxr-mcp/oxr_mcp_server/__main__.py
+++ b/agent-mcp-servers/oxr-mcp/oxr_mcp_server/__main__.py
@@ -18,9 +18,37 @@ Tools (FastMCP, mounted at /mcp)
   position_ahead(distance) → dict {x,y,z}
       World position 'distance' metres along the user's gaze direction.
 
-  position_relative(forward, right, up) → dict {x,y,z}
-      Convert head-relative offsets (metres) to world-space position.
-      forward>0 = ahead, right>0 = right, up>0 = above eye level.
+  position_relative(forward, right, up, origin_x=, origin_y=, origin_z=) → dict {x,y,z}
+      Convert user-frame offsets (metres) to world-space position. Origin
+      defaults to the user's head; pass origin_* to move an existing
+      object user-relatively without doing the vector math yourself.
+
+  place_user_relative(direction, distance) → dict {x,y,z}
+      High-level: world position 'distance' metres in a named direction
+      (front/back/left/right/above/below) from the user. No signs, no origin.
+
+  place_object_relative(origin_x, origin_y, origin_z, direction, distance) → dict {x,y,z}
+      High-level: world position 'distance' metres from an existing object
+      in a named direction (front/back/left/right/above/below/next_to).
+      No signs, no math.
+
+  place_inside_by_id(movee_id, container_x, container_y, container_z) → dict
+      Containment for "put X in Y". Returns {obj_id: movee_id, x, y, z}
+      so the model can feed the result straight into update_primitive
+      without picking which noun's coords to use.
+
+  displace_object(current_x, current_y, current_z,
+                  right=0.0, up=0.0, forward=0.0) → dict {x,y,z}
+      User-frame displacement of an existing object. Add user-frame
+      (right/up/forward) metres to the object's current world position.
+      Handles multi-axis moves in one call ("up and to the left").
+
+  displace_objects(object_ids, current_xs, current_ys, current_zs,
+                   right=0.0, up=0.0, forward=0.0) → dict
+      Batch displacement: same user-frame delta applied to N objects in
+      one call. Returns {items: [{obj_id, x, y, z}, ...]} so the model
+      can fan out to N update_primitive calls without re-deriving each
+      new position.
 
   get_health() → dict
       {status, session_open, open_attempts, last_open_error}.
@@ -37,6 +65,7 @@ import threading
 import time
 from dataclasses import dataclass
 from pathlib import Path
+from typing import Literal
 
 import uvicorn
 import yaml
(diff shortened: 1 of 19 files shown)


nvddr:
🤖 **Automated Summary Review** _(bot reviewer — please verify with a human)_

### Overview
Nice piece of work — the eval delta (40/64 → 58/66) is meaningful and the structure is sensible: a new pure-math `vec-mcp`, richer named-direction helpers on `oxr-mcp`, a substantially expanded prompt with many worked examples, and an eval harness that both grows coverage and adds a "reserved vocabulary" audit to discourage train-on-test leakage. Tools, prompt, and eval moved together — the contract the repo asks for.

### 🚧 Potential Blockers
- **PR is `CONFLICTING/DIRTY` against `main`.** GitHub reports merge conflicts; please rebase onto current `main` before a human reviewer dives in — the actual mergeable diff will likely look different.
- **DCO sign-off missing on all 5 commits.** `CONTRIBUTING.md` is explicit: *"Any contribution which contains commits that are not Signed-Off will not be accepted."* All 5 commit bodies are empty (no `Signed-off-by:` line). Consider an interactive rebase with `git commit --amend -s` (or `git rebase --signoff`) before pushing the rebased branch.

### 🤔 Concerns Worth Discussing
- **Docs-in-same-commit rule (AGENTS.md).** AGENTS.md is firm that *"a change is not done until the docs reflect it"* and that significant decisions go in `docs/changelog.md` (reverse chronological). `DEPENDENCIES.md` and `eval/README.md` are updated, but the following weren't touched — worth confirming whether they should be:
  - `docs/xr-render-demo.md` — the sample's deep-dive doc. A brand-new MCP server in the process stack and a redesigned spatial-tool surface seem in-scope.
  - `docs/changelog.md` — a new MCP-server package + major prompt redesign feels like a "significant decision" by the repo's bar.
- **Greenfield `vec-mcp` package — overall posture looks good.** Pure-math, deterministic, no network egress, SPDX headers on every file, dependency set is minimal (`uvicorn[standard]`, `fastmcp`, `pyyaml`) and aligned with the other MCP servers' dep policy in `DEPENDENCIES.md`. Default bind `0.0.0.0:8250` matches the convention used by `oxr-mcp` and friends, so not a new exposure. Nothing jumps out as a red flag; calling it out only because greenfield packages deserve a second pair of eyes.

### 💡 Nits (all optional)
- `oxr-mcp/__main__.py`: the `_ground_basis` helper is defined inside `build_mcp()` *after* `position_relative`, which is the first caller. Python closure name resolution makes this work at call time, but readers usually expect helpers above their first use — flagging only as a style nit.
- `eval/eval.py:_local_along_direction`: `import math` is inline inside the function. Moving it to the file's import block would match the rest of the module.

### ❓ Questions / Unclear Points
- `oxr-mcp.place_inside_by_id` does no math — it echoes the container's coords back with `obj_id = movee_id`. Was the rationale to give the model a single composable tool call to avoid the "which noun's coords?" disambiguation failure mode? The docstring hints at this; a sentence in the PR description on why this is a tool rather than a prompt rule would help the human reviewer evaluate it.
- The `_recent_moves` log in `processors.py` is capped at 5 and snapshots happen per-turn from the scene block — is 5 the right horizon for "put it back" / "revert" semantics, or did you find longer windows confused the model in practice?

### 👀 Suggested Areas for Human Reviewer Focus
- `agent-mcp-servers/vec-mcp/vec_mcp_server/__main__.py` (entire file) — new package; tools are small but read carefully for tool-name / arg-name stability, since these become part of the agent's prompt-visible API surface.
- `agent-mcp-servers/oxr-mcp/oxr_mcp_server/__main__.py:325-440` — the five new `place_*` / `displace_*` tools and the `_ground_basis` helper. Worth verifying the sign conventions for `place_object_relative` (front→toward user) match what the prompt describes; this is a common place for off-by-180 bugs.
- `agent-samples/xr-render-demo/worker/processors.py:483-501` — new JSON-in-response sanitizer that rewrites the assistant turn to `"Done."` when it looks JSON-shaped. Reasonable defensive measure, but please check it doesn't strip legitimate replies that happen to start with `{` (e.g. quoted strings, code blocks the user asked for).
- `agent-samples/xr-render-demo/worker/processors.py:670-770` — pre-move snapshot + `_recent_moves` write path on `update_primitive`. The same-turn snapshot update (`self._pre_move_positions[obj_id] = new`) is subtle.
- `agent-samples/xr-render-demo/worker/prompts/system.txt` (+1063 lines) — too large to inline-review meaningfully; recommend reading it as a self-contained spec and cross-referencing each new tool's worked examples against the actual tool signatures in `vec-mcp` / `oxr-mcp`. The `_EVAL_VOCAB_COLORS` / `_EVAL_VOCAB_SHAPES` audit in `eval.py` should catch the most likely leakage cases, but spot-check a few worked examples for memorisation patterns.
- `agent-samples/xr-render-demo/eval/eval.py` (new cases + `_local_*` mirrors) — confirm each `_local_*` function matches its server-side counterpart bit-for-bit (rounding, error shapes). Worth a `diff`-by-eye between e.g. `_local_along_direction` and `vec_mcp_server.along_direction`.
- `agent-sdk/xr-ai-models/xr_ai_models/presets/nemotron3_nano.py` — `default_extras.chat_template_kwargs.enable_thinking=False` change. Worth verifying with someone who's loaded that preset elsewhere that it doesn't quietly regress other callers who relied on the previous thinking-on default.

### 🧭 Consistency with Repository
- SPDX headers present on every new file ✓ (`CONTRIBUTING.md` REUSE rule).
- `DEPENDENCIES.md` updated in the same PR ✓ (AGENTS.md hard rule).
- `vec-mcp` follows the established MCP-server layout (FastMCP-only, no API surface beyond `/mcp`, no shared-runtime imports) — matches the `agent-mcp-servers/` pattern in AGENTS.md.
- Eval cases added/updated alongside prompt rules ✓ — addresses AGENTS.md's *"a rule without a case is unverified"* directive. The new vocabulary-leakage audit in `eval.py` operationalises the *"don't train on the test set"* rule, which is a nice systemic improvement beyond the immediate change.
- Gaps: docs-in-same-commit (see Concerns above) and DCO sign-off (see Blockers).

### 🙋 Deferred to Experts
- **Prompt engineering / eval methodology:** someone with context on the xr-render-demo agent's failure modes should sanity-check that the new prompt's added rules are the right rules (not just that they raise the eval score). Eval-driven prompt tuning can over-fit to the harness even with the vocab audit.

---

agent-mcp-servers/vec-mcp/vec_mcp_server/__main__.py
```diff
@@ -0,0 +1,183 @@
...
+
+
+def build_mcp() -> FastMCP:
+    mcp = FastMCP("vec-mcp")
+
+    @mcp.tool()
+    async def between_anchors(
+        ax: float, ay: float, az: float,
+        bx: float, by: float, bz: float,
+    ) -> dict:
```
nvddr, line 81: Argname convention here (`ax, ay, az, bx, by, bz`) diverges from `world_offset` and `along_direction` below, which use prefixed `origin_x/origin_y/origin_z` (and `target_x/target_y/target_z`). Two prompt schemas in one server makes the LLM's job harder.

Stronger suggestion: take positions as `{x, y, z}` dicts across all four tools. Every consumer is an LLM doing tool chaining, and sibling MCPs (oxr-mcp, etc.) already return `{x, y, z}` shaped objects — so the model can pipe outputs straight in without unpacking 3 scalars. The tool surface shrinks too (2-3 args per call instead of 6-7).

Sketch:

```python
@mcp.tool()
async def between_anchors(a: dict, b: dict) -> dict:
    return _round3((a["x"] + b["x"]) / 2,
                   (a["y"] + b["y"]) / 2,
                   (a["z"] + b["z"]) / 2)

@mcp.tool()
async def world_offset(origin: dict, dx: float = 0.0,
                       dy: float = 0.0, dz: float = 0.0) -> dict:
    return _round3(origin["x"] + dx, origin["y"] + dy, origin["z"] + dz)

@mcp.tool()
async def along_direction(origin: dict, target: dict,
                          distance: float = 0.5) -> dict:
    ...
```

FastMCP accepts nested object args via Pydantic, so this is a tool-surface rename, not a protocol change.

blongs-nv: Applied the first change, the second one caused some eval case failures as the prompts need reworking around that version. I'm looking into that but it could be a followup change so it doesn't block this.

blongs-nv:
Updated for comments.

agent-mcp-servers/oxr-mcp/oxr_mcp_server/__main__.py
```diff
@@ -291,43 +320,324 @@ async def position_ahead(distance: float = 1.5) -> dict:
...
+        preference to position_relative + world_offset when placing or moving
+        relative to an existing scene object.
+
+        Direction semantics:
+          front  → toward the user (closer to user's head along their gaze).
+          back   → away from the user (further along their gaze).
+          right  → user's right side at the object's location.
+          left   → user's left side at the object's location.
+          above  → world +Y from the object.
+          below  → world -Y from the object.
```
nvddr: Docstring is misleading here. `front` is computed as **opposite of the user's gaze vector applied at the object's origin**, not as "toward the user" in the sense of `normalize(user_pos − object_pos)`. The two coincide only when the user is gazing directly at the object — they diverge otherwise.

Concrete failure mode: user at `(0, 1.6, 0)` facing `-Z`, cube placed **behind** the user at `(0, 1.6, +3)`. `place_object_relative(cube, "front", 0.5)` computes `(−fx, −fz) = (0, +1)` and lands the new point at `(0, 1.6, +3.5)` — further from the user, not "toward" them. The user-mental-model answer would be `(0, 1.6, +2.5)` (between user and cube).

Sideways gaze has the same issue: user facing `+X`, cube to the side, "front of cube" lands west of the cube along the gaze axis rather than between user and cube.

The actual semantic of this whole tool is *user-frame direction applied at the object's origin*, not *object-frame placement*. `right`/`left`/`above`/`below` are gaze-axis-independent so they're robust; `front`/`back` only do what the docstring claims when the user is looking at the object.

Suggested wording update, no code change:

```
  front  → on the side of the object facing OPPOSITE the user's gaze.
           Coincides with "toward the user" only when the user is
           looking at the object; if the user is gazing away from it,
           this points further along the gaze direction (away from
           "between user and object"). For a true toward-user vector,
           use vec-mcp.along_direction(origin=obj, target=user_pos).
  back   → on the side of the object further along the user's gaze
           direction. Same caveat as `front` when the user is not
           looking at the object.
```

Naming-wise, a longer-term cleanup might rename this to `place_user_frame_at_origin` (or split into a true `place_relative_to_user_from_object` for the toward-user vector and keep this one for user-frame-at-origin) — but a docstring fix is enough for this PR.

agent-samples/xr-render-demo/worker/processors.py
```diff
@@ -361,9 +391,17 @@ async def _drain(self) -> None:
...
+            # starting with "{" passes through.
             if response:
-                self._history.append((text, response))
+                stored = response
+                if _looks_like_leaked_tool_call(response):
+                    logger.warning(
+                        "response looks like a leaked tool call, sanitizing for history: {!r}",
+                        response[:120],
+                    )
+                    stored = "Done."
```
nvddr: The sanitizer is asymmetric: leaked tool-call JSON is stripped from `_history` (line 403: `stored = "Done."`) but the raw JSON is still sent to the user verbatim a few lines below (`await self._send(send_pid, response, ...)`).

The comment above (line 394) says "Strip leaked tool-call JSON from the surface reply", which implies user-facing strip — but the actual code only protects history.

Two intentions are plausible and the choice should be explicit:

1. **History-poisoning prevention only.** The user CAN see the model's failure mode (debug visibility); only future turns are shielded from being primed by the leaked envelope. If this is the intent, the comment at line 394 is misleading — "surface reply" should say "conversation history".

2. **True user-visible strip.** A leaked tool call is noise to the user too. The send at line 409 should use `stored` (or compute `display_text = "Done." if leaked else response`) and push that to both `_send` and the downstream `TextFrame`.

Recommend picking one and aligning the comment + code. If (2) is the intent, something like:

```python
if response:
    display = response
    if _looks_like_leaked_tool_call(response):
        logger.warning("leaked tool call sanitized: {!r}", response[:120])
        display = "Done."
    self._history.append((text, display))
    if len(self._history) > self._history_max:
        self._history.pop(0)

if display and send_pid:
    await self._send(send_pid, display, topic=_AGENT_RESPONSE_TOPIC)
    await self.push_frame(TextFrame(text=display), FrameDirection.DOWNSTREAM)
```

agent-samples/xr-render-demo/worker/prompts/system.txt
```diff
@@ -1,50 +1,984 @@
...
+    with a single shared move.
+
+  OBJECT-RELATIVE placement ("behind the cube", "above the sphere",
+                              "next to the box"):
+    → place_object_relative(origin_x, origin_y, origin_z, direction, distance)
+    origin_* = the named anchor object's CURRENT x, y, z, looked up by id
+               and color/type in the SCENE OBJECTS block. NOT from
+               [Recent moves] — that block is ONLY for undo.
+    direction is ONE of: "front" "back" "left" "right" "above" "below" "next_to"
+    distance is ALWAYS a positive number (≥ 0).
```
nvddr, line 268: This prompt hard-codes the entire tool surface — exact arg names (`origin_x`, `origin_y`, `origin_z`, `current_xs`, `direction=`), positional ordering, enum values, even default-distance hints. Every time a tool signature changes in `oxr-mcp`, `vec-mcp`, `render-mcp`, the prompt has to be edited in lockstep or the model will emit calls against the old signature. We already have one place where this is enforced: MCP tool descriptions (registered via `@mcp.tool()` docstrings). Duplicating them here means two sources of truth that will drift.

The prompt should be doing **routing**, **disambiguation**, and **conversation policy** — not re-stating tool signatures. Concretely:

- ✅ keep: pronoun resolution rules, anchor-selection logic ("the cube" vs "the one in front of me"), "scene block is authoritative", `ONE call per request`, response-style rules, the worked examples that show how to chain tools.
- ❌ drop: tool signature lines (e.g. `→ place_object_relative(origin_x, origin_y, origin_z, direction, distance)`), explicit arg-name enumerations, enum-value listings — the MCP tool descriptions already carry these.

Suggest opening a follow-up issue: **"Make xr-render-demo system prompt less tool-specific; rely on MCP tool descriptions for signatures."** Scope of that issue:

1. Audit `system.txt` and identify every line that names a tool arg, enum value, or default — list them.
2. For each, decide whether the tool's docstring already covers it (delete from prompt) or whether the rule is genuinely cross-cutting (keep but rephrase without arg names).
3. Re-run the eval to confirm the trimmed prompt doesn't regress on the 58/64 baseline.

Not a blocker for this PR — the eval delta is the headline and gutting the prompt now would risk it. But worth tracking before the next round of tool changes (`vec-mcp` argname refactor, `place_object_relative` semantic clarifications) forces a third edit of the same content.

nvddr (approved):

agent-samples/xr-render-demo/worker/prompts/system.txt
blongs-nv: OK, will look into this for follow-up.

agent-samples/xr-render-demo/worker/processors.py
blongs-nv: Fixed with 2.

agent-mcp-servers/oxr-mcp/oxr_mcp_server/__main__.py
blongs-nv: Updated.

PR #161, opened 2026-05-28
Wenxind/glasses additional

wenxind-nvidia:


yanziz-nvidia (commented):

## Summary

Adds `glasses-agent-nat` (NeMo Agent Toolkit variant of the smart-glasses assistant), a gitignored `glasses-agent-vsop` VSOP experiment, YAML configs for `glasses-agent`, `ask_frames` multi-image tool to `vlm-mcp` and `xr-ai-models`, and updates README/AGENTS/DEPENDENCIES/changelog accordingly. `mergeable` and `mergeStateStatus` are both UNKNOWN (PR is ~2 weeks old and likely stale against main — recommend rebasing before merge). No CI result visible from the API.

|     | Finding |
|-----|---------|
| 🚫 | None |
| 💡 | [`agent-mcp-servers/vlm-mcp/vlm_mcp_server/__main__.py:288`](https://github.com/NVIDIA/xr-ai/blob/f562cee3830f9ddadd35a74e8d417db027d09b7e/agent-mcp-servers/vlm-mcp/vlm_mcp_server/__main__.py#L288) — `if not str(p)` is a dead check. `pathlib.Path(s)` never produces a path whose `str()` is falsy for any non-empty input string, and an empty string would have been caught by the caller before reaching this point. The guard gives a false sense of safety and will never fire. Replace with an explicit check on the original string before constructing the `Path`: `if not p.strip(): return "ask_frames: image path is empty."` applied to the raw string elements. |
| 🔍 | [`DEPENDENCIES.md`](https://github.com/NVIDIA/xr-ai/blob/f562cee3830f9ddadd35a74e8d417db027d09b7e/DEPENDENCIES.md) — The `glasses-agent-vsop` table lists `glasses-agent-nat` / `glasses-agent-nat-worker` as the package names. The inline note acknowledges this is intentional until cleanup, which is fine — but the note belongs in the table row itself (or a `> Note:` below it) rather than buried after the service list so a future reader knows at a glance that the table is deliberate rather than a copy-paste error. |

- `agent-mcp-servers/vlm-mcp/vlm_mcp_server/__main__.py:288` — Replace `if not str(p):` with a pre-construction check on the raw string: iterate over the original `image_paths` list and return the error early if any element is empty before building `pathlib.Path` objects.
- `DEPENDENCIES.md` (glasses-agent-vsop table) — Move the "Package names intentionally remain the copied NAT names" note into a `> Note:` line immediately after the dependency table so it is visible inline rather than after the service list.

PR #162, opened 2026-05-29
Pin CloudXR and LOVR to the GPU with the most free VRAM

blongs-nv:
On multi-GPU systems this make the CloudXR and sample app start up on the GPU with the most free VRAM.

Instead of autodetect we could do it via config as in https://github.com/NVIDIA/xr-ai/pull/163 (only merge one of the changes).


nvddr:
Does this actually work? From what I remember cloudxr has a particular GPU it will run on? Something to do with vulkan device/cuda device? I don't remember...

blongs-nv:
> Does this actually work? From what I remember cloudxr has a particular GPU it will run on? Something to do with vulkan device/cuda device? I don't remember...

It fixed OOMs for me. I can't run XR sample without this right now.

PR #166, opened 2026-06-01
feat(conversation): introduce ConversationLoop foundation for shared STT/TTS pipeline

nvddr:
## Summary

Adds the `xr-ai-conversation` SDK package: a shared voice-conversation loop that wires the audio → VAD → STT → voice gate → user `on_query` → TTS → return-audio path so each worker only supplies a single callback.

- New package `agent-sdk/xr-ai-conversation/` exporting `ConversationLoop`, `VadConfig`, `GateBindings`, `QueryHandler`, `QueryResult`, `VoiceState`, `stream_text_to_audio`, and the audio codec helpers (`int16_pcm_to_wav`, `wav_to_chunks`, `chunks_to_wav`, `split_sentences`, `now_us`).
- Owns per-pid VAD lifecycle, `dispatch_lock` + `current_task` for in-flight cancel + `flush_return_audio`, sentence-batched streaming TTS (parallel synth + queued sender preserves audio order), and participant join/leave hooks.
- `on_query` returns either `str` (one-shot reply) or `AsyncIterator[str]` (streamed tokens). Both paths also send the assembled text on a configurable `text_topic` (default `agent.response`).
- 12 new unit tests run on CPU only (no real LiveKit / VAD / STT / TTS); 54 existing voicegate tests still pass.

## Builds on / merge order

- **Base branch: `devdeepr/voicegate-extract` (PR #164).** This PR cannot merge until #164 lands because `xr-ai-conversation` depends on `xr-ai-voicegate`, which lives on that branch.
- No sample migrations in this PR — those land as follow-ups:
  - PR-B: migrate `agent-samples/simple-vlm-example` worker to `ConversationLoop`.
  - PR-C: migrate `agent-samples/xr-render-demo` worker + move shared codec helpers out of `xr-ai-pipecat/audio.py`.

## Implementation notes worth review attention

- `pyproject.toml` uses `path = ...` editable workspace deps, matching the in-tree precedent (`agent-sdk/xr-ai-models`, `agent-sdk/xr-ai-pipecat`) rather than the spec's `workspace = true` shorthand.
- `_handle_stop` order is **cancel + flush + `gate.say_stop_ack` + `on_stop_extra` + data echo**, deliberately differing from the original `simple-vlm-example` which echoed the data message before the stop-ack TTS — this matches PR-A's design.
[...]

PR #167, opened 2026-06-01
feat(simple-vlm-example): migrate to ConversationLoop; brain becomes camera+VLM only

nvddr:
## Summary

PR-B of the voice-pipeline extraction. Replaces the hand-rolled VAD wiring + audio-in callback + STT call + gate.feed + streaming-TTS-out + per-pid cancel/flush bookkeeping in `simple-vlm-example/worker/agent.py` with a single `xr_ai_conversation.ConversationLoop` instance. The sample's brain shrinks to the camera-on-demand state machine, frame tracking, the VLM call, and the data-channel handler.

**Builds on:**
- #164 (`xr-ai-voicegate`)
- #166 (`xr-ai-conversation`) — must merge first

This PR cannot merge until both of the above land.

## Line-count delta

`agent.py` drops from **663 → 432** lines (-231). Engineering-lead's target was ~400-420; landed 12 lines over because the per-pid camera off-scheduling that the original called from four sites is now wired through three brain hooks (`on_stop_extra`, `on_phrase_only`, `on_drop`), each requiring a tiny method. Trimming further would mean inlining the hooks, which obscures the contract.

`simple_vlm_example_worker.py` grew from 161 → 197 lines (+36) to cover the brain + loop wiring + custom greeting + dispatch back-reference assignment.

Two files removed entirely: `worker/voice.py` (loop owns `VoiceState`) and `worker/audio.py` (loop's `xr_ai_conversation.audio` covers both `int16_pcm_to_wav` and `wav_to_chunks`).

## Stop-order behavior

Preserved from PR-A's design: `cancel → flush → say_stop_ack → on_stop_extra → data echo`. The original sample echoed the data message **before** the stop-ack TTS; PR-A's loop runs the stop-ack first so the canned audio starts playing while the data echo travels.

## Surprises discovered during migration

1. **Loop extension needed**: PR #166 didn't expose hooks for `on_phrase_only` / `on_drop`, but the original sample scheduled `camera_off` from both sites — without these, the speculative camera-on from `on_speech_start` has no counterbalancing teardown when the user says the magic phrase without following up, or when the gate drops a non-matching utterance. This PR adds `on_phrase_only_extra` / `on_drop_extra` kwargs to `ConversationLoop.__init__` (small extension, 22 lines), wires them through after the loop's own handlers, and grows the loop's test suite from 12 → 13 cases.

2. **`handle_query` shape**: returns `str` for fast-fail paths (camera unavailable / frame data unavailable) and an inner async generator for the VLM stream. The cancellation envelope wraps both: outer `try/except CancelledError` around frame acquisition (so a cancel during `_wait_for_camera_frame` still runs cleanup), inner generator catches `CancelledError` AND `GeneratorExit` (the gc/aclose path) so the `_camera_held` flag and banner always settle.

3. **VLM HTTPError behavior** (judgment call): original yielded a canned "VLM server unavailable — please retry." message on httpx errors. Preserved in the streaming generator, but only emitted when no tokens have been yielded yet — otherwise the partial reply already went out and concatenating the canned message would speak garbage. This is a small behavior delta from the original (which always sent the canned message on httpx error), surfaced here rather than fixed silently.

4. **Greeting**: the loop's default greeting drops "about what you see" because it's sample-specific; passed a custom `greeting=` callable to preserve the original wording.

## Test plan

- [x] `python -c "from simple_vlm_example_worker import main; from agent import SimpleVlmBrain"` — imports clean
- [x] `pytest test_xr_ai_conversation_loop.py test_xr_ai_voicegate.py test_simple_vlm_example_wire.py` — 75 passed
- [ ] **Board boot-test**: cannot boot full sample against real STT/VLM/TTS in worktree (no GPU). Needs verification before merge.
- [ ] Magic-phrase + STOP smoke test against a real client (verify the gate-driven camera-off scheduling fires correctly)
- [ ] Long VLM streaming response + barge-in cancel (verify the inner async generator's `aclose()` path runs the finally block)


nvddr:
Superseded by upcoming pipecat-unified foundation PR. Board redirected: ONE foundation PR with full framework (voicegate + new pipecat processors + factory) + TWO migration PRs (simple-vlm, xr-render-demo). Reconstruction on a fresh branch.

PR #172, opened 2026-06-01
feat(android-streamkit): public StreamSession.injectVideoFrame API for external frame sources

nvddr:
## Motivation

API parity with iOS. iOS StreamKit already exposes `StreamSession.injectVideoFrame(_: CMSampleBuffer)` so clients can push frames from their own pipelines (external camera adapters, screen capture, synthetic frame generators, glasses video, etc.) into a LiveKit local video track. Android had no equivalent — this PR closes that gap with a generic, framework-shaped public API.

## What changed

New public symbols on the canonical Android StreamKit:

- `StreamSession.injectVideoFrame(i420: ByteBuffer, width: Int, height: Int, timestampUs: Long)` — public suspend method that pushes externally-sourced I420 frames into a lazily-created local video track.
- `StreamingBackend.injectVideoFrame(...)` — matching abstraction on the backend interface so non-LiveKit backends can implement it.
- `LiveKitBackend.injectVideoFrame(...)` — LiveKit implementation. Lazy track creation is mutex-guarded; subsequent frames take the lock-free fast path.
- `InjectedVideoCapturer` (new file) — WebRTC `VideoCapturer` subclass that copies I420 planes into a `JavaI420Buffer` via `YuvHelper.copyPlane` and pushes a `VideoFrame` to the capturer observer.

`LiveKitBackend.stopCamera()` and `tearDown()` are updated to unpublish the injected track when present (instead of, or in addition to, the regular camera). Unpublish is allowed to handle the stop+dispose itself — we deliberately do not call `track.stop()`/`track.dispose()` ourselves.

## MediaCodec race fix

A manual `track.stop()` + `track.dispose()` on the injected track during teardown raced the MediaCodec event handler against its own torn-down thread (`IllegalStateException: Handler on a dead thread`). `unpublishTrack` already runs stop + dispose internally when `stopOnUnpublish=true` (the default), so we now let it do that and avoid the double-stop. The mutex around lazy creation + teardown ensures concurrent first-frame and disconnect calls are serialized.

## Attribution

Ported from `nvddr/xr-ai` branch [`wenxind/rayban-ios-working`](https://github.com/nvddr/xr-ai/tree/wenxind/rayban-ios-working), which inlined its own copy of StreamKit. Per `git log` on the fork branch, all four streamkit-touching commits there are authored by `Devdeep Ray <devdeepr@nvidia.com>` — the same identity as the signer of this PR — so no `Co-Authored-By:` line was added.

The fork's branch name references Ray-Ban (the original consumer of the feature). All Ray-Ban-specific naming and comments have been stripped from the canonical port; the API and implementation are generic.

## Test plan

- [ ] Build verification: **unverified** — no JDK/Gradle on the agent host. Static diffs against the fork's framework files show only the expected package rename (`raybansample` → `streamkitsample`) and comment-level Ray-Ban stripping. Same LiveKit SDK version (`io.livekit:livekit-android:2.7.0`) as the fork, so all referenced SDK types (`LocalVideoTrack`, `VideoCaptureParameter`, `JavaI420Buffer`, `YuvHelper`, `VideoCapturer`, `CapturerObserver`) are guaranteed available.
- [ ] Runtime verification: **needs hardware** — actually injecting frames via a sample app requires a wired-up external source (glasses, screen capture, etc.). The fork branch validated this against Meta Wearables DAT; the canonical port does not include any sample-app integration in this PR (scope: framework only).
- [ ] Reviewer sanity: confirm gradle sync + `./gradlew :app:assembleDebug` is green locally.

### DIFF
--- a/client-samples/android/app/src/main/java/com/nvidia/xrai/streamkitsample/streamkit/StreamSession.kt
+++ b/client-samples/android/app/src/main/java/com/nvidia/xrai/streamkitsample/streamkit/StreamSession.kt
@@ -9,6 +9,7 @@ import com.nvidia.xrai.streamkitsample.streamkit.config.AudioConfig
 import com.nvidia.xrai.streamkitsample.streamkit.config.BackendConfiguration
 import com.nvidia.xrai.streamkitsample.streamkit.config.CameraConfig
 import com.nvidia.xrai.streamkitsample.streamkit.config.SessionConfig
+import java.nio.ByteBuffer
 
 /**
  * Transport-agnostic streaming session — the single public entry-point of StreamKit.
@@ -129,6 +130,30 @@ class StreamSession(private val backend: StreamingBackend) {
      */
     suspend fun stopCamera() {
         backend.stopCamera()
+    }
+
+    /**
+     * Pushes a single externally-sourced I420 video frame to the published
+     * video track. The track is created lazily on the first call and reused
+     * for subsequent frames; [stopCamera] tears it down.
+     *
+     * Used by clients that inject frames from their own pipeline (external
+     * camera adapters, screen capture, synthetic frame sources). Mirror of
+     * iOS `StreamSession.injectVideoFrame(_: CMSampleBuffer)`.
+     *
+     * @param i420         Read-only buffer containing Y, U, V planes back-to-back.
+     * @param width        Y-plane pixel width (must be even).
+     * @param height       Y-plane pixel height (must be even).
+     * @param timestampUs  Source-side presentation timestamp, microseconds.
+     * @throws [StreamError.NotConnected]
+     */
+    suspend fun injectVideoFrame(
+        i420: ByteBuffer,
+        width: Int,
+        height: Int,
+        timestampUs: Long,
+    ) {
+        backend.injectVideoFrame(i420, width, height, timestampUs)
     }
 
     // ── Data channel ──────────────────────────────────────────────────────────
--- a/client-samples/android/app/src/main/java/com/nvidia/xrai/streamkitsample/streamkit/StreamSession.kt
+++ b/client-samples/android/app/src/main/java/com/nvidia/xrai/streamkitsample/streamkit/StreamSession.kt
@@ -148,6 +148,9 @@ class StreamSession(private val backend: StreamingBackend) {
      * @param width        Y-plane pixel width (must be even).
      * @param height       Y-plane pixel height (must be even).
      * @param timestampUs  Source-side presentation timestamp, microseconds.
+     *                     Note: not honored downstream — the WebRTC frame
+     *                     timestamp is overridden internally with the local
+     *                     monotonic clock to keep the encoder's PTS stable.
      * @throws [StreamError.NotConnected]
      */
     suspend fun injectVideoFrame(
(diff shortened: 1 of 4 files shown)


wenxind-nvidia (commented):
## Summary

Adds a public `StreamSession.injectVideoFrame(i420, width, height, timestampUs)` API to the Android StreamKit, plumbed through `StreamingBackend` and implemented in `LiveKitBackend` via a new `InjectedVideoCapturer`. Brings Android to parity with iOS. The diff is clean, well-commented, and SPDX headers are present on the new file. Nice work on the mutex-guarded lazy track creation and the documented MediaCodec/`slice()` rationale.

### Medium

- `LiveKitBackend.kt` (`injectVideoFrame`, fast path) — the lock-free fast path reads `injectedCapturer?.let { ... }` while `stopCamera()` and `tearDown()` clear `injectedCapturer`/`injectedVideoTrack` under the mutex. Because the fast path doesn't take the lock, a frame in flight can race a concurrent `stopCamera()`: the capturer can be disposed (via `unpublishTrack`) between the null-check and `pushI420Frame`. `InjectedVideoCapturer.pushI420Frame` largely guards against this (it bails when `started`/`observer` are cleared in `dispose()`), so the worst case is likely a dropped frame rather than a crash — but it's worth a one-line comment stating that the fast path tolerates a concurrent teardown by design, or a `@Volatile` read into a local before use. Please confirm `pushI420Frame` is safe to call after `dispose()` on the capturer (it appears to be, since `observer` is nulled and the early returns cover it).

### Low / Nit

- `StreamSession.kt` / `InjectedVideoCapturer.kt` — `timestampUs` is part of the documented public signature but is intentionally ignored: `pushI420Frame` overrides it with `System.nanoTime()` (well justified in the comment). The KDoc on `injectVideoFrame` still advertises `timestampUs` as the "source-side presentation timestamp" without noting it's not used for the WebRTC frame timestamp. Consider a one-line note on the public API so callers don't expect their PTS to be honored downstream.
- Test plan is honest that build/runtime are unverified (no JDK/Gradle on the agent host). A reviewer with the toolchain should confirm `./gradlew :app:assembleDebug` is green before merge, as noted.

🤖 Automated review via Claude Code — a maintainer should still review.

nvddr:
Addressed @wenxind-nvidia's review in `298325d` (on top of the keep-both merge of `main` that clears the conflict):

|---|---|---|
| **Medium** — `injectVideoFrame` fast path reads/pushes `injectedCapturer` without the mutex while `stopCamera()`/`tearDown()` clear+dispose it under the lock (use-after-dispose race) | **Fixed** — the fast-path null-check + `pushI420Frame` now run inside `injectedVideoMutex.withLock { ... }`, the same mutex teardown takes, so a concurrent dispose can't run mid-push. Chose the lock over the comment-only option for a hard guarantee rather than relying on `pushI420Frame`'s internal early-returns. | [LiveKitBackend.kt L204-L218](https://github.com/NVIDIA/xr-ai/blob/298325dc00ec6fb20d22af1bbb719cc5dec90006/client-samples/android/app/src/main/java/com/nvidia/xrai/streamkitsample/streamkit/backends/livekit/LiveKitBackend.kt#L204-L218) |
| **Nit** — `timestampUs` documented as the source-side PTS but overridden internally with `System.nanoTime()` | **Fixed (public API, per scope)** — added a KDoc note on `StreamSession.injectVideoFrame` that the timestamp is not honored downstream. Scoped to the public surface; the internal `InjectedVideoCapturer.pushI420Frame` KDoc already documents the override in its comment. | [StreamSession.kt L147-L156](https://github.com/NVIDIA/xr-ai/blob/298325dc00ec6fb20d22af1bbb719cc5dec90006/client-samples/android/app/src/main/java/com/nvidia/xrai/streamkitsample/streamkit/StreamSession.kt#L147-L156) |
| **Nit** — build/runtime unverified (no JDK/Gradle on agent host) | **Acknowledged — on-device gate** — no JDK/Gradle here; `./gradlew :app:assembleDebug` should be confirmed green on a toolchain host before merge, as the review notes. | (no-op) |

CI: all 9 required checks COMPLETED/SUCCESS on `298325d`. Merge state: `CONFLICTING` → `MERGEABLE`; now `BEHIND` only because `main` advanced by docs-only #158 (does not touch the Android files). Outstanding gate: review approval (@yanziz-nvidia).

wenxind-nvidia (approved):

PR #179, opened 2026-06-03
fix(streamkit): resolve remaining Sonar findings in LiveKitBackend

yanziz-nvidia:
## What

Resolves the two remaining Sonar findings in `LiveKitBackend.cpp` that survived the earlier sweep (`86d1760`, already on `main`):

1. **Empty `catch (...)` in `~LiveKitBackend`** (`bug`/`code-smell`: handle the exception or document the empty block) — added a comment explaining the deliberate swallow. A destructor must not propagate, and a failed teardown during destruction has no recoverable action.
2. **`PackedFrameSize` enum verbosity** — added `using enum PixelFormat;` and dropped the `PixelFormat::` qualifiers on the switch cases, matching `MapPixelFormat` directly above it.

Verified with `g++ -std=c++20 -fsyntax-only` in stub mode (no LiveKit SDK).

## Findings intentionally *not* addressed (with reasons)

- **`StreamSession.h` — "Don't mix public and private data members."** The public `std::function` hooks (`on_connection_state_changed`, `on_data_received`, `on_agent_status`) are the documented, assign-directly public API (`session.on_data_received = ...`) and mirror the Swift/Kotlin/JS `StreamSession`. The private members (`backend_`, `connection_state_`) are internal state. Satisfying the rule would mean either exposing internals or wrapping the callbacks in setters, both of which degrade the API. Deliberate design — left as-is.
- **`test_assert.h` — "Use `std::print` instead of `fprintf`" (×2).** `std::print` is **C++23**; this project targets **C++20** (`target_compile_features(streamkit PUBLIC cxx_std_20)`), and the tests inherit that standard. The finding is not actionable without a standard bump, which carries real toolchain-support risk (`<print>` availability across libstdc++/libc++). The `fprintf`-to-`stderr` in this zero-dependency test harness is correct and intentional. Happy to revisit if we decide to move StreamKit to C++23.

### DIFF
--- a/client-samples/native/StreamKit/src/Backends/LiveKit/LiveKitBackend.cpp
+++ b/client-samples/native/StreamKit/src/Backends/LiveKit/LiveKitBackend.cpp
@@ -86,17 +86,18 @@ livekit::VideoBufferType MapPixelFormat(PixelFormat fmt) {
 std::size_t PackedFrameSize(int width, int height, PixelFormat format) {
     const auto pixels =
         static_cast<std::size_t>(width) * static_cast<std::size_t>(height);
+    using enum PixelFormat;
     switch (format) {
-        case PixelFormat::kI420:
-        case PixelFormat::kNV12: {
+        case kI420:
+        case kNV12: {
             const auto chroma_w =
                 (static_cast<std::size_t>(width)  + 1) / 2;
             const auto chroma_h =
                 (static_cast<std::size_t>(height) + 1) / 2;
             return pixels + 2 * chroma_w * chroma_h;
         }
-        case PixelFormat::kRGBA:
-        case PixelFormat::kBGRA:
+        case kRGBA:
+        case kBGRA:
             return pixels * 4;
     }
     return 0;  // unreachable
@@ -165,6 +166,8 @@ LiveKitBackend::~LiveKitBackend() noexcept {
     try {
         TearDown();
     } catch (...) {
+        // Swallowed: a destructor must not propagate, and a failed teardown
+        // here has no recoverable action.
     }
 }
 


gongxinheng (approved):
Looks good. Thanks

wenxind-nvidia (approved):

yanziz-nvidia:

## Summary

Resolves two Sonar findings in `LiveKitBackend.cpp`: documents the intentional empty `catch (...)` in the destructor and removes redundant `PixelFormat::` qualifiers in `PackedFrameSize` via `using enum`.

|     | Finding |
|-----|---------|
| 🚫 | None |
| 💡 | [`LiveKitBackend.cpp:89`](https://github.com/NVIDIA/xr-ai/blob/58d557494aacbee5035ce1d18b0b135e65c6bbe0/client-sdk/src/backends/livekit/LiveKitBackend.cpp#L89) — `return 0; // unreachable` at the end of `PackedFrameSize` suppresses `-Wswitch-enum`; a new `PixelFormat` enumerator would silently return 0-byte frames. Replace with `default: assert(false && "unhandled PixelFormat"); return 0;`. |
| 🔍 | [`LiveKitBackend.cpp:167`](https://github.com/NVIDIA/xr-ai/blob/58d557494aacbee5035ce1d18b0b135e65c6bbe0/client-sdk/src/backends/livekit/LiveKitBackend.cpp#L167) — If a non-throwing logger is available in destructor scope, log before the empty catch body so silent teardown failures are diagnosable. |

- `LiveKitBackend.cpp:89` — Replace trailing `return 0; // unreachable` with `default: assert(false && "unhandled PixelFormat"); return 0;`.
- `LiveKitBackend.cpp:168` — Add a non-throwing log call before the empty catch body.

PR #180, opened 2026-06-03
Remove on-demand camera mode (always-on streaming)

wenxind-nvidia:
## What

Removes the "camera on demand" feature across the entire repo. Clients now stream the camera in **always-on** mode only, and the agent no longer sends `startCamera`/`stopCamera` control signals on the `clientControl` topic.

## Why

The on-demand path added significant complexity (speculative VAD warmup, frame-freshness gating, grace-period stop timers, agent→client signalling, and a per-client toggle on every platform) for a mode that is no longer wanted. Always-on streaming is simpler and more reliable: the agent just consumes the latest frame.

## Platforms touched

- **Server sample (`agent-samples/simple-vlm-example`)** — `SimpleVlmAgent` no longer requests/holds/releases the camera. Removed the `frame_max_age_s`, `camera_on_timeout_s`, `camera_grace_s` config (worker, yaml, docstrings) and all `clientControl` signalling, VAD warmup, freshness gating, and grace-period off timers. `_handle_query` now grabs the latest frame directly; if none exists yet it replies "Camera unavailable, please try again."
- **Android** — removed the `cameraOnDemand` state and "On demand" toggle; `clientControl` is silently dropped.
- **iOS/visionOS** — removed the `cameraOnDemand` property, its persisted `UserDefaults` key, the toggle, and the `sendPing` warmup; `clientControl` is silently dropped.
- **Web + web-xr** — removed the `cameraOnDemand` model field, checkbox UI, render sync, and change listener; `clientControl` is still silently dropped.
- **Docs** — added a `docs/changelog.md` entry (2026-06-03) recording the removal. Historical entries documenting when the feature was added are preserved.

## Note

The manual `startCamera`/`stopCamera` user controls (camera button, StreamKit backends, `StreamSession`) remain unchanged — only the agent-driven on-demand behaviour is removed.

## Verification

- `python3 -m py_compile` passes on the edited worker files.
- Repo-wide grep confirms no remaining `cameraOnDemand`, `camera_on_timeout`, `camera_grace`, `frame_max_age`, `_frame_events`, `_ensure_camera_on`, `_schedule_camera_off`, or `_wait_for_camera_frame`. Remaining `startCamera`/`stopCamera` references are the manual user controls; remaining `clientControl` references are the silently-dropped topic branches.

### DIFF
--- a/agent-samples/simple-vlm-example/worker/agent.py
+++ b/agent-samples/simple-vlm-example/worker/agent.py
@@ -128,6 +128,14 @@ DEFAULT_SYSTEM_PROMPT = (
     "and say nothing else."
 )
 
+# A frame whose latest signal is older than this is treated as "camera
+# unavailable" rather than answered against. With always-on streaming live
+# frames arrive continuously (tens of ms apart), so this only trips when the
+# user has manually stopped the camera and the last signal lingers in
+# ``_latest`` — without it, ``request_frame`` returns an 8x8 placeholder for
+# the dead track and the VLM would answer about a blank image.
+_FRAME_MAX_AGE_US = 5_000_000  # 5s
+
 
 class SimpleVlmAgent:
 
@@ -351,7 +359,9 @@ class SimpleVlmAgent:
         status = "done"
         try:
             sig = self._latest_signal(pid)
-            if sig is None:
+            if sig is None or not self._is_fresh(sig):
+                # No signal, or only a stale one left behind by a manually
+                # stopped track — answering would send an 8x8 placeholder.
                 await self._say(pid, "Camera unavailable, please try again.", pts_us)
                 return
 
@@ -601,6 +611,9 @@ class SimpleVlmAgent:
         # seq restarts from 1 on each camera restart, so the old track's
         # stale entry wins max(seq) for hundreds of frames on the new track.
         return max(candidates, key=lambda s: s.pts_us)
+
+    def _is_fresh(self, sig: FrameSignal) -> bool:
+        return now_us() - sig.pts_us < _FRAME_MAX_AGE_US
 
     async def _on_participant(self, event: ParticipantEvent) -> None:
         if event.joined:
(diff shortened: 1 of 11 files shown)


nvddr:
This PR deletes the agent→client `clientControl` camera-control protocol across the stack: the `SimpleVlmAgent` sample agent and four client platforms (Android, iOS/visionOS, web, web-xr). Structurally, a cross-cutting signalling channel — where the agent drove `startCamera`/`stopCamera` based on VAD warmup, frame-freshness gating, and grace-period timers — is removed, while the `clientControl` topic-handling branch is retained on every client as a silent no-op (so unknown control messages still never leak into the UI message list). All per-platform "camera on demand" state, toggles, and persistence are dropped; manual user camera controls and the underlying StreamKit `startCamera`/`stopCamera` SDK methods are preserved unchanged. `_handle_query` is simplified to consume the latest cached `FrameSignal` directly.

| Category | Finding | Location |
|---|---|---|
| Potential bug | The old `_handle_query` guarded the frame with `_is_fresh`, so a stale `FrameSignal` left in `self._latest` by a *manually*-stopped track was rejected and the user got "Camera unavailable." The new code only checks `sig is None`. Since the PR deliberately keeps manual camera stop, the reachable regression is: user stops the camera → the last (stale) signal lingers in `_latest` until the participant leaves → `request_frame` returns an 8x8 placeholder (not `None`, so the `frame is None` branch does not catch it) → the VLM answers about a blank image instead of replying "Camera unavailable." Edge case off the always-on happy path. | [agent.py#L352-L356](https://github.com/NVIDIA/xr-ai/blob/f12afaef9d0df1dae4e98c84943c3c2d9b2e73fa/agent-samples/simple-vlm-example/worker/agent.py#L352-L356) |

wenxind-nvidia:
Good catch — addressed in 8294f4d.

You're right that dropping the camera-on-demand config also removed the `_is_fresh` guard, and since manual camera stop is intentionally kept, a stopped track leaves a stale `FrameSignal` in `_latest` → `request_frame` returns the 8x8 placeholder (not `None`) → the VLM answers about a blank image.

Fix: reintroduced a freshness check, but decoupled from the removed on-demand knobs — it's now a fixed module constant `_FRAME_MAX_AGE_US = 5_000_000` (5s) rather than the old `frame_max_age_s` config. `_handle_query` now rejects with "Camera unavailable, please try again." when the latest signal is missing **or** stale. 5s is comfortably above the always-on live-frame cadence, so the happy path is unaffected; it only trips on the manual-stop edge case.

nvddr (approved):

PR #181, opened 2026-06-03
Enable NVIDIA NIM as a model backend option (LLM + VLM)

wenxind-nvidia:
## What

Lets each agent sample run its **LLM and VLM on hosted [NVIDIA NIM](https://build.nvidia.com)** instead of locally hosted vLLM, chosen per sample by config. NIM exposes the same OpenAI-compatible API as the local servers, so this rides the existing `xr-ai-models` client layer — **no new client `kind`, no worker code changes**.

Scope is **LLM + VLM** (per the request). STT/TTS stay local because hosted NIM speech (Riva) isn't OpenAI `/v1/audio`-compatible.

## How a user switches a sample to NIM (manual config)

1. Point `models_yaml:` at the shipped `yaml/models.nim.yaml`.
2. Provide `NGC_API_KEY` (already a managed credential — `run_stack` injects it into subprocesses; or `export` it).
3. Omit the now-unused local model-server processes.
4. xr-render-demo only: also point the `vlm-mcp` process at `yaml/vlm_mcp_server.nim.yaml` (its VLM path runs through vlm-mcp).

## Changes

**`xr-ai-models` (one code change):** add `health_check: bool = True` to every spec, threaded config → factory → all four `OpenAICompat*` clients. Hosted NIM has no `/health` route, so NIM entries set `health_check: false` and `health()` returns `True` without probing — otherwise the workers' readiness gate blocks forever. (`api_key_env` → `Authorization: Bearer` was already supported.)

**Samples:** `models.nim.yaml` for both samples + `vlm_mcp_server.nim.yaml` for xr-render-demo; orchestrator comments documenting how to drop local servers.

**Docs:** `docs/ai-services.md` (new "Hosting models on NVIDIA NIM" section), `docs/credentials.md`, `README.md` (both samples), `xr-ai-models/README.md`, `docs/changelog.md`.

**Tests:** `health_check` parsing/default + `health()` short-circuit across clients.

## Notes / caveats

- **Hosted model ids in the overlays are examples** to confirm against build.nvidia.com for your account.
- The xr-render-demo agentic loop is tuned for the local Nemotron stack; a hosted model with tool calling (and reasoning for `agent_llm`) gives the closest behaviour.
- **Self-hosted NIM containers** work via the same mechanism — point `base_url` at the container and set `health_check: true` if it exposes `/v1/health`.
- No `pyproject.toml` changed, so `DEPENDENCIES.md` is untouched.

## Verification

`xr-ai-models` model tests updated; config-parse and `health()` behaviour validated for all four clients and for the vlm-mcp models-block path. CI (pytest 3.11/3.12, uv lock, SPDX, CodeQL) runs on push.


nvddr:
Adds hosted **NVIDIA NIM** as a per-sample model backend for LLM/VLM by leaning entirely on the existing `xr-ai-models` OpenAI-compatible client layer rather than introducing a new client `kind`. The one structural SDK change is a `health_check: bool = True` field threaded through every spec (`config.py`) → `factory.py` → all four `OpenAICompat*` clients (`openai_compat.py`), where `health()` short-circuits to `True` when disabled — needed because hosted NIM has no `/health` route and workers gate readiness on `health()`. Backend selection is a single `model_backend: local|nim` key in each sample's worker YAML: the worker swaps in a shipped `models.nim.yaml` overlay, and the orchestrators (`main.py`) read the same key via a stdlib regex (staying pyyaml-free) to drop the now-unused local model-server `Process` (simple-vlm spawns/removes `vlm`; xr-render-demo relies on `launch_mode="reuse"` so the unstarted local servers are simply skipped), and repoints `vlm-mcp` at a `vlm_mcp_server.nim.yaml`. STT/TTS stay local.

| Category | Finding | Location |
|---|---|---|
| Nit | NIM overlays duplicate the local `stt`/`tts` `base_url` ports (`8103`/`8105`) verbatim instead of inheriting from `models.yaml`; retuning STT/TTS ports in `models.yaml` silently drifts from `models.nim.yaml`. | [models.nim.yaml#L41-L49](https://github.com/NVIDIA/xr-ai/blob/a0536a4259c834cb039c563bc7a68284a6a01e9b/agent-samples/simple-vlm-example/yaml/models.nim.yaml#L41-L49) |
| Nit | `health_check: false` in `vlm_mcp_server.nim.yaml` is inert — vlm-mcp has no startup readiness gate calling `vlm.health()`; the worker's own VLM gate is what `models.nim.yaml`'s flag satisfies. Harmless, but the inline comment implies it's load-bearing. | [vlm_mcp_server.nim.yaml#L25-L26](https://github.com/NVIDIA/xr-ai/blob/a0536a4259c834cb039c563bc7a68284a6a01e9b/agent-samples/xr-render-demo/yaml/vlm_mcp_server.nim.yaml#L25-L26) |
| Nit | PR description's "how to switch" steps describe an earlier manual design; committed code/docs use the automatic `model_backend: nim` key. Only the un-committed PR body is stale. | [docs/changelog.md#L811-L823](https://github.com/NVIDIA/xr-ai/blob/a0536a4259c834cb039c563bc7a68284a6a01e9b/docs/changelog.md#L811-L823) |

Security pass: clean. YAMLs reference `api_key_env: NGC_API_KEY` (env-var name, not a value), `base_url` is the fixed `integrate.api.nvidia.com` (no user-controlled URL/SSRF surface), no `logger`/`print` touches `_api_key`, no hardcoded credentials. AGENTS.md hard rules (no keys in source, all HTTP via xr-ai-models, SPDX on new files, docs-in-same-PR) all satisfied. No `pyproject.toml` change, so `DEPENDENCIES.md` correctly untouched.

yanziz-nvidia:

## Summary

Adds NVIDIA NIM as an opt-in model backend for LLM and VLM inference, selected by `model_backend: nim` in each sample's worker YAML. Core change is a `health_check: bool = True` field propagated through config → factory → all four `OpenAICompat*` clients, with a fast-return in `health()` when `False`.

|     | Finding |
|-----|---------|
| 🚫 | None |
| 💡 | [`tests/test_models_openai_compat.py`](https://github.com/NVIDIA/xr-ai/blob/a0536a4259c834cb039c563bc7a68284a6a01e9b/tests/test_models_openai_compat.py) — `OpenAICompatTTS.health()` short-circuit is untested; `test_vlm_and_stt_health_true_when_health_check_disabled` covers VLM and STT but not TTS. |
| 💡 | [`simple-vlm-example/main.py:92`](https://github.com/NVIDIA/xr-ai/blob/a0536a4259c834cb039c563bc7a68284a6a01e9b/agent-samples/simple-vlm-example/main.py#L92) — `_BACKEND_RE` pattern `[\"']?(\w+)` doesn't consume the closing quote; tighten to `[\"']?(\w+)[\"']?`. Same issue at [`xr-render-demo/main.py:262`](https://github.com/NVIDIA/xr-ai/blob/a0536a4259c834cb039c563bc7a68284a6a01e9b/agent-samples/xr-render-demo/main.py#L262). |
| 💡 | [`yaml/models.nim.yaml:34`](https://github.com/NVIDIA/xr-ai/blob/a0536a4259c834cb039c563bc7a68284a6a01e9b/agent-samples/xr-render-demo/yaml/models.nim.yaml#L34) — `nvidia/cosmos-reason1-7b` listed as VLM without explanation; add a comment explaining why this model fits the render-demo use case. |
| 🔍 | None |

- `tests/test_models_openai_compat.py` — Add `OpenAICompatTTS` assertion to `test_vlm_and_stt_health_true_when_health_check_disabled`.
- `agent-samples/simple-vlm-example/main.py:92` and `agent-samples/xr-render-demo/main.py:262` — Change `_BACKEND_RE` from `[\"']?(\w+)` to `[\"']?(\w+)[\"']?`.
- `agent-samples/xr-render-demo/yaml/models.nim.yaml:34` — Add inline comment explaining the `cosmos-reason1-7b` VLM choice.

yanziz-nvidia:
Is there an easy way to add some test coverage?

agent-samples/xr-render-demo/main.py
```diff
@@ -46,46 +46,72 @@
...
+# With model_backend: nim (in xr_render_demo_worker.yaml) the worker loads
+# models.nim.yaml and vlm-mcp is pointed at vlm_mcp_server.nim.yaml here
+# automatically — run LLM/VLM on hosted NIM and just don't start the local
+# llm / agent-llm / vlm model-servers. STT/TTS stay local. See
+# docs/ai-services.md "Hosting models on NVIDIA NIM".
+def _build_processes(backend: str) -> list[Process]:
+    # The worker reaches the VLM through vlm-mcp, so vlm-mcp must use the same
+    # backend as the worker's models config.
+    vlm_mcp_config = (
+        "yaml/vlm_mcp_server.nim.yaml" if backend == "nim"
```
yanziz-nvidia, line 85: given the naming tradition does it make sense to do vlm_mcp_server.local.yaml?

yanziz-nvidia: or we are trying to keep the existing user flow so we don't change the other name?

README.md
```diff
@@ -321,6 +336,20 @@ cd agent-samples/model-servers
...
+([build.nvidia.com](https://build.nvidia.com)) instead of local vLLM
+(STT/TTS stay local) by setting **one key** in `xr_render_demo_worker.yaml`:
+
+```yaml
+model_backend: nim     # default is "local"
+```
+
+The worker loads `yaml/models.nim.yaml` and the orchestrator points `vlm-mcp`
+at `yaml/vlm_mcp_server.nim.yaml` automatically — no `main.py` edits. Provide
+an `NGC_API_KEY` and just don't start the local `llm` / `agent-llm` / `vlm`
```
yanziz-nvidia: as a global env var? or in yaml?

yanziz-nvidia:
Suggest to also run a security audit prompt with agent to make sure we are good

yanziz-nvidia (approved):

yanziz-nvidia:
LGTM modulo comments

wenxind-nvidia:
Thanks for the review — addressed the actionable items (merged latest `main` first; the only conflict was `docs/changelog.md` + the simple-vlm worker vs the #175 pipecat migration, both resolved keeping #175's structure + the `model_backend` switch):

- **`_BACKEND_RE` regex** — tightened to `[\"']?(\w+)[\"']?` in both `simple-vlm-example/main.py` and `xr-render-demo/main.py`.
- **TTS health coverage** — `OpenAICompatTTS.health()` is now asserted in the `health_check`-disabled test (renamed `test_vlm_stt_tts_health_true_when_health_check_disabled`).
- **`cosmos-reason1-7b` VLM** — added a comment in `xr-render-demo/yaml/models.nim.yaml` noting it mirrors the local Cosmos vlm-server so `ask_image` behaves the same on NIM.
- **Inert `health_check` in `vlm_mcp_server.nim.yaml`** (nvddr) — reworded the comment to say it's inert here (vlm-mcp has no readiness gate) and kept only for parity.
- **README `NGC_API_KEY`** (your "env var or yaml?") — clarified in both sample sections that it's an **environment variable / managed credential, not stored in YAML**; the overlay only names it via `api_key_env: NGC_API_KEY`.

On the `vlm_mcp_server.local.yaml` naming question: I kept the existing `vlm_mcp_server.yaml` as the local default and added `vlm_mcp_server.nim.yaml` as the overlay — renaming the existing file to `.local.yaml` would change the established user flow (and the file other docs reference) for no functional gain, so I left it to preserve backward compatibility. Happy to rename if you'd prefer the explicit `.local`/`.nim` symmetry.

The stt/tts-port-duplication and stale-PR-body nits are noted; the duplication is the intended self-contained-overlay tradeoff, and I'll refresh the PR description.

Security: re-confirmed nvddr's pass — YAMLs carry only the env-var *name* (`api_key_env`), fixed `integrate.api.nvidia.com` base_url (no user-controlled URL), no logging of `_api_key`, no hardcoded secrets, no `pyproject`/`DEPENDENCIES.md` change.

🤖 Automated update via Claude Code.

agent-samples/xr-render-demo/main.py
wenxind-nvidia: Yes, we keep the user flow and follow the naming convention currently in the repo (.yaml default, .[overlay].yaml for any overlay on default)

README.md
wenxind-nvidia: It's env var, prompted in terminal command.

wenxind-nvidia:
## Security audit (per @yanziz-nvidia's suggestion)

Audited the changes this branch introduces. **No exploitable vulnerability found.** Checks run, with evidence:

| Area | Result |
|---|---|
| Secrets in source | None — YAMLs carry only the env-var *name* (`api_key_env: NGC_API_KEY`), never a value |
| Key handling | Read from `os.environ`, sent only as `Authorization: Bearer` header; **never logged** (error logs are status + response body, not headers) |
| Credential storage | `~/.config/xr-ai/credentials.json` written `chmod 0o600` (owner-only) |
| SSRF / URL injection | `base_url` is operator-controlled YAML (fixed `integrate.api.nvidia.com`), not runtime/user input |
| Transport | `httpx.AsyncClient(trust_env=False)` — ignores `HTTP(S)_PROXY`/`.netrc`, so the token can't be redirected via a hostile proxy env |
| Deps | No `pyproject.toml` change → `DEPENDENCIES.md` correctly untouched |

**One defense-in-depth gap fixed.** This feature is the first to ship `api_key_env` over a *configurable* `base_url` (and the self-hosted docs suggest `http://localhost`). If an operator points a key at a **non-loopback `http://`** host, the bearer token would be sent in **cleartext** (CWE-319). The `OpenAICompat*` clients now **warn at construction** in exactly that case; `https` and loopback are exempt, so the shipped `https://integrate.api.nvidia.com` overlays never trip it. Log-only, no behavior change, covered by `test_cleartext_key_transport_detection`.

Pushed as a follow-up commit. CI re-running.

🤖 Automated update via Claude Code.

PR #182, opened 2026-06-04
fix(hub): prefix-safe participant isolation on return-traffic topics

nvddr:
## What

Terminates the participant-id segment of the hub→connector return-traffic ZMQ topics with a trailing `.`, so a connector that owns participant `alice` can no longer byte-prefix-match traffic addressed to `alice2`.

## Why

`docs/architecture.md` and the `test_return_routing` suite assert return traffic reaches **only** the owning connector — "impossible by construction" via pid-scoped topics. ZMQ `SUBSCRIBE` is a *byte-prefix* match, and the connector subscribed without a delimiter after the pid (`return_audio.{pid}`, `return_audio_flush.{pid}`, `return_data.{pid}`). So `alice`'s subscription also matched `alice2`'s topics. The processor *inbound* path already guards this exact case (`_prefixes` in `xr_ai_agent._processor`, with a comment calling out `data.alice` vs `data.alice2.chat`); the connector return path was missing the symmetric guard.

**Severity — latent, not a live leak.** Production runs **one connector per room** (`transport/livekit/connector.py`); that connector subscribes for *every* participant in the room anyway, and `RoomClient` routes each return message to the correct LiveKit participant by payload pid (`destination_identities` / per-pid `xr-hub-return-{pid}` track). The over-match delivered nothing the connector wasn't already entitled to. The defect only manifests in a **connector-per-participant** topology — which is what the test harness constructs. This PR is correctness-by-construction hardening + the missing test, not a fix for observed cross-talk.

## How

- **Hub** publishes `return_audio.{pid}.` and `return_audio_flush.{pid}.` (terminal topics gain the delimiter; `return_data.{pid}.{topic}` was already delimited by its trailing `.{topic}`).
- **Connector** subscribes/unsubscribes with the matching trailing `.`.
- Assumes participant identities contain no `.` — the same assumption the processor path already carries.

## Tests

- New `tests/test_participant_isolation_prefix.py` uses prefix-related identities (`alice` ⊂ `alice2`) and asserts no cross-delivery on **data, audio, and audio-flush**. Verified it **fails on the pre-fix code** (`alice`'s connector receives `alice2`'s return traffic) and passes after.
- Full return-path isolation suite (`test_return_routing`, `test_cross_talk`, `test_return_audio_flush`, `test_return_audio_pacing`, `test_multi_agent`, `test_participant_events`, `test_subscriptions`, `test_hub_data_routing`, `test_audio_routing`) — 42 passed, no regressions.

## Not covered

The LiveKit-transport enforcement layer (`destination_identities`, per-pid `xr-hub-return-{pid}` tracks, `set_track_subscription_permissions`) — the layer production isolation actually rides on — still has **no automated test**; `test_integration_livekit` stops at connect/disconnect. This PR is ZMQ-layer only. A container-based two-participant transport test belongs as a follow-up.


yanziz-nvidia:

## Summary

Fixes a ZMQ topic prefix-collision bug: a subscriber for participant `alice` also received traffic addressed to `alice2` because `SUBSCRIBE` is a byte-prefix match and the pid segment was unterminated. Appends a trailing `.` to both publisher and subscriber sides for `return_audio.*`, `return_audio_flush.*`, and `return_data.*`. Includes a new test file confirmed to fail on pre-fix code.

|     | Finding |
|-----|---------|
| 🚫 | None |
| 💡 | [`ipc/_connector.py:180`](https://github.com/NVIDIA/xr-ai/blob/8597a593f383d6947a64ae5e0bcbe804faf5b546/server-runtime/xr_media_hub/ipc/_connector.py#L180) — Add a comment explaining the trailing `.` terminates the pid segment to prevent prefix-collision, mirroring the comment added for `return_audio`. The PR description's "already delimited by `.{topic}`" framing refers to publisher format, not the subscriber prefix — slightly understates the fix. |
| 💡 | [`test_participant_isolation_prefix.py:45`](https://github.com/NVIDIA/xr-ai/blob/8597a593f383d6947a64ae5e0bcbe804faf5b546/tests/test_participant_isolation_prefix.py#L45) — Add `# allow ZMQ subscriptions to propagate` on both `asyncio.sleep(0.1)` calls to distinguish intentional propagation delays from arbitrary pauses. |
| 🔍 | [`test_participant_isolation_prefix.py:152`](https://github.com/NVIDIA/xr-ai/blob/8597a593f383d6947a64ae5e0bcbe804faf5b546/tests/test_participant_isolation_prefix.py#L152) — Move `_teardown` above `_bring_up_prefix_pair` for top-down readability. |
| 🔍 | [`docs/changelog.md`](https://github.com/NVIDIA/xr-ai/blob/8597a593f383d6947a64ae5e0bcbe804faf5b546/docs/changelog.md) — Clarify that the `return_data` *subscriber* was also changed; "already delimited" could mislead a future reader into thinking no subscriber fix was needed. |

- `server-runtime/xr_media_hub/ipc/_connector.py:180` — Add a comment on the `return_data` setsockopt lines explaining the trailing `.` prevents pid prefix-collision.
- `tests/test_participant_isolation_prefix.py:45-48` — Add `# allow ZMQ subscriptions to propagate` on both `asyncio.sleep(0.1)` calls.
- `tests/test_participant_isolation_prefix.py:152` — Move `_teardown` above `_bring_up_prefix_pair`.
- `docs/changelog.md` — Clarify the `return_data` sentence: the connector subscription was also changed (old `return_data.alice` matched `return_data.alice2.*`).

yanziz-nvidia (approved):

PR #183, opened 2026-06-04
fix: multi-client isolation — each client talks only to the hub (data, audio, per-pid agent audio)

nvddr:
Based on an integration of **#175 + #176** (both on the #174 pipecat foundation) so the agent-side fix lands against the migrated voice pipeline; #174/#175/#176 are left untouched. All multi-client fixes live here per request.

Three independent leaks let two human clients in one room interfere; fixed at the layer each belongs to. All client changes gate on `hubIdentity` (default `xr-hub-connector`; `null`/`nil` restores legacy whole-room behaviour).

## 1. Outbound data — publish-side (web/web-xr/android/iOS)
`send()` addresses data to the hub participant only (`destinationIdentities`/`identities`), so a client's text/ping/custom never reaches peers.

## 2. Inbound data + audio — subscribe-side (web/web-xr/android/iOS)
- `DataReceived` drops messages whose publisher is not the hub.
- Connect with **auto-subscribe disabled** and subscribe only to the hub participant's tracks (track-published event + connect-time sweep). A client no longer receives/plays another participant's mic — only the hub's per-pid return-audio track.

## 3. Agent return audio — per-participant routing (`xr-ai-pipecat` foundation)
The pipecat pipeline was per-pid on input (VAD/STT/brain) but collapsed to a single `_target_participant` (last-join-wins) on output — `XRMediaHubOutputTransport` nulled every frame's `transport_destination`, so **participant A's spoken answer was published on B's return-audio track**. Now: a per-pid `MediaSender` is registered on `ParticipantJoinedFrame` (lazily in `_handle_frame` too), frames keep `transport_destination = pid`, and `write_audio_frame` addresses the chunk at the frame's own pid. Both pipecat samples (simple-vlm, xr-render-demo) inherit it.

## Verification
- SDK symbols verified per platform before use (`livekit-client` `destinationIdentities`/`setSubscribed`/`TrackPublished`; `livekit-android` 2.7.0 `ConnectOptions.autoSubscribe`/`RemoteTrackPublication.setSubscribed`/`Participant.Identity`; `client-sdk-swift` `ConnectOptions.autoSubscribe`/`set(subscribed:)`/`didPublishTrack`/`Participant.Identity(from:)`).
- Web: `node --check` clean.
- Foundation: `test_output_transport_routes_audio_by_frame_pid_not_single_target` — **fails pre-fix** (A & B both addressed to last joiner), passes after. Full `xr-ai-pipecat` suite: **62 passed**.
- **Android/iOS not compiled here** (no toolchain) — need an on-device build.
- **Live two-client repro not yet run.**

## Known follow-ups (degrade only under concurrent speech)
In `xr-ai-pipecat`: `StreamingTts` sentence buffer is still a single shared `_pending`; interruption is still global (`InterruptionFrame` → cancel-all, not pid-scoped); `UserStarted/StoppedSpeaking` are pid-less (camera warmup fans to all). Tracked separately.


yanziz-nvidia:

## Summary

Fixes three multi-client isolation leaks: outbound data was broadcast to all room participants, inbound data/audio subscribed to all participants, and agent return audio was addressed to the last-joined participant instead of the per-frame originator. Applies a `hubIdentity` gate across all four client platforms.

|     | Finding |
|-----|---------|
| 🚫 | [`xr_ai_pipecat/transport.py:218`](https://github.com/NVIDIA/xr-ai/blob/1bcbe670b3724f1eb3568463ade9a22c46e94c26/agent-sdk/xr-ai-pipecat/xr_ai_pipecat/transport.py#L218) — **Race in `_ensure_destination`: `process_frame` and `_handle_frame` can both pass `pid not in self._media_senders` concurrently, leaking a `MediaSender`.** Add an `asyncio.Lock` with a re-check after acquiring; add `self._sender_lock = asyncio.Lock()` in `__init__`. |
| 🚫 | [`tests/test_xr_ai_pipecat_voice_pipeline.py:263`](https://github.com/NVIDIA/xr-ai/blob/1bcbe670b3724f1eb3568463ade9a22c46e94c26/agent-sdk/xr-ai-pipecat/tests/test_xr_ai_pipecat_voice_pipeline.py#L263) — **New test bypasses the `_ensure_destination`/`MediaSender` path entirely** (`start()` never called, `_start_frame` is `None`). Add a test that calls `transport.start()`, pushes a `ParticipantJoinedFrame`, sends an `OutputAudioRawFrame`, and asserts `AudioChunk.participant_id` matches the join pid. |
| 💡 | [`LiveKitBackend.js:209`](https://github.com/NVIDIA/xr-ai/blob/1bcbe670b3724f1eb3568463ade9a22c46e94c26/client-samples/web/StreamKit/Backends/LiveKit/LiveKitBackend.js#L209) — `TrackPublished` listener registered inside the `hubIdentity` guard; move outside and short-circuit in the handler body (matches `DataReceived` pattern). |
| 💡 | [`LiveKitBackend.swift:565`](https://github.com/NVIDIA/xr-ai/blob/1bcbe670b3724f1eb3568463ade9a22c46e94c26/client-samples/ios-visionos/StreamKit/Sources/StreamKit/Backends/LiveKit/LiveKitBackend.swift#L565) — `try?` silently swallows subscription errors; use `do/catch` with a log. |
| 💡 | [`BackendConfiguration.kt:80`](https://github.com/NVIDIA/xr-ai/blob/1bcbe670b3724f1eb3568463ade9a22c46e94c26/client-samples/android/app/src/main/java/com/nvidia/xrai/streamkitsample/streamkit/config/BackendConfiguration.kt#L80) — Hardcoded `"xr-hub-connector"` default causes silent audio failure on identity mismatch; add a startup log when `autoSubscribe` is disabled. |
| 💡 | Android/iOS/Web connect-time participant sweep — not race-free vs. LiveKit's internal state machine; worth a comment for future maintainers. |
| 🔍 | [`docs/changelog.md`](https://github.com/NVIDIA/xr-ai/blob/1bcbe670b3724f1eb3568463ade9a22c46e94c26/docs/changelog.md) — Entry dated `2026-06-04`; update to merge date. |

- `agent-sdk/xr-ai-pipecat/xr_ai_pipecat/transport.py:157` — Add `self._sender_lock = asyncio.Lock()` in `__init__`; wrap the check-and-create block in `_ensure_destination` with `async with self._sender_lock:` plus a re-check after acquiring.
- `agent-sdk/xr-ai-pipecat/tests/test_xr_ai_pipecat_voice_pipeline.py` — Add a test calling `transport.start(start_frame)`, pushing `ParticipantJoinedFrame`, sending `OutputAudioRawFrame`, asserting `AudioChunk.participant_id` equals the join pid.
- `client-samples/ios-visionos/.../LiveKitBackend.swift:565` — Replace `try?` with `do { try await ... } catch { logger.warning(...) }`.
- `client-samples/web/.../LiveKitBackend.js:209` — Move `TrackPublished` registration outside the `hubIdentity` guard; short-circuit inside the handler.
- `client-samples/android/.../LiveKitBackend.kt:113` — Log when `hubIdentity` is non-null and `autoSubscribe` is disabled so a mismatched identity is visible in logcat.

yanziz-nvidia (approved):

PR #177, opened 2026-06-02
Uniform UI on native client

wenxind-nvidia:
Add video frame panel, agent answer panel in ios/android client to match web client
Tested ios, android needs testing

PR #185, opened 2026-06-05
feat(android): synthetic Virtual Camera provider using injectVideoFrame

nvddr:
#172 (the `injectVideoFrame` API this builds on) is merged to `main`; this PR targets `main` with a clean diff of just the virtual-camera additions.

## What
A selectable **"Virtual Camera (synthetic)"** entry in the Android camera dropdown that feeds generated frames through the public `StreamSession.injectVideoFrame` — no physical camera, no CAMERA permission. Useful as a live demo of the inject API and to exercise the video path on an emulator / camera-less device.

## How
- **`SyntheticCamera.kt`** — `SyntheticCameraSource` renders animated I420 (scrolling SMPTE-style colour bars + a bouncing box) into a **reused direct `ByteBuffer`** (`pushI420Frame` copies planes before returning, so reuse is safe; direct buffer required for `YuvHelper`/`GetDirectBufferAddress`).
- **`AppViewModel`** — `selectableCameras = physical + virtual`; selecting the virtual id runs a `Dispatchers.Default` coroutine calling `session.injectVideoFrame(buf, w, h, nanoTime/1000)` at ~30 fps.
- **`MainActivity`** — selector points at `selectableCameras` and skips the CAMERA-permission prompt for the virtual entry.
- **`LiveKitBackend`** — injected track now published with `VideoTrackPublishOptions(source = Track.Source.CAMERA)` so the #177 preview card (reads the CAMERA-source publication) shows the synthetic frames and the hub treats it as the camera feed. Benefits all `injectVideoFrame` callers.

## Lifecycle correctness (the subtle bits)
- Synthetic loop is **`cancelAndJoin`-ed before `stopCamera()`** — otherwise a trailing frame hits the lazy slow-path and *republishes* a fresh injected track, so "Stop" would silently no-op.
- Loop is **gated on a CONNECTED session** and **cancelled on disconnect** (both the state-handler and `disconnect()`), so it never calls `injectVideoFrame` on a torn-down session.
- `stopCamera()` already unpublishes the injected track (`unpublishTrack`, default `stopOnUnpublish=true` → stops+disposes) — reused, not duplicated.

## Verification
- SDK symbols checked against livekit-android 2.7.0 (`VideoTrackPublishOptions.source`, `unpublishTrack(stopOnUnpublish=true)`).
- **Not compiled here** (no Android/Gradle toolchain) — **on-device build + run is the gate**: select Virtual Camera → Start → confirm synthetic frames appear in the preview and reach the agent; Stop removes the track; works with no CAMERA permission and on an emulator.

No new dependencies; `DEPENDENCIES.md` unchanged.

### DIFF
--- a/DEPENDENCIES.md
+++ b/DEPENDENCIES.md
@@ -333,7 +333,7 @@ Jetpack Compose sample app mirroring the web and iOS/visionOS clients feature-fo
 
 | Layer | Language | External deps |
 |---|---|---|
-| StreamKit library | Kotlin | `io.livekit:livekit-android` 2.7.0 |
+| StreamKit library | Kotlin | `io.livekit:livekit-android` 2.7.0 (provides `TextureViewRenderer` used by the in-SDK `CameraPreviewView` composable; no extra `livekit-android-compose-components` dep) |
 | App UI | Kotlin + Jetpack Compose | Compose BOM 2024.11.00, `lifecycle-viewmodel-compose` 2.8.7, `activity-compose` 1.9.3 |
 
 The `gradle-wrapper.jar` is not checked in (binary artifact); Android Studio generates it on first sync.
--- a/README.md
+++ b/README.md
@@ -7,16 +7,15 @@
 
 # xr-ai
 
-Agentic AI for XR — a sample blueprint for multi-modal, real-time conversational AI
-within the CloudXR ecosystem.
+Agentic AI for XR — an open-source foundation for multi-modal, real-time
+conversational AI within the CloudXR ecosystem.
 
-## Early Access Notice
+## Public Beta Notice
 
-This project is currently in early access and is under active development.
-Features, APIs, documentation, and behavior may change without notice.
-Expect bugs, incomplete functionality, and breaking changes as the project
-evolves. Use at your own discretion, and please report issues or feedback
-to help improve the project.
+This project is publicly available in beta and is under active development.
+Features, APIs, documentation, and behavior may change as the project evolves.
+Expect bugs, incomplete functionality, and breaking changes. Use at your own
+discretion, and please report issues or feedback to help improve the project.
 
 ## What is XR-AI?
 
@@ -24,7 +23,7 @@ XR-AI is a developer stack for building powerful XR and AI systems across device
 
 With XR-AI, developers can build agents that see and hear what the user experiences, reason over live physical context, call external tools through MCP, and respond with audio or data in the same XR session. The stack provides an end-to-end foundation for multimodal spatial computing applications: real-time media routing, participant-aware response handling, agent interfaces, AI service integration, remote rendering, and sample applications that show the pieces working together.
 
-The value is speed without lock-in. XR-AI is designed to work quickly with NVIDIA open models for vision, language, speech, and speech synthesis, while still giving developers the flexibility to bring their own models, services, tools, and application logic. Because it is built around NVIDIA GPU infrastructure, the same architecture can be deployed where the workload needs to run: cloud, data center, workstation, or edge.
+The value is speed without lock-in. XR-AI is designed to work quickly with NVIDIA open models for vision, language, and speech, plus swappable speech synthesis services, while still giving developers the flexibility to bring their own models, services, tools, and application logic. Because it is built around NVIDIA GPU infrastructure, the same architecture can be deployed where the workload needs to run: cloud, data center, workstation, or edge.
 
 XR-AI also gives developers a practical path across product categories. Teams can start with AI glasses-style experiences that use live camera, audio, and agent responses, then extend the same framework to richer AR glasses or XR headset experiences that use CloudXR remote rendering. This lets developers build for today's lightweight AI devices while keeping a clear path to immersive, GPU-rendered spatial applications.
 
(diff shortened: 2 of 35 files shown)


wenxind-nvidia (commented):
## Review — synthetic Virtual Camera over `injectVideoFrame`

Nicely scoped and unusually careful on lifecycle (the `cancelAndJoin`-before-`stopCamera`, connected-gating, and inject-first-frame-then-flip-`isCameraActive` ordering are all the right calls, and the rationale comments make them reviewable). `#172` is already merged to `main`, so the base is clean and the diff is just the virtual-camera additions. No High-severity issues found. A few things worth confirming on-device.

### Medium

**1. Buffer reuse is correct only if `injectVideoFrame` copies the planes *before any thread hands-off* — confirm with motion.** `SyntheticCameraSource.renderFrame()` returns the same direct `ByteBuffer` every call (`SyntheticCamera.kt`), and the loop overwrites it ~33 ms later (`AppViewModel.startVirtualCamera`). This is safe iff the copy in the inject path is synchronous w.r.t. the call returning — not merely "`pushI420Frame` copies", but that nothing posts the buffer to another thread before that copy. If the copy is ever deferred, fast motion will tear/corrupt. Please verify on-device that the bouncing box / scrolling bars show no tearing at speed (the scrolling pattern is well chosen for spotting it).

**2. `source = Track.Source.CAMERA` is a global change to the shared inject path, not virtual-camera-only.** It's added in `LiveKitBackend.publish…` (the common `injectVideoFrame` publish path), so **every** `injectVideoFrame` caller — including a future real external-camera adapter — now publishes as the CAMERA source and shows up in the `#177` preview card + is treated as the participant camera by the hub. That appears intended (and is the right default), but it's a behavior change to the `#172` API surface rather than something local to this feature — worth calling out explicitly so a later caller isn't surprised, and confirming no caller wants a non-CAMERA source.

### Low / nits

**3. Per-frame full-image Kotlin loops (`SyntheticCamera.renderFrame`).** ~460k byte writes/frame (luma + both chroma) at 30 fps. Fine for a 640×480 demo, but the bars are static except for `scroll`/box — you could precompute the bar rows once and only stamp the box + apply the scroll offset, cutting most of the per-frame work for low-end devices. Optional.

**4. Error path leaves a possibly-published track.** In `startVirtualCamera`, a mid-loop exception after the first frame sets `isCameraActive = false` but doesn't unpublish; the injected track can linger until the next `stopCamera()`/`disconnect()`. Minor (a subsequent Start re-enters the lazy path), but a `stopCamera()` on the error branch would leave cleaner state.

**5. Disconnect handler uses `cancel()` (not `cancelAndJoin`).** The `onConnectionStateChanged(DISCONNECTED)` branch cancels without joining, so a trailing `injectVideoFrame` can briefly race a tearing-down session — handled by the generic `catch (Exception)`, so harmless, just noting the asymmetry with `disconnect()`/`stopCamera()` which both join.

**6. Stale stacking note (doc only).** The description says "Retarget to `main` once #172 merges" — #172 is now in `main` and the PR already targets `main` with a clean diff, so that note can be dropped.

### Good
SPDX header on the new file, `changelog.md` updated, no new deps (`DEPENDENCIES.md` correctly untouched), camera-less/emulator fallback (`selectedCameraId ?: VIRTUAL_CAMERA_ID`), and CAMERA-permission correctly skipped for the synthetic entry. The on-device build+run gate you noted is the right call given no Android toolchain in CI.

🤖 Automated review via Claude Code — a maintainer should still review.

nvddr:
Disposition of the review (author has tested on-device; works well):

| # | Finding | Disposition |
|---|---|---|
| M1 | Buffer reuse safe only if planes copied before any thread hand-off | **Confirmed.** `pushI420Frame` copies via `YuvHelper.copyPlane` synchronously before `onFrameCaptured` returns, and the loop awaits the suspend `injectVideoFrame` before rendering the next frame. On-device the scrolling bars / bouncing box show no tearing at speed. |
| M2 | `source = CAMERA` is a global change to the shared inject path | **Intended, and the right default** — every `injectVideoFrame` track represents the participant's camera feed, so it should surface in the preview + be treated as the camera by the hub. Flagging explicitly here for future external-camera-adapter callers; none currently want a non-CAMERA source. |
| L3 | Per-frame full-image Kotlin loops | **Deferred (optional).** Fine at 640×480/30fps for a demo; the scroll shifts the per-column bar mapping each frame so a precompute isn't trivial. Left simple/readable. |
| L4 | Error path leaves a possibly-published track | **Fixed** in c059501 — the loop's error branch now `runCatching { session.stopCamera() }` to unpublish before clearing state. |
| L5 | Disconnect uses `cancel()` not `cancelAndJoin()` | **Acknowledged, left as-is.** The `onConnectionStateChanged` handler is a non-suspend callback so it can't join; a trailing `injectVideoFrame` racing teardown is caught by the generic handler (now also unpublishing per L4). `disconnect()`/`stopCamera()` still join. |
| L6 | Stale "retarget once #172 merges" note | **Fixed** — PR description updated; #172 is merged and this targets `main`. |

Still gated on human approval. Android isn't buildable in CI — on-device build+run remains the gate (author has run it).

wenxind-nvidia (approved):

PR #189, opened 2026-06-08
fix(streamkit): fix TOCTOU null-deref race in LiveKitBackend.send()

### LINKED ISSUE #31: what about room
what about room

_Originally posted by @yanziz-nvidia in https://github.com/NVIDIA/xr-ai/pull/16#discussion_r3128275502_

yanziz-nvidia:
## The bug

`LiveKitBackend.send()` read `this.#room` twice: once in its guard
(`if (\!this.#room || this.#room.state \!== 'connected')`) and again at the
`await this.#room.localParticipant.publishData(...)` call.

Because `await` yields the event loop, a concurrent `#tearDown()` can null
`this.#room` between the guard and the dereference, turning the
`publishData()` access into a null-deref crash (TOCTOU race).

## The fix

Snapshot `this.#room` into a local `const room` at the top of `send()` and
use that local everywhere — the guard and the `await`. This is the exact same
snapshot pattern commit `4fedfe0` applied to `stopAudio` / `stopCamera` in
this same file. Change is 4 lines.

Fixes #31

### DIFF
--- a/client-samples/web/StreamKit/Backends/LiveKit/LiveKitBackend.js
+++ b/client-samples/web/StreamKit/Backends/LiveKit/LiveKitBackend.js
@@ -373,7 +373,10 @@ export class LiveKitBackend {
    * @throws {StreamError} `notConnected` if not connected.
    */
   async send(data, { reliable = true, topic } = {}) {
-    if (!this.#room || this.#room.state !== 'connected') {
+    // Snapshot the room so a concurrent #tearDown() nulling this.#room across
+    // the await below can't turn the publishData() deref into a null crash.
+    const room = this.#room;
+    if (!room || room.state !== 'connected') {
       throw StreamError.notConnected();
     }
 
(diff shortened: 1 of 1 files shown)


nvddr:
## Review

Fixes a TOCTOU null-dereference race in `LiveKitBackend.send()` (web StreamKit data-channel publish). `send()` read private `#room` twice — once in the `notConnected` guard, again at `await this.#room.localParticipant.publishData(...)`. The `await` yields the event loop, so a concurrent `#tearDown()` (sets `#room = null`) could slip between check and deref → crash. The fix snapshots `this.#room` into a local `const room` and uses it for both the guard and the publish, mirroring the existing snapshot pattern in `#tearDown()` and commit 4fedfe0. No behavioral change to success/error paths.

|---|---|---|
| _(none)_ | — | Clean, minimal, idiomatic. No blockers, bugs, or nits. |

Note (out of scope): the snapshot only prevents the null-deref crash; if teardown completes mid-publish, `publishData` may still reject — but as a normal rejected promise, not a null crash. That's the intended improvement.

nvddr (approved):

PR #200, opened 2026-06-08
fix(ios): reset isCameraActive when a camera switch fails (#195)

### LINKED ISSUE #195: iOS switchCamera leaves isCameraActive=true after a failed switch (UI shows streaming, nothing published)
## Summary
On iOS/visionOS, `switchCamera` leaves `isCameraActive == true` after a failed camera switch — the UI claims it is streaming while nothing is published.

## Details
`client-samples/ios-visionos/App/AppModel.swift` `switchCamera(to:)` (~L268-276) is only entered when the camera is already active. It calls `session?.startCamera(...)`. The LiveKit backend's `startCamera` first runs `try await stopCamera()` (which unpublishes the old track and sets `localCameraTrack = nil`), then creates and publishes the new track.

If the publish step throws (device busy mid-switch, transient capture error, etc.), the old camera is already torn down and the new one is not published — yet `switchCamera`'s `catch` block only sets `lastError` and never resets `isCameraActive`.

## Consequence
The UI shows "Streaming" with a working "Stop" button and a green status, but no video track is actually published, and the preview goes blank. Contrast with `startCamera()`/`stopCamera()`, which keep `isCameraActive` consistent.

## Suggested fix
In the `catch` of `switchCamera`, set `isCameraActive = false` (the backend has already stopped the previous camera at that point).

---

wenxind-nvidia:
Fixes #195.

## Problem
On iOS/visionOS, `AppModel.switchCamera(to:)` (`client-samples/ios-visionos/App/AppModel.swift`) is only entered when the camera is already active and calls the backend's `startCamera()`. The LiveKit backend's `startCamera()` first `stopCamera()`s the previous track (unpublish + `localCameraTrack = nil`), then publishes the new one.

If the publish step throws (device busy mid-switch, transient capture error), the old camera is already torn down and the new one is not published — but `switchCamera`'s `catch` only set `lastError` and left `isCameraActive == true`. The UI then shows **"Streaming"** with a green status and a working Stop button, while nothing is published and the preview is blank.

## Fix
Set `isCameraActive = false` in the `catch` (the backend has already stopped the previous camera at that point), matching the state consistency `startCamera()`/`stopCamera()` already maintain.

```swift
} catch {
    lastError = error.localizedDescription
    isCameraActive = false   // backend already stopped the old track; nothing is streaming
}
```

## Notes
- One-line behavioral fix in the catch; no API/dependency changes (`DEPENDENCIES.md` untouched).
- No Swift unit-test harness in CI; verification is on-device — trigger a camera switch that fails mid-publish and confirm the UI drops to "Idle" (not stuck on "Streaming").
- Changelog updated.

### DIFF
--- a/client-samples/ios-visionos/App/AppModel.swift
+++ b/client-samples/ios-visionos/App/AppModel.swift
@@ -272,6 +272,10 @@ final class AppModel {
             try await session?.startCamera(config: CameraConfig(position: cameraPosition))
         } catch {
             lastError = error.localizedDescription
+            // The backend's startCamera() stops the previous track before
+            // publishing the new one, so on a failed publish nothing is
+            // streaming — reflect that instead of leaving the UI on "Streaming".
+            isCameraActive = false
         }
     }
 
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -9,6 +9,16 @@ Significant decisions, in reverse-chronological order. Update this whenever a
 non-trivial architectural or design decision is made so the rationale is
 preserved and not re-litigated.
 
+### 2026-06-05 — iOS: reset isCameraActive when a camera switch fails
+
+`AppModel.switchCamera(to:)` only ran on an already-active camera and, on a
+failed publish, set `lastError` but left `isCameraActive = true`. Because the
+LiveKit backend's `startCamera()` stops the previous track before publishing
+the new one, a publish failure mid-switch left nothing streaming while the UI
+still showed "Streaming" with a green status and a working Stop button. The
+`catch` now sets `isCameraActive = false`, matching the consistency that
+`startCamera()`/`stopCamera()` already maintain. Fixes #195.
+
 ### 2026-06-05 — piper voice fetch: catch LocalEntryNotFoundError before EntryNotFoundError
 
 Follow-up to #184. That PR added a dedicated `_EXIT_VOICE_UNAVAILABLE = 3`


nvddr:
## Review

One-line state-consistency fix in the iOS/visionOS sample's `AppModel` (SwiftUI view-model above StreamKit). `switchCamera(to:)` is the only camera path that re-invokes the backend's `startCamera()` while a camera is already active, and that backend method unconditionally tears down the previous track (`stopCamera()` → unpublish + `localCameraTrack = nil`) before publishing the new one. The `catch` previously recorded `lastError` but left `isCameraActive == true`, so a mid-switch publish failure stranded the UI on a green "Streaming" state with nothing published. The fix sets `isCameraActive = false` in the catch, matching the invariant `startCamera()`/`stopCamera()` already hold. Confirmed iOS-only: web and android disable/hide the camera selector while active (`camSelect.disabled`; android shows it only `if (!vm.isCameraActive)`), so neither has an analogous live mid-switch path — cross-client scope satisfied.

|---|---|---|
| Nit | [AppModel.swift#L268](https://github.com/NVIDIA/xr-ai/blob/f5fa4e705846d86e948956023024bf37b46ea2b5/client-samples/ios-visionos/App/AppModel.swift#L268) | Pre-existing (not introduced here): `switchCamera(to:)` lacks the `isCameraStarting` guard `startCamera()` uses, so a switch racing a concurrent start/stop is unguarded. Out of scope for this fix; noted for completeness. |

Clean and correctly scoped.

nvddr (approved):

PR #201, opened 2026-06-08
fix(stt-server): serialize NeMo transcribe() on the shared model (#199)

### LINKED ISSUE #199: STT server runs concurrent NeMo transcribe() on shared model with no lock (magpie serializes, STT doesn't)
## Summary
The STT server runs concurrent `transcribe()` calls on a single shared NeMo model with no serialization, unlike the magpie TTS backend which explicitly serializes.

## Details
In `ai-services/stt-server/stt_server/__main__.py`, `_AsrBackend._lock` guards only model *loading* (used in `_ensure_loaded`, ~L76). The hot path `transcribe` (~L99-109) calls `self._model.transcribe([audio_path])` with **no lock**. The endpoint dispatches each request to the default `ThreadPoolExecutor` via `run_in_executor(None, _run)` (~L167), so two concurrent `POST /v1/audio/transcriptions` requests invoke `transcribe()` on the same NeMo `ASRModel` simultaneously.

NeMo ASR `.transcribe()` is not documented/built as re-entrant or thread-safe (shared model buffers and, on CUDA, shared device state). The magpie TTS backend explicitly acknowledges this and serializes synthesis ("Magpie TTS is not re-entrant — serialize all synthesis calls", magpie `__main__.py` ~L97-101); the STT backend has the equivalent risk but no equivalent guard.

## Consequence
Under concurrent load: corrupted/garbled transcriptions or a crash.

## Suggested fix
Wrap the inference in the existing lock, e.g. `with self._lock: with torch.inference_mode(): results = self._model.transcribe(...)`, matching the magpie backend's pattern.

---
*Filed from an automated code-logic review. Confidence: medium (exact thread-safety of this model version not provable from source; the asymmetry vs. magpie's own stated assumption is the tell). No assignee.*

wenxind-nvidia:
Fixes #199.

## Problem
In `ai-services/stt-server/stt_server/__main__.py`, `_AsrBackend._lock` guards only model **loading** (`_ensure_loaded`). The hot path `transcribe()` called `self._model.transcribe([audio_path])` **lock-free**. The endpoint dispatches each `POST /v1/audio/transcriptions` to the default `ThreadPoolExecutor` via `run_in_executor(None, _run)`, so two concurrent requests invoke `transcribe()` on the same NeMo `ASRModel` simultaneously.

NeMo ASR `.transcribe()` is not re-entrant/thread-safe (shared model buffers and, on CUDA, shared device state). The magpie TTS backend explicitly serializes synthesis for the same reason; STT had the equivalent risk with no guard → garbled transcriptions or a crash under concurrent load.

## Fix
Wrap inference in the existing `self._lock`, mirroring magpie:

```python
self._ensure_loaded()
with self._lock:
    with torch.inference_mode():
        results = self._model.transcribe([audio_path], verbose=False)
```

`_ensure_loaded()` runs **before** acquiring the lock — `_lock` is non-reentrant and `_ensure_loaded()` takes it for the one-time load (post-load fast path is lock-free), so there's no deadlock and concurrent first-calls serialize correctly.

## Notes
- Mirrors the established pattern in `ai-services/tts/magpie/magpie_tts_server/__main__.py` (`synthesize`).
- No API/dependency changes (`DEPENDENCIES.md` untouched).
- Serialization is the correct trade-off here: a single shared model can't run true concurrent inference safely anyway; this removes the corruption/crash risk. Throughput-via-replicas is a separate, larger change.
- Changelog updated.

### DIFF
--- a/ai-services/stt-server/stt_server/__main__.py
+++ b/ai-services/stt-server/stt_server/__main__.py
@@ -100,8 +100,15 @@ class _AsrBackend:
         """Synchronous. Call from a thread pool."""
         self._ensure_loaded()
         import torch
-        with torch.inference_mode():
-            results = self._model.transcribe([audio_path], verbose=False)
+        # NeMo ASR .transcribe() is not re-entrant/thread-safe (shared model
+        # buffers and, on CUDA, shared device state), and the endpoint
+        # dispatches each request to a thread pool — serialize inference on the
+        # shared model, mirroring the magpie TTS backend. _ensure_loaded() runs
+        # BEFORE acquiring the lock because _lock is non-reentrant and
+        # _ensure_loaded() takes it itself (the post-load fast path is lock-free).
+        with self._lock:
+            with torch.inference_mode():
+                results = self._model.transcribe([audio_path], verbose=False)
         # NeMo returns a list of strings (or Hypothesis objects).
         if not results:
             return ""
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -9,6 +9,18 @@ Significant decisions, in reverse-chronological order. Update this whenever a
 non-trivial architectural or design decision is made so the rationale is
 preserved and not re-litigated.
 
+### 2026-06-05 — STT: serialize NeMo transcribe() on the shared model
+
+`_AsrBackend._lock` guarded only model *loading*; the hot path `transcribe()`
+ran `self._model.transcribe(...)` lock-free. The endpoint dispatches each
+`POST /v1/audio/transcriptions` to a thread pool, so concurrent requests
+invoked inference on the same NeMo `ASRModel` simultaneously — which is not
+re-entrant/thread-safe (shared model buffers, shared CUDA device state),
+risking garbled transcriptions or a crash under load. Inference is now wrapped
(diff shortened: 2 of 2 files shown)


nvddr:
## Review

Thread-safety fix in the in-process NeMo ASR backend of the OpenAI-compatible STT server (`ai-services/stt-server`). Each `POST /v1/audio/transcriptions` is dispatched to the default `ThreadPoolExecutor` via `run_in_executor(None, _run)`, so concurrent requests could call `ASRModel.transcribe()` on the same shared model simultaneously — NeMo ASR inference isn't re-entrant (shared model buffers + CUDA device state). The existing `_AsrBackend._lock` previously guarded only one-time load in `_ensure_loaded()`; the fix extends it to wrap the inference call, serializing GPU work on the single shared model, mirroring the magpie TTS backend pattern. Lock ordering is correct: `_lock` is non-reentrant and `_ensure_loaded()` runs before `transcribe()` acquires it (and `_serve` preloads the model before opening the port), so no self-deadlock — the only real contention is between concurrent inferences, which is the intent. Coarse whole-call serialization is the documented trade-off (replica throughput out of scope). No API/dep changes; changelog updated.

No findings — clean.

nvddr (approved):

PR #202, opened 2026-06-08
fix(hub): release held slots before closing a re-registered ring (#197)

### LINKED ISSUE #197: Hub use-after-close: connector re-registration closes ring while _latest_slots still references its buffer
## Summary
Hub use-after-close: when a connector re-registers, the hub closes its ring buffer while `_latest_slots` may still reference slots backed by that ring's mmap.

## Details
In `server-runtime/xr_media_hub/ipc/_hub.py`, on `MsgType.CONNECTOR_REGISTER` for an already-known `connector_id`, `_handle_registration` (~L325-328) calls `self._ring_registry[reg.connector_id].close()` and replaces the entry. But `_latest_slots` may still hold `SlotView`s into that ring's mmap buffer.

`ShmRingBuffer.close()` calls `self._buf.release()`, which raises `BufferError` (silently swallowed in `xr_ai_agent/_shm.py` ~L172-175) because the slice memoryview held by the live `SlotView` is still exported — so the old ring is left half-closed but still referenced by `_latest_slots`. On the next `FRAME_SIGNAL` for that participant, ~L218-220 pops the stale slot and calls `prev[0].release_slot(...)` on that old ring, writing through a buffer whose owner has had `close()` attempted.

## Trigger
A connector crashes/reconnects and re-sends its registration while frames are still held.

## Consequence
Write to a released/closing mmap (undefined behavior, potential crash) or silent corruption.

## Suggested fix
In `_handle_registration`, before closing/replacing the old ring, release any `_latest_slots` entries whose ring is the one being replaced — mirror the cleanup in the `PARTICIPANT_EVENT` leave branch.

---
*Filed from an automated code-logic review. Confidence: medium (edge-triggered on re-registration). No assignee.*

wenxind-nvidia:
Fixes #197.

## Problem
In `server-runtime/xr_media_hub/ipc/_hub.py`, on `MsgType.CONNECTOR_REGISTER` for an already-known `connector_id`, `_handle_registration` calls `self._ring_registry[id].close()` and replaces the entry — but `_latest_slots` may still hold `SlotView`s backed by that ring's mmap.

A live `SlotView` keeps a *sliced* memoryview exported into the ring's buffer, so `ShmRingBuffer.close()`'s `self._buf.release()` raises `BufferError` (it's the `release()` on line 169, ahead of the `try/except` that only wraps `_shm.close()`). The old ring is left half-closed but still referenced by `_latest_slots`; the next `FRAME_SIGNAL` for that participant pops the stale entry and `release_slot()`s on that ring → write through a released/closing mmap (undefined behavior / corruption).

**Trigger:** a connector crashes/reconnects and re-sends its registration while frames are still held.

## Fix
Before closing the old ring, release every `_latest_slots` entry backed by it — `view.data.release()` then `release_slot()` then drop the entry — then `close()`. This matches the teardown order already used in `HubEndpoint.close()` (release the exported memoryview *before* closing so the mmap has no exported pointers):

```python
old_ring = self._ring_registry[reg.connector_id]
for key in [k for k, (ring, _) in self._latest_slots.items() if ring is old_ring]:
    _, view = self._latest_slots.pop(key)
    view.data.release()
    old_ring.release_slot(view.signal.slot)
old_ring.close()
```

## Test
`tests/test_participant_events.py::test_connector_reregistration_releases_held_slots` — register, join, push a frame (slot held), re-register the same connector, and assert the held slot was released (no stale `_latest_slots` entry) and that a fresh frame still lands afterward. Mirrors the existing `test_participant_leave_releases_held_slots` (#143).

## Notes
- No API/dependency changes (`DEPENDENCIES.md` untouched). Changelog updated.
- Same single-process, single-loop assumptions as the rest of the hub (no new locking introduced).

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -9,6 +9,20 @@ Significant decisions, in reverse-chronological order. Update this whenever a
 non-trivial architectural or design decision is made so the rationale is
 preserved and not re-litigated.
 
+### 2026-06-05 — Hub: release held slots before closing a re-registered ring
+
+On `CONNECTOR_REGISTER` for a known `connector_id`, `_handle_registration`
+closed and replaced the ring buffer while `_latest_slots` could still hold
+`SlotView`s backed by that ring's mmap. Because a live `SlotView` keeps a
+sliced memoryview exported, `ShmRingBuffer.close()`'s `self._buf.release()`
+raises `BufferError` — leaving the old ring half-closed but still referenced,
+so the next `FRAME_SIGNAL` writes through a released/closing buffer
+(use-after-close). `_handle_registration` now releases the memoryview and slot
+for every `_latest_slots` entry backed by the old ring before closing it,
+matching the teardown order already in `close()`. Triggered by a connector
+crash/reconnect that re-sends its registration while frames are held.
+Regression test: `test_connector_reregistration_releases_held_slots`. Fixes #197.
+
 ### 2026-06-05 — piper voice fetch: catch LocalEntryNotFoundError before EntryNotFoundError
 
 Follow-up to #184. That PR added a dedicated `_EXIT_VOICE_UNAVAILABLE = 3`
--- a/server-runtime/xr_media_hub/ipc/_hub.py
+++ b/server-runtime/xr_media_hub/ipc/_hub.py
@@ -329,7 +329,19 @@ class HubEndpoint:
     def _handle_registration(self, reg: ConnectorRegistration) -> None:
         if reg.connector_id in self._ring_registry:
             logger.warning("Connector {} re-registered — replacing ring buffer", reg.connector_id)
-            self._ring_registry[reg.connector_id].close()
+            old_ring = self._ring_registry[reg.connector_id]
+            # Drop any frames still held in the old ring BEFORE closing it.
+            # A live SlotView keeps a sliced memoryview exported into the ring's
+            # mmap; closing with that outstanding makes ShmRingBuffer.close()'s
+            # self._buf.release() raise BufferError, and leaves _latest_slots
+            # referencing a half-closed ring whose slot the next FRAME_SIGNAL
+            # would write through (#197). Release the memoryview before the slot
+            # and close, matching the teardown order in close().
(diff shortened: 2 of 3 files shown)


nvddr:
## Review

Fixes a use-after-close in the server-runtime IPC hub (`HubEndpoint`). When a connector re-registers under a known `connector_id` (crash/reconnect with frames in flight), `_handle_registration` previously closed and replaced the connector's `ShmRingBuffer` without first dropping the `SlotView`s in `_latest_slots` still backed by that ring's mmap. A live `SlotView` holds a sliced memoryview exported from the ring's buffer, so `ShmRingBuffer.close()`'s `self._buf.release()` raised `BufferError`, leaving the old ring half-closed but still referenced — the next `FRAME_SIGNAL` would then `read_slot`/`release_slot` through a released mmap. The fix releases the exported memoryview (`view.data.release()`) and frees the slot for every `_latest_slots` entry backed by the old ring *before* `old_ring.close()`, matching the teardown order in `HubEndpoint.close()`. Symbols verified against `_shm.py` at head SHA; the regression test genuinely exercises the path (held view's memoryview truly exported from the hub's old ring), and the key-materializing list comprehension makes pop-during-iteration safe.

|---|---|---|
| Nit | [_hub.py#L348](https://github.com/NVIDIA/xr-ai/blob/59de46d91353716010d1a2c8965adb4f4a02fc62/server-runtime/xr_media_hub/ipc/_hub.py#L348) | Pre-existing (not introduced here), adjacent to the change: if the subsequent `ShmRingBuffer(name=reg.shm_name, create=False)` reopen raises, `_ring_registry[reg.connector_id]` is never reassigned and keeps pointing at the just-closed `old_ring`; that connector then fails every later `FRAME_SIGNAL` and never recovers. Now degrades to a caught error rather than #197's corruption, but the stale entry persists — consider `pop`-ing it on reopen failure. Out of scope; flagging for awareness. |

Otherwise clean — teardown order, comment, changelog, and regression test all accurate.

nvddr (approved):

PR #203, opened 2026-06-08
fix(piper-tts): valid empty WAV on empty/whitespace input (#194)

### LINKED ISSUE #194: Piper TTS returns HTTP 500 (wave.Error) on empty/whitespace input
## Summary
Piper TTS returns an unhandled HTTP 500 on empty or whitespace-only input.

## Details
In `ai-services/tts/piper/piper_tts_server/__main__.py`, `_PiperBackend.synthesize` (~L178-184, reached via the `/v1/audio/speech` handler ~L221-224) opens `wave.open(buf, "wb")` and calls `self._voice.synthesize_wav(text, wf)`. Piper's `synthesize_wav` only sets the WAV format params (`setnchannels`/`setsampwidth`/`setframerate`) inside its loop over synthesized chunks, on the first chunk.

For empty/whitespace input, `phonemize("")` yields no phonemes, the loop body never runs, the format params are never set, and no frames are written. When the `with wave.open(...)` block exits, `wave.close()` raises `wave.Error: # channels not specified`. That exception propagates out of `run_in_executor` into the async handler with no try/except, surfacing as an unhandled HTTP 500.

(Magpie does not share this — `sf.write` emits a valid header even for empty audio.)

## Suggested fix
Reject empty `req.input` with a 400 in the handler, or set `wf.setnchannels(1); wf.setsampwidth(2); wf.setframerate(self.sample_rate)` before `synthesize_wav` so an empty-but-valid WAV is produced.

---

wenxind-nvidia:
Fixes #194.

## Problem
`_PiperBackend.synthesize` (`ai-services/tts/piper/piper_tts_server/__main__.py`, reached via `/v1/audio/speech`) returned an unhandled **HTTP 500** on empty/whitespace input. Piper's `synthesize_wav` sets the WAV format params (`setnchannels`/`setsampwidth`/`setframerate`) only inside its per-chunk loop, on the first chunk. For empty input there are no chunks, so the params are never set and `wave.close()` raises `wave.Error: # channels not specified` — which propagates out of `run_in_executor` into the async handler with no try/except → 500.

(Magpie doesn't share this: `sf.write` emits a valid header even for empty audio.)

## Fix
Short-circuit empty/whitespace input to a valid, empty (silent) WAV — set the header params and write no frames — matching magpie's behavior. The non-empty path is byte-for-byte unchanged:

```python
with wave.open(buf, "wb") as wf:
    if text.strip():
        self._voice.synthesize_wav(text, wf)
    else:
        wf.setnchannels(1)
        wf.setsampwidth(2)
        wf.setframerate(self.sample_rate)
```

I chose "return a valid empty WAV" over "400 on empty input" because the sample TTS pipelines batch synthesis per sentence and can occasionally emit whitespace; a graceful empty WAV avoids turning that into client-visible errors, and matches the magpie contract.

## Test
`tests/test_piper_tts.py::test_piper_tts_smoke` now also POSTs whitespace-only input and asserts a `200` + `RIFF` WAV header (`_post_speech` asserts status 200, so a 500 fails the test). Runs end-to-end against the real server, under the same voice-availability gating as the rest of the smoke test.

## Notes
- No API/dependency changes (`DEPENDENCIES.md` untouched). Changelog updated.

### DIFF
--- a/ai-services/tts/piper/piper_tts_server/__main__.py
+++ b/ai-services/tts/piper/piper_tts_server/__main__.py
@@ -180,7 +180,17 @@ class _PiperBackend:
         self._ensure_loaded()
         buf = io.BytesIO()
         with wave.open(buf, "wb") as wf:
-            self._voice.synthesize_wav(text, wf)
+            if text.strip():
+                self._voice.synthesize_wav(text, wf)
+            else:
+                # Empty/whitespace input yields no synthesized chunks, and
+                # Piper's synthesize_wav sets the WAV format params only on the
+                # first chunk — so they'd never be set and wave.close() would
+                # raise "# channels not specified" (→ unhandled HTTP 500, #194).
+                # Emit a valid, empty (silent) WAV instead, matching magpie.
+                wf.setnchannels(1)
+                wf.setsampwidth(2)
+                wf.setframerate(self.sample_rate)
         return buf.getvalue()
 
 
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -9,6 +9,19 @@ Significant decisions, in reverse-chronological order. Update this whenever a
 non-trivial architectural or design decision is made so the rationale is
 preserved and not re-litigated.
 
+### 2026-06-05 — Piper TTS: valid empty WAV on empty/whitespace input
+
+`_PiperBackend.synthesize` returned an HTTP 500 for empty/whitespace input.
+Piper's `synthesize_wav` sets the WAV format params only on the first
+synthesized chunk; empty input produces no chunks, so the params were never
+set and `wave.close()` raised `wave.Error: # channels not specified`, which
+propagated out of the `/v1/audio/speech` handler unhandled. `synthesize` now
+short-circuits empty/whitespace input to a valid, empty (silent) WAV — header
+params set, zero audio frames — matching the magpie backend (whose `sf.write`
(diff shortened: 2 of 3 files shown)


nvddr:
## Review

Fixes an unhandled HTTP 500 in the Piper TTS backend (`/v1/audio/speech`). Piper's `synthesize_wav` only sets the WAV format params (`setnchannels`/`setsampwidth`/`setframerate`) inside its per-chunk loop, so empty/whitespace input — which yields zero chunks — leaves the params unset and `wave.close()` raises `wave.Error: # channels not specified`, surfacing from the executor as a 500. The fix short-circuits empty/whitespace input (`text.strip()` falsy) to write a valid zero-frame mono 16-bit WAV header, leaving the non-empty path byte-for-byte unchanged. This matches the magpie backend's output contract (magpie's `sf.write` already emits a valid header for empty audio). Untrusted input is the user-supplied `req.input`; the change strictly narrows the failure surface. A regression assertion is added to the smoke test (POST whitespace → expect 200 + `RIFF`). Verified the magpie parity claim and that the `pcm` response path also handles the empty WAV cleanly (`sf.read` → 0-length int16 array → empty body, no error).

|---|---|---|
| _(none)_ | — | Clean, minimal, well-tested. No blockers, bugs, or nits. |

nvddr (approved):

PR #205, opened 2026-06-08
fix(token-server): swallow uvicorn SystemExit so shutdown stays graceful (#192)

### LINKED ISSUE #192: TokenServer.stop() re-raises uvicorn SystemExit on bind failure, aborting graceful shutdown
## Summary
`TokenServer.stop()` can re-raise uvicorn's `SystemExit` (raised on a bind failure), aborting the rest of graceful shutdown.

## Details
In `server-runtime/xr_media_hub/transport/livekit/_token_server.py`, the task runs `self._server.serve()` directly (~L87). uvicorn calls `sys.exit(1)` on a bind failure (e.g. port already in use), so the task terminates with `SystemExit` (a `BaseException`). In `stop()`, `await self._task` (~L98) re-raises that `SystemExit` into the caller (`LiveKitConnector.stop()`), aborting the remaining graceful-shutdown steps.

The sibling `WebServer` already guards against exactly this with a `_serve_safe()` wrapper (`_web_server.py` ~L139-148) and an explanatory comment; `TokenServer` is missing the equivalent guard.

## Reproduction
Start the token server on an already-bound port, then trigger shutdown.

## Suggested fix
Wrap `self._server.serve()` in a `_serve_safe`-style coroutine that catches `SystemExit`, matching `WebServer`.

---

wenxind-nvidia:
Fixes #192.

## Problem
In `server-runtime/xr_media_hub/transport/livekit/_token_server.py`, the task runs `self._server.serve()` directly. uvicorn calls `sys.exit(1)` on a bind failure (e.g. port already in use), so the task terminates with `SystemExit` (a `BaseException`). In `stop()`, `await self._task` re-raises that `SystemExit` into the caller (`LiveKitConnector.stop()`), aborting the remaining graceful-shutdown steps.

The sibling `WebServer` already guards this with a `_serve_safe()` wrapper (`_web_server.py`); `TokenServer` was missing the equivalent.

## Fix
Route the task through a `_serve_safe()` coroutine that catches `SystemExit` and logs a clear error, mirroring `WebServer._serve_safe`:

```python
self._task = asyncio.create_task(self._serve_safe())
...
async def _serve_safe(self) -> None:
    try:
        await self._server.serve()
    except SystemExit as exc:
        logger.error(
            "Token server failed to start on port {} — is it already in use? (exit code {})",
            self._cfg.token_server_port, exc.code,
        )
```

`stop()` is unchanged: `await self._task` now sees a normally-completed task, so graceful shutdown proceeds.

## Notes
- Mirrors the established `WebServer._serve_safe` pattern in the same package; no API/dependency changes (`DEPENDENCIES.md` untouched). Changelog updated.
- No automated test: reproducing requires binding the token-server port and triggering connector shutdown (a live LiveKit transport path); the fix is a direct mirror of the already-present `WebServer` guard. On-stack verification: start the hub with the token-server port already occupied and confirm shutdown completes without the `SystemExit` traceback.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -86,7 +86,7 @@ still showed "Streaming" with a green status and a working Stop button. The
 `catch` now sets `isCameraActive = false`, matching the consistency that
 `startCamera()`/`stopCamera()` already maintain. Fixes #195.
 
-### 2026-06-05 — TokenServer: swallow uvicorn SystemExit so shutdown stays graceful
+### 2026-06-05 — TokenServer: fail startup loudly on bind error; keep shutdown graceful
 
 `TokenServer` ran `self._server.serve()` directly as its task. uvicorn calls
 `sys.exit(1)` on a bind failure (e.g. port already in use), so the task ends
@@ -94,7 +94,13 @@ with `SystemExit` (a `BaseException`); `stop()`'s `await self._task` then
 re-raised it into `LiveKitConnector.stop()`, aborting the remaining
 graceful-shutdown steps. Wrapped the task in a `_serve_safe()` coroutine that
 catches `SystemExit` and logs a clear "port in use?" error — mirroring the
-sibling `WebServer._serve_safe`, which already guards this. Fixes #192.
+sibling `WebServer._serve_safe`, which already guards this. To stop a bind
+failure from looking healthy, `start()` now awaits the bind (polls
+`uvicorn.Server.started` until the serve task either binds or finishes) and
+raises `RuntimeError` if the server never bound within `_STARTUP_TIMEOUT_S` —
+the token server is the browser-facing auth/signaling entry point, so a dead
+endpoint must abort connector startup rather than silently swallow the error.
+Fixes #192.
 
 ### 2026-06-05 — STT: serialize NeMo transcribe() on the shared model
 
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -9,6 +9,16 @@ Significant decisions, in reverse-chronological order. Update this whenever a
 non-trivial architectural or design decision is made so the rationale is
 preserved and not re-litigated.
 
+### 2026-06-05 — TokenServer: swallow uvicorn SystemExit so shutdown stays graceful
+
+`TokenServer` ran `self._server.serve()` directly as its task. uvicorn calls
+`sys.exit(1)` on a bind failure (e.g. port already in use), so the task ends
+with `SystemExit` (a `BaseException`); `stop()`'s `await self._task` then
+re-raised it into `LiveKitConnector.stop()`, aborting the remaining
+graceful-shutdown steps. Wrapped the task in a `_serve_safe()` coroutine that
+catches `SystemExit` and logs a clear "port in use?" error — mirroring the
+sibling `WebServer._serve_safe`, which already guards this. Fixes #192.
+
 ### 2026-06-05 — piper voice fetch: catch LocalEntryNotFoundError before EntryNotFoundError
 
 Follow-up to #184. That PR added a dedicated `_EXIT_VOICE_UNAVAILABLE = 3`
--- a/server-runtime/xr_media_hub/transport/livekit/_token_server.py
+++ b/server-runtime/xr_media_hub/transport/livekit/_token_server.py
@@ -84,12 +84,25 @@ class TokenServer:
             scheme = "http"
 
         self._server = uvicorn.Server(uvicorn.Config(**uv_cfg))
-        self._task = asyncio.create_task(self._server.serve())
+        self._task = asyncio.create_task(self._serve_safe())
         logger.info(
             "Token server → {}://{}:{}  room={!r}",
             scheme, self._cfg.token_server_host, self._cfg.token_server_port,
             self._cfg.room_name,
         )
+
+    async def _serve_safe(self) -> None:
+        # uvicorn calls sys.exit(1) on bind failure; SystemExit is a BaseException
+        # that ``await self._task`` in stop() would re-raise into the caller,
+        # aborting the rest of graceful shutdown. Swallow it here (matches
+        # WebServer._serve_safe).
+        try:
+            await self._server.serve()
+        except SystemExit as exc:
+            logger.error(
+                "Token server failed to start on port {} — is it already in use? (exit code {})",
+                self._cfg.token_server_port, exc.code,
+            )
 
     async def stop(self) -> None:
         if self._server:
--- a/server-runtime/xr_media_hub/transport/livekit/_token_server.py
+++ b/server-runtime/xr_media_hub/transport/livekit/_token_server.py
@@ -24,6 +24,11 @@ from . import _lk_proxy
 from ._token import make_client_token
 from .config import LiveKitConnectorConfig
 
+# Max seconds start() waits for uvicorn to bind before treating startup as
+# failed. Binding the token port is near-instant (no model load); the bound is
+# a backstop for a startup that neither binds nor exits.
+_STARTUP_TIMEOUT_S = 10.0
+
 
 def build_app(cfg: LiveKitConnectorConfig) -> FastAPI:
     lk_internal_http = f"http://127.0.0.1:{cfg.lk_port_ws}"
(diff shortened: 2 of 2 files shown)


nvddr:
## Review

Fixes a graceful-shutdown bug in the LiveKit transport's browser-facing `TokenServer`. Its uvicorn instance runs as an asyncio task; on a bind failure uvicorn calls `sys.exit(1)`, so the task completes with `SystemExit` (a `BaseException`). `LiveKitConnector.stop()` awaits `TokenServer.stop()` (which awaits the task) before its final `await docker_task`, so the re-raised `SystemExit` would abort the remaining shutdown steps. The fix wraps `serve()` in a `_serve_safe()` coroutine that catches `SystemExit` and logs a clear "port in use?" error — an exact mirror of the existing `WebServer._serve_safe` guard in the sibling module. Verified the data flow in `connector.py` and confirmed against uvicorn 0.29 source (`startup()` → `OSError` → `sys.exit(1)`) that the caught exception and `exc.code` are correct. No untrusted input.

|---|---|---|
| Nit | [_token_server.py#L94](https://github.com/NVIDIA/xr-ai/blob/98e80ac6edf9c716bf2eeb0a1ccb2845a6be7efc/server-runtime/xr_media_hub/transport/livekit/_token_server.py#L94) | Swallowing `SystemExit` also absorbs a startup bind failure: `start()` still logs "Token server → …" and the connector proceeds through `register()`/`connect()` as healthy while the token endpoint is dead, so browser clients silently fail to fetch tokens with only a logged error. Matches the intentional degrade-don't-crash behavior `WebServer` already has — pattern note for awareness, not a regression or blocker. |

Correct as written. No blockers or potential bugs.

yanziz-nvidia (approved):

nvddr:
## Follow-up — escalating the silent-startup-failure (correcting my earlier "nit")

I under-rated this in my first pass. Framing the silent bind failure as "matches `WebServer`'s degrade-don't-crash pattern" was wrong — *another component swallowing it too is a second bug, not a justification*, and the token server is too critical for it. If the port is in use, `_serve_safe` logs one line and returns, `start()` has already logged `"Token server → …"` and returns normally, and `LiveKitConnector` proceeds through `register()`/`connect()` reporting healthy — while **every browser client silently fails to fetch a token**. The whole browser entry point is down with no failed health state and nothing but a buried `logger.error`. For the browser-facing auth/signaling entry point, startup failure must be loud.

The PR's actual goal is sound and should stay: catching `SystemExit` in `_serve_safe` so a re-raise through `await self._task` in `stop()` doesn't abort graceful shutdown. The bug is conflating that with *masking a startup failure*. Separate the two: swallow `SystemExit` only to keep shutdown clean, but make `start()` detect that the server never bound and fail loudly.

This is deterministic to detect — `uvicorn.Server.started` stays `False` on bind failure (`startup()` calls `sys.exit(1)` before setting `self.started = True`). Suggested shape:

```python
async def start(self) -> None:
    ...
    self._server = uvicorn.Server(uvicorn.Config(**uv_cfg))
    self._task = asyncio.create_task(self._serve_safe())

    # Wait until the server is actually bound (or the task died trying).
    while not self._server.started and not self._task.done():
        await asyncio.sleep(0.05)
    if self._task.done():
        # serve() exited during startup → bind failed; surface it loudly so
        # the connector aborts instead of coming up with a dead token endpoint.
        raise RuntimeError(
            f"Token server failed to bind {self._cfg.token_server_host}:"
            f"{self._cfg.token_server_port} (already in use?)"
        )
    logger.info("Token server → {}://{}:{} room={!r}", ...)
```

`_serve_safe` keeps the `SystemExit` catch as-is (it still prevents the shutdown re-raise). Net: graceful shutdown stays fixed, and a bind failure now aborts connector startup with a clear error instead of a silently-dead token endpoint. (A timeout bound on the wait loop would make it fully robust against a startup that neither binds nor exits.)

|---|---|---|
| Blocker | [_token_server.py#L99](https://github.com/NVIDIA/xr-ai/blob/98e80ac6edf9c716bf2eeb0a1ccb2845a6be7efc/server-runtime/xr_media_hub/transport/livekit/_token_server.py#L99) | Bind failure on the browser-facing token server is swallowed: `start()` returns healthy and the connector continues, leaving the token/signaling entry point dead with only a logged error → all browser clients silently fail to connect. Detect non-binding in `start()` (`await` until `self._server.started` or `self._task.done()`; raise if the task finished) and abort startup loudly. Keep the `SystemExit` catch in `_serve_safe` for the shutdown path only. |

nvddr (changes requested):
Please fix blocker

wenxind-nvidia:
Thanks @nvddr — fixed the blocker exactly as suggested, plus the timeout bound you noted.

`start()` now waits for the bind to actually happen before reporting healthy: it polls `self._server.started` until the serve task either binds or finishes, and raises `RuntimeError` if the server never bound within `_STARTUP_TIMEOUT_S` (10s). So a port-in-use (or any startup that neither binds nor exits) now aborts `LiveKitConnector` startup loudly with a clear error instead of coming up with a dead token endpoint.

`_serve_safe` is unchanged — it still swallows `SystemExit` so the *shutdown* re-raise through `await self._task` in `stop()` stays clean. The `logger.info("Token server → …")` line now runs only after the bind is confirmed.

```python
self._server = uvicorn.Server(uvicorn.Config(**uv_cfg))
self._task = asyncio.create_task(self._serve_safe())

loop = asyncio.get_running_loop()
deadline = loop.time() + _STARTUP_TIMEOUT_S
while not self._server.started and not self._task.done():
    if loop.time() >= deadline:
        break
    await asyncio.sleep(0.05)
if not self._server.started:
    raise RuntimeError(
        f"Token server failed to start on "
        f"{self._cfg.token_server_host}:{self._cfg.token_server_port} "
        "— port already in use, or startup timed out."
    )
logger.info("Token server → …")  # only after bind confirmed
```

Changelog entry updated to describe the loud-startup-failure behavior. Pushed.

nvddr (approved):

nvddr:
## Re-review — blocker resolved ✅

Thanks for the fix. The graceful-shutdown win is intact (`_serve_safe()` still swallows uvicorn's bind-failure `SystemExit` so `stop()`'s `await self._task` doesn't re-raise), and `start()` now adds the startup-failure detection: it polls `while not self._server.started and not self._task.done()` (bounded by `_STARTUP_TIMEOUT_S = 10s`) and raises `RuntimeError` if the server never bound. Since `LiveKitConnector.start()` awaits `self._token.start()` *before* `register()`/`connect()`, a bind failure now aborts connector startup loudly instead of coming up healthy with a dead token endpoint. **The silent-startup-failure blocker is resolved.** Detection verified correct against uvicorn source: `startup()` calls `sys.exit(1)` on bind `OSError` (`server.py:182`) before `self.started = True` (`server.py:195`), so on failure the task finishes with `started` still False — no race, sane 10s backstop.

|---|---|---|
| Nit | [_token_server.py#L100-L113](https://github.com/NVIDIA/xr-ai/blob/c96dc175c7c76a4171da5bc8668c9e4081a1b69f/server-runtime/xr_media_hub/transport/livekit/_token_server.py#L100-L113) | Pathological-timeout branch only: if startup neither binds nor exits within 10s, `start()` raises while the serve task is still running and never cancels/awaits it (orphaned task). The real bind-failure path is clean (task is `done()`). Backstop case; non-blocking. Optional: `self._task.cancel()` before raising on timeout. |
| Nit | [_token_server.py#L130-L138](https://github.com/NVIDIA/xr-ai/blob/c96dc175c7c76a4171da5bc8668c9e4081a1b69f/server-runtime/xr_media_hub/transport/livekit/_token_server.py#L130-L138) | `_serve_safe` only catches `SystemExit`. A non-`SystemExit` serve failure ends the task, `start()` raises its generic `RuntimeError` (masking the real cause) and the task's exception is never retrieved ("exception was never retrieved" warning). Edge case; non-blocking. Optional: catch `Exception` too and surface it. |

No blockers or potential bugs remain.

wenxind-nvidia:
Thanks for the thorough re-review @nvddr — both nits addressed.

**Nit 1 — orphan serve task on timeout.** The not-started branch now distinguishes the two cases. The real bind-failure path leaves the task `done()`, so nothing to clean up. The pathological *neither-binds-nor-exits* path cancels and awaits the task before raising, then drops the reference so a later `stop()` doesn't re-await (and re-raise) the cancelled task:

```python
if not self._server.started:
    if not self._task.done():
        self._task.cancel()
        try:
            await self._task
        except asyncio.CancelledError:
            pass
    self._task = None
    self._server = None
    raise RuntimeError(...) from self._serve_error
```

**Nit 2 — non-`SystemExit` failure masked / exception never retrieved.** `_serve_safe` now also catches `Exception` (not `BaseException` — `CancelledError` is intentionally left to propagate so the timeout-cancel above works cleanly) and records it in `self._serve_error`. So the task's exception is always retrieved (no warning), and `start()` chains the real cause instead of masking it:

```python
except SystemExit as exc:
    self._serve_error = exc
    logger.error("… is it already in use? (exit code {})", port, exc.code)
except Exception as exc:
    self._serve_error = exc
    logger.error("Token server crashed on port {}: {!r}", port, exc)
```

`raise RuntimeError(...) from self._serve_error` chains the captured cause (a populated `SystemExit`/`Exception` on real failures, `None` on a pure timeout). Changelog updated; rebased on main, CI green.

PR #210, opened 2026-06-09
fix(ios): guard switchCamera with startCamera's isCameraStarting re-entrancy flag (#208)

### LINKED ISSUE #208: iOS sample: switchCamera(to:) lacks startCamera's isCameraStarting re-entrancy guard
### Summary

`AppModel.switchCamera(to:)` (iOS/visionOS sample) lacks the `isCameraStarting` guard/flag that `startCamera()` uses, so a camera switch racing a concurrent start/stop is unguarded and can interleave backend calls.

### Context

Surfaced during review of #200 (which fixed the `isCameraActive` state-consistency bug in `switchCamera(to:)`'s `catch`). That fix is correct and merged; this is a separate, **pre-existing** concern noted for follow-up — out of scope for the one-line #200 change.

`startCamera()` guards re-entrancy with `isCameraStarting`; `switchCamera(to:)` re-invokes the backend's `startCamera()` (which tears down the active track before publishing the new one) without taking that same guard. A switch overlapping a concurrent start/stop is therefore unserialized.

### Location

`client-samples/ios-visionos/App/AppModel.swift` — `switchCamera(to:)` (~L268).

### Suggested fix

Apply the same `isCameraStarting` guard/flag pattern `startCamera()` uses to `switchCamera(to:)`, so concurrent start / stop / switch are serialized.

### Severity

Low / nit — iOS-visionOS sample app only; requires a concurrent start/stop racing a switch to trigger. No security or data-integrity impact.

wenxind-nvidia:
Fixes #208. Follow-up to #200 (flagged by @nvddr in that review).

## Problem
`AppModel.switchCamera(to:)` (`client-samples/ios-visionos/App/AppModel.swift`) re-invokes the backend's `startCamera()` — which tears down the active track before publishing the new one — but, unlike `startCamera()`, took no `isCameraStarting` re-entrancy guard. So a switch overlapping a concurrent start/stop could interleave backend calls across `await` suspension points on the `@MainActor` model.

## Fix
Apply the same guard/flag pattern `startCamera()` already uses:

```swift
func switchCamera(to position: CameraConfig.Position) async {
    cameraPosition = position
    guard isCameraActive, !isCameraStarting else { return }
    isCameraStarting = true
    defer { isCameraStarting = false }
    do {
        try await session?.startCamera(config: CameraConfig(position: cameraPosition))
    } catch {
        lastError = error.localizedDescription
        isCameraActive = false   // (#200) backend stopped the old track; nothing is streaming
    }
}
```

Now `startCamera()` (`guard !isCameraStarting`) and `switchCamera` both refuse to start while the other is in flight — start and switch are serialized. The persisted `cameraPosition` is still updated up front (unchanged), and the #200 `catch` state reset is preserved.

## Notes
- Sample-only, low severity (needs a concurrent start/stop racing a switch to trigger; no security/data-integrity impact). Mirrors the established `startCamera()` pattern.
- No automated test — iOS Swift, no Swift test harness in CI; on-device verification is rapidly toggling start / switch. Changelog updated; no dependency changes.

### DIFF
--- a/client-samples/ios-visionos/App/AppModel.swift
+++ b/client-samples/ios-visionos/App/AppModel.swift
@@ -267,7 +267,13 @@ final class AppModel {
 
     func switchCamera(to position: CameraConfig.Position) async {
         cameraPosition = position
-        guard isCameraActive else { return }
+        // Serialize against a concurrent start/switch via the same flag
+        // startCamera() uses. switchCamera re-invokes the backend's
+        // startCamera() (which tears down the active track before publishing
+        // the new one), so an overlapping start must not interleave (#208).
+        guard isCameraActive, !isCameraStarting else { return }
+        isCameraStarting = true
+        defer { isCameraStarting = false }
         do {
             try await session?.startCamera(config: CameraConfig(position: cameraPosition))
         } catch {
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -9,6 +9,17 @@ Significant decisions, in reverse-chronological order. Update this whenever a
 non-trivial architectural or design decision is made so the rationale is
 preserved and not re-litigated.
 
+### 2026-06-05 — iOS sample: guard switchCamera against concurrent start/switch
+
+Follow-up to #200. `AppModel.switchCamera(to:)` re-invokes the backend's
+`startCamera()` (which tears down the active track before publishing the new
+one) but, unlike `startCamera()`, took no `isCameraStarting` re-entrancy guard
+— so a switch overlapping a concurrent start could interleave backend calls
+across `await` suspension points on the main actor. Applied the same
+`guard … !isCameraStarting` + `isCameraStarting = true` / `defer` pattern
+`startCamera()` uses, so start and switch are serialized. Sample-only, low
+severity. Fixes #208.
+
 ### 2026-06-05 — Native StreamKit: Android NDK C++20 portability
 
 Native StreamKit no longer depends on C++20 `<format>`. Some Android NDK


nvddr:
## Review

Sample-only iOS concurrency hardening (fixes #208). `AppModel` is `@MainActor @Observable`, so the only interleaving points are `await` suspensions — no true parallelism — which makes a re-entrancy flag the correct, sufficient mechanism. `switchCamera(to:)` re-invokes the backend's `startCamera()` (tear-down-then-republish), so it shares the hazard `startCamera()` already guards. The fix copies `startCamera()`'s `guard !isCameraStarting` + `isCameraStarting = true` / `defer` pattern verbatim, serializing start and switch against each other. The success path's reliance on `isCameraActive` already being true (it never sets it) mirrors the inverse of `startCamera`, unchanged from prior behavior. Changelog updated, no dep changes. Correctly addresses the root cause.

|---|---|---|
| Nit | [AppModel.swift#L269](https://github.com/NVIDIA/xr-ai/blob/871c9f6eb7e7df4b2dbcb007362562cb710a91c9/client-samples/ios-visionos/App/AppModel.swift#L269) | `cameraPosition = position` is persisted *before* the `guard … !isCameraStarting`, so a switch issued while a start/switch is in flight saves the new position but returns without applying it — persisted position can desync from the live stream until the next start. Acknowledged in the PR body and unchanged from prior behavior; noted for completeness. |

No blockers or potential bugs. (The pre-existing `stopCamera`/`disconnect`-racing-a-switch window is not introduced or affected by this diff — out of scope.)

wenxind-nvidia:
Thanks @nvddr. On the nit (`cameraPosition = position` persisted before the `!isCameraStarting` guard): leaving it as-is intentionally.

The early assignment is load-bearing for the **camera-off** path — when `!isCameraActive`, the picker selection must still persist (via `cameraPosition`'s `didSet` → UserDefaults) so the next `startCamera()` honours it. Moving the assignment after the guard would drop that selection whenever the camera is off, which is the common case for choosing a default.

The desync you flagged is confined to the narrow in-flight (`isCameraStarting == true`) window and self-heals on the next start. Closing it properly means coalescing the latest requested position and re-applying it when the in-flight start/switch completes — extra state and a re-check loop in two methods, which is more machinery than this sample warrants for a transient, self-correcting UI inconsistency. Noted in the PR body as known/by-design.

Also brought the branch up to date with `main`; CI is green.

nvddr (approved):

PR #217, opened 2026-06-09
feat(nemo): opt-in docker backend for STT + magpie to escape host cuDNN/CUDA

nvddr:
## Why

The NeMo GPU services (stt-server, magpie TTS) run in-venv and `import torch`, so they inherit the host's cuDNN/CUDA via `LD_LIBRARY_PATH`. On a host whose system cuDNN differs from torch's bundled one they abort at import:

```
RuntimeError: cuDNN version incompatibility: PyTorch was compiled against (9, 20, 0) but found runtime version (9, 13, 1)
```
(observed on an L40S cloud box). The vLLM services already escape this by running in NGC containers; this gives the NeMo services the same option.

**Relationship to #216:** #216 (already merged) strips a conflicting `libcudnn` dir off `LD_LIBRARY_PATH` in the launcher, which fixes the common in-venv case. This PR is the **complete** isolation: an opt-in container backend immune to *all* host CUDA/cuDNN/NCCL/driver-userspace mismatches, for hosts where the strip isn't enough or where full isolation is wanted. Default stays `pip` — **no behavior change unless opted in.**

## What

- **New stdlib-only package `utils/xr-ai-nemo-runtime/`** — a bespoke NeMo container runner (separate from `xr_ai_vllm`, which is untouched). Builds the `docker run` argv against an NGC NeMo image, waits on `/health`, touches the ready-file, streams container logs, and self-stops+removes its container on SIGINT/SIGTERM (mirrors the #215 vLLM signal handler) so Ctrl-C during a slow pull/download cleans up.
- **Opt-in `backend: pip|docker`** on `stt_server.yaml` / `magpie_tts_server.yaml` (+ the 3 per-profile stt copies). Default `pip` = today's in-venv path, unchanged. `docker` routes to the runner.
- **In-container packaging:** the NeMo image has torch+nemo+cuDNN but not our server, so the run command bind-mounts the repo read-only, puts the server package + `xr-ai-logging` on `PYTHONPATH`, `pip install`s only the light deps the image lacks (`fastapi`, `uvicorn`, `hf_transfer`, `loguru`, `pyyaml`, + per-server extras) — **never** the server's own pyproject (which would drag `nemo_toolkit`/torch and conflict) — then runs `python -m <server> --_serve --config <mounted>`. `--_serve` short-circuits the backend dispatch so the in-container process can't recursively spawn another container.
- **Lifecycle:** NeMo containers **die with the stack** (no cross-restart persistence — NeMo reloads in ~20 s from cached weights). On `--stop` the in-container server is killed and the wrapper's idle loop then stop+removes the (now-dead) container; a stale container from a prior run is cleared before `docker run` so relaunch never hits a name clash. The reaper in `xr_ai_vllm.stop_persistent_servers` filters on `xr-ai-vllm.port=`, so it can't reap these `xr-ai-nemo.port=` containers — hence the self-cleanup.
- piper TTS (CPU/ONNX) is **not** in scope.

## Tests

- `tests/test_nemo_docker.py` — argv-builder (image, name, `xr-ai-nemo.port=` label, `--network`/`--ipc host`, `--gpus`, repo `:ro` + model-cache mounts, HF/NeMo env, light-dep install + `python -m <server> --_serve`, never the pyproject, PYTHONPATH to mounts) + backend toggle for both servers (default-pip skips the runner; `docker` routes to it with the in-venv path NOT invoked; `--_serve` bypasses dispatch).
- `cd tests && uv run pytest test_nemo_docker.py test_vllm_docker.py test_launcher_stack.py -q` → **69 passed** (hermetic; no docker/GPU).
- `DEPENDENCIES.md` updated for the new path package (same commit).

## On-device gate

The real path — NeMo NGC image pull + cuDNN-clean model load + which of the light deps the image already ships + the exact `nvcr.io/nvidia/nemo:<tag>` for your account — can only be confirmed on a GPU host with docker + NGC access. `nemo_image` is a configurable YAML key with an example default (`nvcr.io/nvidia/nemo:25.04`) clearly marked "confirm a tag available to your account". Not asserting it pulls/loads in CI.

### DIFF
--- a/DEPENDENCIES.md
+++ b/DEPENDENCIES.md
@@ -105,6 +105,19 @@ xr-ai-vllm  (utils/xr-ai-vllm/)
     container.  Imported by the four vllm wrappers and by the orchestrator
     `--stop` flow.
 
+xr-ai-nemo-runtime  (utils/xr-ai-nemo-runtime/)
+    └── (stdlib only — zero runtime deps)
+    Opt-in NGC NeMo container backend for the two in-process NeMo servers
+    (stt-server, magpie TTS).  When a server's YAML sets `backend: docker` its
+    `run()` delegates here, which `docker run`s an NGC NeMo image, bind-mounts
+    the repo, pip-installs only the light deps the NeMo image lacks
+    (fastapi/uvicorn/hf_transfer/loguru/pyyaml + per-server extras), and runs
+    `python -m <server> --_serve` off PYTHONPATH — never the server's own
+    pyproject, so nemo_toolkit/torch are NOT reinstalled.  Stays stdlib-only so
+    docker mode adds nothing to the wrapper's venv.  Bespoke and separate from
+    `xr-ai-vllm` (different in-container command + packaging).  Imported by
+    stt-server and magpie-tts-server.
+
 xr-ai-vad  (utils/xr-ai-vad/)
     └── numpy >=1.24
     └── silero-vad >=5.1  (pulls torch + onnxruntime transitively)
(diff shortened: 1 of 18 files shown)


nvddr:
Closing — the host dependency-isolation need this addressed is already handled by #216 (merged), which strips a conflicting system cuDNN dir off `LD_LIBRARY_PATH` before spawning so each venv's bundled cuDNN wins. That resolves the original L40S `cuDNN version incompatibility` crash for the in-venv path without new container/build infrastructure.

A full NeMo container backend (this PR) is heavier (image build/caching, in-container packaging) than the problem currently warrants. Parking the approach; if we later want immunity from the whole host userspace (CUDA runtime / NCCL / system libs, not just cuDNN), the cheaper next step is a proactive "prepend the venv's nvidia/*/lib dirs to LD_LIBRARY_PATH" tweak before reaching for containers.

PR #218, opened 2026-06-10
fix(render-mcp): close LOVR launch context on respawn (fixes #196)

### LINKED ISSUE #196: render-mcp leaks ManagedProcess contexts on LOVR respawn (parked in app-lifetime AsyncExitStack)
## Summary
render-mcp leaks a `ManagedProcess` context on every LOVR respawn — dead contexts (with their pipe tasks and open log file) accumulate in the process-lifetime `AsyncExitStack` until the whole server shuts down.

## Details
In `agent-mcp-servers/render-mcp/render_mcp/__main__.py`, each successful `start_lovr_once` does `await self._stack.enter_async_context(ManagedProcess(...))` (~L219-221). `self._stack` is the **process-lifetime** `AsyncExitStack` from `_serve` (~L540).

When LOVR exits, `_watch` resets `_lovr_started = False` (~L229); a later `start_xr` re-enters a **new** `ManagedProcess` into the **same** stack. The previous `ManagedProcess`'s `finally` (terminate + pipe-task cancel + sink close) never runs until whole-process shutdown.

## Consequence
Across N LOVR restarts, N-1 dead `ManagedProcess` contexts accumulate, each holding two leaked `_forward` pipe tasks and an open log-file handle. Slow resource creep (the previous child is already exited, so no orphaned process).

## Suggested fix
Don't park the per-launch `ManagedProcess` in the app-lifetime stack. Use a dedicated `AsyncExitStack` (or explicit `__aexit__`) per launch and close it inside `_watch` after the child exits, before allowing respawn.

---

wenxind-nvidia:
## Problem

Fixes #196. `render-mcp` leaked a `ManagedProcess` context on every LOVR respawn.

In `agent-mcp-servers/render-mcp/render_mcp/__main__.py`, each successful `start_lovr_once` did `await self._stack.enter_async_context(ManagedProcess(...))`, where `self._stack` is the **process-lifetime** `AsyncExitStack` from `_serve`. When LOVR exits, `_watch` reset `_lovr_started = False` and a later `start_xr` entered a **new** `ManagedProcess` into the **same** stack — so the previous context's teardown (its two `_forward` pipe tasks + open log-file handle) never ran until whole-process shutdown. Across N restarts, N-1 dead contexts accumulated.

## Fix

- Each LOVR launch gets its **own** `AsyncExitStack` (`self._launch_stack`); the `ManagedProcess` is entered into that, not the app-lifetime stack.
- `_watch` closes the per-launch stack **as soon as the child exits**, before resetting `_lovr_started` (i.e. before a respawn is allowed) — running the `ManagedProcess` teardown (cancel pipe tasks, close log sink). This is safe post-exit: `ManagedProcess` skips `terminate()` when `returncode` is already set.
- A single `_aclose_live_launch` callback is registered **once** on the app-lifetime stack to cover the shutdown-while-LOVR-still-running case, so nothing accumulates per launch.

Net: the app-lifetime stack now holds one callback instead of one dead `ManagedProcess` per restart.

## Test

Added `test_lovr_respawn_closes_previous_launch_context` (in the existing local/`gpu`-marked `tests/test_local_render_mcp.py`, which already stubs LOVR). It asserts:
- after a simulated LOVR exit + `_watch`, the previous launch context is closed (`__aexit__` ran) and `_launch_stack` is cleared;
- a second `start_lovr_once` opens a fresh context with **no accumulation** (exactly one closed context);
- the still-live context is closed when the app-lifetime stack unwinds at shutdown.

No `pyproject.toml` change (stdlib only), so no `DEPENDENCIES.md` update.

> Note: the regression test is in the `gpu`-marked local suite (not in the CI matrix) and I couldn't run it in this environment (no local pytest/zmq). The change + test were verified by `py_compile` and inspection; please run `uv run pytest tests/test_local_render_mcp.py -k respawn` on a dev box to confirm.

### DIFF
--- a/agent-mcp-servers/render-mcp/render_mcp/__main__.py
+++ b/agent-mcp-servers/render-mcp/render_mcp/__main__.py
@@ -177,6 +177,15 @@ class SceneDispatcher:
         self._render_drops: int = 0
         self._spawn_error: str | None = None
         self._watch_task: asyncio.Task | None = None
+        # Per-launch context for the current LOVR child. Each launch gets its
+        # own AsyncExitStack so its ManagedProcess teardown (pipe-task cancel +
+        # log-sink close) can run on respawn instead of piling up in the
+        # app-lifetime stack until whole-process shutdown (issue #196).
+        self._launch_stack: contextlib.AsyncExitStack | None = None
+        # Safety net: close whatever launch context is live when the server
+        # shuts down (LOVR still running). Registered once on the app-lifetime
+        # stack so it never accumulates per launch.
+        self._stack.push_async_callback(self._aclose_live_launch)
 
         # Scene state: { id → { type, position, color, scale } }
         self._objects: dict[str, dict] = {}
@@ -216,9 +225,11 @@ class SceneDispatcher:
             logger.info(
                 "render-mcp: starting LOVR  bin={}  app={}", cfg.lovr_bin, cfg.xr_app_dir,
             )
-            lovr_proc = await self._stack.enter_async_context(
+            launch_stack = contextlib.AsyncExitStack()
+            lovr_proc = await launch_stack.enter_async_context(
                 ManagedProcess("lovr", lovr_cmd, cwd=cfg.xr_app_dir)
             )
+            self._launch_stack = launch_stack
 
             async def _watch() -> None:
                 rc = await lovr_proc.wait()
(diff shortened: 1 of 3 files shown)


nvddr:
## Review

Fixes a per-respawn context leak in render-mcp's LOVR child lifecycle (`SceneDispatcher`). Previously every successful `start_lovr_once` registered its `ManagedProcess` into the process-lifetime `AsyncExitStack`, so on each respawn the prior context's teardown (its two `_forward` pipe tasks + the open log-file sink) never ran until whole-process shutdown — leaking N-1 dead contexts across N restarts. The fix gives each launch its own `AsyncExitStack` (`self._launch_stack`), closes it inside `_watch` the moment the child exits (before `_lovr_started` is reset and a respawn is allowed), and registers a single app-lifetime `_aclose_live_launch` callback for the shutdown-while-LOVR-running case. Fits the `ManagedProcess`/stack model cleanly: `__aexit__` skips `terminate()` once `returncode` is set and always cancels pipe tasks + closes the sink, so post-exit `aclose()` is safe; the stack swaps are identity-guarded and `disp.close()` cancels `_watch` before the app stack unwinds, so no concurrent/double `aclose`. Verified against the real `xr_ai_launcher._processes.ManagedProcess` at head SHA; the regression test asserts the no-accumulation property (`sum(c.exited) == 1`) and the shutdown-time close.

|---|---|---|
| _(none)_ | — | Clean; addresses the root cause of #196 and is covered by a test. |

nvddr (approved):

PR #219, opened 2026-06-10
fix(render-mcp): block scene resync until LOVR connects (fixes #198)

### LINKED ISSUE #198: render-mcp scene resync after LOVR respawn silently dropped (PUSH+NOBLOCK with no connected peer)
## Summary
render-mcp's scene resync after a LOVR (re)spawn is silently dropped, because it sends on a PUSH socket with `NOBLOCK` before LOVR has connected.

## Details
In `agent-mcp-servers/render-mcp/render_mcp/__main__.py`, `start_lovr_once` sets `self._lovr_started = True` (~L233) then immediately `await self._resync_scene()` (~L237), which calls `forward("scene.add", …)` for every stored primitive. `forward` sends with `zmq.NOBLOCK` (~L324).

At that instant LOVR was just spawned via `create_subprocess_exec` and has **not yet connected** to the bound PUSH socket. A ZMQ PUSH socket with **zero connected peers** does not enqueue up to SNDHWM — it returns `EAGAIN` immediately under `NOBLOCK`. So every resync message raises `zmq.Again` and is dropped (counted only as `{"ok": False, "reason": "backpressure"}`).

The `SNDHWM=256` comment ("room for ~5 s … while LOVR starts up") is misleading: HWM buffering only applies **after** a peer connects, not before.

## Consequence
The point of `_resync_scene` — restoring primitives after a LOVR restart — fails. After a crash/respawn the scene comes back empty even though `_objects` still holds everything.

## Suggested fix
Defer resync until LOVR has actually connected: have LOVR send a ready/hello message the dispatcher waits for before resyncing, retry on a short backoff until sends succeed, or send the resync blocking (without `NOBLOCK`) so it queues once the peer attaches.

---
*Filed from an automated code-logic review. Confidence: medium. No assignee.*

wenxind-nvidia:
## Problem

Fixes #198. After a LOVR (re)spawn, `render-mcp`'s scene resync was silently dropped.

`start_lovr_once` sets `_lovr_started = True` then `await self._resync_scene()`, which re-pushed every stored primitive via `forward()`. `forward()` sends with `zmq.NOBLOCK`. But at that instant LOVR was just spawned and has **not yet connected** its PULL socket — and a ZMQ PUSH with **zero connected peers** does not enqueue up to `SNDHWM`; it returns `EAGAIN` immediately under `NOBLOCK`. So every resync message raised `zmq.Again` and was dropped (counted as `backpressure`). The `SNDHWM=256` "room while LOVR starts up" comment is misleading — HWM buffering only applies **after** a peer connects.

**Consequence:** after a crash/respawn the scene came back empty even though `_objects` still held everything.

## Fix

`_resync_scene` no longer routes through the live NOBLOCK `forward()` path. It now sends the restore as **blocking** sends:
- the first send waits until LOVR's PULL connects;
- the rest queue (up to `SNDHWM`) and return fast.

An overall `_RESYNC_TIMEOUT_S` (10s) deadline via `asyncio.wait_for` keeps a LOVR that never connects from wedging the spawn path (the spawn lock is held during resync). On timeout it logs an incomplete-resync warning instead of hanging.

The live `forward()` path is **unchanged** — NOBLOCK drop-on-backpressure remains correct for the 50 Hz scale-message stream.

## Tests

Added to the existing local/`gpu`-marked `tests/test_local_render_mcp.py` (real ZMQ ipc sockets):
- `test_resync_delivers_after_late_peer_connect` — starts resync with **no peer**, asserts it blocks (doesn't drop), then connects a PULL late and asserts **all** `scene.add` messages arrive. This is the exact scenario the old NOBLOCK path failed.
- `test_resync_is_bounded_when_lovr_never_connects` — with a short monkeypatched timeout, asserts resync returns promptly instead of hanging when no peer ever connects.

No `pyproject.toml` change (stdlib/zmq only), so no `DEPENDENCIES.md` update.

> Note: the regression tests live in the `gpu`-marked local suite (not in the CI matrix) and I couldn't run them in this environment (no local pytest/zmq). Verified via `py_compile` + inspection; please run `uv run pytest tests/test_local_render_mcp.py -k resync` on a dev box to confirm.

### DIFF
--- a/agent-mcp-servers/render-mcp/render_mcp/__main__.py
+++ b/agent-mcp-servers/render-mcp/render_mcp/__main__.py
@@ -235,11 +235,20 @@ class SceneDispatcher:
 
             self._watch_task = asyncio.create_task(_watch(), name="lovr-watch")
 
-            self._lovr_started = True
             logger.info("render-mcp: LOVR spawned (xr.start handled)")
-            # Resync current scene state into LOVR's ZMQ receive buffer so any
-            # previously-added primitives survive a LOVR restart.
+            # Restore the scene BEFORE advertising started. While resync's first
+            # (blocking) send is parked waiting for LOVR's PULL to attach,
+            # ``_lovr_started`` stays False so concurrent live ``forward()`` ops
+            # keep fast-dropping as "not_started" instead of queueing behind the
+            # parked send on the shared PUSH socket (PR #219 review nit). It also
+            # makes the flag mean what it says: LOVR connected AND scene restored.
             await self._resync_scene()
+            # Only advertise started if LOVR is still alive — it may have exited
+            # during the resync wait, in which case ``_watch`` has already run to
+            # completion (its post-wait body has no awaits) and cleared the flag,
+            # so we must not re-set it True.
+            if self._watch_task is not None and not self._watch_task.done():
+                self._lovr_started = True
             return {"status": "started"}
 
     async def _resync_scene(self) -> None:
--- a/agent-mcp-servers/render-mcp/render_mcp/__main__.py
+++ b/agent-mcp-servers/render-mcp/render_mcp/__main__.py
@@ -46,6 +46,11 @@ from xr_ai_logging import setup_logging
 
 _DEFAULT_YAML = Path(__file__).resolve().parent.parent / "render_mcp.yaml"
 
+# Max seconds the post-spawn scene resync waits for LOVR to connect its PULL
+# socket. LOVR normally attaches within a second or two of spawn; the bound
+# stops a LOVR that never connects from wedging the spawn path indefinitely.
+_RESYNC_TIMEOUT_S = 10.0
+
 
 # ── Config ─────────────────────────────────────────────────────────────────────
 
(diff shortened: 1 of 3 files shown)


nvddr:
## Review

Fixes a silent scene-loss bug in `render-mcp`, the MCP adapter that mirrors a 3D primitive scene and forwards mutations to the LOVR OpenXR app over a ZMQ PUSH socket. In the post-spawn resync path, `start_lovr_once` set `_lovr_started=True` and `_resync_scene` re-pushed every stored primitive through `forward()` (a `zmq.NOBLOCK` send). A PUSH socket with zero connected peers doesn't buffer to `SNDHWM` — it returns `EAGAIN` immediately — so the entire restore was dropped at the exact moment LOVR hadn't yet attached its PULL socket, and the scene came back empty after a crash/respawn. The fix routes only the resync off the live NOBLOCK path: `_resync_scene` hand-builds the identical `{op, value}` msgpack frame and does **blocking** `self._push.send(...)`, so the first send parks (ZMQ mute state) until LOVR's PULL connects and the rest queue to `SNDHWM`, with a `_RESYNC_TIMEOUT_S=10s` budget via `asyncio.wait_for` so a never-connecting LOVR can't wedge the spawn path. The live 50 Hz `forward()` keeps its NOBLOCK drop-on-backpressure semantics. Wire format, deadline accounting, and partial-send logging are consistent; the blocking-send-parks-on-peerless-socket premise is correct ZMQ mute-state behavior and is asserted by `test_resync_delivers_after_late_peer_connect`.

|---|---|---|
| Nit | [render_mcp/__main__.py#L298](https://github.com/NVIDIA/xr-ai/blob/8a2ca04951dec8f6f1f439cf300d36daf88c5dba/agent-mcp-servers/render-mcp/render_mcp/__main__.py#L298) | During the resync wait window `_lovr_started` is already `True`, so concurrent live `forward()` NOBLOCK sends hit the same `_push` socket while the first resync send is parked. pyzmq's asyncio socket serializes sends through an internal queue, so those live sends can queue behind the parked send rather than fast-dropping — a brief deviation from the NOBLOCK drop guarantee. Bounded (LOVR attaches in ~1–2 s; clients poll `get_health` before drop-intolerant ops) and self-clears on connect — noted for completeness, effectively a non-issue. |

No blockers or potential bugs — clean, well-scoped fix.

nvddr (approved):

wenxind-nvidia:
Thanks @nvddr — addressed the nit (and rebased onto main now that #196 landed).

**Fix for the nit:** `start_lovr_once` now runs `_resync_scene` **before** flipping `_lovr_started`, so during the resync wait window the flag stays `False` and concurrent live `forward()` ops fast-drop as `not_started` (their existing gate) instead of queueing behind the parked blocking send on the shared PUSH socket. The flag also now means what it says — *LOVR connected and scene restored*.

```python
await self._resync_scene()
# Only advertise started if LOVR is still alive — it may have exited during the
# resync wait, in which case _watch already ran to completion and cleared the
# flag, so we must not re-set it True.
if self._watch_task is not None and not self._watch_task.done():
    self._lovr_started = True
return {"status": "started"}
```

The `_watch_task.done()` guard handles the LOVR-exits-during-resync race: `_lovr_started = False` is `_watch`'s final statement with no `await` after it, so "not done" at the check always means `_watch` will still set `False` afterward — it can't end wrongly `True`.

Added `test_live_forward_fast_drops_during_resync_window` asserting `forward()` returns `not_started` while resync is parked and that the flag flips `True` only after LOVR connects.

Also merged main (resolved the changelog + this test file's additive overlap with #196's lifecycle test) — `__main__.py` cleanly combines #196's per-launch stack with this PR's blocking resync + deferred flag.

PR #235, opened 2026-06-10
ci(nightly): pause/restart GPU host via Brev CLI + alert on failure

yanziz-nvidia:
## What

The nightly `gpu`-marked pytest suite now **boots and pauses its GPU host** (a [Brev](https://docs.nvidia.com/brev/cli/cli-overview) instance) so the box isn't billing between runs — and **alerts via a GitHub issue** if a run fails or times out.

## Job graph

```
start (brev-controller)  ->  pytest (gpu)  ->  stop (brev-controller)
                                              \-> notify (on failure)
```

The Brev lifecycle commands run on a **dedicated always-on "controller" VM** (a small self-hosted runner labelled `brev-controller`) that already has the Brev CLI installed and is `brev login`'d. This avoids a CLI token in CI (the headless `--token` path didn't work) and solves the bootstrap problem — the GPU runner is asleep before `start` and is being paused by `stop`, so it can't run its own lifecycle commands.

- **start** — `brev start` the instance (idempotent — skips if already RUNNING), then poll until it reports `RUNNING` so the self-hosted `gpu` runner has time to re-register.
- **pytest** — the existing test steps, unchanged (CUDA discovery, `uv`, container purge, `pytest -m gpu`). Now gated on `needs: start`.
- **stop** — `brev stop` with `if: always()`, so a failed test or a broken `start` never leaves the instance billing (also idempotent). A `keep_running` `workflow_dispatch` input skips this to hold the box up for debugging.
- **notify** — `if: failure()` on `ubuntu-latest`: opens (or comments on) a `nightly-failure` tracking issue so a broken/timed-out nightly is visible without anyone polling the Actions tab. Update-don't-stack; skipped on PRs.

## Runner setup

| Label | Where | Purpose |
|-------|-------|---------|
| `gpu` | the GPU Brev instance | runs the pytest suite |
| `brev-controller` | a separate always-on VM (Brev CLI installed + logged in) | runs the start/stop lifecycle jobs |

## Required repo config

Settings -> Secrets and variables -> Actions:

| Kind | Name | Value |
|------|------|-------|
| Secret | `BREV_INSTANCE_NAME` | name of the GPU Brev instance to start/stop |
| Secret | `BREV_ORG` | *(optional)* org the instance lives in; set only if the controller isn't already pinned to it |

No token secret is needed — the controller VM's persistent `brev login` handles auth, and the `notify` job uses the built-in `GITHUB_TOKEN` (job-level `permissions: issues: write`). If the org defaults workflow tokens to read-only, enable issue writes under Settings -> Actions -> General -> Workflow permissions.

## Notes

- The `pull_request` trigger (on edits to this file) is kept, so the full start->test->stop flow gets exercised end-to-end before merge. `notify` is skipped on PRs.
- The RUNNING poll greps `brev ls` by instance name so other orgs' instance names stay out of the logs.

### DIFF
--- a/.github/workflows/nightly-xr-ai-test.yml
+++ b/.github/workflows/nightly-xr-ai-test.yml
@@ -82,6 +82,9 @@ jobs:
         # but failing here gives a clearer error than a stuck queue).
         run: |
           set -euo pipefail
+          echo "--- brev ls (raw, for diagnosis) ---"
+          brev ls || true
+          echo "------------------------------------"
           for i in $(seq 1 60); do
             line=$(brev ls | grep -iw "$BREV_INSTANCE_NAME" || true)
             echo "[$i] ${line:-<$BREV_INSTANCE_NAME not found in brev ls>}"
--- a/.github/workflows/nightly-xr-ai-test.yml
+++ b/.github/workflows/nightly-xr-ai-test.yml
@@ -96,20 +96,18 @@ jobs:
         # but failing here gives a clearer error than a stuck queue).
         run: |
           set -euo pipefail
-          echo "--- brev ls (raw, for diagnosis) ---"
-          brev ls || true
-          echo "------------------------------------"
           for i in $(seq 1 60); do
-            line=$(brev ls | grep -iw "$BREV_INSTANCE_NAME" || true)
-            echo "[$i] ${line:-<$BREV_INSTANCE_NAME not found in brev ls>}"
-            if echo "$line" | grep -qi "RUNNING"; then
-              echo "Instance $BREV_INSTANCE_NAME is RUNNING."
+            # grep by name keeps other orgs' instance names out of the
+            # log; report just this instance's status field.
+            status=$(brev ls | grep -iw "$BREV_INSTANCE_NAME" | grep -oiE 'RUNNING|STOPPED|STARTING|STOPPING|FAILED' | head -n1 || true)
+            echo "[$i] $BREV_INSTANCE_NAME status: ${status:-<not found>}"
+            if [[ "$status" == "RUNNING" ]]; then
+              echo "Instance is RUNNING."
               exit 0
             fi
             sleep 15
           done
           echo "::error::Timed out waiting for $BREV_INSTANCE_NAME to reach RUNNING"
-          brev ls
           exit 1
 
   pytest:
--- a/.github/workflows/nightly-xr-ai-test.yml
+++ b/.github/workflows/nightly-xr-ai-test.yml
@@ -19,7 +19,7 @@ name: nightly XR AI test
 #
 # Required repo config (Settings -> Secrets and variables -> Actions):
 #   secret  BREV_API_TOKEN      — token for headless `brev login --token`
-#   var     BREV_INSTANCE_NAME  — name of the GPU Brev instance to manage
+#   secret  BREV_INSTANCE_NAME  — name of the GPU Brev instance to manage
 
 on:
   schedule:
@@ -44,7 +44,7 @@ concurrency:
   cancel-in-progress: false
 
 env:
-  BREV_INSTANCE_NAME: ${{ vars.BREV_INSTANCE_NAME }}
+  BREV_INSTANCE_NAME: ${{ secrets.BREV_INSTANCE_NAME }}
 
 jobs:
   start:
--- a/.github/workflows/nightly-xr-ai-test.yml
+++ b/.github/workflows/nightly-xr-ai-test.yml
@@ -53,7 +53,11 @@ jobs:
     timeout-minutes: 20
     steps:
       - name: Install Brev CLI
-        run: sudo bash -c "$(curl -fsSL https://raw.githubusercontent.com/brevdev/brev-cli/main/bin/install-latest.sh)"
+        # The installer drops the binary in $HOME/.local/bin (no sudo);
+        # add it to $GITHUB_PATH so later steps in this job find `brev`.
+        run: |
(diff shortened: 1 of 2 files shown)


yanziz-nvidia (commented):

## Summary

Adds Brev CLI lifecycle jobs (`start`/`stop`) around the nightly GPU test so the instance is paused between runs to save cost, and opens/updates a tracking issue on failure. CI state depends on `BREV_INSTANCE_NAME` and `BREV_ORG` secrets being set; no dependency file changes in this PR.

|     | Finding |
|-----|---------|
| 🚫 | None |
| 💡 | [`nightly-xr-ai-test.yml:67`](https://github.com/NVIDIA/xr-ai/blob/48d0ce0a007a06cd3ca41d4095fd2cdf5b76dd44/.github/workflows/nightly-xr-ai-test.yml#L67) — **`Select org` step uses `[[ ]]` correctly, but the bare `run:` block has no `set -euo pipefail`.** If `brev set` fails silently the subsequent `brev ls` / `brev start` will target the wrong org. Add `set -euo pipefail` to both `Select org` steps (lines 67 and 245). |
| 🔍 | [`nightly-xr-ai-test.yml:300`](https://github.com/NVIDIA/xr-ai/blob/48d0ce0a007a06cd3ca41d4095fd2cdf5b76dd44/.github/workflows/nightly-xr-ai-test.yml#L300) — `notify` depends on `stop` which depends on `pytest`, so the `needs: [start, pytest, stop]` chain means `notify` never fires if `start` fails (because `stop` is skipped when `start` fails and `if: always()` on `notify` resolves to `skipped`, not `failure`). Consider `needs: [start, pytest, stop]` with `if: contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled')` to catch start-job failures too. |

- `nightly-xr-ai-test.yml:67` and `nightly-xr-ai-test.yml:245` — Add `set -euo pipefail` as the first line of both `Select org` step `run:` blocks so a failed `brev set` surfaces immediately rather than silently targeting the wrong org.
- `nightly-xr-ai-test.yml:300` — Change `notify` job's `if:` condition to `contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled')` so a `start`-job failure also triggers the alert (current `needs` chain causes `notify` to be skipped when `stop` is skipped).

nvddr:
PR #235 wires Brev-CLI GPU-host lifecycle management around the existing nightly GPU pytest job. It adds a `start` job (on an always-on `brev-controller` runner) that idempotently boots the Brev instance and polls `brev ls` until RUNNING before the `pytest` job lands on the `gpu` runner, a `stop` job (`if: always() && inputs.keep_running != true`) to pause the instance afterward so it stops billing, and a `notify` job that opens/comments a tracking issue when any prior job fails or is cancelled. It also adds a `keep_running` workflow_dispatch input and a CI-only `pytest-timeout` per-test guard via `uv run --with`.

| Severity | Location | Issue | Suggested fix |
|---|---|---|---|
| **Blocker** | nightly-xr-ai-test.yml (stop job, "Stop instance") | The guard `if brev ls \| grep -iw NAME \| grep -qi RUNNING` collapses two failure modes into a silent "nothing to stop", both exiting 0 and leaving the GPU billing: (1) `brev ls` itself failing (auth/network/CLI) → no RUNNING match → skips stop; (2) a transient `STARTING`/`STOPPING` state (e.g. the start job's wait timed out *after* `brev start` was issued) is not `RUNNING` → skip → it finishes booting and bills indefinitely. Both exit 0, so `notify` (keyed off `needs.*.result`) never fires — the exact cost-leak-without-alert this PR aims to prevent. | Fail loudly if `brev ls` returns nonzero (capture its exit status outside the piped `if` condition, which `set -e` ignores), and treat any non-`STOPPED` state (RUNNING/STARTING/STOPPING) as "must stop and re-verify STOPPED", re-polling after `brev stop`. |
| Potential bug | env block vs start/stop error messages | `env:` reads `secrets.BREV_INSTANCE_NAME`/`secrets.BREV_ORG`, but the guard messages say the `BREV_INSTANCE_NAME` repo **variable** is not set. If configured as a `vars.*` variable rather than a secret, env resolves empty → start fails (safe: pytest skipped) but never boots. Reconcile source of truth. | Pick `vars.*` (keep "variable" wording) or `secrets.*` (fix messages to "secret"); document where the operator sets it. |
| Nit | top of file (no top-level `permissions:`) | Only `notify` declares `permissions: issues: write`; start/pytest/stop run with default token scope, not least-privilege. | Add top-level `permissions: contents: read`; let `notify` keep its `issues: write` override. |
| Nit | notify body | Mixes runner-default `$GITHUB_EVENT_NAME`/`${GITHUB_SHA:0:7}` with the explicitly-set `RUN_URL` env — cosmetic inconsistency. | Optionally surface them via the `env:` block alongside `RUN_URL`. |

Verified as NOT issues: `brev set`/`start`/`stop`/`ls` are real subcommands; `set -euo pipefail` does not abort the step on a non-matching grep inside an `if` condition; `inputs.keep_running != true` correctly evaluates true on schedule/PR events, so the stop job isn't accidentally skipped on the nightly cron.

yanziz-nvidia:
Addressed above concerns.

- Blocker — stop job fails loudly on brev ls errors and stops any non-STOPPED state with a re-poll
- Potential bug — message wording reconciled with secrets.*, documented at the env block
- Nit — top-level contents: read least-privilege token
- Nit — notify env-block consistency
- Plus a docs/changelog.md entry per AGENTS.md

nvddr (approved):

PR #237, opened 2026-06-11
fix(client-samples): camera re-entrancy + reconnect reset; doc StreamKit parity

nvddr:
Cleanup pass over `client-samples/` from the repo audit (WP-A8/A9, WP-B21/B22, WP-F7, WP-M3).

## Changes
- **A8 (Android)** — `startCamera()` physical path now guards with a synchronous `isCameraStarting` check-and-set *before* `viewModelScope.launch` (the iOS pattern). Setting it inside the coroutine would let two rapid taps both pass before either ran.
- **A9 (Android + iOS)** — reset camera state on `RECONNECTING`, matching the web client. The LiveKit session persists across a reconnect, so both platforms delegate to their existing `stopCamera()` (cancels synthetic feed on Android, unpublishes the track, clears `isCameraActive`).
- **B21 (web)** — removed the unused `_startCamera` param from `sendPing` in `core.js` and updated both call sites (`web`, `web-xr`).
- **B22 (web-xr)** — deleted the permanently no-op `dbg` logger and all `dbg.*` calls (`window.__dbg` was never set). `CloudXRStream` keeps its own internal no-op fallback, so dropping the passed-in `dbg` is safe.
- **F7 (Android)** — rewrote the `LiveKitBackend` teardown comment to state the actual mechanism: cancelling the connection scope stops the `RoomEvent` collector, so the unconditional emit is the single DISCONNECTED notification per teardown.
- **M3** — added `client-samples/STREAMKIT_PARITY.md` documenting four cross-platform StreamKit divergences (send() topic param, AudioConfig fields, CameraConfig.default facing, stopAudio throws), each tagged idiom or to-fix.

## Verification
- `python3 .github/scripts/check_spdx_headers.py` → exit 0 (new parity doc carries the `<!-- -->` SPDX header).
- web/web-xr JS verified by reading + `node --check` (syntax OK); no Python in scope, pytest e2e skipped.
- **Android (Kotlin) and iOS (Swift) edits were made by static reading only and need an on-device build to verify** — they cannot be compiled in this environment.

### DIFF
--- a/client-samples/STREAMKIT_PARITY.md
+++ b/client-samples/STREAMKIT_PARITY.md
@@ -1,0 +1,19 @@
+<!--
+  SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+  SPDX-License-Identifier: Apache-2.0
+-->
+
+# StreamKit cross-platform API parity
+
+The three StreamKit client SDKs (web `web/StreamKit/`, iOS/visionOS
+`ios-visionos/StreamKit/`, Android `android/.../streamkit/`) are intended to
+mirror one another. A full-repo audit found the divergences below. Each is
+tagged **idiom** (intentional — follows the platform's idiom; leave as-is) or
+**to-fix** (a real gap to close in future native-client work).
+
+| Surface | web | iOS / visionOS | Android | Status | Notes |
+|---|---|---|---|---|---|
+| `send()` topic param | `send(data, { reliable, topic })` — supports a per-message `topic` | `send(_ data:, reliable:)` — no topic | `send(data, reliable)` — no topic | **to-fix** | Only the web XR client publishes on a named topic (`xr.session.started`, consumed by render-mcp). Native samples send untopiced `ping`/custom data, so the gap is currently invisible — but any native client that needs topic-routed signaling will require the param. |
+| `AudioConfig` fields | `mode`, `echoCancellation` | `mode`, `highpassFilter`, `typingNoiseDetection` | `mode` only | **idiom** | Each platform exposes the DSP knobs its capture stack actually supports (WebRTC `echoCancellation`; AVAudioSession `highpassFilter`/`typingNoiseDetection`; Android relies on the `MicrophoneMode` preset). `mode` is the shared cross-platform contract. |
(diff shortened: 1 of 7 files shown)


yanziz-nvidia (commented):

## Summary

Fixes camera re-entrancy on Android (synchronous `isCameraStarting` guard mirroring iOS) and adds camera reset on `RECONNECTING` to both Android and iOS. Also cleans up dead `dbg` logging in `web-xr/app.js`, drops the unused `_startCamera` param from `sendPing`, and adds a new `STREAMKIT_PARITY.md` cross-platform audit doc. CI state unknown from diff alone; note that the commit message flags "Android/iOS changes by static reading only — require on-device build to verify."

|     | Finding |
|-----|---------|
| 🚫 | None |
| 💡 | [`AppViewModel.kt:357`](https://github.com/NVIDIA/xr-ai/blob/abc74f784e8cbcd522f138ddd3ff0b6a8fce46dd/client-samples/android/app/src/main/java/com/nvidia/xrai/streamkitsample/AppViewModel.kt#L357) — `stopCamera()` never resets `isCameraStarting`. If a reconnect fires while `startCamera()` is mid-flight (coroutine launched, `isCameraStarting = true`), `stopCamera()` clears `isCameraActive` but leaves `isCameraStarting = true`, permanently blocking future taps until the in-flight coroutine's `finally` runs. The window is narrow but real. Add `isCameraStarting = false` inside `stopCamera()`'s coroutine, or reset it at the top of the block before the try. |
| 🔍 | [`app.js:255`](https://github.com/NVIDIA/xr-ai/blob/abc74f784e8cbcd522f138ddd3ff0b6a8fce46dd/client-samples/web-xr/App/app.js#L255) — `catch { /* stop failures surface via the CloudXR onStateChange handler */ }` — the comment is accurate but the original `catch (err)` block with `dbg.err` was deleted because `dbg` itself was removed. The remaining comment is good; consider whether a `console.warn` is warranted here now that the `dbg` safety net is gone, since `stopXR` failures would otherwise be fully silent in production. |

- `client-samples/android/app/src/main/java/com/nvidia/xrai/streamkitsample/AppViewModel.kt:357` — Add `isCameraStarting = false` at the start of `stopCamera()`'s coroutine body (before the try) so a mid-flight `startCamera()` guard is cleared when a reconnect-triggered stop runs.

yanziz-nvidia (approved):

nvddr:
Addressed in 6ebe6cb:

- **💡 `isCameraStarting` in `stopCamera()`**: already fixed in a prior commit — `isCameraStarting = false` is set at the top of the coroutine body with a comment explaining why (verified against current HEAD vs the reviewed commit `abc74f78`).
- **🔍 `app.js:255` silent catch**: replaced `catch { /* comment */ }` with `catch (err) { console.warn(...) }` — failures still surface via CloudXR onStateChange but are now also visible in production devtools.
- Branch rebased onto main.

PR #239, opened 2026-06-11
fix(server-runtime): hub & transport bug fixes + cleanups

nvddr:
Part of the repo cleanup batch (CLEANUP_PLAN.md WP-A/WP-G), server-runtime slice.

**Bug fixes:** A1 (WebServer.start reports success on failed port bind → mirror TokenServer bind-poll), A2 (guard encoder recreation in `_rotate_chunk`, set `enc.failed`), A3 (pop encoders inside the lock in `close_participant`), A4 (per-callback try/except for audio/data/participant loops).

**Cleanups:** B6/B7/B8 (dead `_stop` Event, empty TYPE_CHECKING block, always-true `if not self._room` guards), G2 (read IPC addrs from config), G4 (`get_running_loop` + FastAPI lifespan), G5 (`_maybe_start_track`), G6 (shared `_serve_safe`), G7 (done-callback logging on `_spawn`), G8 (`logger.debug` not bare pass), G10/G11 (top-level imports).

Excluded by design: G1 (now_us — cross-module follow-up), G3 (token-server default — human-gated), G9/J4 (P2 splits).

**Verification:** CPU suite + SPDX local; A2's NVENC-failure path needs GPU → flag for human run (`tests/run_local_gpu_tests.sh`). Base is `main`. Do not merge — human gate.

### DIFF
--- a/server-runtime/xr_media_hub/__main__.py
+++ b/server-runtime/xr_media_hub/__main__.py
@@ -25,8 +25,6 @@ from xr_media_hub._errors import StartupError
 from xr_media_hub.ipc import AudioChunk, DataMessage, HubEndpoint, ParticipantEvent, SlotView
 from xr_media_hub.transport.livekit import LiveKitConnector, make_client_token
 
-PULL_ADDR      = "ipc:///tmp/xr_hub_in"
-PUB_ADDR       = "ipc:///tmp/xr_hub_pub"
 STATS_INTERVAL = 5.0
 
 _frame_counts: dict[str, int] = collections.defaultdict(int)
@@ -89,13 +87,14 @@ async def main(ready_file: Path | None = None) -> None:
 
     setup_logging("hub")
 
-    hub = HubEndpoint(pull_addr=PULL_ADDR, pub_addr=PUB_ADDR)
+    cfg = load_config()
+
+    hub = HubEndpoint(pull_addr=cfg.hub_push_addr, pub_addr=cfg.hub_sub_addr)
     hub.on_frame(on_frame)
     hub.on_audio(on_audio)
     hub.on_data(on_data)
     hub.on_participant(on_participant)
 
-    cfg = load_config()
     connector = LiveKitConnector(cfg)
     await connector.start()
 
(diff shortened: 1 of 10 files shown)


yanziz-nvidia (commented):

## Summary

Fixes four real bugs (WebServer silent bind failure, encoder null-deref after `_rotate_chunk`, encoder pop outside lock in `close_participant`, per-callback exception swallowing) and refactors dead code / deprecated API usage (`get_event_loop` → `get_running_loop`, dead `_stop` Event, empty `TYPE_CHECKING` block, always-true guards). All CPU CI checks pass; GPU path (A2 encoder failure) is flagged for manual run per the PR description.

|     | Finding |
|-----|---------|
| 🚫 | None |
| 💡 | [`_token_server.py:57`](https://github.com/NVIDIA/xr-ai/blob/59fbd821dbfd1a637da43e7e9535eb6b57fb0531/server-runtime/xr_media_hub/transport/livekit/_token_server.py#L57) — `_proxy_client_lifespan` is named with a leading underscore (module-private convention) yet is imported directly by `_web_server.py`. If both servers share this helper long-term, consider promoting it to a small `_utils.py` or `_serve_utils.py` module so the cross-import is intentional rather than reaching into an internal name. |
| 🔍 | [`_room_client.py:247`](https://github.com/NVIDIA/xr-ai/blob/59fbd821dbfd1a637da43e7e9535eb6b57fb0531/server-runtime/xr_media_hub/transport/livekit/_room_client.py#L247) — The `if not self._room: return` guards in `send_return_data` / `send_return_audio` were removed as "always-true." Worth a one-line comment confirming `self._room` is guaranteed non-None by the time these methods are called (e.g., callers only reach them post-`connect()`), so future readers don't wonder if the guard was dropped accidentally. |

- `server-runtime/xr_media_hub/transport/livekit/_token_server.py:57` — Move `_proxy_client_lifespan`, `serve_safe`, and `wait_until_bound` to a new `_serve_utils.py` module so `_web_server.py` can import them without crossing a private-name boundary.
- `server-runtime/xr_media_hub/transport/livekit/_room_client.py:247` — Add a brief comment above `send_return_data` and `send_return_audio` stating that `self._room` is guaranteed non-None at call time (post-`connect()`), to document why the removed guards were safe to drop.

yanziz-nvidia (approved):

PR #241, opened 2026-06-11
refactor(xr-ai-pipecat): dedup PCM/WAV helpers, drop dead code

nvddr:
Repo cleanup batch — xr-ai-pipecat slice (WP-B1/B2/B3/B5, C2/C3/C4, F1-F4). Dead-code deletions grep-confirmed unused; PCM/WAV conversion consolidated into audio.py; pts_us standardized to time_ns()//1_000; docstring history narration stripped. CPU suite + SPDX local. Base main — do not merge (human gate).

### DIFF
--- a/agent-sdk/xr-ai-pipecat/xr_ai_pipecat/transport.py
+++ b/agent-sdk/xr-ai-pipecat/xr_ai_pipecat/transport.py
@@ -40,6 +40,7 @@ from xr_ai_agent import (
     Subscribe,
 )
 
+from .audio import float32_to_int16, int16_to_float32
 from .frames import ParticipantJoinedFrame, ParticipantLeftFrame
 
 _HUB_PUB  = "ipc:///tmp/xr_hub_pub"
@@ -48,16 +49,6 @@ _HUB_PUSH = "ipc:///tmp/xr_hub_in"
 SAMPLE_RATE            = 16_000
 NUM_CHANNELS           = 1
 TTS_NATIVE_SAMPLE_RATE = 22_050
-
-
-def _float32_to_int16(data: bytes) -> bytes:
-    f32 = np.frombuffer(data, dtype=np.float32)
-    return np.clip(f32 * 32767.0, -32768, 32767).astype(np.int16).tobytes()
-
-
-def _int16_to_float32(data: bytes) -> bytes:
-    i16 = np.frombuffer(data, dtype=np.int16)
-    return (i16.astype(np.float32) / 32767.0).tobytes()
 
 
 def _hub_pcm_to_mono_16k(pcm_int16: bytes, channels: int, sample_rate: int) -> bytes:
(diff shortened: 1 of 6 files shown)


yanziz-nvidia (commented):

## Summary

Consolidates two duplicated `float32 ↔ int16` conversion functions (previously defined separately in `transport.py`, `voice_gate.py`, and `streaming_tts.py`) into `audio.py`, and introduces `wav_to_output_frames` to eliminate the repeated WAV-decode + dtype-conversion + frame-construction loop that existed verbatim in both processors. Also strips `stream_sentences_to_audio` (superseded by `StreamingTtsProcessor`) and trims narrative comments per CLAUDE.md's comment rules. No `pyproject.toml` changes, so `DEPENDENCIES.md` is unaffected.

|     | Finding |
|-----|---------|
| 🚫 | None |
| 💡 | [`transport.py:375`](https://github.com/NVIDIA/xr-ai/blob/44942a4b3d4874a10afd23d5e8a5402f6e3323e1/agent-sdk/xr-ai-pipecat/xr_ai_pipecat/transport.py#L375) — `XRMediaHubTransport.cleanup_participant` now delegates to `self._output.target_participant` via the new property, but `XRMediaHubInputTransport` has no symmetric cleanup. If a future caller passes a pid that was never set as the output target, the guard silently no-ops while input-side state (e.g. VAD detector map in `VadSttProcessor`) may still hold a reference. Not a regression from this PR, but the asymmetry is now more visible — consider a follow-up to route cleanup through the input transport as well, or document that input cleanup is the caller's responsibility. |
| 🔍 | [`audio.py:26`](https://github.com/NVIDIA/xr-ai/blob/44942a4b3d4874a10afd23d5e8a5402f6e3323e1/agent-sdk/xr-ai-pipecat/xr_ai_pipecat/audio.py#L26) — `int16_to_float32` divides by `32767.0` (INT16_MAX), matching the old private function. This is correct and conventional, but means `-32768` maps to `≈ -1.000030` (slightly outside `[-1, 1]`). No clip is applied on the decode path. Upstream WAV data almost certainly never contains `INT16_MIN`, but noting it for completeness. |

- `agent-sdk/xr-ai-pipecat/xr_ai_pipecat/transport.py:381` — Add a comment (or a follow-up task) noting that `cleanup_participant` only clears output-transport state; callers are responsible for cleaning up input-side per-pid structures (e.g. VAD maps in `VadSttProcessor`).

agent-sdk/xr-ai-pipecat/xr_ai_pipecat/transport.py
```diff
@@ -184,6 +175,10 @@ def __init__(self, ep: ProcessorEndpoint, params: TransportParams, **kwargs):
         # be created on demand (they need it to .start()).
         self._start_frame: StartFrame | None = None
 
+    @property
+    def target_participant(self) -> str:
```
yanziz-nvidia, line 179: i might missed this somewhere - who consumes this?

yanziz-nvidia (approved):

agent-sdk/xr-ai-pipecat/xr_ai_pipecat/transport.py
nvddr: `target_participant` is consumed as the send-target fallback when a frame reaches the output with no explicit `transport_destination`/pid: `streaming_tts.py` (TTS target routing, ~L272/L308), and in xr-render-demo `processors.py` (L258/L268) and `agent.py` (L136/L160). The XRMediaHubTransport property at L367 just delegates to `self._output.target_participant` (the B5 cleanup). So it's live, not dead.

PR #245, opened 2026-06-11
ci: add ruff lint; dedup subprocess test helpers; smoke-test oxr/vec MCP

nvddr:
Repo cleanup batch — tooling/CI slice (WP-K1/K2/K3, M2). Conservative ruff.toml (E,F,W,I) + lint.yml `ruff check`, tuned to land green without touching module source; repo-wide `ruff format` left as a human-gated follow-up. Shared tests/_helpers_subprocess.py; CPU smoke tests for vec-mcp and oxr-mcp. Base main — do not merge (human gate).


yanziz-nvidia (commented):

## Summary

Introduces `ruff check` CI (check-only, conservative E/F/W/I ruleset, pre-existing violations grandfathered via `per-file-ignores`), deduplicates three near-identical port-selection/health-probe helpers into a new `tests/_helpers_subprocess.py`, and adds subprocess smoke tests for the vec-mcp and oxr-mcp (self-skipping) servers. All CI checks pass (pytest 3.11/3.12, ruff, SPDX, CodeQL, uv lock).

|     | Finding |
|-----|---------|
| 🚫 | None |
| 💡 | [`tests/test_oxr_mcp.py:20-28`](https://github.com/NVIDIA/xr-ai/blob/0730a69f8c2cf9a6778c3e65f691c41b599161cb/tests/test_oxr_mcp.py#L20-L28) — Double-skip: `pytest.importorskip` at module scope already skips the whole file on CPU; the explicit `pytest.skip(...)` inside `test_oxr_mcp_placeholder` is unreachable on CI and misleading on a GPU host where `isaacteleop` is installed (the test body runs but immediately bails rather than exercising anything). On a GPU host the intent presumably is to run a real check, not a no-op. Consider either removing the inner `pytest.skip` or replacing it with a `pytest.mark.skip(reason=...)` on the test until the real smoke test is implemented. |
| 🔍 | [`tests/test_piper_tts.py:27`](https://github.com/NVIDIA/xr-ai/blob/0730a69f8c2cf9a6778c3e65f691c41b599161cb/tests/test_piper_tts.py#L27) — `import socket` is still present for `_port_open` (which uses `socket.settimeout`), so it is legitimate — but worth a brief comment that this is intentionally not replaced, since the surrounding dedup might cause a future reader to wonder why it wasn't removed alongside `_pick_port`. |

- `tests/test_oxr_mcp.py:28` — Remove the `pytest.skip(...)` call inside `test_oxr_mcp_placeholder` (the module-level `importorskip` already handles the CPU-skip path; the inner skip makes the test a no-op even on a GPU host where `isaacteleop` resolves).

yanziz-nvidia (approved):

nvddr:
Addressed in 4d55356:

- **`test_oxr_mcp_placeholder` double-skip** (💡): replaced inline `pytest.skip()` with `@pytest.mark.skip(reason=...)` on the test function. Skip now declared at collection time; no silent no-op on a GPU host with `isaacteleop` installed.
- **`import socket` in `test_piper_tts`** (🔍): added inline comment explaining `socket` is kept for the local `_port_open` helper — `_pick_port` moved to `_helpers_subprocess` but `_port_open` stays in this file.

PR #261, opened 2026-06-23
fix(ipc): drop stale ring on failed re-registration

### LINKED ISSUE #209: IPC hub: connector re-register left bound to closed ring if ShmRingBuffer reopen fails
### Summary

In `HubEndpoint._handle_registration`, if a connector re-registers and the new `ShmRingBuffer(name=reg.shm_name, create=False)` **reopen raises**, `self._ring_registry[reg.connector_id]` is never reassigned and keeps pointing at the just-closed `old_ring`. That connector then fails every subsequent `FRAME_SIGNAL` (`read_slot` on a released buffer, caught by `_dispatch`) and never recovers — a stuck, stale-closed registry entry.

### Context

Surfaced during review of #202 (which fixed the use-after-close / `BufferError` by releasing exported `SlotView` memoryviews before `old_ring.close()`). That fix is correct; this is a separate, **pre-existing** edge case noted for follow-up — out of scope for #202.

Post-#202 the failure mode degrades from #197-style corruption to a *caught* error, but the connector is still permanently broken until it re-registers again, because the registry entry points at a closed ring.

### Location

`server-runtime/xr_media_hub/ipc/_hub.py` — `_handle_registration`, around the `ShmRingBuffer(..., create=False)` reopen (~L348).

### Suggested fix

On reopen failure, `pop` the stale `_ring_registry[reg.connector_id]` entry (and any related per-connector state) so the connector isn't left bound to a closed ring — or reopen into a local first and only swap into the registry once construction succeeds.

### Severity

Low — requires a reopen failure during connector re-registration (crash/reconnect path). No data corruption (caught), but the affected connector is stuck until it registers anew.

WilliamK112:
## Summary

- remove the stale connector ring entry before closing the old ring during re-registration
- leave the connector without a registered ring if opening the replacement shared-memory segment fails, instead of retaining a closed ring in `_ring_registry`
- add a regression test for the failed-reopen path that does not depend on an IPC socket fixture

Fixes #209.

## Validation

Rebased onto current `NVIDIA/xr-ai@37822349729f9c67f492f9aac29b7eeb803d1dd9` on 2026-08-09. The only conflict was the upstream SDK rename from `xr_ai_agent` to `xr_ai_hub`; the refreshed test uses the current package.

- focused failed-reopen regression: 1 passed
- complete `tests/test_participant_events.py`: 5 passed
  - run with a short `TMPDIR` on macOS to stay below the Unix-domain socket path limit
- Ruff 0.15.16 on both changed Python files: passed
- repository SPDX checker on both changed files: passed
- DCO sign-off present
- `git diff --check upstream/main...HEAD`: passed

### DIFF
--- a/services/xr-media-hub/xr_media_hub/ipc/_hub.py
+++ b/services/xr-media-hub/xr_media_hub/ipc/_hub.py
@@ -461,7 +461,7 @@ class HubEndpoint:
     def _handle_registration(self, reg: ConnectorRegistration) -> None:
         if reg.connector_id in self._ring_registry:
             logger.warning("Connector {} re-registered — replacing ring buffer", reg.connector_id)
-            old_ring = self._ring_registry[reg.connector_id]
+            old_ring = self._ring_registry.pop(reg.connector_id)
             # Drop any frames still held in the old ring BEFORE closing it.
             # A live SlotView keeps a sliced memoryview exported into the ring's
             # mmap; closing with that outstanding makes ShmRingBuffer.close()'s
--- a/tests/test_participant_events.py
+++ b/tests/test_participant_events.py
@@ -11,8 +11,9 @@ from __future__ import annotations
 import asyncio
 
 import pytest
+import xr_media_hub.ipc._hub as hub_module
 
-from xr_ai_hub import ParticipantEvent, PixelFormat
+from xr_ai_hub import ConnectorRegistration, ParticipantEvent, PixelFormat
 
 pytestmark = pytest.mark.asyncio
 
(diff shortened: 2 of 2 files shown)


WilliamK112:
Refreshed this PR onto current `main` (`37822349`) at head `e6f8d5f4`. The only conflict was the upstream SDK rename, so the regression test now imports `ConnectorRegistration` from `xr_ai_hub`. Current validation: all 5 participant-event tests pass with a short macOS `TMPDIR`; focused regression, Ruff 0.15.16, SPDX, DCO, and diff checks are clean.

WilliamK112:
Refreshed again onto current `main` (`6de6ca73`) after XR-Media-Hub moved from `server-runtime/` to `services/`. Git followed the rename cleanly; the PR remains one signed commit and the diff is still limited to the registry `pop` plus its regression test. GitHub now reports `behind_by: 0`, `ahead_by: 1`, and mergeable. Validation on head `ed61f3f2`: all 5 `test_participant_events.py` tests pass, Ruff 0.15.16 passes, SPDX and DCO checks pass, and `git diff --check` is clean.

WilliamK112:
Refreshed onto current `main` (`b641bf5`) at signed head `13cb7135`. The six upstream commits replayed cleanly, including the concurrent `xr-ai-hub-client` changes; the PR diff remains limited to the stale-ring `pop` and its regression test. Validation: all 5 `test_participant_events.py` tests pass, Ruff passes, both changed files pass the SPDX checker, the commit retains its DCO sign-off, and `git diff --check` is clean. GitHub reports the PR mergeable.

nvddr:
Thank you for this contribution, and for keeping it current through the repository reorganizations. We consolidated this fix with the related reliability changes in #408, which is now merged. The stale-ring cleanup was adapted to the renamed DeviceIOHub layout and extended to purge held slots safely during failed re-registration.

We also preserved your authorship: GitHub credits @WilliamK112 as a co-author on the merged consolidation commit (`a6d079f`). This PR is now superseded by #408. Thanks again!

PR #266, opened 2026-06-25
refactor(agent-sdk): extract reusable VisionModule; modularize vision samples

wenxind-nvidia:
simple-vlm-example and xr-render-demo each re-implemented the same live-camera VLM Q&A (frame tracking, freshness, camera-on-demand startCamera/stopCamera + grace, the VLM call) and each carried a copy of pixels.py. Factor that into the shared xr-ai-pipecat SDK so samples reuse it instead of duplicating it.

Behaviour is unchanged from main — only the implementation moved.

xr-ai-pipecat (new shared modules):
- pixels.py — frame → PIL → JPEG data URL (was copied per vision worker).
- vision.py — VisionModule: live-camera VLM Q&A with camera-on-demand. Two call styles over one frame-acquisition path: ask() streams tokens (for TTS), perceive() returns a string and raises VisionUnavailable (for agentic tool loops). Declares Pillow (used by pixels).

simple-vlm-example:
- SimpleVlmBrain is now thin glue over VisionModule (handle_query → vision.ask, warmup/release hooks, data-channel path). Worker + YAML are unchanged from main; local worker/pixels.py deleted (use xr_ai_pipecat.pixels).

xr-render-demo:
- look_at_current_frame delegates to VisionModule.perceive (mapping VisionUnavailable onto its existing {"error","spoken"} tool result); the duplicated frame/camera helpers and worker/pixels.py removed. The agentic render loop and MCP path-based vision tools are unchanged.

A hardware-capability layer (HardwareProfile/AgentFeatures → auto-gated pipeline) was prototyped and deferred; see agent-samples/MODULARIZATION.md.

Verified: py_compile across the SDK + both samples; no references to removed symbols. DEPENDENCIES.md updated.


nvddr:
## Review

**Goal alignment: strong.** `VisionModule` is exactly the right abstraction — live-camera VLM Q&A is no longer per-sample boilerplate. The `ask`/`perceive` dual API is well-designed (streaming for TTS, one-shot for tool loops). Both samples are meaningfully slimmer. CI green. Tests updated without `# noqa: SLF001`.

---

### 💡 Issues

**1. `asyncio.get_event_loop()` in `_wait_for_camera_frame` (vision.py:149, 155)**

```python
deadline  = asyncio.get_event_loop().time() + timeout   # line 149
remaining = deadline - asyncio.get_event_loop().time()  # line 155
```

`get_event_loop()` is deprecated since Python 3.10 and emits `DeprecationWarning` on 3.12. All other async code in the same file correctly uses `asyncio.get_running_loop()` (e.g. line 222). Fix: replace both calls with `asyncio.get_running_loop()`.

**2. `ask()` and `perceive()` open with identical 3-line preamble**

Both start with:
```python
old = self._camera_off_timers.pop(pid, None)
if old and not old.done():
    old.cancel()
self._camera_held.add(pid)
```

This duplicated preamble belongs in a small private helper (or folded into `_acquire_image_url`, which is already the shared frame-acquisition path). Minor but it is exactly the kind of repetition this PR is designed to eliminate.

---

### 🔍 Nits

**3. `print_task_done_banner` silently dropped**

The old `_stream_query` logged call timing (`duration_s`, `status`, query preview). The new `VisionModule` has no equivalent, so per-call VLM latency/status is no longer visible in logs. Worth a comment in the PR confirming this is intentional (e.g. timing belongs in the brain layer) rather than an accidental omission.

**4. `set_status` asymmetry between `ask()` and `perceive()`**

`ask()` calls `set_status("processing")` / `set_status("idle")` around the VLM stream. `perceive()` does not. The agent-status badge stays correct today because xr-render-demo only calls `perceive()` and manages its own status, but the asymmetry is a trap for future callers who expect both methods to maintain the same status contract.

nvddr:
## Architectural follow-up

The PR is directionally correct and the `VisionModule` pattern is exactly what we want. But it surfaces some structural questions worth resolving before or right after this lands.

---

### `pixels.py` is in the wrong package

`xr_ai_pipecat/pixels.py` only imports `xr_ai_agent.FrameData` and `PixelFormat` — it has zero dependency on the pipecat voice pipeline. It's a codec utility: hub `FrameData` → PIL → JPEG. Its natural home is `xr-ai-agent` (which owns `FrameData`) or a thin vision extra on that package. Placing it in `xr-ai-pipecat` forces Pillow + numpy as a transitive dependency of the voice pipeline layer, and it signals that `xr-ai-pipecat` is becoming a catch-all rather than staying focused on the voice pipeline (VAD, STT, TTS, `BrainProcessor`).

---

### `xr-ai-pipecat` scope is drifting

The package started as "the pipecat voice pipeline." It's now also home to VisionModule, pixels, and (per MODULARIZATION.md) potentially more capability modules. That's fine if intentional — but should be explicit. The alternative is a separate `xr-ai-capabilities` package (or `xr_ai_pipecat.capabilities` subpackage) with a clear boundary: pipecat = voice pipeline plumbing; capabilities = reusable agent building blocks.

---

### MCP tool wiring is not modularized

`VisionModule` is the right pattern but it stops halfway. The MCP side hasn't been touched. `RenderSceneProcessor` still takes five separate `McpClient` constructor args, and `_execute_tool` is a hardcoded routing switch:

```python
if tool in _OXR_TOOLS:   → oxr client
if tool in _VEC_TOOLS:   → vec client
if tool in _VLM_TOOLS:   → vlm client
if tool in _VIDEO_TOOLS: → video client
else:                    → render client (catch-all)
```

Adding a sixth MCP server to any future sample requires: new constructor arg, new frozenset, new branch in the switch. Nothing is reusable across samples. The natural next step is an `MCPToolset` abstraction — something that pairs a client with the tool names it owns, so a brain accepts a list of them and auto-routes:

```python
brain = AgentBrain(
    transport    = transport,
    llm          = llm,
    toolsets     = [
        MCPToolset(oxr_client,    _OXR_TOOLS),
        MCPToolset(vec_client,    _VEC_TOOLS),
        MCPToolset(render_client),   # catch-all
    ],
    capabilities = [VisionModule(transport, vlm)],
)
```

A new sample that needs three tools instead of five just drops the rest. That's the "sub-tools composed into a sample" goal. This doesn't block this PR, but it should be a named follow-up — not just the vague "hardware-capability gating" deferral currently in MODULARIZATION.md.

---

### What is `MODULARIZATION.md` and where do these learnings belong?

Should the architectural decisions in `agent-samples/MODULARIZATION.md` live in `AGENTS.md` (the authoritative conventions doc) or `docs/source/guides/`? A standalone file under `agent-samples/` is somewhat hidden. If this captures the roadmap for how agent samples should be built, it's more durable and discoverable as a section in `AGENTS.md` under "Agent sample architecture" or as a proper guide page.

nvddr (approved):

PR #275, opened 2026-06-30
Fix shared-memory frame size guard

fallintoplace:
## Summary

Reject frames that exceed a shared-memory slot before claiming or writing to the ring buffer. The error includes the frame dimensions, pixel format, byte count, and configured slot capacity.

This also adds a defensive read-side check for oversized `FrameSignal.data_sz` values, and covers the failure mode with focused regression tests that verify rejected frames do not advance the writer or corrupt the next slot header.

## Validation

- `PYTHONPATH=server-runtime:utils/xr-ai-logging uv run --python 3.12 --project agent-sdk --with pytest --with pytest-asyncio --with loguru pytest tests/test_shm_ring_buffer.py`
- `PYTHONPATH=server-runtime:utils/xr-ai-logging TMPDIR=/tmp uv run --python 3.12 --project agent-sdk --with pytest --with pytest-asyncio --with loguru pytest --basetemp=/tmp/xr-ai-pytest tests/test_participant_events.py`
- `uvx ruff check agent-sdk/xr_ai_agent/_shm.py tests/test_shm_ring_buffer.py`
- `git diff --check`

### DIFF
--- a/agent-sdk/xr_ai_agent/_shm.py
+++ b/agent-sdk/xr_ai_agent/_shm.py
@@ -57,6 +57,19 @@ _STATE_READY   = 2
 _STATE_OFFSET = 4
 
 
+def _frame_nbytes(data: bytes | memoryview) -> int:
+    return data.nbytes if isinstance(data, memoryview) else len(data)
+
+
+def _frame_size_error(width: int, height: int, fmt: PixelFormat, n: int, max_frame_bytes: int) -> str:
+    fmt_name = getattr(fmt, "name", str(fmt))
+    return (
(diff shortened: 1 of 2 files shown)


nvddr:
Thank you for this contribution. We consolidated the shared-memory frame-size guard into #408, which is now merged. During review, it was expanded with canonical slot-header validation, malformed-signal rejection and slot recovery, non-contiguous buffer handling, and additional shared-memory regressions.

We also preserved your authorship: GitHub credits @fallintoplace as a co-author on the merged consolidation commit (`a6d079f`). This PR is now superseded by #408. Thanks again!

PR #277, opened 2026-06-30
Bound return-audio pacing queue

fallintoplace:
## Summary
- add a per-participant return-audio backlog cap (`return_audio_max_buffer_s`, default 3 seconds)
- drop the oldest queued audio frames when a TTS burst exceeds the cap, keeping newer audio responsive
- expose queued/dropped frame and duration counters and rate-limit backpressure warnings
- extend pacing tests for overflow, drop-oldest behavior, flush metrics, and invalid buffer limits

## Tests
- `PYTHONPATH=server-runtime:utils/xr-ai-logging uv run --project agent-sdk --with pytest --with pytest-asyncio --with loguru --with numpy --with livekit --with livekit-api --with pyyaml --with fastapi --with 'uvicorn[standard]' --with httpx --with websockets --with cryptography pytest tests/test_return_audio_pacing.py`\n- `uvx ruff check server-runtime/xr_media_hub/transport/livekit/_room_client.py server-runtime/xr_media_hub/transport/livekit/config.py tests/test_return_audio_pacing.py`

### DIFF
--- a/server-runtime/xr_media_hub.yaml
+++ b/server-runtime/xr_media_hub.yaml
@@ -41,6 +41,11 @@ web_client_dir:    ../client-samples/web
 # ── Optional: token server (legacy plain-HTTP entry point) ────────────────────
 enable_token_server: true
 
+# ── Return audio pacing ───────────────────────────────────────────────────────
+# Maximum queued TTS audio per participant before oldest queued frames are
+# dropped to keep latency and memory bounded.
+return_audio_max_buffer_s: 3.0
+
 # ── Video recording (NVENC, optional) ─────────────────────────────────────────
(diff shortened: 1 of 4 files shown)


nvddr:
Thank you for identifying the unbounded return-audio backlog. We consolidated the fix into #408, which is now merged. During review, it evolved into a hard per-participant queued-audio duration bound combined with producer-side pacing and participant-scoped interruption handling, preserving normal long replies while bounding faulty-producer bursts.

We also preserved your authorship: GitHub credits @fallintoplace as a co-author on the merged consolidation commit (`a6d079f`). This PR is now superseded by #408. Thanks again!

PR #279, opened 2026-07-02
Add XR render demo support to visionOS client

blongs-nv:
Adds XR streaming with the render demo to the AVP client, along with changes to the media lifecycle to prevent CloudXR and LiveKit from disrupting each other, so audio and camera survive XR starting and stopping.


nvddr:
## Summary review

Thanks for this — it's a substantial, carefully-commented addition and the inline rationale made it a pleasure to read through.

### What this does / how it fits

Adds a CloudXR render-streaming path to the visionOS sample, running **alongside** the existing LiveKit agent channel as a second, independent transport. The core design:

- **Two parallel transports.** LiveKit (mic/camera/data via the hub) and CloudXR (XR render frames via CloudXRKit) are decoupled; a new `XRState` machine (derived from `CloudXRKit.Session.state`) drives a small "XR Stream" row in the UI. Connecting/disconnecting one never tears down the other, except deliberately (an unexpected LiveKit drop stops XR; `disconnect()` stops XR first; closing the immersive space stops XR).
- **render-mcp handshake.** On CloudXR `.connected` the app publishes an empty message on the `xr.session.started` LiveKit topic, mirroring the web-xr client so `render-mcp` gates the LOVR launch identically. (I confirmed the topic string matches its consumers exactly — `web-xr/App/app.js`, the native `main.cpp` client, and the worker's `_XR_SESSION_STARTED_TOPIC` in `agent-samples/xr-render-demo/worker/agent.py` — so the gate signal is wired correctly.) This required threading an optional `topic` through `StreamingBackend.send` / `StreamSession.send` / `LiveKitBackend` (default impl drops the topic, so other backends are unaffected).
- **Media lifecycle hardening.** A bounded mic-recovery loop (settle → stop → start → verify, generation-guarded) re-arms LiveKit capture after OS interruptions / media-services resets that XR entry/exit triggers; `prepareAudio()` pre-warms `AVAudioSession`; camera intent is preserved across LiveKit reconnects; and a new always-on `MediaSessionDiagnostics` logger captures untethered interruption/route events.
- **Wiring:** `RealityView` hosts the `CloudXRSessionComponent`; new SPM dep on `NVIDIA/cloudxr-framework` 6.1.0; entitlement (`low-latency-streaming`), usage strings, and `DEPENDENCIES.md` updated in-PR.

Scope is client-samples-only (no server code), and it's a clean, parity-driven extension of the existing StreamKit sample.

### Findings

| Severity | Location | Issue | Suggested fix |
|---|---|---|---|
| Potential bug | `App/AppModel.swift` — `publishXRStartedWhenConnected()` / `onConnectionStateChanged` `.connected` | The `xr.session.started` publish is only triggered on the CloudXR `.connected` edge and retries a bounded ~4×500ms while *both* transports are up. If LiveKit is momentarily disconnected/reconnecting during that whole window (each attempt hits `guard … connectionState == .connected … else { return }`), the loop exits, `hasPublishedXRStarted` stays `false`, and nothing re-arms — CloudXR won't re-emit `.connected` while it stays connected. Net effect: the signal is silently lost and `render-mcp` never launches LOVR, with no error surfaced. | Re-attempt the XR-started publish when LiveKit returns to `.connected` (e.g. call `publishXRStartedWhenConnected()` from the `.connected` case when `xrState` is `.streaming`), so the gate signal survives a transient LiveKit reconnect that overlaps XR connect. |
| Nit | `StreamKit/.../LiveKitBackend.swift` — `prepareAudio()` | `AVAudioApplication.requestRecordPermission { _ in }` discards the grant result; a denied mic permission surfaces only later as the generic "Microphone couldn't start" toast. | Optional: log the denied case (`mediaLog`) so the untethered diagnostics trail explains why capture never comes up. |
| Nit | `App/AppModel.swift` — `.reconnecting` handling | Camera intent is explicitly preserved/restored across a LiveKit reconnect, but the mic has no equivalent reconnect handling (recovery is driven only by `AVAudioSession` interruptions/resets). This is very likely deliberate — LiveKit auto-republishes standard audio tracks on reconnect, whereas the camera uses a custom ARKit `FrameInjectable` provider that needs manual re-init — so flagging only to confirm it's intentional, not as a suspected defect. | No change expected; a one-line comment noting *why* the mic is not special-cased here would save the next reader the same double-take. |

The `withCheckedContinuation`/waiter teardown in `stopXR()`, the generation-guarded `recoverMic()` loop, and the `withObservationTracking` re-arm all look correct under `@MainActor` isolation as far as I can trace statically — but the concurrency here is subtle and the mic-recovery + XR-teardown timing is best validated on-device, which I can't do. Worth a human eye (and ideally a Vision Pro run) on those paths.

I'm a static reviewer without a CloudXRKit environment, so the SDK-facing assumptions (reusing a session across connect/disconnect, `connect()` succeeding from a `disconnected(.failure)` state, `CloudXRSessionComponent` lifecycle) are taken on faith from the code and comments — a reviewer with device access should confirm those.

— devdeepr-bot

nvddr:
@blongs-nv PTAL at the potential bug identified by the bot reviewer.

nvddr (approved):

blongs-nv:
> @blongs-nv PTAL at the potential bug identified by the bot reviewer.

Updated.

PR #285, opened 2026-07-20
feat(nat): add native spatial-math functions

nvddr:
## Stack

P00 of 3. This is the base PR and targets `main`.

- **P00:** native spatial-math functions
- P01: native text-memory functions and MCP export
- P02: native vision functions

## What changed

- Registers spatial math directly through its capability-owned `nat.plugins` entry point; there is no package-wide registration module.

- Adds the `xr-ai-nat` package and registers an `xr_spatial_math` NAT function group.
- Exposes six typed coordinate calculations with explicit anchors, reference frames, and metre-based argument names:
  - `compute_gaze_target`
  - `compute_user_relative_position`
  - `compute_position_relative_to_anchor`
  - `offset_position_in_user_frame`
  - `compute_position_toward_or_away_from_reference`
  - `compute_midpoint`
- Every native operation returns `Vector3` and never creates, moves, or associates a scene object.
- Shares the pure spatial implementation with the transitional Vec and OpenXR MCP compatibility surfaces.
- Removes duplicate position-result schemas and keeps containment with the application-level compatibility adapter.
- Adds generated-schema and behavior tests and updates dependency, notice, and public architecture documentation.

## Why

Spatial calculations are in-process deterministic functionality. They should be directly composable as NAT functions instead of requiring an MCP process boundary. Tracking and OpenXR pose acquisition remain separate concerns and pass explicit frames into the math functions.

Calculation-oriented names also prevent an agent from mistaking coordinate helpers for scene mutations. Inputs name whether the user, an object, or another position is the anchor, and distance arguments state their units.

## Compatibility and impact

- Existing Vec MCP tool names and behavior are preserved.
- Existing OpenXR MCP tool names and pose ownership are preserved.
- No sample is migrated or removed in this PR.
- This establishes the bottom-up native function layer used by the following PRs.

## Validation

- Full local CPU presubmit on Python 3.12: 436 passed, 1 skipped, 37 GPU-only deselected.
- Focused native spatial-math and Vec MCP tests: 5 passed.
- Strict Sphinx documentation build passed.
- Ruff, SPDX, DCO, compilation, and diff checks passed.
- GitHub Actions passes on Python 3.11 and Python 3.12, including strict docs, lock verification, SPDX, Ruff, DCO, and CodeQL.

### DIFF
--- a/agent-sdk/xr-ai-nat/xr_ai_nat/functions/spatial_math/_math.py
+++ b/agent-sdk/xr-ai-nat/xr_ai_nat/functions/spatial_math/_math.py
@@ -1,0 +1,170 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Pure calculations shared by native functions and compatibility adapters."""
+
+from __future__ import annotations
+
+import math
+from typing import Literal
+
+from .schemas import SpatialFrame, Vector3
+
+
+def _position(x: float, y: float, z: float) -> dict[str, float]:
+    return {"x": round(x, 3), "y": round(y, 3), "z": round(z, 3)}
+
+
+def _require_distance(distance: float) -> None:
+    if distance < 0:
+        raise ValueError("distance must be non-negative; flip the direction instead")
+
+
+def _ground_basis(frame: SpatialFrame) -> tuple[tuple[float, float], tuple[float, float]]:
+    fx, fz = frame.forward.x, frame.forward.z
+    magnitude = math.hypot(fx, fz)
+    if magnitude < 1e-6:
+        rx, rz = frame.right.x, frame.right.z
+        right_magnitude = math.hypot(rx, rz)
+        if right_magnitude < 1e-6:
+            fx, fz = 0.0, -1.0
+        else:
+            fx, fz = rz / right_magnitude, -rx / right_magnitude
+    else:
+        fx, fz = fx / magnitude, fz / magnitude
+    return (fx, fz), (-fz, fx)
+
+
+def offset(
+    frame: SpatialFrame,
+    origin: Vector3,
+    *,
+    forward: float = 0.0,
+    right: float = 0.0,
+    up: float = 0.0,
+) -> dict[str, float]:
+    """Apply a gravity-aligned frame-relative offset to an origin."""
+
+    (fx, fz), (rx, rz) = _ground_basis(frame)
+    return _position(
+        origin.x + fx * forward + rx * right,
+        origin.y + up,
+        origin.z + fz * forward + rz * right,
+    )
+
+
+def world_offset(
+    origin: Vector3,
+    *,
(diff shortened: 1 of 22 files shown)


agent-sdk/xr-ai-nat/xr_ai_nat/functions/spatial_math/_math.py
```diff
@@ -0,0 +1,170 @@
...
+        "front": (-distance, 0.0, 0.0),
+        "back": (distance, 0.0, 0.0),
+        "left": (0.0, -distance, 0.0),
+        "right": (0.0, distance, 0.0),
+        "above": (0.0, 0.0, distance),
+        "below": (0.0, 0.0, -distance),
+    }
+    forward, right, up = offsets[direction]
+    return offset(frame, anchor, forward=forward, right=right, up=up)
+
```
wenxind-nvidia, line 127: Can place_user_relative and place_object_relative be combined into one function given that they share very similar code?

wenxind-nvidia (commented):
## Review — native spatial-math functions (P00)

**Verdict:** Solid, mergeable. I traced numerical parity for every migrated tool (oxr-mcp `position_ahead` / `position_relative` / `place_user_relative` / `place_object_relative` / `displace_object` / `displace_objects`, and vec-mcp `between_anchors` / `world_offset` / `along_direction`) against the new pure-math core and they match, including the ground-plane axis fallback, the `above`/`below` no-pose identity-frame path, and the `along_direction` toward/away sign handling. Deps, THIRD_PARTY_NOTICES, DEPENDENCIES.md, AGENTS.md, README, changelog, and SPDX headers are all updated in-commit. No blocker found.

### Blocker
none

### Suggestion
- `agent-mcp-servers/oxr-mcp/oxr_mcp_server/__main__.py:325` (`position_ahead`) — now delegates to `compute_gaze_target`, which calls `_require_non_negative_distance` and raises `ValueError` on a negative distance. Its sibling tools (`place_user_relative:421`, `place_object_relative:480`) guard with `if distance < 0: return {"error": ...}` and the pre-migration `position_ahead` returned a coordinate, so a negative distance now surfaces as an uncaught MCP tool error instead of a friendly `{"error"}` result. Fix: add the same `if distance < 0` guard (or explicitly document that negatives are rejected) so error behavior is consistent across the pose tools.

### Nit
- `THIRD_PARTY_NOTICES.md:150` — `pydantic` is listed at pinned `2.10.0` while `agent-sdk/xr-ai-nat/pyproject.toml:14` requires `>=2.10` (a floor, not a pin). Minor; consider noting it as a minimum to avoid implying an exact version.

wenxind-nvidia (approved):

nvddr:
Addressed both follow-ups in e4e7d04:

- `position_ahead` now rejects a negative distance with the same friendly `{error: ...}` response as its sibling pose tools, before acquiring a pose. Added a GPU-host regression test. The native NAT function remains strictly schema-validated (`ge=0`); the compatibility adapter translates invalid input to the MCP response convention.
- `THIRD_PARTY_NOTICES.md` now records `pydantic` as `>=2.10`, matching the package requirement floor.

PR #286, opened 2026-07-20
feat(nat): add native text-memory functions and MCP export

nvddr:
## Stack

P01 of 3. This PR is stacked on #285 and targets `migration/00-native-spatial-math`.

- P00: #285 — native spatial-math functions
- **P01:** native text-memory functions and MCP export
- P02: native vision functions

Review this PR as the incremental diff from P00. Its base will move to `main` after #285 merges.

## What changed

- Adds a direct text-memory `nat.plugins` entry point alongside spatial math instead of extending a central registration module.

- Adds the `xr_text_memory` NAT function group for persistent timestamped text history.
- Moves transcript JSONL storage, source identity handling, querying, and statistics into the native function implementation.
- Adds a generic `create_mcp_server()` adapter that exports an explicit list of NAT functions as typed MCP tools.
- Rebuilds transcript MCP as a thin compatibility process over the native functions while retaining its existing public tool names and response contracts.
- Adds direct text-memory and generic NAT-to-MCP adapter tests.

## Why

Transcript persistence is repository-native capability logic, not an MCP-specific implementation. NAT applications should invoke it directly in process. MCP remains an optional compatibility boundary for non-NAT consumers and should only republish selected native functions.

## Compatibility and impact

- Existing transcript MCP tool names, JSONL data, source naming, and FastMCP response behavior are preserved.
- The adapter exports only functions explicitly selected by an application.
- No existing sample is migrated or removed in this PR.
- P00 remains independently usable and reviewable.

## Validation

- Full CPU presubmit on Python 3.11: 441 passed, 1 skipped, 37 GPU-only deselected.
- Full CPU presubmit on Python 3.12: 441 passed, 1 skipped, 37 GPU-only deselected.
- Existing transcript MCP end-to-end test passed.
- Existing Vec MCP regression test passed.
- Native text-memory and generic adapter tests passed.
- Ruff, SPDX, DCO, dependency resolution, compilation, and diff checks passed.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -55,6 +55,10 @@ deps/               # Gitignored downloaded binaries (e.g. LOVR AppImage)
 - **Agentic functions are NAT-first and in-process.** Reusable deterministic
   functions live in `xr-ai-nat` as typed NAT function groups. Existing MCP
   servers remain compatibility surfaces while their capabilities migrate.
+- **A process boundary does not imply MCP.** `xr-ai-nat[mcp]` may expose an
+  application's explicit native-function list to MCP-only agents, but native
+  applications invoke the functions directly. Text-memory owns transcript
+  JSONL storage; transcript MCP only republishes that capability.
 - **No API keys or tokens in source files** — use env vars or
   `xr_media_hub.yaml`. See `docs/credentials.md`.
 
@@ -124,6 +128,8 @@ them. They split across SDK packages by what they depend on:
 Typed agent functions live in `xr-ai-nat`. `SpatialMathFunctionsConfig`
 registers deterministic coordinate operations that receive an explicit spatial
 frame; tracking and process boundaries remain outside the math functions.
+`TextMemoryFunctionsConfig` provides persistent timestamped text without a
+network boundary.
 
 The **voice pipeline** lives in `xr-ai-pipecat` (it depends on pipecat):
 
--- a/DEPENDENCIES.md
+++ b/DEPENDENCIES.md
@@ -109,13 +109,16 @@ xr-ai-models  (agent-sdk/xr-ai-models/)
 xr-ai-nat  (agent-sdk/xr-ai-nat/)
     └── nvidia-nat-core ==1.8.0
     └── pydantic >=2.10
+    └── [mcp] fastmcp >=3.4,<4
     Typed, in-process NeMo Agent Toolkit functions for XR capabilities. The
     ``xr_spatial_math`` function group accepts explicit coordinate frames and
     performs deterministic spatial calculations without OpenXR, model, or MCP
-    dependencies. Each capability module is its own ``nat.plugins`` discovery
-    entry point; there is no package-wide registration aggregator. The pure
-    math core is also used by the transitional Vec and OpenXR MCP compatibility
-    surfaces.
+    dependencies. ``xr_text_memory`` owns persistent per-source JSONL text
+    history. Its optional MCP adapter exports an application's explicit native
+    function list without routing native composition through MCP. Each
+    capability module is its own ``nat.plugins`` discovery entry point; there
+    is no package-wide registration aggregator. The spatial pure math core is
+    also used by the transitional Vec and OpenXR MCP compatibility surfaces.
 
 xr-ai-launcher  (utils/xr-ai-launcher/)
     └── (stdlib only — zero runtime deps)
@@ -169,10 +172,12 @@ xr-media-hub  (server-runtime/)
 
 transcript-mcp-server  (agent-mcp-servers/transcript-mcp/)
     └── uvicorn[standard] >=0.29
-    └── fastmcp >=2.0
     └── pyyaml >=6.0
-    Pure FastMCP — every operation is an MCP tool at /mcp (no REST).
-    Storage: JSONL files per participant in configurable transcripts_dir.
+    └── xr-ai-logging [editable: ../../utils/xr-ai-logging]
+    └── xr-ai-nat[mcp] [editable: ../../agent-sdk/xr-ai-nat]
+    Thin MCP compatibility process at /mcp (no REST). It republishes the four
+    native ``xr_text_memory`` functions under their existing MCP tool names;
+    JSONL storage and source identity handling live in xr-ai-nat.
 
 vlm-mcp-server  (agent-mcp-servers/vlm-mcp/)
     └── uvicorn[standard] >=0.29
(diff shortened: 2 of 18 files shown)


wenxind-nvidia (commented):
## Review: PR #286 — native text-memory functions + MCP export

**Verdict:** Solid, faithful increment. The `xr_text_memory` group reproduces transcript-mcp's `add_transcript`/`query_transcripts`/`list_sources`/`get_transcript_stats` semantics (JSONL storage, `.identity` sidecar disambiguation, corrupt-line skipping+logging, ordered query, error-as-data), and the generic `create_mcp_server` adapter carefully preserves the legacy MCP tool names and response shapes (the `untyped_outputs` set keeps `add_transcript`'s unwrapped `{"ok": true}` and the bare-list contracts intact). Deps, THIRD_PARTY_NOTICES, DEPENDENCIES.md, AGENTS.md, changelog, and tests are all updated in-commit. No blockers.

### Blocker
none

### Suggestion

- **`agent-mcp-servers/transcript-mcp/transcript_mcp_server/__main__.py:292-316` — NAT `Function` objects are used after the `WorkflowBuilder` context exits, and no unit test covers that path.** `build_mcp` captures `functions = await group.get_all_functions()` inside `async with WorkflowBuilder()` but then builds `exports`/`create_mcp_server(...)` *after* the block closes. This works only because the function closures keep the underlying `TextMemoryStore` alive past builder teardown — a lifetime assumption the new unit tests never exercise (`tests/test_text_memory_functions.py:71-79` `_functions()` yields and uses the functions *inside* the context). The PR notes the existing transcript-mcp e2e passed, so it holds at runtime, but the pattern is fragile. Fix: either keep the builder open for the server's lifetime, or add a test that invokes an exported function after the `WorkflowBuilder` context has exited so a future NAT teardown change can't silently break the server.

- **`agent-mcp-servers/transcript-mcp/transcript_mcp_server/__main__.py:211,521 — reaches into the private `_store` module and builds a throwaway store just to read `.directory`.** `run()`/`serve()` construct `TranscriptStore(directory)` (which `mkdir`s), pass `store.directory` into `TextMemoryFunctionsConfig`, and then the function group internally constructs a *second* `TextMemoryStore(config.directory)` (mkdir again). The first store's only purpose is to hand its directory back. This forces importing the capability's private `xr_ai_nat...text_memory._store` across the package's public boundary. Fix: pass the resolved directory string straight to `TextMemoryFunctionsConfig(directory=directory)` and drop both the `_store` import and the redundant store instantiation.

### Nit

- **`agent-sdk/xr-ai-nat/xr_ai_nat/adapters/mcp.py:19-21` — depends on FastMCP non-public internals.** `from fastmcp.tools.base import ToolResult`, subclassing `fastmcp.tools.Tool`, and overriding `run()`/calling `convert_result` are internal-surface APIs. The `>=3.4,<4` pin bounds the blast radius, but a short comment flagging the coupling (or a thin isinstance/attribute guard) would make the next FastMCP bump less surprising.

- **`agent-sdk/xr-ai-nat/xr_ai_nat/functions/text_memory/functions.py:942 — `list_sources` takes an `_EmptyInput` model while its siblings take bare annotated params.** Presumably a NAT requirement for zero-argument functions (the test asserts empty `properties`), but the asymmetry is non-obvious; a one-line comment would help.

_Not independently run: relied on the PR's reported presubmit (441 passed on 3.11/3.12) and existing transcript-mcp e2e; findings are diff/contract-based._

nvddr:
Addressed in 7dc242f (rebased on `main`):

- Replaced the custom FastMCP `Tool` subclass and private `ToolResult`/`run()` path with the public `FunctionTool` API while retaining the NAT-derived schemas and compatibility response handling.
- `transcript-mcp` now passes the resolved directory directly into `TextMemoryFunctionsConfig`; the private store import, re-export, redundant store construction, and duplicate directory creation are gone.
- Added a regression test that builds the transcript MCP server, then invokes its exported tools after `WorkflowBuilder` has exited. It verifies the adapter-level wrapped list result; the existing end-to-end test continues to cover the legacy MCP response.
- Documented why `list_sources` uses an explicit empty request model.

Focused native/transcript tests, Ruff, SPDX, and DCO pass locally; the fresh CI run is in progress.

wenxind-nvidia:

This re-review checks each structural choice for **convergence** toward the migration target end-state, classifying it CONVERGES / INTERMEDIATE-OK (target keeps the behavior but reorganizes later) / DEVIATES (target rejects that direction).

**Verdict: No blockers.** Clean, well-tested increment. Conventions fully satisfied (SPDX on every new file, DCO on both commits, `DEPENDENCIES.md`/`AGENTS.md`/`docs/changelog.md`/`THIRD_PARTY_NOTICES.md` updated in-commit). Two choices head away from the target and are cheapest to correct now (PR 1 of 8).

### Direction vs target

**1. MCP export lives in `xr_ai_nat/adapters/mcp.py` — DEVIATES (low cost).**
Target places the MCP export at `xr_ai_nat/mcp/server.py` (with `mcp/__init__.py` re-exporting `create_mcp_server`) and reserves `adapters/` for framework adapters — in the target `adapters/` holds only `voice.py`, no `adapters/mcp.py`. The README + transcript-mcp + tests now document/import `xr_ai_nat.adapters.mcp.create_mcp_server` as the public path. The function name/signature **converges**; only the module location deviates. Relocating to `mcp/server.py` now is nearly free and avoids churning a published import path later.

**2. `text_memory/` split into `_store.py` + `schemas.py` + `functions.py` — INTERMEDIATE-OK.**
Target folds all of this into a single `text_memory/functions.py` (`__init__.py` + `functions.py` only). The split mirrors the existing `spatial_math/` layout and folds down trivially later — a reasonable stepping stone, not a divergence.

**3. Function I/O contract — mixed; the error-as-data part DEVIATES.**
Positional `Annotated` args match the `spatial_math` convention (request-model refactor is INTERMEDIATE-OK). But bare `list[...]` returns and `OperationResult | TextMemoryError` **error-as-data** unions DEVIATE: the target validates inputs instead (`text: Field(min_length=1)` + `field_validator`), wraps outputs in result models (`QueryTranscriptsResult`, `TranscriptStatsResult`), and returns a zero-count stats result (with `earliest_us/latest_us = None`) for a missing source rather than an error object. This PR's choice is what forces the extra `untyped_outputs` / `x-fastmcp-wrap-result` machinery in `adapters/mcp.py` that the target's adapter does not need. Cost to unwind is one module + the adapter — align before it propagates to later PRs.

**4. transcript-mcp rewritten as a thin republisher — INTERMEDIATE-OK.** Matches the "MCP servers remain compatibility surfaces during migration" story. Note the target ultimately **deletes** `agent-mcp-servers/transcript-mcp/` entirely, so the `untyped_outputs` escape hatch is transitional scaffolding — don't invest further in it.

### Findings (normal review)
Nothing at/above the reporting bar. Low-confidence nits:
- `_store.py` — `_resolve_or_create` compares identity contents exactly while `_resolve_existing` compares `.strip()`ed; two `source_id`s differing only by surrounding whitespace are distinct to the writer but can collide on the reader. Use the same comparison (drop `.strip()`). (~55)
- `_store.py::list_sources` doesn't verify the `.jsonl` exists for an identity file; target guards with `identity.with_suffix(".jsonl").exists()`. (~40)
- `get_transcript_stats` returns two "no data" shapes (missing → `TextMemoryError`; empty → zero-count with `0` bounds); ties into direction item 3. (~45)

Tests are solid (persistence + ordering, `.identity` disambiguation, wrapped/unwrapped adapter output shapes, builder-teardown survival, adapter validation-error paths).

wenxind-nvidia (approved):

PR #287, opened 2026-07-20
feat(nat): add native vision functions

nvddr:
## Stack

P02 of 3. This PR is stacked on #286 and targets `migration/01-native-text-memory`.

- P00: #285 — native spatial-math functions
- P01: #286 — native text-memory functions and MCP export
- **P02:** native vision functions

Review this PR as the incremental diff from P01. Its base will move down the stack as the earlier PRs merge.

## What changed

- Adds a direct vision `nat.plugins` entry point alongside the inherited capabilities instead of extending a central registration module.

- Adds the `xr_vision` NAT function group for asking a VLM questions about an acquired local image.
- Injects the repository `VLMService`, normalizes image input off the event loop, and keeps all model I/O behind `xr-ai-models`.
- Keeps image acquisition separate so live-frame and future video-memory functions can compose with the same vision function.
- Rebuilds VLM MCP as a thin export of `vision__ask_image` through the generic NAT-to-MCP adapter.
- Preserves the existing MCP `ask_image` name, arguments, result shape, error behavior, and model request payload.

## Why

Vision reasoning is reusable native agent functionality. It should not own participant, hub, recording, or MCP state. Accepting an already acquired image path keeps frame acquisition and model reasoning independently composable while allowing MCP-only clients to use the same implementation.

## Compatibility and impact

- Existing VLM MCP configuration and public tool interface are preserved.
- Existing model presets and OpenAI-compatible request behavior remain in `xr-ai-models`.
- No existing sample is migrated or removed in this PR.
- P00 and P01 remain separate review and merge units.

## Validation

- Full CPU presubmit on Python 3.11: 441 passed, 1 skipped, 37 GPU-only deselected.
- Full CPU presubmit on Python 3.12: 441 passed, 1 skipped, 37 GPU-only deselected.
- Existing VLM MCP contract suite: 10 passed.
- Combined native spatial-math, text-memory, adapter, and vision tests passed.
- Existing transcript and Vec MCP regression tests passed.
- Ruff, SPDX, DCO, dependency resolution, compilation, and diff checks passed.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -59,6 +59,10 @@ deps/               # Gitignored downloaded binaries (e.g. LOVR AppImage)
   application's explicit native-function list to MCP-only agents, but native
   applications invoke the functions directly. Text-memory owns transcript
   JSONL storage; transcript MCP only republishes that capability.
+- **Image acquisition and vision reasoning stay separate.** The native vision
+  function accepts an acquired image path and calls a `VLMService` from
+  `xr-ai-models`; it does not own hub, participant, recording, or MCP state.
+  VLM MCP republishes this function only for MCP clients.
 - **No API keys or tokens in source files** — use env vars or
   `xr_media_hub.yaml`. See `docs/credentials.md`.
 
@@ -129,7 +133,9 @@ Typed agent functions live in `xr-ai-nat`. `SpatialMathFunctionsConfig`
 registers deterministic coordinate operations that receive an explicit spatial
 frame; tracking and process boundaries remain outside the math functions.
 `TextMemoryFunctionsConfig` provides persistent timestamped text without a
-network boundary.
+network boundary. `VisionFunctionsConfig` adds image-question answering over an
+injected `xr-ai-models` VLM while leaving frame acquisition to its own
+capability.
 
 The **voice pipeline** lives in `xr-ai-pipecat` (it depends on pipecat):
 
--- a/DEPENDENCIES.md
+++ b/DEPENDENCIES.md
@@ -110,6 +110,7 @@ xr-ai-nat  (agent-sdk/xr-ai-nat/)
     └── nvidia-nat-core ==1.8.0
     └── pydantic >=2.10
     └── [mcp] fastmcp >=3.4,<4
+    └── [vision] httpx >=0.27, Pillow >=10.0, xr-ai-models [editable: ../xr-ai-models]
     Typed, in-process NeMo Agent Toolkit functions for XR capabilities. The
     ``xr_spatial_math`` function group accepts explicit coordinate frames and
     performs deterministic spatial calculations without OpenXR, model, or MCP
@@ -119,6 +120,8 @@ xr-ai-nat  (agent-sdk/xr-ai-nat/)
     capability module is its own ``nat.plugins`` discovery entry point; there
     is no package-wide registration aggregator. The spatial pure math core is
     also used by the transitional Vec and OpenXR MCP compatibility surfaces.
+    ``xr_vision`` normalizes an acquired local image and calls an injected
+    xr-ai-models VLM; it does not acquire frames itself.
 
 xr-ai-launcher  (utils/xr-ai-launcher/)
     └── (stdlib only — zero runtime deps)
@@ -181,16 +184,14 @@ transcript-mcp-server  (agent-mcp-servers/transcript-mcp/)
 
 vlm-mcp-server  (agent-mcp-servers/vlm-mcp/)
     └── uvicorn[standard] >=0.29
-    └── fastmcp >=2.0
     └── pyyaml >=6.0
-    └── Pillow >=10.0
-    └── httpx >=0.27   (imported directly to catch httpx.HTTPError from xr-ai-models)
     └── xr-ai-logging  [editable: ../../utils/xr-ai-logging]
     └── xr-ai-models   [editable: ../../agent-sdk/xr-ai-models]
-    Pure FastMCP — one tool at /mcp (no REST). Reads a local image file,
-    encodes it as a JPEG data URL, and calls vlm-server via xr-ai-models
-    ``OpenAICompatVLM``. Back-compat: legacy ``vlm_server:`` URL key is
-    still accepted with a deprecation warning.
+    └── xr-ai-nat[mcp,vision] [editable: ../../agent-sdk/xr-ai-nat]
+    Thin MCP compatibility process with one tool at /mcp (no REST). It
+    republishes ``vision__ask_image`` under the existing ``ask_image`` name.
+    Image normalization and the VLM call live in the native vision function;
+    the legacy ``vlm_server:`` URL key remains accepted with a warning.
 
 video-mcp-server  (agent-mcp-servers/video-mcp/)
     └── uvicorn[standard] >=0.29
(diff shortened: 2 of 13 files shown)


wenxind-nvidia (commented):
## Review — feat(nat): add native vision functions (P02)

**Verdict:** Solid, contract-faithful increment. The `xr_vision` group reproduces vlm-mcp's `ask_image` (same JPEG data-url normalization, `<think>` stripping, error strings, and OpenAI-compat payload), routes all model I/O through `xr-ai-models` `VLMService` with no raw httpx HTTP call, and keeps deps/docs/SPDX in sync. No blockers — a couple of minor items below.

Contract checks that pass:
- `ask_image` name preserved via `tool_names={vision__ask_image: "ask_image"}`; args `question`/`image_path`, str result, and error-string behavior unchanged.
- `VLMService.ask_image(image, question, *, system_prompt="")` accepts `system_prompt`, so passing `config.system_prompt` (default `""`) is payload-equivalent to the old `vlm.ask_image(data_url, question)` — no request-shape drift (golden wire-trace test still passes).
- Model routing stays behind `xr-ai-models`; `httpx` is only imported to catch `HTTPError` from its transport and is declared in the new `[vision]` extra.
- No `ask_frames` / multi-image tool exists on the base branch (`migration/01-native-text-memory` vlm-mcp has only `ask_image`), so there is no multi-image path to reproduce — nothing missing.
- DEPENDENCIES.md, THIRD_PARTY_NOTICES (Pillow/HPND), AGENTS.md, README, changelog, and SPDX headers all updated in-change.

### Blocker
none

### Suggestion
- **MCP tool description regressed for MCP-only agents.** `agent-mcp-servers/vlm-mcp/vlm_mcp_server/__main__.py` (build) + `agent-sdk/xr-ai-nat/xr_ai_nat/functions/vision/functions.py:73-77`. The old FastMCP `ask_image` exposed a rich multi-section docstring (acquire-a-frame-first flow, "never invent/guess a path", worked `question` examples) as the tool description the agent reads to drive the tool. `create_mcp_server` publishes `function.description` verbatim, which is now two sentences. The PR preserves name/args/result/error/payload but not the description, and no test covers it, so this drops silently. For MCP-only clients this can reduce tool-use quality. Consider enriching the native function's `description=` (it becomes the single source for both native and MCP surfaces) or noting the intentional trim in the changelog entry.

### Nit
- **Private cross-package re-export kept only for a legacy test.** `agent-mcp-servers/vlm-mcp/vlm_mcp_server/__main__.py` — `from xr_ai_nat.functions.vision._images import load_jpeg_data_url as _load_jpeg_data_url  # noqa: F401`. This shim reaches into an underscore-private module purely so `tests/test_vlm_mcp.py::test_load_jpeg_data_url_emits_data_url` can still import `_load_jpeg_data_url` from `vlm_mcp_server.__main__`. Cleaner to point that test at the native `_images` helper (or fold it into `tests/test_vision_functions.py`) and drop the re-export.
- **Roundabout HTTP-error catch.** `functions.py:48-55` catches broad `except Exception`, does a deferred `import httpx`, then `isinstance`-checks and re-raises non-HTTP errors. Functionally identical to `except httpx.HTTPError as exc:`, which is clearer given `httpx` is already a required `[vision]` dependency.

wenxind-nvidia:

Checking each structural choice for convergence toward the migration target (CONVERGES / INTERMEDIATE-OK / DEVIATES).

**Verdict:** Mechanism converges; the tool *surface* is a deliberate transitional shim. Ship-able as an intermediate, but one hypothesis is **disproven** — `_images.py` is not a pre-rename of the target's `_pixels.py` (different responsibility) — and there's one real MCP tool-surface regression worth fixing before the stack lands.

### Direction vs target
The target's `functions/vision/` has **no** `image_path`-based tool. It exposes `StreamingVisionConfig` (`xr_streaming_vision`) and `VisionToolsConfig` (`xr_vision_tools`), whose tools take `participant_id` + a question and pull frames from a live `LiveFrameSource` or the `video_memory` group. This PR's entire `xr_vision` / `ask_image(question, image_path)` surface is **replaced**, not evolved, in the target.

| Choice | Classification | Evidence |
|---|---|---|
| `_images.py` = `load_jpeg_data_url(path)` — loads a file from disk | **DEVIATES (responsibility), not a rename** | Target `_pixels.py` = `frame_to_pil(FrameData)` + `encode_image(img)`; decodes raw hub pixel buffers (I420/NV12/RGBA/BGRA/RGB24), never touches a path. `_images.py` will be superseded/removed, not renamed to `_pixels.py`. |
| `VisionFunctionsConfig` (`xr_vision`) with single `ask_image(question, image_path)` | **DEVIATES (end-state) / INTERMEDIATE-OK (transitional)** | No `VisionFunctionsConfig`/`xr_vision`/`ask_image` in target. Justified here as byte-for-byte preservation of the vlm-mcp contract during migration. |
| Injected VLM via `Field(exclude=True, repr=False)` + `arbitrary_types_allowed` | **CONVERGES** | Identical to target's `StreamingVisionConfig`/`VisionToolsConfig`. |
| MCP republish via generic `create_mcp_server(...)` + `tool_names` rename | **CONVERGES** | The generic NAT→MCP adapter is the established mechanism. |
| Error handling returns `"ask_image: ..."` strings; regex-strips `<think>` | **INTERMEDIATE-OK** | Target instead `raise`s `FrameUnavailable` and does no `<think>` strip in `functions.py`; kept here for MCP compat. |

Net: plumbing is on-target; the tool shape + helper module are a compatibility layer the target discards. Defensible for PR 2/8 — just don't expect `_images.py` to survive as `_pixels.py`.

### Findings by severity
**Important — MCP `ask_image` tool description regressed, unguarded (conf 80).** `agent-mcp-servers/vlm-mcp/vlm_mcp_server/__main__.py`: the MCP tool now inherits the native function's two-sentence description. The previous `ask_image` docstring — the actual MCP tool description agents see — included the load-bearing anti-hallucination guidance *"`image_path` MUST be the `path` value returned by a prior call … Never invent or guess a path."* That is gone. Unifying native/MCP descriptions is good, but the surviving text is materially weaker for agent tool-use, and `test_wire_trace_golden_matches_pre_migration_shape` only asserts the HTTP body, not the tool description — so this is unguarded. Enrich the native `ask_image` description (and per-field `Field(description=...)`) to retain the "don't invent a path" guidance, or note the reduction in the changelog if intentional.

**Minor (noted for the stacked review):**
- `build_mcp` invokes the function after its `WorkflowBuilder` context exits. Works only because this group has no teardown; the target's frame-source groups run `finally: config._frames = None` on generator close, so copying this pattern to a later frame-source group would break.
- Convoluted HTTP-error catch in `functions/vision/functions.py` (`except Exception … if not isinstance(exc, httpx.HTTPError): raise`) — equivalent to `except httpx.HTTPError`; harmless indirection.

**Conventions — clean.** SPDX on all new files; `DEPENDENCIES.md` in-commit (`[vision]` extra on `xr-ai-nat`; vlm-mcp's fastmcp/Pillow/httpx removed in favor of `xr-ai-nat[mcp,vision]`); `THIRD_PARTY_NOTICES.md` adds Pillow (HPND); lazy `PIL`/`httpx` imports keep the entry point importable without the extra. Gap: no direct unit test for the native fn's empty-`image_path` or httpx-error branches (covered only transitively via `test_vlm_mcp.py`).

nvddr:
## Disposition — re-review vs. target

Accepted the target classification: `xr_vision` / `ask_image` and `_images.py` are a transitional compatibility surface. `_images.py` will be replaced or removed as the modular frame-source design lands; it is not the future `_pixels.py`, whose raw-frame decoding responsibility is distinct.

Addressed in `aa2d516` (P02 is now rebased on the updated P01 head, `7dc242f`):

- Restored the MCP-facing contract in the native function and schema: acquire an image first, pass that exact returned path, and never invent or guess one. An MCP-client test now asserts the published tool and field descriptions.
- Removed the private `_images` re-export from the legacy MCP entry point; its JPEG helper coverage now lives with the native vision functions.
- Simplified the VLM HTTP handling to a direct `httpx.HTTPError` catch while keeping the optional dependency import lazy.
- Added direct native-function coverage for an empty `image_path` and an HTTP failure.

For the builder-lifetime concern: this path-only compatibility group has no teardown resources, and the existing MCP invocation test already calls its function after `build_mcp()` has returned. I have recorded the boundary for the final modular design: future live-frame/frame-source groups must retain builder lifetime ownership and must not copy this transitional pattern.

Validation: `20 passed` across the focused vision/MCP/text-memory tests, plus Ruff and `git diff --check`; commit hooks also passed Ruff, SPDX, and DCO.

wenxind-nvidia (approved):

PR #288, opened 2026-07-20
feat(nat): add native XR tracking service

nvddr:
## Summary

- add the xr_tracking NAT function group and private correlated service RPC
- move headless OpenXR session ownership into services/openxr-service
- preserve the existing OXR MCP port and tool schemas as a compatibility adapter
- wire the XR render sample through the typed service without changing its worker API

## Why

Native NAT agents should consume tracking as a typed function even when the hardware capability is long-running. MCP remains available for non-NAT consumers but no longer owns OpenXR capability logic.

## Validation

- focused XR tracking and spatial math tests: 7 passed
- Ruff on all changed Python files
- SPDX and DCO pre-commit hooks
- uv lock/import checks for xr-ai-nat, oxr-mcp, openxr-service, and tests

## Stack

Depends on #287.


wenxind-nvidia (commented):
## Review — native XR tracking service (draft, formative feedback)

Nice, coherent slice. The pose/transform contract is genuinely preserved: `services/openxr-service/openxr_service/pose.py` carries the `rotate_vector` / yaw / pitch / forward-right-up derivation over verbatim from the old `oxr-mcp`, the OpenXR Y-up conventions and the `(x,y,z,-1 forward)` basis are unchanged, and the MCP adapter now just reshapes the typed `HeadPose` (`timestamp_ms → ts`). Deps are declared in both `pyproject.toml` files and `DEPENDENCIES.md` in-commit, the DEALER/ROUTER correlated RPC looks sound, and dead OpenXR/isaacteleop code was cleanly lifted out of `oxr-mcp`. Scope reads as on-topic for the migration step — the 33 files are mostly docs. No blockers; a few things to consider before marking ready.

### Blocker
none

### Suggestion
- **The one piece of math with the highest regression risk is the one piece with no test.** `pose.py` (`rotate_vector`, yaw/pitch, basis vectors) *is* the coordinate-convention reproduction this migration is meant to preserve, and it lives in `services/openxr-service/` which ships no tests. `tests/test_xr_tracking_functions.py` covers RPC correlation, error-code propagation, and the typed `get_user_frame` round-trip, but feeds a hand-built `HeadPose` — it never exercises quaternion→frame. A small CPU-only regression test pinning a known quaternion → expected `forward/right/up/yaw_deg/pitch_deg` would lock the contract and catch any future drift. (`services/openxr-service/openxr_service/pose.py`)
- **`health()` reads session state outside the lock.** `HardwarePoseSource.health()` reads `_oxr_session` / `_open_attempts` / `_last_open_error` without `self._lock`, while `get_pose()` mutates them under the lock in a worker thread (service dispatch calls `get_pose` via `asyncio.to_thread`, `health` inline). GIL-atomic reads mean no crash, but health can report torn/stale values. Taking the lock in `health()` would make the snapshot consistent. (`services/openxr-service/openxr_service/session.py`)

### Nit
- `OpenXRClient.is_available()` (`agent-sdk/xr-ai-nat/xr_ai_nat/functions/xr_tracking/_client.py`) is defined but unused in this PR — fine if a later stacked PR consumes it, otherwise dead code.
- MCP `get_head_pose` success output now carries an extra `error: null` field (`HeadPose.model_dump` always includes `error`), a small drift from the pre-migration schema that omitted it on the valid path. If strict tool-schema parity matters, drop a `None` error in the adapter before returning. (`agent-mcp-servers/oxr-mcp/oxr_mcp_server/__main__.py`, `PoseSource.get_pose`)
- `OpenXRService.dispatch` validates `EmptyRequest` before dispatching on `operation`, so a `get_head_pose` call with unexpected args surfaces as generic `internal_error` rather than a targeted `invalid_request`. Minor. (`services/openxr-service/openxr_service/service.py`)

wenxind-nvidia:

Checking each structural choice for convergence (CONVERGES / INTERMEDIATE-OK / DEVIATES).

**Verdict:** Correct incremental PR — RPC transport and pose math are sound (math verified on CPU: identical to the target's `_rotate_vec`/yaw/pitch). Converges functionally but takes three structural detours the target unwinds, and ships the flagged pose-math untested. No blocking bugs.

### Direction vs target

**1. RPC layer `_rpc/{__init__,client,protocol,server}.py` → target `_service/rpc.py` — INTERMEDIATE-OK (rename pending).** The four-file split is logic-identical to the target's single `_service/rpc.py` (which is `protocol.py`+`client.py`+`server.py` concatenated; `PROTOCOL_VERSION`/`pack`/`unpack` → private `_PROTOCOL_VERSION`/`_pack`/`_unpack`). Two convergence gaps to preserve for the consolidating PR:
- Package name `_rpc` → `_service` (imports become `xr_ai_nat.functions._service.rpc`).
- This PR's `RPCClient` drops the `__aenter__`/`__aexit__` the target's client keeps — a real API regression vs target if anything later does `async with RPCClient(...)`.

**2. `xr_tracking/schemas.py` → target has none — DEVIATES.** Target inlines contracts into `_client.py` and derives them from shared bases: requests from `_models._StrictRequest` (`HeadPoseRequest`/`OpenXRHealthRequest`), results from `types.ServiceResult`. PR adds a standalone `schemas.py` with a shared `EmptyRequest(extra="forbid")` and plain `BaseModel` results. Field-name divergence too: PR `HeadPose.timestamp_ms` vs target `HeadPose.ts`.

**3. `services/openxr-service/` layout — DEVIATES (layout) + coupling concern.** PR ships five modules (`__init__`,`__main__`,`pose.py`,`service.py`,`session.py`); target flattens all of it into a single `__main__.py`. More material: PR's `pose/session/service` **import the full `HeadPose`/`OpenXRHealth` pydantic models from `xr_ai_nat.functions.xr_tracking`** and construct them service-side, which drags NAT plugin-registration machinery into service import. Target keeps the wire as plain **dicts** — service imports only the request types for validation and emits dicts; response models live client-side only ("service speaks dicts, client owns types"). Combined with the schema-sharing, this points away from the target's boundary.

### Findings by severity
**Important — pose math untested; the known series gap is still open (conf ~90).** `services/openxr-service/openxr_service/pose.py` (`rotate_vector`, `from_native_pose`, yaw/pitch) is pure and CPU-testable, but `tests/test_xr_tracking_functions.py` covers only RPC correlation/error-code propagation and constructs `HeadPose` directly. Worse, the math lives in `xr-openxr-service`, which is **not** a dependency of `tests/pyproject.toml`, so it isn't importable from the suite. Add table-driven tests (identity quat; 90° yaw about +Y → forward `-Z→-X`, right `+X→-Z`, yaw `90.0`; ±90° pitch clamp via the `asin` bound) and put `pose.py` on the test path. Logic is correct today — purely a regression-safety hole.

**Important — service↔schema cross-package coupling (conf ~82).** As in Direction #3; if the service stays multi-file, at minimum construct dicts service-side (as the target does) rather than importing the response models.

**Conventions — clean.** SPDX on every new file; DCO present; `DEPENDENCIES.md` in-commit (new `[services]` extra = `msgpack>=1.0, pyzmq>=27.0`, `xr-openxr-service` at port 8330, oxr-mcp adapter re-point). No correctness bugs in the RPC client/server (correlation, timeout cleanup, error round-trip, socket teardown) or the pose conversion.

tests/test_xr_tracking_functions.py
```diff
@@ -0,0 +1,105 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
```
wenxind-nvidia, line 3: Add tests for services/openxr-service/openxr_service/pose.py

nvddr:
## Disposition — XR tracking review

Addressed in `bd109e0`:

- Added CPU-only, table-driven pose conversion tests for identity, +90° yaw, and ±90° pitch (including the `asin` clamp boundary).
- Moved XR request/result contracts into the client, aligned the wire field to `ts`, and removed the standalone tracking schemas module.
- Made the OpenXR service emit and validate plain dictionaries; the client remains the typed boundary. Health snapshots now take the hardware lock.
- Added `RPCClient` async context management, removed unused availability code, preserved MCP schema parity by omitting a null `error`, and return `invalid_request` for unexpected service arguments.

This moves P03 toward the private modular target without importing its later broad type consolidation. The `_rpc` → `_service/rpc` rename and service-file flattening remain an explicit later consolidation; the current split keeps the pure pose conversion directly CPU-testable.

Validation: 8 focused tests passed (plus one expected hardware-only skip), Ruff, diff check, and lock dry-run all pass.

tests/test_xr_tracking_functions.py
nvddr: Addressed in bd109e0. The test suite now imports the pure pose module without hardware bindings and covers identity, +90-degree yaw, and +/-90-degree pitch, including the clamped asin path. It also verifies the dict-only service wire and targeted invalid_request validation.

wenxind-nvidia (approved):

PR #291, opened 2026-07-20
feat(samples): migrate Simple VLM to native vision

nvddr:
## What changed

- add a native streaming live-camera vision function to `xr-ai-nat`
- migrate `simple-vlm-example` from `xr-ai-capabilities.VisionModule` to that NAT function
- preserve the existing Pipecat voice, wake-word, interruption, model-profile, and text-channel behavior
- add CPU coverage for live frame acquisition and streaming invocation

## Why

This is the first sample cutover from a framework-local capability object to an ordinary in-process NAT function. It establishes the native invocation path before XR render migrates its larger capability set.

## Impact

Simple VLM now composes its vision operation with `WorkflowBuilder`; MCP is not involved. Existing launch commands and configuration remain unchanged.

## Validation

- `uv run pytest -q test_vision_functions.py test_simple_vlm_example_wire.py`
- focused Pipecat query, supersede, and participant cleanup tests
- `uvx ruff check .`
- SPDX header check
- pre-commit hooks and DCO sign-off

PR #292, opened 2026-07-20
P07: compose XR render with native NAT functions

nvddr:
## Summary
- build scene, XR tracking, spatial math, video memory, vision, live vision, and text memory as one native NAT workflow in the XR render worker
- remove capability MCP processes and sample-specific MCP configuration from the live demo path
- derive the existing model tool schemas from NAT Functions while preserving all legacy spatial tool names and parameters
- use one native-toolbox assembly for both the live worker and prompt eval; the eval no longer starts or discovers MCP adapters
- persist only final user/assistant turns through native text memory
- keep Pipecat voice behavior, prompts, quick acknowledgements, and the current model-service tool loop unchanged

## Stack
- depends on #291 (migration/06-simple-vlm-native-vision)
- NativeToolbox, render_spatial, and the existing single model loop are transitional compatibility scaffolding. Their native function-group registrations are the durable substrate. #293 adds the models-backed NAT LLM provider and a read-only consumer; the supervisor/subagent migration and replacement of the flat toolbox remain separate follow-up work.

## Validation
- uv run --project tests pre-commit run --all-files
- uv run --project tests pytest -q -m "not gpu" — 452 passed, 1 skipped, 38 deselected
- strict Sphinx build with warnings treated as errors
- native eval schema smoke test — 23 model-facing tools, including live perception; lifecycle tools excluded
- native/legacy spatial tool schema comparison — all 12 tool names and property sets match


wenxind-nvidia:

This PR is the most important **direction** check in the series (branch name = "modular-render-subagents"; the target decomposes the render worker into subagents). Classifying each choice CONVERGES / INTERMEDIATE-OK / DEVIATES.

**Verdict:** Approve direction, with one thing to confirm. The load-bearing work (MCP clients → in-process native NAT function groups; config/deps/process-list rewrite) is exactly the target's substrate and converges cleanly. The net-new glue this PR stacks on top (`NativeToolbox` adapter + `render_spatial` convenience group + a monolithic model-loop) is transitional scaffolding the target **deletes** — acceptable for PR 7 of 8 **only if PR 8 (#293) performs the supervisor + subagents decomposition**. Worth calling out so `capabilities.py` isn't mistaken for durable design.

### Direction vs target (the headline)
Target end-state: package `xr_render_demo_worker/` = `supervisor.py` + `agents/{appearance,memory,object,placement,vision}/agent.py`, each a NAT `ToolCallAgentWorkflowConfig`. `capabilities.py` is a **split personality** — its front half converges, its back half is torn down. Net = **INTERMEDIATE-OK, leaning converge**: the expensive/risky migration is done right; what the target re-does is comparatively mechanical and builds *on top of* this PR rather than undoing it.

**CONVERGES (durable — the target is built on exactly this):**
- `build_native_toolbox()` registers the native groups `scene_state / scene_updates / scene_objects / scene_control / tracking / spatial_math / video_memory / vision` (+ `text_memory`). Target's `app.py` registers the *same* groups via `add_function_group(...)` and the subagents consume them by `FunctionGroupRef` (e.g. target `object/agent.py` refs `scene_updates, scene_objects, tracking, spatial_math`; `vision/agent.py` refs `vision`). This is the exact substrate the subagents need.
- Config swap `render_mcp/oxr_mcp/vlm_mcp/video_mcp/vec_mcp` → `scene_endpoint/openxr_endpoint/video_memory_endpoint/text_memory_dir` matches target's `config.py` field-for-field.
- Process-list/deps: deleting the oxr/video/vlm MCP servers + yaml, dropping `xr-ai-capabilities`+`fastmcp`, adding `xr-ai-nat[services,vision]`+`xr-render-scene` — matches the target worker's `pyproject.toml`. DEPENDENCIES.md updated in-commit.

**DEVIATES (throwaway — the subagent architecture removes it):**
- `NativeToolbox` (the `ToolDef` adapter: `.definitions(exclude=...)`/`.invoke()`) exists only to flatten every native function into one flat tool list for a **single** custom model loop (`processors.py::RenderSceneProcessor`, `_MAX_LOOP=10`). Target worker has **no `NativeToolbox`, no `ToolDef`, no `RenderSceneProcessor`** — it delegates the tool loop to NAT's per-subagent `ToolCallAgentWorkflowConfig` coordinated by `supervisor.py`. The monolithic single-LLM loop is precisely what the 5-subagent split replaces.
- `RenderSpatialToolsConfig` / `render_spatial_tools` — a ~250-line group of ~12 hand-authored convenience wrappers (`get_head_pose`, `position_ahead`, `place_object_relative`, `displace_objects`, `between_anchors`, `along_direction`, `scale_value`, …). Target has **no `render_spatial`**; it points the object subagent at the *primitive* `spatial_math`+`tracking` groups directly, plus a tiny per-agent `agents/object/size.py::ScaleObjectSizeConfig`. This group — the single largest chunk of net-new code in `capabilities.py` — is deleted/re-authored; only the primitives survive.
- `_WORKER_MANAGED_TOOLS = {"start_xr","get_health"}` exclusion + `_PERCEPTION_TOOL_DEF` append are flat-tool-list curation. Target consumes these control-plane functions directly from `scene_control` via `XRSessionController`, not by filtering a model-visible flat list.

**What must be re-done to reach target:** delete `capabilities.py`'s `NativeToolbox` + `render_spatial` group and `processors.py`'s monolithic loop; keep the `add_function_group(...)` registrations (they move to `app.py` verbatim). The native-function *wiring* is not re-done — that's why this is a legitimate intermediate, not a detour.

**Confirm:** `capabilities.py` as a named module/concept does **not** survive to target. Fine if #293 (or a follow-up) does the subagent decomposition; if the stack stopped here it would read as a monolith. Please flag `NativeToolbox`+`render_spatial` as deliberately transitional in the stack description.

### Findings by severity
**Critical/Important:** none blocking. The substantive feedback is the throwaway-surface point above (design/expectations, not a correctness bug).

**Conventions:** SPDX on `capabilities.py`; both commits carry DCO; DEPENDENCIES.md updated in-commit. All pass.

**Nits (below bar):** residual `render-mcp` strings in `scene/lovr/*.lua` log tags + one `eval.py` comment (cosmetic); `main.py::_build_processes(backend)` immediately `del backend` (param kept for signature stability — could drop it); default endpoints differ stylistically (`tcp://127.0.0.1:…` vs target `tcp://localhost:…`), worker YAML supplies the real values.

nvddr:
Addressed the architectural disposition from the modularization review: the PR Stack section now explicitly marks NativeToolbox, render_spatial, and the existing flat model loop as transitional compatibility scaffolding. The direct native function-group registrations are the durable substrate. #293 adds the models-backed NAT LLM provider and a read-only consumer; the supervisor/subagent migration and flat-loop replacement remain separate follow-up work.

wenxind-nvidia:
### Code review — two findings

**1. (Low / fragility) `processors.py:953-961` — perception no-frame detection by prose prefix-matching.**
`_look_at_current_frame` distinguishes a graceful "no frame" from a real answer by testing `answer.startswith(...)` against three literal strings (`"No camera frame available"`, `"Frame data unavailable"`, `"VLM server unavailable"`). These match today's `FrameUnavailable`/`StreamingVision` messages exactly, so it works now. Failure scenario: if any of those messages is reworded (capitalization, punctuation, wording) in `xr-ai-nat`/`xr_ai_agent`, a genuine no-frame turn silently returns `{"answer": "<error text>"}` instead of raising `_PerceptionUnavailableError` — so the model relays the raw error to the user as if it were a description of the scene, rather than speaking `_NO_FRAME_MSG`. Fix: have the native live-vision function report unavailability structurally (a status/sentinel on `VisionResult`) and branch on that instead of prose.

**2. (Low / pre-existing) `processors.py:814-816, 931` — agent-LLM failure returns a success-sounding reply.**
If `self._agent_llm.chat(...)` raises, the loop logs and `break`s, then falls through to `return "Done."` — the user hears "Done." for a hard model/transport failure (a silent failure). This is unchanged from `main` (same `break`→`"Done."` shape), so it's not a regression, but since the agentic loop is this PR's centerpiece it's worth fixing while here: return a distinct "something went wrong" string on the exception path (as `_run_turn`'s outer handler already does) rather than the same terminal `"Done."`.

wenxind-nvidia (approved):

nvddr:
Addressed the review feedback in 31ae92d:

- VisionResult now reports status ok or unavailable, and the render worker branches on that typed result instead of matching failure prose. The test deliberately varies the unavailable text to cover the regression.
- An agent-LLM exception now returns the same retry message as the outer error path rather than falling through to Done.
- Removed the unused build-processes argument and updated residual renderer naming in the LOVR diagnostics and eval comment.

I left the 127.0.0.1 endpoint spelling unchanged: it is consistent between the worker defaults and the shipped YAML; the target localhost form is stylistic and does not require a networking change here.

Validated with the focused vision/render tests (33 passed) and pre-commit (passed).

PR #295, opened 2026-07-23
ci: fix notify job failing with 'not a git repository'

yanziz-nvidia:
## Summary

- The `notify` (alert on failure) job runs on a fresh `ubuntu-latest` runner with **no checkout step**
- `gh` CLI falls back to running git to detect the repo, which fails with `fatal: not a git repository (or any of the parent directories): .git`
- Fix: add `GH_REPO: ${{ github.repository }}` to the step's env block — `gh` honors this env var and skips git detection entirely


wenxind-nvidia (approved):

yanziz-nvidia:
oh.. we need to fix the test first..

PR #300, opened 2026-07-24
fix(pipecat): align worker ready files

### LINKED ISSUE #299: fix(launcher): make worker ready files represent request readiness
## Problem

The ready-file contract says a managed process should create its file only when it is fully initialized and able to serve requests. The affected agent workers currently touch the file immediately after downstream service health checks:

- `simple-vlm-example` still has to construct the transport, brain, and Pipecat pipeline.
- `xr-render-demo` does so before opening its MCP clients and discovering their tools, as well as before building and starting the transport pipeline.

The launcher therefore prints `All processes ready` before the worker can receive and process client traffic. This is an early liveness signal, not request readiness.

## Desired outcome

- Define one explicit worker readiness boundary: its service dependencies have passed health checks, required startup work is complete, and the Pipecat input/IPC transport is active enough to receive requests.
- Touch the worker ready file only at that boundary.
- Keep process management in `xr-ai-launcher`; workers must not import launcher internals.
- Provide a reusable hook or barrier for Pipecat workers so samples do not recreate timing logic.
- Add CPU-only coverage proving the ready file is not released before the endpoint/transport startup path is active.

## Multi-client state contract

One hub serves many clients. The ready-file transition therefore represents worker/agent availability shared by those clients, not the state of an individual request. Client status must distinguish:

- shared availability: `loading`, `ready`, and eventually `failed`, latched and snapshotted to each joining client;
- per-client interaction: `processing` and `idle`, targeted only to the participant whose request is active.

The readiness barrier must not cause one client's interaction to change another client's UI, and a client joining after startup should receive the current availability snapshot rather than a stale startup sequence.

## Context

This surfaced while reviewing #296, which adds a client-facing loading/ready status. The launcher banner and client availability signal should eventually derive from the same logical worker readiness transition, while preserving the hub's ability to tell early-joining clients that startup is still in progress.

nvddr:
## Summary

- Add `run_voice_pipeline`, which starts a Pipecat worker and releases a callback only after inbound hub IPC is receiving.
- Move both Pipecat sample workers' ready-file touches to that shared readiness boundary, after startup work such as render-demo MCP discovery.
- Document the worker ready-file contract and add CPU-only coverage for the runner and transport barriers.

## Why

The launcher previously treated completed dependency health checks as worker readiness, although each worker still had to construct and start its request path. The shared callback now represents agent availability for all clients connected to the hub, not any individual participant's interaction state.

Closes #299

## Validation

- `pre-commit run --files agent-sdk/xr_ai_agent/_processor.py agent-sdk/xr-ai-pipecat/xr_ai_pipecat/transport.py agent-sdk/xr-ai-pipecat/xr_ai_pipecat/pipeline.py agent-sdk/xr-ai-pipecat/xr_ai_pipecat/__init__.py agent-samples/simple-vlm-example/worker/simple_vlm_example_worker.py agent-samples/xr-render-demo/worker/xr_render_demo_worker.py tests/test_xr_ai_pipecat_voice_pipeline.py docs/process-model.md docs/source/components/agent-sdk.md docs/changelog.md`
- `uv run --project tests pytest tests/test_xr_ai_pipecat_voice_pipeline.py tests/test_simple_vlm_example_wire.py tests/test_xr_render_demo_wire.py -q` (91 passed)


wenxind-nvidia:
### Code review

**Correctness: solid.** I traced `run_voice_pipeline`'s branches and they hold up:
- Normal path (input starts first): `on_ready` fires only after `wait_until_started`, then it awaits the runner to completion. ✓
- Runner exits/raises **before** input starts: `await runner_task` re-raises an error, or a clean early-exit raises `RuntimeError("exited before its input transport started")`. ✓ — good defensive check.
- `except BaseException` cancels the runner and awaits it (suppressing `CancelledError`) before re-raising — proper cooperative teardown, including outer cancellation. ✓
- `start()` awaits `ep.wait_until_running()`, and `run()` sets `_running_event` synchronously at the top (no preceding await), so the barrier can't deadlock and adds negligible latency. ✓

**1. (Minor) `_running_event` is never cleared on `stop()`.** `run()` does `clear()`→`set()`; `stop()` sets `_running = False` but leaves the event set. After a run→stop cycle, `wait_until_running()` would return immediately on a stale event. Harmless for the single-lifecycle worker, but the event no longer tracks `_running`. Consider clearing it in `stop()` for consistency.

**2. (Minor — coverage) the failure/teardown branches of `run_voice_pipeline` are untested.** The two new tests cover the happy path (`test_run_voice_pipeline_releases_ready_after_input_starts`) and the transport barrier ordering — good. But the function's reason to exist is careful ordering/teardown, and none of these are exercised: the `RuntimeError` on early clean exit, the runner-raises path, and the `except BaseException` → `worker.cancel()` teardown. A stub worker that records `cancel()` plus a runner that raises would lock in the teardown contract.

**3. (Nit) `on_ready` exceptions tear down the worker.** If `ready_file.touch()` raises (e.g. permissions), it propagates into `except BaseException` and cancels the pipeline after it's essentially up. Defensible (a launcher that can't see the ready file would hang anyway), just noting the coupling.

Conventions are clean: `__init__` docstring + `__all__` updated, and changelog / process-model / agent-sdk docs updated in-commit; DCO on both commits.

yanziz-nvidia (approved):

yanziz-nvidia (commented):

## Summary

Adds `run_voice_pipeline` as a shared runner that gates the ready-file callback on ZMQ receive-loop start, replaces premature `ready_file.touch()` calls in both sample workers, and introduces stateful agent-status tracking with periodic re-announcement for late/reconnecting clients.

|     | Finding |
|-----|---------|
| 🚫 | None |
| 💡 | [`agent-sdk/xr_ai_agent/_processor.py:338`](https://github.com/NVIDIA/xr-ai/blob/a581efec5371523f310f262acaba04d0515c6e58/agent-sdk/xr_ai_agent/_processor.py#L338) — `set_status(status, participant_id)` silently drops the call when `participant_id not in self._participants`. The docstring doesn't mention this constraint, so callers hitting a tight join-vs-status race will lose the update silently and fall back to the default. Add a note: "Has no effect if `participant_id` is not currently in `connected_participants`; call after the join callback fires." |
| 💡 | [`AGENTS.md:157`](https://github.com/NVIDIA/xr-ai/blob/a581efec5371523f310f262acaba04d0515c6e58/AGENTS.md#L157) — `run_voice_pipeline` is now the standard run step for every Pipecat worker but AGENTS.md still ends the voice-pipeline bullet at "hands it to the factory." DEPENDENCIES.md requires worker-boilerplate docs to track `agent-sdk/` API changes. Add a sentence: "then runs it with `run_voice_pipeline(worker, transport, on_ready=ready_file.touch)`." |
| 🔍 | [`tests/test_xr_ai_pipecat_voice_pipeline.py:1966`](https://github.com/NVIDIA/xr-ai/blob/a581efec5371523f310f262acaba04d0515c6e58/tests/test_xr_ai_pipecat_voice_pipeline.py#L1966) — Docstring says "The transport barrier waits for the endpoint's initial roster" but roster catch-up is async and doesn't gate the barrier. Replace with "The startup barrier is released once the endpoint receive loop has started, independent of roster convergence." |
| 🔍 | [`agent-sdk/xr_ai_agent/_codec.py:283`](https://github.com/NVIDIA/xr-ai/blob/a581efec5371523f310f262acaba04d0515c6e58/agent-sdk/xr_ai_agent/_codec.py#L283) — `lambda _m` / `lambda _p` changed to `lambda m` / `lambda p`, dropping the unused-parameter signal. Restore the underscore prefix. |

- `agent-sdk/xr_ai_agent/_processor.py:338` — Add to `set_status` docstring: "Has no effect if `participant_id` is not currently in `connected_participants`; call after the join callback fires."
- `AGENTS.md:157` — After "hands it to the factory", add: "then runs it with `run_voice_pipeline(worker, transport, on_ready=ready_file.touch)`."
- `tests/test_xr_ai_pipecat_voice_pipeline.py:1966` — Replace docstring with "The startup barrier is released once the endpoint receive loop has started, independent of roster convergence."
- `agent-sdk/xr_ai_agent/_codec.py:283` — Restore `lambda _m: []` and `lambda _p: RosterRequest()`.

PR #304, opened 2026-07-27
refactor(nat): move the MCP export to xr_ai_nat.mcp

wenxind-nvidia:
Second `xr-ai-nat` alignment PR (independent of #303). Pure relocation — no signature, behavior, or dependency change.

## What changes
- Move the generic native-function → MCP publisher from `xr_ai_nat/adapters/mcp.py` to a top-level `xr_ai_nat/mcp/server.py`, with `xr_ai_nat/mcp/__init__.py` re-exporting `create_mcp_server`.
- `mcp/` now owns exposing native capabilities to MCP-only agents; `adapters/` is reserved for framework adapters (e.g. the voice adapter in a later phase).
- Importers rewired: `transcript-mcp`, `vlm-mcp`, `tests/test_text_memory_functions.py`, and the `xr-ai-nat` README.

## Why it's safe
- Git detects it as a pure rename (`adapters/mcp.py => mcp/server.py`, 0 content changes); `create_mcp_server`'s signature/behavior are identical.
- Packaging is `packages = ["xr_ai_nat"]` (hatchling includes subpackages), so no `pyproject.toml` / dependency change.

## Verification
- `ruff check` (CI-pinned v0.15.16) clean.
- `pytest` green (17 passed): `test_text_memory_functions`, `test_vlm_mcp`, `test_transcript_mcp` (all exercise `create_mcp_server` through the new path).

## Coordination note
Any in-flight branch importing `xr_ai_nat.adapters.mcp` needs the one-line rewrite to `xr_ai_nat.mcp` on rebase.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #305, opened 2026-07-27
refactor(nat): inline video-memory contracts into the typed client

wenxind-nvidia:
Phase-1 `xr-ai-nat` capability alignment (video_memory). Folds the
`functions/video_memory/schemas.py` model definitions into `_client.py` and
adopts the target's renamed, `_StrictRequest`-based model surface — while
preserving main's merged behavior and a deprecated forwarding alias for the old
public names.

## What changes
- Request/result models inlined into `_client.py`. Renames: `EmptyRequest` →
  `VideoHealthRequest` + `ListRecordedParticipantsRequest`; `ParticipantsResult`
  → `ListRecordedParticipantsResult`; `VideoMemoryHealth` → `VideoHealthResult`.
  Requests extend `_StrictRequest`.
- `functions/video_memory/schemas.py` is **retained as a deprecated forwarding
  alias**: it re-exports the renamed models and keeps the old names as
  deprecated aliases (`EmptyRequest` → `ListRecordedParticipantsRequest`,
  `ParticipantsResult` → `ListRecordedParticipantsResult`, `VideoMemoryHealth` →
  `VideoHealthResult`; same data contracts) and emits a `DeprecationWarning` on
  import. The package-level `ParticipantsResult` export is retained too.
- `_client.py`: `list_recorded_participants(request=None)` and
  `get_health(request=None)` build their (field-less) typed request internally,
  so both the intent-named typed NAT Function **and** the no-argument call site
  keep working.
- `functions.py`: retains the detailed agent-facing function/config descriptions
  (part of the model's tool-use contract) + `VideoMemoryClient.health()`;
  `__init__.py` re-exports the new surface.
- `xr_video_memory_control` is **deferred** — removed until it has a real
  readiness-orchestration consumer (registering an unused group would be dead
  surface). `get_health` / `health` are retained (the shim depends on them).
- Rename rippled to consumers: the video-memory-service dispatcher/validators,
  the video-mcp compat shim, and the tests.

## Rebased onto current main (#303 / #304 merged)
- Imports the canonical `xr_ai_nat.functions._service.rpc` `RPCClient` (main
  deleted `_rpc` in #303); resolved the `docs/changelog.md` overlap with the
  merged #303 / #304 entries.

## Behavior preserved (do-not-reverse audit)
- **#289** — the video-mcp shim's conditional tool sets and outage fallback are
  unchanged (`VideoHealthResult` keeps `recording_enabled`, which
  `_recording_enabled()` reads); `select_decoded_frame` untouched.
- Kept main's field constraints/descriptions + `QueryVideoRequest` /
  `HistoricalFrameRequest` validators (`gt=0` on `reference_time_us`,
  agent-facing descriptions) rather than the target's leaner models.
- Kept `get_health(request=None)` optional + the 2 s probe timeout so the shim's
  no-arg health probe keeps working.

## Verification
- `ruff check` (v0.15.16) clean across `agent-sdk/xr-ai-nat`,
  `services/video-memory-service`, `agent-mcp-servers/video-mcp`, `tests`.
- `pytest test_video_memory_functions.py` → **21 passed**, including the
  real-signature `list_recorded_participants` happy-path, the no-argument
  compatibility call, the deprecated-alias / legacy-name regression tests, and
  the #289 conditional-tool-set / outage-fallback tests.


nvddr:
High-level review: the typed video-memory direction is right and it correctly preserves main stricter contracts instead of copying every fork change. There is a runtime regression to fix before merge: Video MCP still calls list_recorded_participants with no request, but the new client requires ListRecordedParticipantsRequest, so participant discovery raises TypeError before RPC handling. Please pass the empty request (or retain a no-argument compatibility path) and add a real-signature MCP happy-path test. Also decide whether removed schemas and old names need a compatibility shim, and either wire, document, and test xr_video_memory_control or defer that currently unused group to the readiness-orchestration phase.

wenxind-nvidia:
**Addressed in `43a0459`.**

**Problem:** a real runtime regression — the Video MCP `list_recorded_participants` tool called `client.list_recorded_participants()` with no argument, but the typed client now requires `ListRecordedParticipantsRequest`, so participant discovery raised `TypeError` before any RPC handling. The existing tests missed it because their stub client had a no-argument signature. Separately, `xr_video_memory_control` was registered but had no consumer or test on main.

**Edit:**
- The shim now passes `ListRecordedParticipantsRequest()` (imported from `xr_ai_nat.functions.video_memory`).
- Added `test_video_mcp_lists_recorded_participants_over_a_real_client` — it stands up an in-memory `RPCServer`, points a **real** `VideoMemoryClient` at it, and calls the MCP tool. It fails on the pre-fix code with exactly `TypeError: ... missing 1 required positional argument: 'request'` and passes after. Also fixed the pre-existing `_UnavailableRecordedClient` stub whose no-arg signature would otherwise mask the regression.
- Deferred the unused control group: removed `VideoMemoryControlFunctionsConfig` + its registration (it lands with its readiness-orchestration consumer later). `VideoMemoryClient.get_health()`/`health()` are retained — the shim's `_recording_enabled()` depends on them.

**Resolution:** fully resolved — the `TypeError` is fixed and now guarded by a real-signature happy-path test; 20 tests pass.

nvddr:
## Follow-up review

### Blockers

1. **Rebase onto current `main`.** #305 now conflicts after #303/#304 landed. Reconcile the RPC imports with the canonical `xr_ai_nat.functions._service.rpc` path, resolve any changelog overlap, and rerun the focused and CI checks. The current green checks predate that integration.

2. **Restore the detailed agent-facing function and config descriptions in `video_memory/functions.py`.** The request-field constraints and descriptions remain strong, but the function descriptions were shortened enough to drop essential guidance: discovery must precede recorded queries; statistics establish the absolute Unix-microsecond range; clip windows are absolute; and recorded-frame extraction is never live-camera access. Keep the endpoint and timeout descriptions as well. These descriptions are part of the model's tool-use contract, not cosmetic prose.\n\n### Suggestions\n\n- Deferring `xr_video_memory_control` is correct while no caller consumes it; registering it now would be dead surface. When readiness orchestration lands, add the group with its real consumer and retain `get_health` / `is_available` as native functions. If the model itself needs to distinguish “recording disabled” from “no history yet,” include the structured `get_health` result (including `recording_enabled`) in its native toolbox or injected capability context. Do not route that through MCP.\n\n### Nits\n\nNone.

wenxind-nvidia:
**Addressed in `918e0c0` (branch rebased onto current main).**

**Blocker 1 — rebase.** Rebased onto current `main`. The RPC client, service, and tests now import the canonical `xr_ai_nat.functions._service.rpc` (main deleted `_rpc` in #303); resolved the `docs/changelog.md` overlap with the merged #303/#304 entries. Re-ran ruff + the focused suite green post-integration.

**Blocker 2 — restore the agent-facing descriptions.** Restored the detailed function/config descriptions in `video_memory/functions.py` — they are part of the model's tool-use contract:
- `list_recorded_participants`: "List exact participant identities with persisted camera history. Call this before requesting recorded statistics, clips, or frames." (discovery precedes recorded queries)
- `get_video_stats`: "Return a participant's recorded Unix-epoch microsecond range and storage statistics. Use the range to validate absolute clip windows."
- `query_video`: "Write an H.264 clip overlapping an absolute Unix-epoch microsecond window and return its local path." (windows are absolute)
- `get_frame_from_time`: "Extract the recorded PNG frame nearest reference_time_us minus second_ago whole seconds. This never accesses a current live camera frame."
- Config: restored `endpoint` ("Private msgpack/ZMQ endpoint of video-memory-service.") and the dropped `timeout_s` description ("Maximum time to wait for an H.264 query or frame extraction."). Field constraints were already intact.

**Nits — mixed imports.** Consolidated the alias compat test so `_client` and `schemas` are each imported one way (dropped the `import … as` / `from … import` duplication), clearing both code-quality threads.

**Suggestion — `get_health` in the native toolbox.** Agreed and noted for the readiness-orchestration phase: `xr_video_memory_control` stays deferred until it has a real consumer, and when it lands the structured `get_health` (incl. `recording_enabled`) will be surfaced through the native toolbox / injected capability context, not routed through MCP. No change in this PR.

Verification: `ruff@0.15.16` clean; `test_video_memory_functions.py` → 21 passed (incl. the real-signature `list_recorded_participants` happy-path test and the consolidated alias test).

nvddr:
## Current review disposition

### Blockers

1. **Preserve the deprecated video-memory public surface.** The retained `video_memory.schemas` alias still omits `EmptyRequest`, `ParticipantsResult`, and `VideoMemoryHealth`; the package-level `ParticipantsResult` export is gone as well. That is not a complete forwarding alias. `ParticipantsResult` and `VideoMemoryHealth` have the same data contracts as their renamed replacements, so keep deprecated aliases for them (and keep the legacy empty request import on the deprecated module).

   Also retain the compatible no-argument `VideoMemoryClient.list_recorded_participants()` call by creating `ListRecordedParticipantsRequest()` internally, as `get_health()` already does. The native Function can continue to expose the intent-named typed request. Add regression coverage for the legacy names and the no-argument call.

2. **Correct the checked-in changelog.** It still says the video-memory client imports `.._rpc` and that `_service.rpc` is a separate open PR. The rebased code correctly imports `.._service.rpc`, and #303 is merged.

### Suggestions

- Update the PR description to match the current branch: `schemas.py` is retained as a deprecated alias, `xr_video_memory_control` is deferred, and #303/#304 are merged.

### Nits

None.

wenxind-nvidia:
**Both blockers fully resolved in `392f041`.** You were right on both — my earlier edits were incomplete.

### Blocker 1 — deprecated public surface (my alias was incomplete)
What was wrong: I re-exported only the *unchanged-name* models and documented `EmptyRequest`/`ParticipantsResult`/`VideoMemoryHealth` as "renamed → breaking, not aliased." But a rename with an **identical data contract** can and should be aliased — I gave up on it too early. I also fixed the no-argument call at the wrong layer (the shim) instead of the client, and dropped the package-level `ParticipantsResult` export.

Now:
- **Complete forwarding alias.** `video_memory.schemas` re-exports the full legacy surface: unchanged names **plus** the renamed ones as aliases (contracts verified identical) — `ParticipantsResult = ListRecordedParticipantsResult`, `VideoMemoryHealth = VideoHealthResult`, and the legacy `EmptyRequest = ListRecordedParticipantsRequest` (both field-less strict requests).
- **Package-level `ParticipantsResult` export restored** (as a deprecated alias).
- **No-argument client call retained at the client:** `VideoMemoryClient.list_recorded_participants(request=None)` builds `ListRecordedParticipantsRequest()` internally, exactly like `get_health()`. The native Function still exposes the intent-named typed request; the Video MCP shim reverts to the no-argument call.
- **Regression coverage added:** the legacy-name aliases (`schemas.ParticipantsResult/VideoMemoryHealth/EmptyRequest`), the package-level `ParticipantsResult`, and a real-server no-argument `client.list_recorded_participants()` call.

### Blocker 2 — changelog
Corrected: the entry now states the client imports the canonical `xr_ai_nat.functions._service.rpc` (#303, merged), and describes the renamed models as **kept as deprecated aliases** (identical contracts) rather than breaking removals.

### Suggestion
PR description updated to match the current branch (schemas retained as a deprecated alias, `xr_video_memory_control` deferred, #303/#304 merged).

Verification: `ruff@0.15.16` clean; `pytest test_video_memory_functions.py` → 21 passed.

nvddr:
## Follow-up blocker — preserve the no-input NAT tool contract

The restored no-argument compatibility on `VideoMemoryClient.list_recorded_participants(request: ListRecordedParticipantsRequest | None = None)` is appropriate for legacy Python callers, but `video_memory_functions` registers that method directly. The NAT-generated schema therefore now exposes an optional nullable wrapper:

```json
{"request": {"anyOf": ["ListRecordedParticipantsRequest", null], "default": null}}
```

rather than the original strict empty-object input schema for this no-input operation. An agent should simply call `list_recorded_participants` with `{}`; it should not need to know about the Python compatibility parameter. This also means the claim that the native Function still exposes the intent-named typed request is not true in the generated tool schema.

Please keep the compatibility behavior on `VideoMemoryClient`, but register a thin strict native wrapper whose parameter is a required `ListRecordedParticipantsRequest` and forwards it to the client. Add a regression test that inspects the generated input schema and asserts there is no optional/nullable `request` wrapper. That keeps legacy direct callers working without leaking the migration detail into the agent-facing contract.

wenxind-nvidia:
**Resolved in `48bba09`.**

You're right — registering the client method directly leaked its Python compatibility parameter into the agent-facing schema (and it was actually a do-not-reverse: main had a thin wrapper with a required request, which my fold dropped).

Fix:
- **Client keeps the compatibility form** — `VideoMemoryClient.list_recorded_participants(request=None)` still builds the request internally for legacy Python callers (and the Video MCP shim's no-argument call).
- **Native function is now a thin strict wrapper** — `async def list_recorded_participants(request: ListRecordedParticipantsRequest) -> ListRecordedParticipantsResult` forwarding to the client. So the NAT-generated tool schema is a strict empty object again; an agent calls it with `{}` and never sees the compat parameter. The intent-named typed request is genuinely what the tool exposes now.
- **Regression test** `test_list_recorded_participants_tool_schema_is_strict_empty` inspects the generated `input_schema.model_json_schema()` and asserts `properties == {}` with no optional/nullable `request` wrapper.

Verification: `ruff@0.15.16` clean; `pytest test_video_memory_functions.py` → 22 passed.

nvddr (approved):

PR #306, opened 2026-07-27
refactor(nat): typed text-memory API + native conversation recall

wenxind-nvidia:
Phase-1 `xr-ai-nat` capability alignment (text_memory), rebased on current main.

## What changes
- Adopt the typed request/result text-memory API: request models extend `_StrictRequest` (input validation replaces error-as-data), functions return typed result models. `_store.py` folds into `functions.py`.
- Rebased onto main: the transcript-mcp shim and tests import the canonical `xr_ai_nat.mcp` (#304, merged).
- The `xr_conversation_memory` group / `recall_conversation` / `ConversationMemoryFunctionsConfig` are **not** in this PR — deferred (no producer on main) and re-landing with their producer in the R5 voice-adapters PR (#309).

## Deprecated public surface preserved
`text_memory/schemas.py` is retained as a deprecated forwarding alias (warns on import) — it is **not deleted**:
- `TranscriptSegment` — unchanged, re-exported.
- `TranscriptStats` — renamed to `TranscriptStatsResult` but **field-identical** (only `earliest_us`/`latest_us` widened `int` → `int | None`), so kept as a deprecated alias.
- `OperationResult` / `TextMemoryError` — **genuine removals** (error-as-data → validation + typed results), not aliased.
The private `_store.py` is removed outright (underscore-private, no compatibility surface).

## Do-not-reverse
- Restored main's detailed agent-facing function/config descriptions and per-field request descriptions (the model's tool-use contract).
- The `#294` path-escape guard is preserved: every `.identity`/`.jsonl` path is `_check()`-wrapped before use; the symlink-guard test is retained and passes.

## Verification
- `ruff@0.15.16` clean; `pytest test_text_memory_functions.py test_transcript_mcp.py` → **8 passed**, incl. the symlink guard and the `schemas` alias test (`TranscriptSegment` + `TranscriptStats`).


nvddr:
High-level review: the typed text-memory surface and conversation-recall boundary match the eventual modular memory agent, and the branch correctly preserves the stronger path-escape checks that the fork lacks. Please do not leave recall as an unconnected contract on current main: the render worker currently writes both turns under the bare participant ID with role text prefixes, while recall_conversation reads participant:user and participant:agent sources. If wired today it returns an empty history. Either defer the group until its producer migration, or transition the writer and support existing transcripts, with an end-to-end recall test. Also preserve the legacy Transcript MCP result/error wire shape inside the compatibility shim (or explicitly version that API), since query and sources now return wrapper objects rather than the prior bare lists.

wenxind-nvidia:
**Addressed in `307df47`.**

**Problem (2 parts):**
1. `recall_conversation` reads sources named `{participant_id}:{role}`, but nothing on current main produces them — the render worker writes transcripts under the bare `participant_id`. So the group is an unconnected contract that returns empty history if wired.
2. Dropping `untyped_outputs` in the compatibility shim changed `query_transcripts`/`list_sources` MCP outputs from bare lists to typed wrapper objects (`{"segments": …}` / `{"sources": …}`).

**Edit:**
1. **Deferred** the conversation capability to its producer: removed `ConversationMemoryFunctionsConfig`, `conversation_memory_functions`, `RecallConversationRequest`/`Result`, and `ConversationEntry` from `functions.py` + `__init__.py`. It will land alongside the `record_voice_transcripts` writer (which will produce the `:user`/`:agent` sources) with an end-to-end recall test. The core typed surface and the `_check` path-escape guard are untouched — the symlink-guard test still passes.
2. **Explicitly versioned** the shape change (your offered alternative): documented in the transcript-mcp module docstring and the changelog that `query_transcripts`/`list_sources` now return typed objects rather than bare lists — a deliberate change with the typed native API. Transcript MCP is a transitional compatibility process with no in-tree consumer, so I versioned the change rather than reproduce the legacy bare-list shape (which would require reverting the typed outputs the tests now assert).

**Resolution:** part 1 fully resolved (deferred to its producer migration). Part 2 resolved via explicit versioning rather than legacy-shape preservation; if a bare-list-consuming client turns up before transcript-mcp is retired, I can add a shim-local unwrap then.

wenxind-nvidia:
**Blocker resolved in `d7fed33`; PR description updated.**

**Blocker — incomplete `text_memory.schemas` alias.** You're right, and it's the same incompleteness from #305. `TranscriptStats` → `TranscriptStatsResult` is a same-contract rename (field-identical, only `earliest_us`/`latest_us` widened `int` → `int | None`), so it stays as a deprecated alias:
- `schemas.py` now imports `TranscriptStatsResult` and aliases `TranscriptStats = TranscriptStatsResult`, added to `__all__`.
- The schemas-alias test asserts `schemas.TranscriptStats is TranscriptStatsResult` (and `__all__ == ["TranscriptSegment", "TranscriptStats"]`).
- `OperationResult` / `TextMemoryError` stay **genuine removals** (error-as-data → validation + typed results, no consumers) — not aliased, as you noted.
- Changelog updated to describe `TranscriptStats` as kept-as-alias rather than a breaking rename.

**Suggestion — PR description.** Rewritten to match the branch: `schemas.py` is a deprecated alias (not deleted), `xr_conversation_memory`/`recall_conversation` is deferred (re-lands in the R5 voice-adapters PR, not here), the branch is rebased and imports the canonical `xr_ai_nat.mcp`, and the suite is 8 passed.

Verification: `ruff@0.15.16` clean; `pytest test_text_memory_functions.py test_transcript_mcp.py` → 8 passed.

nvddr (changes requested):
## Review disposition

### Blocker

- **Rebase onto current `main`.** PR #305 has merged, and GitHub now reports this PR as `CONFLICTING` / `DIRTY`. The integration conflict is the adjacent `docs/changelog.md` entries; preserve both the merged video-memory decision and this text-memory decision, then re-run the focused text-memory/transcript-MCP tests.

### Assessment

The typed native text-memory direction is sound: it preserves the stronger path-escape guard that the modular fork lacks, keeps MCP as a compatibility adapter, and correctly defers conversation recall until the producer exists. The explicit transcript-MCP object-output change is documented and acceptable for the transitional shim. No other blockers found in the current diff.

nvddr (approved):

PR #271, opened 2026-06-30
fix(vision): remove camera-on-demand re-introduced by PR #266

nvddr:
## Summary

PR #180 intentionally removed always-on-demand camera control (`startCamera`/`stopCamera`) in favour of always-on streaming. PR #266 accidentally re-introduced it by extracting `VisionModule` from `xr-render-demo`, which still had the code.

- **`VisionModule`**: remove `warmup()`, `_ensure_camera_on()`, `_schedule_camera_off()`, `_camera_session()`, `_client_control()`; remove `_camera_on`/`_camera_held`/`_camera_off_timers` state; replace `camera_on_timeout_s`/`camera_grace_s` constructor params with `frame_timeout_s`
- **`simple-vlm-example`**: drop `warmup()` call from `on_user_started_speaking`, update constructor args and YAML config keys
- **`xr-render-demo`**: update `VisionModule` constructor args

Camera is now always-on streaming; `VisionModule` simply waits up to `frame_timeout_s` for a fresh frame.

## Test plan

- [x] `simple-vlm-example` runs without error; VLM answers use live frames
- [x] `xr-render-demo` `look_at_current_frame` tool still works
- [x] No `startCamera`/`stopCamera` messages sent to clients during a session
- [x] `VisionModule` raises `VisionUnavailable` if no frame arrives within `frame_timeout_s`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #307, opened 2026-07-27
feat(nat): adopt native xr_vision_tools current/past-frame surface

wenxind-nvidia:
`xr-ai-nat` vision capability migration — rewires the #292 render worker onto a native perception surface.

## What changes
- **SDK vision module:** replace the path-based `VisionFunctionsConfig`/`xr_vision`/`ask_image` with a native `VisionToolsConfig` (`xr_vision_tools`) exposing `look_at_current_frame` (live) and `look_at_past_frame` (recorded, via the `video_memory` group). Rename `_images.py` → `_pixels.py` (`frame_to_pil`/`encode_image`). `StreamingVisionConfig` (`xr_streaming_vision`) is preserved with its richer streaming/status/`FrameUnavailable` behavior (simple-vlm depends on `.astream`).
- **#292 worker rewiring:** `capabilities.py` registers `VisionToolsConfig(endpoint=…, vlm=…, video_memory=FunctionGroupRef("video_memory"))` and returns the vision config so the worker can wire `on_participant_left → release`. `processors.py` routes `_look_at_current_frame`/`_look_at_past_frame` through the native tools, injecting `participant_id` (and the utterance `reference_time_us` for recorded lookups), which the model never supplies.
- **Trimmed model-facing perception schemas:** the native request models (`LiveVisionRequest`/`HistoricalVisionRequest`) expose the injected `participant_id`/`reference_time_us` — context the worker supplies, not the model. The worker and eval therefore **exclude** the two native perception tools from the toolbox definitions and append hand-crafted trimmed `ToolDef`s (`look_at_current_frame` → `{question}`, `look_at_past_frame` → `{question, second_ago}`), preserving main's `_PERCEPTION_TOOL_DEF` contract so the model is never told to fill a `participant_id` it cannot know.
- **vlm-mcp compat:** since the native surface no longer offers a path tool, the file-based `ask_image` is re-homed into the shim itself (`vlm_mcp_server/_ask_image.py`, self-contained NAT group with the identical description/error-string/think-strip contract). vlm-mcp is deleted in a later phase.

## Notable improvement
The old worker detected "no frame" by prose prefix-matching the VLM answer. Now `look_at_current_frame` **raises** `FrameUnavailable` on no-frame, which the processor catches → graceful spoken message. This structurally removes the fragile string match.

## Do-not-reverse audit
- **#271 always-on:** no camera-on-demand/`startCamera` path reintroduced; vision stays always-on streaming (`test_perception_no_frame_yields_graceful_message` asserts no client camera-control sends).
- **Per-turn client status at the executor boundary:** the render turn brackets itself with `set_status("processing"/"idle")` — the `processing` publish sits inside the `try` and `idle` clears in `finally`, so it survives failure and barge-in cancellation (incl. cancellation during the initial publish); the native vision functions stay UI-free. `HistoricalVisionRequest.reference_time_us` is required `gt=0` to match the forwarded `HistoricalFrameRequest`.
- **Trimmed perception contract:** restores main's strict `{question}` model-facing schema (guarded by `test_model_facing_perception_schema_is_trimmed`).
- **Rebased onto current `main`** (#303–#306 merged): vlm-mcp now imports the canonical `from xr_ai_nat.mcp import create_mcp_server` (#304); the `docs/changelog.md` entries from the merged PRs are preserved alongside this one. SDK vision keeps main's stable `from xr_ai_agent import FrameUnavailable, LiveFrameSource, …` paths.

## Verification
- `ruff check` (v0.15.16) clean across the SDK, vlm-mcp, both samples, and tests.
- `pytest test_vision_functions.py test_vlm_mcp.py test_xr_render_demo_wire.py test_simple_vlm_example_wire.py` → **57 passed** (warning-free after de-deprecating the `video_memory` test import). `git diff --check` clean.
- Public docs updated in the same change (README, `DEPENDENCIES.md`, `AGENTS.md`, eval README) to describe `VisionToolsConfig` / `look_at_current_frame` / `look_at_past_frame` and VLM MCP's file-path role.


nvddr:
High-level review: replacing the local two-step perception path with typed native current and recorded vision tools is the right direction, and retaining main richer streaming behavior is preferable to copying the fork unchanged. Please keep live capture truly independent of video memory: VisionToolsConfig eagerly resolves the video-memory group even for look_at_current_frame, so a live-only consumer cannot construct the tool without configuring recorded storage. Split or lazily compose the historical capability instead. The fork has the same coupling, so it is not a reason to retain it. Please also finish the public schemas with descriptions for every request/result field and correct the render prompt wording that says the tool turns the camera on; the implementation uses the always-on frame path.

wenxind-nvidia:
**Addressed in `0a139ec`.**

**Problem (3 parts):**
1. `VisionToolsConfig`/`vision_tools` eagerly resolved the video-memory group at group setup, so a live-only consumer could not construct or call `look_at_current_frame` without also configuring recorded storage.
2. Some public request/result fields lacked descriptions.
3. The render system prompt said `look_at_current_frame` "turns the camera on," but the implementation uses the always-on live-frame path.

**Edit:**
1. Made the recorded dependency **lazy**: removed the eager resolution at setup; `video_memory` / `get_frame_from_time` are now resolved (and cached) inside `look_at_past_frame` on first invocation. `look_at_current_frame` never touches them, so it builds and runs with no video-memory group present. Added `test_look_at_current_frame_builds_and_runs_without_video_memory` (constructs `VisionToolsConfig` with no video-memory group and successfully invokes the live tool) to lock in the decoupling. Behavior is identical when video-memory *is* configured.
2. Added `Field(description=...)` to the five fields that lacked one: `VisionChunk.text`, `LiveVisionResult.answer`, `HistoricalVisionRequest.participant_id`/`question`/`reference_time_us`.
3. Reworded `system.txt` to "look_at_current_frame inspects the user's current live camera frame — the camera is always on, so you do not need to capture a frame or pass a file path." (A grep confirmed this was the only "turns the camera on"-style wording.)

**Resolution:** all three fully resolved; ruff clean and the vision/vlm/render/simple-vlm test suites pass (52).

wenxind-nvidia:
**Blocker (perception schema) resolved in `38ecc2a`; PR description updated via REST PATCH.**

**Blocker — model-facing perception schema leaked `participant_id`/`reference_time_us`.** Correct, and it's a do-not-reverse regression of main's stricter contract plus an agent-facing interface change. The native `LiveVisionRequest`/`HistoricalVisionRequest` carry the participant/reference context the worker *injects* (`processors._look_at_current_frame`/`_look_at_past_frame`) — the model never supplies it — but registering those functions directly fed their generated schemas straight to the LLM through `NativeToolbox.definitions()`, since `_WORKER_MANAGED_TOOLS` only excluded `start_xr`/`get_health`. So the model was told to fill a required `participant_id` it cannot know and whose value is discarded.

Fix (mirrors main's `_PERCEPTION_TOOL_DEF` pattern):
- Added trimmed model-facing `ToolDef`s in `processors.py`: `look_at_current_frame` -> `{question}` (required), `look_at_past_frame` -> `{question, second_ago}` (both required).
- Worker **and** eval now exclude the two native perception tools from `toolbox.definitions(...)` and `extend` the trimmed defs (main's eval did the same `definitions.append(_PERCEPTION_TOOL_DEF)`, which this PR had dropped — fixed there too).
- Native SDK invocation is unchanged: the processor still injects `participant_id`/`reference_time_us` when it calls `toolbox.invoke(...)`.

**Regression test.** `test_model_facing_perception_schema_is_trimmed` asserts the *raw* native schemas expose `participant_id` (proving the trim is meaningful) and that the assembled model-facing perception schemas do **not** — trimmed to exactly the model-supplied fields (`{question}` and `{question, second_ago}`). It fails against the pre-fix assembly.

**PR description.** Rewritten to describe the trimmed-perception handling and corrected to **53 passed** (my new test brings it to 53). Applied through `gh api ... --method PATCH` so it actually persists.

Verification: `ruff@0.15.16` clean; `pytest test_vision_functions.py test_vlm_mcp.py test_xr_render_demo_wire.py test_simple_vlm_example_wire.py` -> 53 passed.

wenxind-nvidia:
**github-code-quality — `_PERCEPTION_TOOL_DEFS` "unused global" (processors.py):** false positive from single-file analysis. The tuple is consumed cross-module — `xr_render_demo_worker.py` and `eval/eval.py` both import it and append it onto the model-facing tool list (the same import pattern main used for `_PERCEPTION_TOOL_DEF`). No change.

nvddr (changes requested):
Full current review of `38ecc2a` against today's `main`:

The native design is now on the right track: current-frame vision is independent of video memory, historical lookup is lazy, the worker injects participant/timing context while the model receives only its own fields, and the file-path VLM MCP remains a compatibility adapter. `_PERCEPTION_TOOL_DEFS` is used by both the worker and eval, so the code-quality report is a false positive.

**Blockers**

1. **Rebase onto current `main` before merge.** The branch is behind `#303`–`#306` and GitHub reports it as dirty. The rebase conflicts in VLM MCP's import and `docs/changelog.md`; retain the canonical `from xr_ai_nat.mcp import create_mcp_server` import and preserve the merged changelog entries. Update the PR description too: after this rebase its “no relocation coupling” statement and “53 passed” result are no longer current.

2. **Update the public guidance in the same change.** This PR deletes `VisionFunctionsConfig` / `xr_vision` / `vision__ask_image`, but `agent-sdk/xr-ai-nat/README.md`, both affected entries in `DEPENDENCIES.md`, the vision section of `AGENTS.md`, and `agent-samples/xr-render-demo/eval/README.md` still tell users or agents to use that removed path. Describe `VisionToolsConfig` (`look_at_current_frame` / `look_at_past_frame`) and VLM MCP's narrow file-path compatibility role instead.

**Suggestions**

- `tests/test_vision_functions.py` now imports the deprecated `video_memory.schemas` alias. Use the canonical package exports so the focused suite is warning-free.
- Update the touched perception-test docstring: live capture is always on; the tool no longer “turns the camera on.”

Validation on the rebased integration state: the focused vision/VLM/render/simple-VLM suite passes (**52 passed**); its only project warning is the deprecated test import above. `git diff --check` is clean.

wenxind-nvidia:
All blockers and suggestions addressed; both are valid.

**Blocker 1 — rebase onto current `main`.** Rebased onto `main` (with #303–#306 merged). Two conflicts, resolved as directed:
- `agent-mcp-servers/vlm-mcp/vlm_mcp_server/__main__.py`: kept the canonical `from xr_ai_nat.mcp import create_mcp_server` (#304) and dropped main's `VisionFunctionsConfig` import (this PR removes it; the shim uses its own self-contained `AskImageConfig`).
- `docs/changelog.md`: kept all merged entries and placed this PR's "Native vision exposes current- and recorded-frame tools" entry at the top (reverse-chronological).
PR description updated: replaced the now-stale "no relocation coupling" note with the rebase summary and corrected the count to **52 passed**.

**Blocker 2 — public guidance updated in the same change.** The removed `VisionFunctionsConfig`/`xr_vision`/`vision__ask_image` references now describe `VisionToolsConfig` (`look_at_current_frame`/`look_at_past_frame`) and VLM MCP's file-path compatibility role:
- `agent-sdk/xr-ai-nat/README.md` — rewrote the Vision section.
- `DEPENDENCIES.md` — both entries (xr-ai-nat capability description; vlm-mcp self-contained file-path `ask_image`, `[vision]` extra still needed for Pillow).
- `AGENTS.md` — the reusable-modules description and the "image acquisition vs vision reasoning" rule (native tools own frame acquisition; the file-path adapter is separate in VLM MCP).
- `agent-samples/xr-render-demo/eval/README.md` — stubbed-queries note.

**Suggestions.**
- `tests/test_vision_functions.py` now imports `HistoricalFrameRequest`/`HistoricalFrameResult` from the canonical `xr_ai_nat.functions.video_memory` — the focused suite is warning-free (only the third-party pipecat `audioop` warning remains).
- Corrected the `look_at_current_frame` test docstring: live capture is always-on; the tool pulls the current live frame rather than turning the camera on.

Verification: `ruff@0.15.16` clean; focused vision/VLM/render/simple-VLM suite → 52 passed; `git diff --check` clean.

nvddr:
**Follow-up — correcting the status recommendation in my review.**

The prior rebase/docs blockers are resolved. The remaining status gap is real, but the correct owner is the top-level render turn, **not** `VisionToolsConfig`: `_agent.status` is a per-client UI lifecycle signal, while native functions must stay reusable and UI-free.

**Blocker — restore per-turn status at the executor boundary.** `#307` removes the old `StreamingVisionConfig` path that emitted status, and `RenderSceneProcessor._run_turn` does not replace it, so no status currently reaches the client. Wrap the complete turn for the originating `pid` with `processing` at entry and `idle` in `finally` (including failure/cancellation). Do not add status calls back into `look_at_current_frame` or `look_at_past_frame`.

The render loop already sends per-tool descriptions on `agent.progress`; add developer-owned labels there for the two vision tools (for example, “Looking at your camera…” and “Checking the recording…”). This produces useful tool-specific UI text today without adding a new cross-client status payload or treating LLM tool descriptions as UI copy.

**Blocker — make the public historical contract valid.** `HistoricalVisionRequest.reference_time_us` defaults to `0`, but the forwarded `HistoricalFrameRequest` requires it to be greater than zero. Make it required with `gt=0` and add tests for omitted/zero timestamps. The render worker already injects a valid utterance timestamp, so this does not change its behavior.

**Suggestion:** mention in the vision README that live-only `VisionToolsConfig` consumers may omit `video_memory`; the lazy composition is an important supported contract.

A richer structured tool-status envelope is worthwhile later, but it needs a deliberate backward-compatible SDK/client protocol change and should not expand this focused migration PR.

wenxind-nvidia:
Both blockers fixed in `75eed6f`; you're right on both, and thanks for the status-ownership correction.

**Blocker — per-turn status at the executor boundary.** Restored, and kept off the native functions. `RenderSceneProcessor._run_turn` now brackets the whole turn: `set_status("processing", pid)` on entry and `set_status("idle", pid)` in a `finally`, so it always clears — including the failure and barge-in-cancellation paths (I refactored the turn body into `_run_turn_body` so the bracket wraps every yield). `_agent.status` stays owned by the render turn; `look_at_current_frame`/`look_at_past_frame` remain reusable and UI-free. For per-tool UI text I added developer-owned `agent.progress` labels in `_TOOL_PROGRESS` ("Looking at your camera…", "Checking the recording…") rather than treating LLM tool descriptions as UI copy. Tests assert `processing→idle` on a normal turn and that `idle` still fires on failure and on barge-in cancellation.

**Blocker — valid historical contract.** `HistoricalVisionRequest.reference_time_us` is now required `gt=0` (was `default=0, ge=0`), matching the forwarded `HistoricalFrameRequest`. Tests cover both omitted and zero. The render worker injects a real utterance timestamp, so its behavior is unchanged.

**Suggestion.** README now notes the lazy composition: a live-only `VisionToolsConfig` consumer may omit `video_memory` (and need not register that group) as long as it never calls `look_at_past_frame`.

I agree a richer structured status envelope is a separate, protocol-level change and out of scope here.

Verification: `ruff@0.15.16 check .` clean; `pytest test_vision_functions.py test_vlm_mcp.py test_xr_render_demo_wire.py test_simple_vlm_example_wire.py` → **56 passed** (warning-free aside from the third-party pipecat `audioop` notice).

nvddr (changes requested):
Full updated review of `75eed6f`:

The native vision surface, lazy recorded-frame composition, trimmed model schema, and public docs are now in good shape. The focused vision/VLM/render/simple-VLM suite passes (56 tests).

**Blockers**

1. **Close the initial-status cancellation window.** `_run_turn` awaits `set_status("processing")` before entering its `try/finally`. If cancellation lands while that first publish is in progress, the `finally` is never entered and the client can remain stuck in `processing`. I reproduced that path with an endpoint that accepts the processing publish and is then cancelled: it receives only `processing`, never `idle`. Put the initial publish inside the protected region and add a regression test that cancels during that await.

2. **Keep this migration focused.** Remove the new `look_at_current_frame` / `look_at_past_frame` entries from `_TOOL_PROGRESS` (and the associated PR-description wording). Tool-specific UI progress is a useful separate design, but it is not required to preserve the per-turn lifecycle status restored here. It should not be bundled into the native-vision migration.

**Nit**

- The PR description currently contains two overlapping “Per-turn client status” bullets; consolidate them after the code change.

wenxind-nvidia:
Both blockers and the nit are addressed in `601afcb`.

**Blocker — initial-status cancellation window.** Good catch. The `set_status("processing")` publish now sits *inside* `_run_turn`'s `try`, so a cancellation that lands while that first publish is in flight still runs the `finally` and clears the client to `idle` — it can no longer get stuck in `processing`. Added a regression test that starts a turn, blocks inside the `processing` publish, cancels the turn mid-await, and asserts `idle` still fires.

**Blocker — keep the migration focused.** Removed the `look_at_current_frame` / `look_at_past_frame` entries from `_TOOL_PROGRESS` (and the PR-description wording). Restoring the per-turn lifecycle status doesn't need per-tool progress text, so that stays out of this change; the PR is now just the status bracket plus the historical-request contract fix.

**Nit.** Consolidated the PR description to a single status bullet.

Verification: `ruff@0.15.16 check .` clean; `pytest test_vision_functions.py test_vlm_mcp.py test_xr_render_demo_wire.py test_simple_vlm_example_wire.py` → **57 passed** (only the third-party pipecat `audioop` warning).

nvddr (approved):

PR #308, opened 2026-07-28
feat(voice): introduce xr-ai-voice package alongside xr-ai-pipecat

wenxind-nvidia:
Introduce the new `xr-ai-voice` SDK package **alongside** the existing `xr-ai-pipecat`, made safe for the repository's one-hub/many-clients model. Pipecat is untouched, no sample migrates onto voice, and there are no consumers yet.

## What lands
- `agent-sdk/xr-ai-voice/` (14 modules + README + pyproject): `VoiceSession`, the `VoiceHandler`/`VoiceQuery`/`VoiceResponse`/`VoiceTurn` handler surface, `HubVoiceTransport`, `TextMessageInput`, `VadConfig`, and the `_processors/{vad_stt,voice_gate,streaming_tts,handler}` + `_readiness`/`_pipeline`/`_audio`/`_frames`/`_session`/`_transport` internals. Public `__all__`
- Registered in `tests/pyproject.toml`, `ruff.toml` (per-file ignores only), `AGENTS.md`, `DEPENDENCIES.md`, `docs/changelog.md`. Ported `test_voice_{audio,pipeline,session}.py`.

## Multi-participant safety
- **Per-participant TTS state.** `StreamingTtsProcessor` keys pending text, the synth/order queue, the sender task, interruption, and hub flush by pid, so concurrent participants never share a buffer (which spliced their words) or a sender (which misrouted audio). An `InterruptionFrame` scopes to its `transport_source` pid — draining/flushing only that participant; the handler's supersede interrupt now carries that pid.
- **Lifecycle cleanup.** A departing participant's transport `MediaSender` is released on `ParticipantLeftFrame`, and every TTS sender task is torn down on pipeline `EndFrame`/`CancelFrame` — no state retained across join/leave churn or shutdown.
- **Supersede semantics.** `on_query_superseded` fires only when a new query actually replaces a still-in-flight turn — not for a queued follow-up or a query after the previous turn finished.
- Tests: interleaved two-participant TTS stays separate; an interrupt is participant-scoped (only that pid flushed, the other's audio survives); a departing participant's `MediaSender` is released; `on_query_superseded` does not fire on serial/queued queries.

## Hub-client imports
On main the hub client is `xr_ai_agent`; every symbol used (`AudioChunk`, `DataMessage`, `ParticipantEvent`, `ProcessorEndpoint`, `Subscribe`) is imported from it.

## Prerequisite: additive `xr-ai-voicegate` helpers (backward-compatible for pipecat)
The voice `VoiceGateProcessor` needs wake-gate helpers not yet on main — `begin_utterance`, `wake_ack_enabled`, `matches_magic_phrase`, `could_match_magic_phrase`, and `play_chime` returning a `bool`. `utils/xr-ai-voicegate/{gate,config}.py` are in this PR (bundled with their only consumer, per the "add surface with its consumer" principle). **Verified backward-compatible for the existing pipecat pipeline:**
- The new `feed()` clause `or pid in self._followup_started` is **inert** unless a consumer calls the new `begin_utterance()` — which only `xr-ai-voice` does; pipecat never does, so `_followup_started` stays empty and `feed()` behavior is unchanged.
- `play_chime`'s new bool return is ignored by pipecat's caller.
- Regression: `test_xr_ai_voicegate` + `test_xr_ai_pipecat_voice_pipeline` + `test_xr_ai_vad` → **126 passed**.
- `config.py` is a docstring clarification only. No main-only voicegate fix is reversed.

## Scope note
 `test_voice_session::test_simple_vlm_config_loads_prompt_without_media_imports` — was omitted: it subprocess-imports the *migrated* simple-vlm worker package, which is sample-migration work not in this PR. It tests the sample, not the voice package.

## Verification (before push)
- Rebased onto current `main` (changelog conflict resolved, both entries preserved).
- `ruff@0.15.16 check .` clean.
- `pytest` → **84 passed** (voice package) + **126 passed** (pipecat/voicegate/vad regression), including the new multi-participant tests. No pipecat source file modified.


wenxind-nvidia:
**github-code-quality — three items on the xr-ai-voice package, none actionable:**
- `handler.py:33` / `streaming_tts.py:47` unused `HubVoiceTransport` import — false positives: both are under `if TYPE_CHECKING:` and used in the string annotation `transport: "HubVoiceTransport | None"`. Removing them breaks the forward-reference type.
- `vad_stt.py:399` empty `except` — the doubled `try/await task/except CancelledError` probe teardown is verbatim from the source this package was ported from; I'm leaving it unchanged to keep the port aligned rather than trimming one arm of it.

No edits.

nvddr (changes requested):
Full review of `9c20114`:

The public `VoiceSession` direction is sensible, and the package cleanly keeps Pipecat behind a private implementation boundary. However, it is not safe for the repository’s one-hub/many-clients model yet. The existing tests and CI pass, but they do not exercise simultaneous participants.

**Blockers**

1. **Rebase onto current `main`.** GitHub currently marks this PR `DIRTY`; the outstanding conflict is in `docs/changelog.md`. Resolve it before the behavioral fixes so the final diff and CI result are reviewable.

2. **Make TTS state participant-scoped.** `StreamingTtsProcessor` has one global `_pending` text buffer, sender queue, and sender task. Concurrent token streams can be combined and routed to the wrong person: I reproduced Alice emitting `"Alice"` and Bob emitting `"Bob."`; the processor dispatched `"AliceBob."` to Bob. Its interrupt path also drains the global queue and flushes the single fallback `target_participant`, so Alice stopping can cancel or flush Bob’s audio. Model pending text, synthesis/order state, interruption, and hub flush by pid (or explicitly serialize all clients, which would conflict with the hub contract). Add an interleaved two-participant regression test.

3. **Finish participant and pipeline lifecycle cleanup.** Per-pid `MediaSender`s are created on join but never released on leave, and the TTS sender task has no normal EndFrame/CancelFrame cleanup path. A long-lived hub with join/leave churn will retain sender tasks/buffers; a normal pipeline shutdown can retain the TTS task. Tear down the departing pid’s state and cancel/await the TTS background work on pipeline shutdown.

4. **Make queueing semantics match the public callback.** With `queue_queries=True`, a new query is queued but `on_query_superseded` still fires; `_seen_query` also makes later queries look “superseding” after the active task has finished. A consumer of that callback can reasonably cancel UI or audio state even though no turn was superseded. Fire it only for an actual replacement, or expose a separately named queued/follow-up callback.

**Suggestions**

- Add a compact `VoiceSession` usage example and define the Unix/PTS semantics of the public `timestamp_us` fields.
- Remove the duplicated cancelled-task await in `VadSttProcessor._cancel_probe_task`; mirroring it from the modular branch is not a reason to retain redundant teardown code.
- Trim the new historical/debugging commentary to the non-obvious invariants; several modules are materially harder to read than the implementation requires.

wenxind-nvidia:
All four blockers and the three suggestions are addressed in `e61d057` (on top of the rebase). Thanks — the concurrency review was right; the package wasn't safe for the one-hub/many-clients model.

**Blocker 1 — rebase.** Rebased onto current `main`; the `docs/changelog.md` conflict is resolved keeping every merged entry plus this one. CI now reflects the integrated state.

**Blocker 2 — participant-scoped TTS.** `StreamingTtsProcessor` now holds a `_TtsPidState` per participant — pending text, the synth/order queue, the sender task, interruption, and hub flush are all keyed by pid. Two participants can no longer share a pending buffer (which produced `"AliceBob."`) or a sender. An `InterruptionFrame` scopes to its `transport_source` pid and drains/flushes only that participant; a pid-less frame falls back to draining all. The handler's supersede `InterruptionFrame` now carries `transport_source=pid` (it previously went out unlabeled). New test drives interleaved Alice/Bob token streams and asserts each is synthesized separately and addressed only to its own participant; another asserts an Alice interrupt flushes only Alice and leaves Bob's audio intact.

**Blocker 3 — lifecycle cleanup.** A departing participant's `MediaSender` is released on `ParticipantLeftFrame` (cancelled + removed), and every TTS sender task is torn down on pipeline `EndFrame`/`CancelFrame`. Tests cover the sender release and the post-`EndFrame` teardown.

**Blocker 4 — supersede semantics.** `on_query_superseded` now fires only when a new query actually replaces a still-in-flight turn. A follow-up that is queued (`queue_queries`) or that arrives after the previous turn finished no longer fires it. The rewritten tests assert no supersede on serial/queued queries; the existing in-flight-replacement tests still assert it fires (and interrupts) there.

**Suggestions.** Dropped the duplicated cancelled-task await in `VadSttProcessor._cancel_probe_task`; documented the Unix-epoch-µs semantics of `VoiceQuery.timestamp_us` / `VoiceTurn.timestamp_us` and added a compact `VoiceSession` usage example to the README; the `StreamingTtsProcessor` rewrite trims the verbose interrupt commentary to the load-bearing invariants.

Verification: `ruff@0.15.16 check .` clean; **84 passed** (voice package, incl. the new multi-participant tests) + **126 passed** (pipecat/voicegate/vad regression).

wenxind-nvidia:
Two correctness gaps fixed in `8d99a6a` (raised on the stacked PR, but they belong here).

**Speech-onset timestamps.** Confirmed: the input transport dropped the hub's `AudioChunk.pts_us`, and the gate stamped wall clock *after* STT — so every turn was timestamped with VAD hangover plus transcription latency baked in, and that error would persist into stored transcripts and time-relative recorded-frame lookups. Fixed along the whole path:
- The transport forwards the hub capture time on pipecat's standard `Frame.pts` (nanoseconds) rather than discarding it.
- `VadSttProcessor` records the pts of the audio at speech onset and stamps it onto the resulting transcript — on both the final-utterance and the early-probe paths.
- `VoiceGateProcessor` uses that value for the dispatched query's `pts_us`, falling back to wall clock only for a transcript with no originating audio frame (e.g. an injected one).
- Per-pid onset state is evicted on participant leave.

**Pipeline shutdown.** Also confirmed — both processors fell through on `EndFrame`/`CancelFrame`. `_VoiceHandlerProcessor` now cancels **and awaits** every in-flight turn: awaiting matters because a cancelled turn still runs its `finally` (turn observer, end frame), so returning early is exactly how a transcript write escapes the session. `VadSttProcessor` tears down all pending early-STT probe tasks so a probe can't push a transcript after shutdown.

Tests: transcript carries the onset pts; the query carries it as `pts_us`, with the wall-clock fallback covered; the in-flight turn and the pending probe are both torn down by the pipeline `EndFrame`.

Verification: `ruff@0.15.16 check .` clean; voice + voicegate + pipecat regression suites → **209 passed**.

nvddr (changes requested):
Updated review of 8d99a6a:\n\nThe rebase, per-participant TTS routing, scoped normal interruption, queueing semantics, timestamp propagation, and EndFrame/CancelFrame shutdown work are now in good shape. CI is green, and I reran the focused multi-client/timestamp/shutdown regressions (5 passed).\n\n**Blockers**\n\n1.  still has no cleanup path in . The VAD, gate, handler, and output transport clean up, but the TTS processor retains that pid's pending text/sender task. If stale synthesis later emits audio, the output transport's lazy routing can recreate the just-released sender for the departed pid. Handle leave in TTS by cancelling and removing that pid's state (without affecting other clients), and add a regression covering both teardown and no later output/sender recreation.\n\n2. The pid-less interruption fallback drains every TTS state but flushes hub audio only for . Audio already paced for other active participants can continue after the global interruption. Capture the active pids and flush each of them; add a two-participant regression.\n\n**Suggestion**\n\nAlign 's docstring with the implementation and test: cancelled turns intentionally do not emit the marker, so the current claim that every cancelled turn produces one is incorrect.

nvddr (commented):
Correction to my just-submitted review: shell quoting stripped several identifiers. The intended disposition is unchanged.

The rebase, per-participant TTS routing, scoped normal interruption, queueing semantics, timestamp propagation, and EndFrame/CancelFrame shutdown work are now in good shape. CI is green, and I reran the focused multi-client, timestamp, and shutdown regressions: 5 passed.

Blockers:

1. ParticipantLeftFrame still has no cleanup path in StreamingTtsProcessor. The VAD, gate, handler, and output transport clean up, but TTS retains that pid's pending text and sender task. If stale synthesis later emits audio, the output transport's lazy routing can recreate the just-released sender for the departed pid. Handle leave in TTS by cancelling and removing that pid's state without affecting other clients, and add a regression covering teardown and no later output or sender recreation.

2. The pid-less interruption fallback drains every TTS state but flushes hub audio only for target_participant. Audio already paced for other active participants can continue after the global interruption. Capture the active pids and flush each of them; add a two-participant regression.

Suggestion: align AssistantResponseEndFrame's docstring with the implementation and test. Cancelled turns intentionally do not emit the marker, so the current claim that every cancelled turn produces one is incorrect.

wenxind-nvidia:
Both blockers and the suggestion are fixed in `6b61c75`. Thanks for catching the leave path — the interaction with the output transport's lazy routing is a good call.

**Blocker 1 — `ParticipantLeftFrame` had no cleanup path in `StreamingTtsProcessor`.** Confirmed: every other processor cleaned up, but TTS retained the departed pid's pending text and sender task. TTS now releases that pid's state (and only that pid's) when the frame passes through. The ordering works in our favour — TTS sits upstream of the output transport, so the synthesis teardown lands *before* the transport drops the media sender, which closes exactly the window you describe where a lingering synth task emits audio afterwards and the transport's lazy routing recreates the sender it just released. Leaving is not an interruption, so no hub flush is issued for a participant who is already gone.

Regression: after a leave, the pid's state is gone, **no audio is emitted for that pid afterwards** (so nothing can trigger sender recreation), no flush is issued, the other participant's stream is untouched, and the frame still reaches the rest of the pipeline.

**Blocker 2 — pid-less interruption flushed only the fallback target.** Confirmed and fixed: the fallback drains every participant, so it now flushes every active participant's hub audio too, plus the fallback target (audio pushed with no destination is paced under that pid and has no `_by_pid` entry of its own). Regression asserts both participants are flushed.

**Suggestion — `AssistantResponseEndFrame` docstring.** You're right, it was factually wrong: `_run_query` guards on `if not cancelled and is_current`, so a cancelled turn deliberately emits no marker. The docstring now says so, with the reason — a cancelled turn's response is partial, and echoing it as the turn's final data would publish text the agent was interrupted out of saying.

Verification: `ruff@0.15.16 check .` clean; voice + voicegate + pipecat regression suites → **211 passed**. The stacked PR has been rebased onto this tip.

nvddr (approved):

PR #309, opened 2026-07-28
feat(nat): voice adapters and conversation recall

wenxind-nvidia:
Add the voice adapters that bridge native NAT functions to a voice session, and re-land the `xr_conversation_memory` capability **with the producer that makes it work**.

## What lands
- **`xr_ai_nat/adapters/voice.py`** — `as_voice_handler(function, *, request, response, streaming=False)` maps a `VoiceQuery` onto one native function's request model and its result back to text (with `streaming=True` forwarding `astream` chunk by chunk); `record_voice_transcripts(add_transcript)` is a turn observer that persists completed turns under `{participant_id}:user` / `{participant_id}:agent`.
- **Both adapters are exported from `xr_ai_nat.adapters`** — the documented import path. The re-export is lazy (PEP 562 `__getattr__`) because `adapters/voice` imports `xr-ai-voice`: importing the package without the optional extra still succeeds and stays cheap, and only attribute access raises, naming the extra to install. It also keeps the deprecated `adapters.mcp` alias from warning on an unrelated import.
- **`xr-ai-nat[voice]` optional extra** — the adapters need `xr-ai-voice`, so they are gated behind an extra rather than a hard dependency.
- **`xr_conversation_memory` re-landed with its producer.** This group was deliberately deferred out of the typed text-memory change: `recall_conversation` reads `{pid}:{role}` sources that nothing produced, so wiring it then would have returned empty history. `record_voice_transcripts` is that producer, so the group returns here with an end-to-end record→recall test.

## Agent-facing contract
The recall surface is consumed by agents, so it is fully described: every `ConversationEntry` field and `RecallConversationResult.entries` carry descriptions, `role` is constrained to `Literal["user", "agent"]` (the only values the producer writes), and the config's `text_memory` reference documents which group recall reads from. A regression asserts the generated request/result schemas keep their descriptions and the `role` enum.

Ordering is part of the contract: a user turn and the agent turn answering it share one timestamp (both carry the originating query's time), and recall orders that tie user-before-agent. The record→recall test uses one shared timestamp per exchange to pin it.

## Do-not-reverse
- The four existing text-memory functions keep their detailed descriptions unchanged.
- `conversation_memory_functions` registers through the existing `xr_ai_nat_text_memory` plugin entry point — no new entry point, no dead surface.
- Nothing moved or was deleted, so no deprecation alias is needed.

## Documentation
`xr-ai-nat/README.md` gains a **Conversation recall** section (the group, its composition of `text_memory`, the tie ordering, and that recall is empty without a producer) and a **Voice adapters** section (the `[voice]` extra, both adapters, and that recording is an observer rather than a side effect of invoking a function). `DEPENDENCIES.md` names the conversation group and both adapters.

## Verification
- `ruff@0.15.16 check .` clean; `git diff --check` clean.
- Full CPU suite: **631 passed / 2 skipped / 10 deselected**.
- `uv lock --project` resolves for all 34 projects.


wenxind-nvidia:
**Resolved in `da3ab84`.**

Confirmed and fixed. #309 was stacked (merge `e376780`) on #306 at `df1bedc` — one commit *before* the alias completion `d7fed33` — so its `text_memory/schemas.py` still had `__all__ = ["TranscriptSegment"]` with no `TranscriptStats = TranscriptStatsResult` forwarding, and the alias test was rolled back to match. Landing after #306 would have re-dropped `TranscriptStats`, reintroducing the incomplete-deprecated-surface blocker.

Fix: merged #306's current tip (`d7fed33`) forward into the branch, continuing the existing merge-based stack. #309's own commits touch only `text_memory/__init__.py` / `functions.py` (conversation-recall producer), not `schemas.py` or the alias test, so the merge was conflict-free. The branch now:
- carries `TranscriptStats = TranscriptStatsResult` in `schemas.py` with `__all__ = ["TranscriptSegment", "TranscriptStats"]`,
- restores the alias test asserting `schemas.TranscriptStats is TranscriptStatsResult`,
- `git branch --contains d7fed33` now includes this branch.

Verification: `ruff@0.15.16` clean; `pytest test_adapters_voice.py test_text_memory_functions.py test_transcript_mcp.py test_voice_audio.py test_voice_pipeline.py test_voice_session.py` → 93 passed.

wenxind-nvidia:
**github-code-quality — four items:**
- `test_adapters_voice.py:40` unnecessary lambda — **valid, fixed in `a904e76`**: `response=lambda result: str(result)` → `response=str`.
- `handler.py:33` / `streaming_tts.py:47` unused `HubVoiceTransport` — false positives (both `TYPE_CHECKING`-only, used in the string annotation `transport: "HubVoiceTransport | None"`). Same disposition as #308.
- `vad_stt.py:399` empty `except` — verbatim from the ported source; left unchanged for alignment. Same disposition as #308.

wenxind-nvidia:
Both points are valid and addressed.

**Section H — the branch carried the unfixed voice package.** Correct — it can't be fixed inside this PR, so I re-stacked. #308's participant-scoping fix has landed on `nat-align/06-xr-ai-voice`, and this branch is rebuilt on top of it (new #308 base → merge current text-memory base → this PR's adapters + conversation-recall commit). `StreamingTtsProcessor` in this tree now holds per-participant `_TtsPidState` (no global `_pending`/`_sender_queue`/`_sender_task`), so the cross-talk, mis-scoped interrupt, and cleanup issues are gone here too. When #308 merges to `main`, this branch gets the usual one-step rebase onto `main`.

**Principle 8 — changelog / DEPENDENCIES referenced an external source.** Rewritten in terms of this repo: the `DEPENDENCIES.md` voice entry now reads "Readiness is health-based, split across the `_readiness` / `_session` modules," and the changelog entry drops the "mirrors the upstream …" / "brought up to the … revision" phrasings for repo-local wording. (These live in #308's introduce commit, so the fix rides in via the re-stack.)

Verification: `ruff@0.15.16 check .` clean; `pytest test_adapters_voice.py test_voice_pipeline.py test_voice_session.py test_voice_audio.py test_text_memory_functions.py test_transcript_mcp.py` → **96 passed**.

wenxind-nvidia:
All four blockers and the suggestion are addressed.

**Blocker 1 — restack.** Done. The stacked base's blockers were fixed in place (not duplicated here), and this branch is rebuilt on the current base plus this PR's own two commits. Since the typed text-memory work has landed on `main` and the base is rebased onto it, the ancestry is now a single clean line — no `docs/changelog.md` conflict. Now `MERGEABLE`.

**Blocker 2 — inherited correctness blockers.** Fixed in the base PR, as directed, so conversation recall now persists accurate speech timestamps and cannot receive a transcript write after shutdown. Summary: the hub capture time is carried on the frame's presentation timestamp and stamped at speech onset instead of wall-clock-after-STT; and `EndFrame`/`CancelFrame` now cancels *and awaits* in-flight handler turns and tears down pending early-STT probes. Details on that PR.

**Blocker 3 — the documented adapter API didn't exist.** Correct, and a real break: the package exported nothing while the docs pointed at `xr_ai_nat.adapters.as_voice_handler`. Both adapters are now re-exported from `xr_ai_nat.adapters`. The re-export is **lazy** (PEP 562 `__getattr__`) because `adapters/voice` imports `xr-ai-voice`, which is only present with the optional `[voice]` extra — so importing `xr_ai_nat.adapters` without the extra still succeeds and stays cheap, and only attribute access raises, with an error naming the extra. It also keeps the deprecated `adapters.mcp` alias from emitting its warning on an unrelated import of the package. Tests now import through the public namespace and assert `__all__` / `dir()` plus the `AttributeError` for unknown names.

**Blocker 4 — package README.** Added, per the same-task documentation rule. `xr-ai-nat/README.md` gains a **Conversation recall** section (`xr_conversation_memory`, its composition of `text_memory`, the tie ordering, and that recall is empty without a producer) and a **Voice adapters** section (the `[voice]` extra, both adapters, and that recording is an observer rather than a side effect of invoking a function). `DEPENDENCIES.md`'s descriptive paragraph now names the conversation group and both adapters.

**Suggestion — shared timestamp in the record→recall test.** Taken. The user and agent turns of an exchange now share one timestamp, matching how a session stamps both from the originating query, and the test pins user-before-agent ordering for that tie. It imports the adapters from `xr_ai_nat.adapters`.

Verification: `ruff@0.15.16 check .` clean; adapters + voice + text-memory + transcript-MCP suites → **104 passed**.

wenxind-nvidia:
**Rebased onto `main`.** With `#308` merged (squashed as `46121f5`), this branch no longer needs to stack — it is now `main` plus only its own three commits (voice adapters + conversation recall, the lambda fix, and the adapter-export/documentation fix). The four `#308` commits it was carrying are dropped, since that content is in `main` now; all three of this PR's commits cherry-picked cleanly with no conflicts.

The diff is now just this PR's work: `adapters/voice.py`, the lazy `xr_ai_nat.adapters` export, `xr_conversation_memory`, the `[voice]` extra, the README/DEPENDENCIES sections, and `tests/test_adapters_voice.py`.

Verification on the rebased state: `ruff@0.15.16 check .` clean, `git diff --check` clean, `uv lock --project` succeeds for all 34 projects, full CPU suite **629 passed / 2 skipped / 10 deselected**.

nvddr (changes requested):
Current review of ca82289 after the rebase onto main (46121f5).

Blocker:
- Complete the public recall_conversation result contract. The request fields are documented, but ConversationEntry.timestamp_us, role, and text plus RecallConversationResult.entries expose no field descriptions; role is also an unconstrained string although the only valid values are user and agent. These schemas are part of the native function surface that agents consume. Add precise descriptions, constrain role to the two supported values, and add a regression covering the generated contract. Please also describe the new text_memory function-group reference in the conversation-memory config.

Suggestion:
- Refresh the PR title and body. They still say this is stacked on two open PRs and instruct a merge order, but #308 is now merged and this branch is main plus its three own commits.

Nit:
- The one remaining github-code-quality thread is test-import style only. Please clean it up and resolve the thread, but it is not a design concern.

The rebase itself is correct: the diff is now only the nine intended files, with no inherited #308 code. Focused adapter, text-memory, and MCP-compatibility validation passed locally (14 tests); all CI checks are green.

wenxind-nvidia:
All three items addressed in `84c54c4`.

**Blocker — complete the `recall_conversation` result contract.** Valid: the request models were described but the result surface was bare. Fixed:
- Every `ConversationEntry` field (`timestamp_us`, `role`, `text`) and `RecallConversationResult.entries` now carry descriptions.
- `role` is constrained to `Literal["user", "agent"]`. That is provably the full domain — `recall` iterates exactly `("user", "agent")` to build the role-scoped source ids, so an open `str` only invited the model to invent a third value.
- The `entries` description states the ordering guarantee, including that a user turn and the agent turn answering it share a timestamp and the user turn is ordered first.
- `ConversationMemoryFunctionsConfig.text_memory` is now described: which group recall reads from, and that it reads the role-scoped `{participant_id}:user` / `{participant_id}:agent` sources rather than touching storage directly.
- Regression `test_recall_conversation_generated_contract_is_fully_described` inspects the *generated* schemas: every request field described, the result and nested entry model described, and `role`'s enum exactly `["user", "agent"]`. A second test pins the config field's description.

README and changelog updated to match.

**Suggestion — refresh the title and body.** Done. The title no longer claims a stack, and the body is rewritten for the current state: this branch is `main` plus its three own commits, with no merge-order instructions. It now leads with what the change actually is and documents the agent-facing contract above.

**Nit — code-quality thread.** Fixed: the namespace test imported `xr_ai_nat.adapters` with both `import` and `from … import`; it now uses a single `from xr_ai_nat import adapters`.

Verification: `ruff@0.15.16 check .` clean; `git diff --check` clean; full CPU suite **631 passed / 2 skipped / 10 deselected**.

nvddr (approved):
Follow-up review of 84c54c4: approved.

The public recall_conversation contract is now complete: result and nested entry fields are described, role is constrained to user or agent, and the generated-schema regression verifies those guarantees. The text-memory reference is documented, the PR metadata now reflects the unstacked state, and the remaining code-quality thread is resolved.

The branch is current with main. I re-reviewed the full nine-file diff; focused adapter, text-memory, and MCP-compatibility validation passed locally (16 tests), and all CI checks are green. No remaining blocker.

PR #311, opened 2026-07-29
Apply low-risk fixes from WSL2 setup report

blongs-nv:
Included:
- Require HF_TOKEN by default
- STT: return 5xx on backend failure
- Fix vLLM container log-streamer race
- Document token_server_port
- Document the two HF cache roots
- Document WSL2 support status

### DIFF
--- a/DEPENDENCIES.md
+++ b/DEPENDENCIES.md
@@ -353,7 +353,10 @@ xr-ai-tests  (tests/)
     tests, the vlm-mcp adapter, and the sample-local scene native groups plus
     render-mcp adapter surface (LOVR is stubbed). oxr-mcp is not
     included: it needs native isaacteleop + a CloudXR runtime, so its
-    smoke test self-skips on CPU (see tests/README.md).
+    smoke test self-skips on CPU (see tests/README.md). Root pytest adds
+    ai-services/stt-server to its Python path (not a dependency) so the
+    endpoint tests can import its FastAPI app with a mocked backend,
+    avoiding a test-time NeMo installation.
 
     Tests marked `@pytest.mark.gpu` are the local-only set (skipped by
     `-m "not gpu"` in CI). They spawn real ai-services via `uv run` (e.g.
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ endpoint and no local GPU is required for the agent or hub.
 
 | Requirement | Version | Notes |
 |---|---|---|
-| OS | Linux | Ubuntu 22.04 / 24.04 recommended |
+| OS | Linux | Ubuntu 22.04 / 24.04 recommended; WSL2 is not officially supported (see **Windows (WSL2)** below) |
 | Python | 3.11 or 3.12 | 3.10 and 3.13 are not supported |
 | [uv](https://docs.astral.sh/uv/) | latest | dependency manager used by all samples |
 | NVIDIA driver | 570+ | required for local model inference |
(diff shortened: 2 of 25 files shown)


nvddr (changes requested):
Full review of ad367f6:

The credential flow, Docker log-supervisor tests, documentation updates, and focused test suite look reasonable. I ran the launcher, STT endpoint, and vLLM tests: 63 passed.

Blockers:

1. Rebase onto current main and resolve the outstanding conflicts before further review. The branch is five commits behind, GitHub reports DIRTY and CONFLICTING, and the full CI suite has not run on an integrated head.

2. Do not return str(exc) from the STT endpoint as the HTTP 500 detail. This service binds broadly and backend exceptions can expose internal paths, runtime state, or provider details. Log the exception server-side, return a stable generic 5xx detail to callers, and update the endpoint test so it verifies the status and generic contract rather than the CUDA error text.

The existing code-quality finding around an empty exception branch is non-blocking; the shutdown path is intentionally best-effort.

blongs-nv:
> Full review of [ad367f6](https://github.com/NVIDIA/xr-ai/commit/ad367f6476d7a00ae1372c4f19386ae618a6f8e2):
> 
> The credential flow, Docker log-supervisor tests, documentation updates, and focused test suite look reasonable. I ran the launcher, STT endpoint, and vLLM tests: 63 passed.
> 
> Blockers:
> 
> 1. Rebase onto current main and resolve the outstanding conflicts before further review. The branch is five commits behind, GitHub reports DIRTY and CONFLICTING, and the full CI suite has not run on an integrated head.
> 2. Do not return str(exc) from the STT endpoint as the HTTP 500 detail. This service binds broadly and backend exceptions can expose internal paths, runtime state, or provider details. Log the exception server-side, return a stable generic 5xx detail to callers, and update the endpoint test so it verifies the status and generic contract rather than the CUDA error text.
> 
> The existing code-quality finding around an empty exception branch is non-blocking; the shutdown path is intentionally best-effort.

Updated

nvddr:
## AI review

Reviewed current head `2ee4a5a` against current `main` (`8fe916a`).

### Findings

1. **[P2] The Docker log reattach cursor can lose container output.**  
   In `utils/xr-ai-vllm/xr_ai_vllm/_docker.py`, `_since` is captured only after `proc.wait()` returns. Logs emitted after the old `docker logs -f` follower disconnects but before that timestamp is recorded are earlier than the next `--since` value and will be omitted permanently. Please use a conservative/overlapping cursor (duplicates are preferable to losing the startup diagnostics this supervisor is intended to preserve).

2. **[P2] The documented HuggingFace cache roots are not correct for all bundled configurations.**  
   `docs/ai-services.md` and its source mirror hard-code `models/hub` and `models/huggingface/hub`, but the physical paths depend on the YAML's resolved `model_cache`. The sample profiles resolve to repository-root `models`, while the standalone service YAMLs resolve to `ai-services/models`. As written, the manual `hf download` examples can download tens of GB into a cache the selected service will not use. Please document these as `<resolved model_cache>/hub` and `<resolved model_cache>/huggingface/hub`, with examples for both layouts.

3. **[P2] The WSL2 token-port workaround points to a config that the affected sample does not use.**  
   `simple_vlm_example` launches the hub with `agent-samples/simple-vlm-example/yaml/xr_media_hub.yaml`; editing `server-runtime/xr_media_hub.yaml` does not change that run. Please add `token_server_port` to the active sample configuration or explicitly instruct users to add it there.

4. **[P3] The changelog should state the generic STT error contract.**  
   `docs/changelog.md` says backend exceptions propagate with “the error detail,” while the endpoint now deliberately logs the exception server-side and returns only the stable generic `"transcription failed"` detail. Please clarify this so the changelog does not contradict the security fix.

### Validation

- The earlier requested fixes are otherwise present: exception text is no longer exposed by the STT endpoint, and the prior conflict is resolved.
- A synthetic merge with current `main` is conflict-free and the changed Python files compile.
- GitHub reports the PR workflows green (tests, lint, docs, CodeQL, DCO, SPDX, and lock-check).
- Focused local pytest execution was not available on this macOS host because the repository test environment requires a platform-specific `PyNvVideoCodec` wheel; the green Linux workflow provides execution coverage.

Requesting changes for the three P2 findings above.

nvddr:
## AI re-review

All blocking findings from the previous AI review are resolved at head `855255b`. One non-blocking documentation issue remains:

- **[P3] Align the cache-root introduction with the revised section.** `docs/ai-services.md` still says all model weights land in repository-root `models/`, immediately before explaining that standalone service configurations resolve `model_cache` to `ai-services/models/`. Please correct that introductory sentence and the mirrored `docs/source/components/ai-services.md` text for consistency.

The branch is current with `main`, mergeable, and all GitHub workflows are green. This note does not block approval.

nvddr (approved):
AI re-review complete at `855255b`. All blocking findings are resolved; approving. The remaining cache-root wording note is non-blocking and is recorded in the top-level AI review comment.

PR #312, opened 2026-07-29
refactor(models): privatize xr-ai-models implementation modules

wenxind-nvidia:
Privatize the `xr-ai-models` implementation modules; the public API is unchanged. Standalone (no dependency on other open NAT-alignment PRs).

## What changes
- Rename the four implementation modules to underscore-private names:
  `config.py→_config.py`, `factory.py→_factory.py`, `openai_compat.py→_openai_compat.py`, `protocols.py→_protocols.py`.
- `__init__.py` now imports from the private modules; **the public API is unchanged** — every name previously re-exported from `xr_ai_models` is still importable from the package root (`from xr_ai_models import VLMService, load_models_config, make_vlm, …`).
- In-tree consumers move to the canonical root import (`vlm-mcp`, `xr-ai-pipecat/services.py`, tests). `vlm-mcp` drops the redundant explicit `kind=KIND_OPENAI_COMPAT` (already the `VLMSpec` default); the `_is_cleartext_key_transport` test reaches into `_openai_compat` directly.

## Do-not-reverse / lessons applied
- **Deprecated forwarding aliases** for all four old public module paths: each emits a `DeprecationWarning` and re-exports the canonical objects, so external `from xr_ai_models.config import …` / `.protocols` / `.openai_compat` / `.factory` keep working. The config names not exposed at the package root — `KIND_OPENAI_COMPAT`, `ModelKind`, `Category`, `Spec` — stay importable from `xr_ai_models.config` (a complete forwarding alias).
- **Behavior unchanged** — pure module moves; the `kind: openai_compat` config *value* is untouched.
- **Scope-limited to privatization** — the new `_config` spec models (`AdapterSpec`/`DeploymentSpec`/`EndpointSpec`) are a separate config refactor and are **not** included here.
- `ruff.toml`'s grandfather snapshot entries follow the rename (module contents unchanged); `DEPENDENCIES.md` + README reference `_factory.py::make_*`.

## Verification
- `ruff@0.15.16` clean.
- `pytest test_models_deprecated_aliases.py test_models_config.py test_models_openai_compat.py test_models_protocols.py test_nat_models_llm.py test_vlm_mcp.py` → **73 passed**, warning-free.
- Consumers: `test_xr_render_demo_wire.py test_pipecat_services_wire.py test_simple_vlm_example_wire.py` → **40 passed**.
- Smoke: root `import xr_ai_models` is clean under `-W error::DeprecationWarning`; each old module path warns and forwards.

[...]

PR #313, opened 2026-07-29
refactor(hub): rename xr_ai_agent -> xr_ai_hub (xr-ai-hub-client)

wenxind-nvidia:
Rename the agent-side hub SDK toward the modular architecture. This remains standalone and does not depend on other open NAT-alignment PRs.

## What changes

- **Module:** `xr_ai_agent` → `xr_ai_hub`.
- **Distribution:** `xr-ai-agent` at the `agent-sdk/` root → `xr-ai-hub-client` in `agent-sdk/xr-ai-hub-client/`; `agent-sdk/` becomes a plain package container.
- All in-tree consumers import `xr_ai_hub` and depend on `xr-ai-hub-client`; dependency declarations and editable paths are updated across 10 `pyproject.toml` files.
- The test and nightly workflow cache globs track `agent-sdk/xr-ai-hub-client/pyproject.toml` instead of the deleted root manifest.

## Compatibility and scope

- **The distribution rename is breaking:** dependency declarations must use `xr-ai-hub-client`. No compatibility distribution is shipped because the package is unpublished and all in-tree editable consumers migrate in this PR.
- **The import remains temporarily compatible:** `xr-ai-hub-client` ships a deprecated `xr_ai_agent` forwarding package that warns once and re-exports the canonical `xr_ai_hub` objects.
- `_live_frames.py` keeps current `main` behavior because `video-mcp` still calls `participants()` in two places.
- `_processor.py` and `_types.py` are unchanged beyond the rename. The only implementation adjustments are the focused type-narrowing changes in `_codec.py` and `_shm.py`.

## Verification

- Rebased onto current `main`, including #316.
- `ruff@0.15.16 check .` passes.
- Both edited workflow files parse as YAML; `git diff --check` passes.
- Full CPU suite: **649 passed, 2 skipped, 10 GPU tests deselected**.
- Alias regression tests verify warning behavior, matching `__all__`, and object identity for every public export.


wenxind-nvidia:
github-code-quality (unused `AudioChunk` import in `test_multi_agent.py` / `test_return_routing.py`) — **valid, fixed** in `1c5a485`: both only use `DataMessage`, so `AudioChunk` is dropped from the import. Also burned down the now-unneeded `F401` grandfather entries for both files (`I001` retained). ruff clean; the two suites pass.

nvddr (changes requested):
Full review of 1c5a485:

The hub-client rename direction is aligned with the modular architecture. The canonical xr_ai_hub import and deprecated top-level xr_ai_agent alias behave correctly in the focused checks. I ran the codec, routing, participant, vision, video-memory, and render-wire tests: 113 passed, plus both import-compatibility smoke checks.

Blocker:

Rebase onto current main before this can be reviewed for merge. GitHub reports DIRTY and CONFLICTING against the merged #307 native-vision work. This is not a mechanical conflict: main deletes the old vision _images module and changes vision functions and tests, while this branch renames imports in that former surface. Preserve the current native current/past-frame design, do not revive the deleted path-based code, then apply the hub import and dependency rename to the integrated state. Resolve the AGENTS, DEPENDENCIES, changelog, VLM MCP, render-worker, and test conflicts at the same time, and rerun the full CI suite.

No additional design blockers from this pass.

wenxind-nvidia:
All four blockers and the suggestion are addressed in `cf855d5` — with one item I could not push, flagged at the end.

**Blocker 1 — rebase without reversing #307.** Rebased onto current `main`. All three do-not-reverse traps handled exactly as described:
- `vision/_images.py` stays **deleted**; `_pixels.py` is carried forward and its hub import updated to `xr_ai_hub`. (That import was the live half of the trap — `_pixels.py` came from #307 and still referenced the old name.)
- `tests/test_vision_functions.py` keeps #307's `VisionToolsConfig` tests verbatim; only the hub import changed. The pre-#307 `VisionFunctionsConfig` file was not restored.
- `docs/changelog.md` keeps both entries.

Full matrix re-run locally since CI had only done Analyze/DCO: `ruff@0.15.16 check .` clean, `git diff --check` clean, `uv lock --project` succeeds for **all 33 projects**, and the full CPU suite is **540 passed / 2 skipped / 10 deselected**.

**Blocker 2 — documented scaffold no longer resolved.** Correct: the dependency name was updated but the path was not, and this PR removes `agent-sdk/pyproject.toml`, so a worker created from the template could not resolve. `docs/adding-a-sample.md` and the `docs/source` mirror now use `../../../agent-sdk/xr-ai-hub-client`, matching the shape the adjacent `xr-ai-models` line already had. `DEPENDENCIES.md` no longer attributes the hub-client dependency rule (or the API-change row) to `agent-sdk/`, which is now only a container.

**Blocker 3 — the rename is breaking at the distribution level.** Agreed — "public API unchanged" conflated the import API with the distribution API. The wheel's metadata name is only `xr-ai-hub-client`, so the bundled `xr_ai_agent` import package only helps *after* a consumer has switched its dependency declaration. Took the documentation route as suggested: the changelog now states the dependency rename as **breaking** and says why no compatibility distribution ships (unpublished package; every consumer is an in-tree editable path updated here), and `agent-sdk/xr-ai-hub-client/README.md` gains a migration section covering the dependency change, the path, and the temporary import alias.

**Blocker 4 — fork references removed.** The entry now argues from this repository's own state: the name described the agent rather than the hub client, and the package occupied the `agent-sdk/` root so the directory holding every SDK package was itself one of them. `LiveFrameSource` is justified by its real constraint — the video-mcp live-frame exporter calls `participants()` in two places, so slimming it would break a live consumer.

**Suggestion — automated alias regression.** Added `tests/test_hub_client_deprecated_alias.py`, mirroring the models-alias pattern: importing `xr_ai_hub` is warning-free (guards the transitive case), importing `xr_ai_agent` warns exactly once, `__all__` lists match, and — the load-bearing one — every name in `xr_ai_hub.__all__` resolves to the *identical object* on the alias, which is what catches a public name missing from `__all__` failing to forward through `import *`. 7 tests.

**One item I could not push:** the `tests.yml:43` and `nightly-xr-ai-test.yml:164` cache keys still list the removed `agent-sdk/pyproject.toml`. The one-line fix is `agent-sdk/xr-ai-hub-client/pyproject.toml` in both, but pushing `.github/workflows/**` needs workflow-scope permission my push was rejected for. Happy for someone with that scope to apply it, or I can hand over the exact diff.

wenxind-nvidia:
**Rebase note (update to the previous comment).** `#308` merged to `main` while I was pushing, so I rebased again onto the new tip. Three conflicts, all additive and resolved keeping both sides: `ruff.toml` (kept the new xr-ai-voice grandfather entries alongside the renamed hub-client ones), `AGENTS.md` (kept main's "Six packages" count with the renamed package), and `docs/changelog.md` (both entries preserved).

That surfaced one real item: the newly merged `xr-ai-voice` package imports the hub client under the **old** name, so it would have resolved only through the deprecated alias — a package added in the same release starting life on a compatibility shim. Carried the rename into it: four modules and two voice test files now import `xr_ai_hub`, and its dependency/uv source point at `xr-ai-hub-client` at the new path, with the DEPENDENCIES.md entry following.

Re-ran the full matrix on the final state: `ruff@0.15.16 check .` clean, `git diff --check` clean, `uv lock --project` succeeds for **all 34 projects**, full CPU suite **631 passed / 2 skipped / 10 deselected** (now including the voice suite). Now `MERGEABLE`.

The workflow cache-key one-liner is still outstanding — it needs workflow-scope push permission.

wenxind-nvidia:
**Rebased onto current `main`** (now including `#312`). Four conflicts, all the same additive shape — `main` added the `[voice]` extra / entries while this branch renames the hub dependency — resolved by keeping both sides:
- `agent-sdk/xr-ai-nat/pyproject.toml`: the renamed `xr-ai-hub-client` in `[vision]` **and** main's `voice = ["xr-ai-voice"]` extra, plus both `uv.sources` entries.
- `DEPENDENCIES.md`: same, in the xr-ai-nat block.
- `docs/changelog.md` (twice, once per commit that touches it): every entry preserved, and the second pass dropped the superseded paragraph so the breaking-rename wording is the only version left — no fork references remain.

Re-verified on the new base: no stray `xr_ai_agent` / `xr-ai-agent` references outside the intentional deprecated alias, its packaging, and the migration docs. `ruff@0.15.16 check .` clean, `git diff --check` clean, `uv lock --project` resolves for all 34 projects, full CPU suite **643 passed / 2 skipped / 10 deselected**.

Still outstanding and unchanged: the `tests.yml:43` / `nightly-xr-ai-test.yml:164` cache keys reference the removed `agent-sdk/pyproject.toml`. The one-line fix in each is `agent-sdk/xr-ai-hub-client/pyproject.toml`; pushing `.github/workflows/**` needs workflow-scope permission I don't have.

nvddr (changes requested):
Follow-up review of 40ff290.

Blocker:

- Update cache-dependency-glob in both .github/workflows/tests.yml and .github/workflows/nightly-xr-ai-test.yml to replace the deleted agent-sdk/pyproject.toml with agent-sdk/xr-ai-hub-client/pyproject.toml. The setup-uv cache currently no longer tracks the active hub-client dependency manifest, so a hub-client dependency change may reuse a stale cache. This is an in-scope one-line update in each workflow; please arrange a workflow-authorized push before merge.

The canonical xr_ai_hub import, deprecated xr_ai_agent alias, dependency paths, and documentation are otherwise aligned. I found no further actionable issue in this pass.

nvddr (approved):
Approved after fresh re-review: current-main rebase, cache-key correction, compatibility alias, dependency migration, docs, and CPU/GPU checks are all in order.

PR #314, opened 2026-07-29
refactor(render-demo): package the worker and extract the scene-context seam

wenxind-nvidia:
Give the xr-render-demo worker real internal boundaries before it grows further. Structural only — no behavior change, no agent split.

## Why
The worker's brain is six top-level modules sharing one `sys.path` entry, with a **1066-line `processors.py`** at the centre holding turn execution, the reasoning loop, scene-context assembly, perception routing, and transport helpers. Tests and the eval reach in by pushing `worker/` onto `sys.path` and importing bare module names. That layout is at its limit: there is nowhere to put a new collaborating module, and nothing in the brain can be unit-tested without standing up a pipeline.

## What changes

**Package layout**
- `worker/*.py` → the installed **`xr_render_demo_worker` package**; modules import each other relatively and `prompts/` ships inside the package.
- `xr_render_demo_worker.py` splits into `app.py` (startup + pipeline assembly) and `__main__.py`, with `__init__.py` re-exporting `run`/`main` for the console script.
- `agent.py` → **`xr_session.py`**, naming the responsibility it already had: XR/LOVR session lifecycle, launch-failure notices, text-input intake.
- `pyproject.toml` moves from listing every module in `only-include` to `packages = ["xr_render_demo_worker"]` — new modules no longer need a packaging edit to ship.
- **The console-script name is unchanged**, so the orchestrator's `Process(...)` entry and every launch command work as before.

**Scene seam**
- The per-turn context builder moves out of `processors.py` into **`scene.py`** as `build_turn_context(call_tool, …) -> SceneContext`. It was already a self-contained unit (pre-fetch scene state, head pose, and the 1.5 m-ahead position concurrently, then render them plus the move log and conversation history) but reached into four brain attributes to do it. As a function over an injected tool-caller it is independently testable and reusable; `processors.py` keeps a thin delegate that assigns the returned scene snapshot.
- `processors.py`: 1066 → 978 lines.

## Behavior preserved
Same prompts, same reasoning loop, same pipecat pipeline, same model-facing tool surface, same launch commands. The rendered context block is byte-identical — the new tests pin its exact strings.

## Scope boundary
This is the structural groundwork only. Splitting the single flat reasoning loop into focused agents, and any voice-runtime migration, are **not** in this PR.

## Verification
- `ruff@0.15.16 check .` clean; `git diff --check` clean.
- Full CPU suite: **537 passed, 2 skipped, 10 deselected**.
- New `tests/test_render_scene_context.py` (4 tests) pins the context contract — pre-fetch calls, scene/pose rendering, empty-scene and invalid-pose fallbacks, move log, conversation history, and the `position_ahead` fallback — with no pipeline and no LLM.
- Entry points smoke-checked: console script (`xr_render_demo_worker:run`), `python -m xr_render_demo_worker`, and the eval's imports all resolve.
- `ruff.toml` grandfather keys follow the moved paths; `DEPENDENCIES.md` and `AGENTS.md` updated in the same commit (AGENTS.md now documents when a worker graduates from flat modules to a package).


wenxind-nvidia:
Thanks — agreed on both, and on the disposition for each.

**Finding 1 — shared per-turn state. Valid, and deliberately not fixed here.** `_history`, `_recent_moves`, and `_pre_move_positions` are single instance attributes, so two participants share one conversation history, one move log, and one move baseline. It's pre-existing and it is a behavior change to turn state, so folding it into a structural refactor is exactly the scope creep to avoid. As you note, this PR does make the eventual fix cheaper: the hidden `self._history` / `self._recent_moves` reads are now explicit `history=` / `recent_moves=` parameters, so the fix becomes pid-keyed storage in the brain plus one call-site change rather than reworking the context builder. I have a follow-up issue drafted (problem statement, the three concrete cross-participant symptoms, and a suggested pid-keyed fix with a two-participant regression test) — say the word and I'll file it, or feel free to file it yourself.

**Finding 2 — loosened parameter contract. Valid, fixed in `6b94b6b`.** `pid` was required on the original method and became an optional `""`; `ref_us` kept its default. Since the whole point of the seam is reuse, silently dropping the `Participant:` / `Reference time` lines is the wrong default — reference time in particular is what anchors recorded-frame lookups. Both are now **required keyword arguments**; a caller that genuinely wants the block rendered without them passes them explicitly. `recent_moves` / `history` stay optional, since a caller with no prior turns genuinely has neither. The docstring states the reasoning, and the tests pass the values explicitly.

Verification after the change: `ruff@0.15.16 check .` clean; `test_render_scene_context.py` + `test_xr_render_demo_wire.py` → 29 passed.

wenxind-nvidia:
All three items addressed in `88754f6` (then rebased onto current `main`, which picked up `#308`).

**Blocker — the prompt relocation broke the eval workflow.** Confirmed and fixed. You were right that this is functional, not a docs nit: `--prompt` defaults to the moved path and `eval.py` reads it immediately after argparse, so the documented 66-case run died before contacting a model.

Took the stronger fix you suggested — the literal is gone rather than rewritten. The package now declares `PROMPTS_DIR` / `SYSTEM_PROMPT`, and `app.py`, `eval.py`, and the wire tests all resolve from there, so there is a single source of truth that cannot drift on the next move. `eval_watch.sh` and both `xr-render-demo` doc pages are updated. New test `test_bundled_system_prompt_resolves` asserts the bundled prompts exist, so a future relocation fails in CI instead of at run time. Verified the default now resolves and reads (60,547 chars).

**Suggestion 1 — keep `__init__.py` lightweight.** Valid, and it did undercut the PR's own argument, so I did it here rather than defer. `__init__.py` no longer imports `app` — it declares only the prompt paths — and the console script resolves through `xr_render_demo_worker.__main__:run`. The script *name* is unchanged, so the orchestrator's `Process(...)` entry still resolves. New test asserts that importing `xr_render_demo_worker.scene` loads none of `pipecat` / `nat` / `xr_ai_models` / `xr_ai_pipecat` / `xr_ai_nat` (it runs in a subprocess so the check is real). `worker/pyproject.toml:34` and the AGENTS.md guidance are updated, with the rationale recorded.

**Suggestion 2 — enforce the byte-identical claim.** Fair hit: the description promised byte-identical while every assertion was a substring check. Added `test_build_turn_context_renders_the_exact_expected_block`, asserting the complete rendered string for a representative fixture (scene + pose + participant + reference time + move log + history), so an added line, a reordered section, or whitespace drift now fails. The substring tests stay for the fallback cases.

**Suggestion 3 — per-participant turn state.** Agreed it is pre-existing and out of scope here. I have the follow-up written up (the three cross-participant symptoms and a pid-keyed fix with a two-participant regression test) but have not filed it, since opening an issue on the tracker isn't mine to do unprompted — happy to post it if you'd like, or it's yours to file.

Verification: `ruff@0.15.16 check .` clean; `git diff --check` clean; full CPU suite **631 passed / 2 skipped / 10 deselected** on the rebased base.

PR #316, opened 2026-07-30
fix(video): require PyNvVideoCodec 2.2 packet API

### LINKED ISSUE #302: Nightly XR AI test is failing
Nightly run **failed** (event: schedule, commit `70ed909`).

- Run: https://github.com/NVIDIA/xr-ai/actions/runs/30145161565

wenxind-nvidia:
## Summary

- require PyNvVideoCodec 2.2+ in both `xr-media-hub` and `video-memory-service`
- consume only the 2.2 encoded-packet contract in every recorder encode and flush path
- use the same packet handling in the historical-video GPU fixture
- ensure type-contract errors fail the GPU fixture instead of being mislabeled as unavailable NVENC hardware
- add packet-contract regression coverage and update dependency documentation

Fixes #302

## Root cause

PyNvVideoCodec 2.2 changed `Encode()` and `EndEncode()` from returning `bytes`
to returning a list of packet dictionaries with the bitstream under `data`.
The recorder passed that list directly to `bytearray.extend()`, raising
`TypeError: 'dict' object cannot be interpreted as an integer`. Frame errors
were logged and dropped, so no H.264 chunks or sidecars were written.

This PR adopts the 2.2 packet API as the repository's only supported contract;
it does not retain support for the pre-2.2 byte return shape.

## Testing

- `uv run pytest -q test_video_recorder.py` — 6 passed
- targeted GPU tests on NVIDIA L40S with PyNvVideoCodec 2.2.0 — 3 passed
  - synthetic recorder
  - resolution-change recorder
  - historical NVENC → NVDEC → PNG
- `uv run pytest -q -m 'not gpu'` — 642 passed, 2 skipped, 10 deselected
- standalone lock resolution and package builds for both affected projects — passed
- Ruff 0.15.16, SPDX validation, and `git diff --check` — passed

### DIFF
--- a/DEPENDENCIES.md
+++ b/DEPENDENCIES.md
@@ -15,11 +15,9 @@
 ## Python version
 
 Every `pyproject.toml` in this repo pins `requires-python = ">=3.11,<3.13"` by
-convention. The upper bound exists because `PyNvVideoCodec` (used by
-`xr-media-hub` and `video-memory-service` for NVENC encode / NVDEC decode) does not
-yet publish wheels for Python 3.13. With the cap in place, `uv sync` will pick
-3.12 even on a host where 3.13 is also installed. Loosen the upper bound only
-after `PyNvVideoCodec` ships 3.13 wheels.
+convention. PyNvVideoCodec 2.2 publishes Python 3.13 wheels, but the rest of the
+repository dependency graph and test matrix have not been qualified on 3.13.
+Loosen the upper bound only as a coordinated repository-wide change.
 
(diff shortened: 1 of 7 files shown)


nvddr:
Confirmed the intended direction: support PyNvVideoCodec 2.2+ only, with the dependency floor raised to match and the recorder consuming packet-list output directly. This is simpler and avoids carrying legacy bytes-output compatibility code. The current update takes that approach; no further action requested from this point.

nvddr (approved):

PR #317, opened 2026-07-30
docs(architecture): add dependency and folder ownership guide

wenxind-nvidia:
## Summary

- add a runtime and import dependency diagram for clients, the hub, native agent workers, NAT functions, typed services, model services, CloudXR, and MCP compatibility adapters
- document ownership boundaries for every top-level folder
- define the expected Agent SDK and agent-sample layouts and placement rules for new components
- keep the GitHub-facing and Sphinx architecture pages aligned

## Validation

- `git diff --check`
- `uv run --with-requirements docs/requirements.txt sphinx-build -W --keep-going -b html docs/source <output-dir>`

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -23,7 +23,109 @@ agent-mcp-servers/  # Optional MCP compatibility adapters for non-NAT consumers
 agent-samples/      # End-to-end agent demos
 tests/              # Multi-client / multi-agent integration tests
 docs/               # Design docs and topic deep-dives
+models/             # Gitignored model-weight cache
+deps/               # Gitignored downloaded runtime binaries
 ```
+
+## Dependency diagram
+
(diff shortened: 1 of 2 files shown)


nvddr (commented):
**Blocker — update the SDK package map after the rebase.**

`xr-ai-capabilities` was retired by #318, but both architecture pages still present it as a current Agent SDK distribution. Current `main` contains `xr-ai-hub-client`, `xr-ai-models`, `xr-ai-nat`, `xr-ai-voice`, and `xr-ai-pipecat`. Because this page is the placement and dependency guide for future components, documenting a nonexistent package would send new work to the wrong boundary. Please remove that entry from both `docs/architecture.md` and `docs/source/overview/architecture.md`.

The branch is now rebased onto current `main` at `d33fe78`; `make -C docs strict` passes on that rebased head.

PR #319, opened 2026-07-30
refactor(simple-vlm): package worker and adopt native voice runtime

wenxind-nvidia:
## Summary

- package the simple-VLM worker with explicit entry-point, application, configuration, and prompt-resource boundaries
- replace sample-owned Pipecat assembly with `VoiceSession` and the NAT `as_voice_handler` adapter over `StreamingVisionConfig`
- preserve live-frame questions, typed text and ping input, voice-gate behavior, streaming output, participant cleanup, superseded-turn interruption, readiness, ready-file, signal, and shutdown semantics
- make native voice startup safe around readiness failures and ignore typed messages outside an active session
- exclude the broken NLTK 3.10.1 release at both Pipecat-owning SDK boundaries so in-project virtual environments remain importable
- update focused sample, SDK, dependency, and contributor documentation with contract tests

## Scope

This PR is rebased directly on `main`, including the model deployment profiles from #325. The packaged worker now consumes the same `models_config` selection as the stdlib-only orchestrator. `models.local.json`, `models.hosted.json`, `models.omni.json`, their ownership semantics, server configurations, and orchestrator process selection are unchanged by this PR.

The packaged system prompt preserves the former no-config fallback contract, including direct second-person address, and the shipped YAML now uses that same package resource instead of maintaining a divergent copy. The worker keeps its explicit `xr-ai-hub-client` dependency as required by `AGENTS.md` and `DEPENDENCIES.md`.

NLTK 3.10.1 incorrectly rejects its own dependencies when they are installed below the process working directory, which is the normal uv in-project environment used by CI and sample workers. The `nltk!=3.10.1` compatibility constraint lives in both `xr-ai-voice` and `xr-ai-pipecat`, the two packages that directly own Pipecat, rather than in the test environment. Future fixed NLTK releases remain eligible.

No transcript persistence, MCP client, or MCP invocation is introduced. No `xr-render-demo` files change, and `xr-ai-pipecat` remains for its active consumers.

## Validation

- full non-GPU suite on Python 3.11: 693 passed, 2 skipped, 10 deselected
- full non-GPU suite on Python 3.12: 693 passed, 2 skipped, 10 deselected
- fresh dependency resolution selects Pipecat 1.7.0 with NLTK 3.10.0 and imports both SDK voice surfaces from an in-project virtual environment
- focused simple-VLM and voice-session tests: 14 passed
- packaged prompt direct-address contract test
- shared launcher/deployment-profile tests: 17 passed
- Ruff 0.15.16 repository check
- SPDX header audit: 490 files
- strict Sphinx build with warnings as errors
- worker wheel build and installed-wheel packaged-prompt/profile load
- explicit no-MCP and no-`xr-render-demo` scope checks


blongs-nv:
Review by my bot:
Blockers:
  1. Shipped YAML defeats the prompt packaging (agent-samples/simple-vlm-example/yaml/simple_vlm_example_worker.yaml:28) — system_prompt_file: ../worker/simple_vlm_example_worker/prompts/system.txt reaches back into the worker source tree, so the packaged prompt default is never actually used, and an installed wheel without an adjacent checkout fails at startup with
  FileNotFoundError before the ready file. The packaged default is byte-identical; delete the key.
  2. The sample-authoring guide prescribes two contradictory structures (docs/adding-a-sample.md:155, mirrored in docs/source/guides/adding-a-sample.md:155) — the section is retitled <snake_name>_worker/__main__.py with "Follow this structure exactly", but the template body is still the old flat raw-IPC agent, contradicting the __main__.py/app.py/config.py split table
  introduced earlier in the same file. Update the template or label it as the raw-IPC variant and add a voice-runtime one.
  3. The prompt/config forwarding into the vision function is untested (tests/test_simple_vlm_example_worker.py:270) — StreamingVisionConfig is monkeypatched and only endpoint is asserted, so deleting system_prompt=config.system_prompt or the frame knobs from app.py leaves every test green. Assert kwargs["system_prompt"], frame_max_age_s, and frame_timeout_s.

Should-fix issues:

1.   xr-ai-hub-client is declared but never imported (worker/pyproject.toml:13) — nothing under simple_vlm_example_worker/ imports xr_ai_hub; the worker only touches the hub via session.transport.endpoint, and tests/test_simple_vlm_example_worker.py:139 cements the unused dep. Either drop it or state the pin-it-explicitly rule in AGENTS.md.
2.   Empty system_prompt: key becomes the literal string "None" (worker/simple_vlm_example_worker/config.py:52) — a user who blanks the key gets str(None) as the VLM system prompt. Treat None as unset and fall through to the file/default path.
3.   A data message arriving outside session.run() can kill the process (app.py:82) — TextMessageInput registers on the endpoint at construction, before run() sets _handler_processor; in that window enqueue_query raises RuntimeError, which the endpoint's callback error path escalates to os._exit(1). Have TextMessageInput drop messages when the session isn't running.
4.   Transport sockets open before readiness (app.py:57) — VoiceSession.__init__ eagerly builds the ZMQ transport; if wait_for_services loops forever or raises, __aexit__ never runs and sockets/httpx clients leak. The old worker created the transport only after readiness. Cheapest fix is SDK-side (build the transport in __aenter__).
5.   Stale pipecat-first prose left in AGENTS.md — the tree comment (AGENTS.md:20) still says xr-ai-voice has "no consumers yet", AGENTS.md:41-43 and :188-191 still present xr-ai-pipecat as the voice package, and the new-sample checklist (AGENTS.md:130-138) doesn't list __init__.py or the per-sample README this PR introduces.
6.   Config edge cases untested (config.py:41-61) — no test covers load_config(None)/missing path (what run() does with no --config), nor the packaged-prompt fallback when both system_prompt and system_prompt_file are absent. One load_config(None) test closes most of this.
7.   Second-person style rule silently dropped (worker/simple_vlm_example_worker/prompts/system.txt) — the old in-code DEFAULT_SYSTEM_PROMPT line "Speak directly to me in second person… never refer to 'the user' in the third person" is now gone from the repo. Shipped behavior is unchanged, but confirm dropping it from the fallback is intentional.

blongs-nv (approved):

nvddr:
**Blocker:** the packaged `prompts/system.txt` is not behavior-equivalent to the previous `DEFAULT_SYSTEM_PROMPT`. It drops the second-person rule (“Speak directly to me in second person where natural … Never refer to ‘the user’ in the third person”) and now itself refers to “the user”. That silently changes the sample’s response style and contradicts this PR’s byte-for-byte-equivalent claim. Please restore the rule, or explicitly make and document this as an intentional behavior change.

The other earlier review findings are addressed. Separately, the two pytest jobs currently fail during Pipecat/NLTK collection (`regex` safe-path import) before these focused tests run; that appears broader than this PR, but required CI still needs to be rerun or repaired.

PR #320, opened 2026-07-31
refactor(models): separate model deployment specs

wenxind-nvidia:
## Summary

- add `AdapterSpec`, `EndpointSpec`, and `DeploymentSpec`, and compose every role-specific model spec from them
- keep model loading compatible with flat YAML/JSON entries, direct role mappings, `models` roots, nested profiles, presets, and legacy `health_check` values
- preserve the complete public legacy construction surface for `LLMSpec`, `VLMSpec`, `STTSpec`, and `TTSSpec`, including positional/keyword `deployment`, plus read-only flat attribute aliases
- make factories and VLM MCP consume structured specs without changing their public service behavior
- keep `xr-ai-launcher` stdlib-only while deriving process ownership and required credential names from wrapped nested JSON deployment profiles
- explicitly reserve YAML and flat JSON profiles for worker-only `xr-ai-models` loading; shared launcher profiles are validated as wrapped nested `.json`

This branch is rebased directly on the merged #319 and #325 state. The superseded stacked Simple VLM commits are gone, PR 319's packaged-prompt/blank-prompt behavior is retained, and no render-owned files change.

## Validation

- `cd tests && uv run pytest -q -m "not gpu"` — 715 passed, 2 skipped, 10 deselected
- focused model, launcher, client, VLM MCP, and Simple VLM tests — 126 passed
- `uvx --from ruff==0.15.16 ruff check --no-cache .`
- `python3 .github/scripts/check_spdx_headers.py`
- `uv run --with-requirements docs/requirements.txt -- make -C docs strict`
- `uv lock` for `agent-sdk/xr-ai-models`, `utils/xr-ai-launcher`, and `tests`
- `git diff --check`
- verified no `agent-samples/xr-render-demo/` files change

PR #321, opened 2026-07-31
refactor(services): consolidate model-serving projects under services

wenxind-nvidia:
## Summary

- move the VLM, STT, three LLM, and two TTS projects into direct-child `services/` directories
- update model-servers, Simple VLM, launcher references, presets, copied YAML comments, GPU tests, CI/tooling paths, notices, and dependency documentation
- preserve package names, commands, ports, and serving behavior
- normalize standalone reference YAMLs to the repository-root `models/` cache and document how existing offline installations can merge ignored caches left under `ai-services/`
- document and test the direct-child service convention, including a tracked-text guard against retired service paths

## Cache migration

Git cannot move ignored model weights. Before starting relocated services offline, users of the old standalone VLM, STT, Nemotron 3 Nano, Nemotron Omni, Magpie TTS, or Piper TTS configurations should merge `ai-services/models/` and `ai-services/tts/models/` into `models/`. The exact non-clobbering procedure is documented in `docs/ai-services.md` and linked from `services/README.md`.

Sample-owned YAMLs and the standalone Llama Nemotron YAML already resolved to repository-root `models/` and need no migration.

## Scope

- MCP adapters remain under `agent-mcp-servers/`
- `xr-render-demo` changes are limited to mechanical model-service path replacements in `main.py` and one stale path correction in a YAML comment
- no render worker, scene, prompt, eval, or architecture changes

## Validation

- `uv run --project tests pytest -q tests -m 'not gpu'` — 692 passed, 2 skipped, 10 deselected
- focused service-layout, model-config, and launcher tests — 53 passed
- `uv run pytest -q test_piper_tts.py` — 1 passed
- `uv lock --project` for all seven moved projects
- `uvx ruff check --no-cache .`
- strict Sphinx build with warnings as errors
- SPDX header check and `git diff --check`

Stacked on PR #320.


blongs-nv:
Reviewed at `75709b87db4d7122d221d134aaca5f49ada7d669` by a bot.

I verified the path arithmetic independently for all eight moved projects: every `[tool.uv.sources]` path, every `model_cache` YAML value and code default (changed and 100%-similarity-renamed alike), the launcher `Process()` directories, the four-deep profile YAMLs, `.gitignore`, ruff, sonar, and the CI workflow all resolve correctly at the new depth. I also reproduced the focused validation in a fresh clone (the service-layout, model-servers, and launcher-stack test files: 38 passed). No blockers.

## Suggestions

1. **The "Adding a server to a sample" instructions now mis-resolve the cache when followed literally.** The doc has you copy the standalone YAMLs into a sample's `yaml/` directory ([`docs/ai-services.md:133`](https://github.com/NVIDIA/xr-ai/blob/75709b87db4d7122d221d134aaca5f49ada7d669/docs/ai-services.md#L133)), but those YAMLs now carry `model_cache: ../../models`, which from `agent-samples/<sample>/yaml/` resolves to `agent-samples/models`, a gitignored directory that `resolve_model_cache` silently creates, so the user re-downloads tens of GB with no error. The note at [`docs/ai-services.md:420-421`](https://github.com/NVIDIA/xr-ai/blob/75709b87db4d7122d221d134aaca5f49ada7d669/docs/ai-services.md#L420-L421) also contradicts the cp block: it says to copy to the *sample root* (where `../../models` does resolve to the repo root), not `./yaml/`. Add an explicit "then set `model_cache` to `../../../models`" step after the cp block, matching every real sample YAML (e.g. [`agent-samples/simple-vlm-example/yaml/vlm_server.yaml:9`](https://github.com/NVIDIA/xr-ai/blob/75709b87db4d7122d221d134aaca5f49ada7d669/agent-samples/simple-vlm-example/yaml/vlm_server.yaml#L9)), and reconcile the two sections.

2. **Migrate the nightly GPU runner's cache before this lands, or keep the legacy hub dir for one release.** [`tests/test_gpu_llm_servers.py:58-61`](https://github.com/NVIDIA/xr-ai/blob/75709b87db4d7122d221d134aaca5f49ada7d669/tests/test_gpu_llm_servers.py#L58-L61) drops `ai-services/models/hub` from `_HF_HUB_DIRS`, so on a runner whose weights still sit in the legacy location `_hf_any_cached()` finds nothing, `HF_HOME` falls back to the default, and the Nemotron tests cold-download 15-30 GB inside the 1800 s startup budget: a likely nightly timeout until someone runs the migration snippet on that box.

3. **The rewritten launcher paths, the largest surface of this PR, have no automated coverage.** No test resolves the `Process(..., "../../services/<name>", ...)` strings in [`agent-samples/model-servers/main.py:53-90`](https://github.com/NVIDIA/xr-ai/blob/75709b87db4d7122d221d134aaca5f49ada7d669/agent-samples/model-servers/main.py#L53-L90), [`agent-samples/simple-vlm-example/main.py:44-59`](https://github.com/NVIDIA/xr-ai/blob/75709b87db4d7122d221d134aaca5f49ada7d669/agent-samples/simple-vlm-example/main.py#L44-L59), or [`agent-samples/xr-render-demo/main.py:92-110`](https://github.com/NVIDIA/xr-ai/blob/75709b87db4d7122d221d134aaca5f49ada7d669/agent-samples/xr-render-demo/main.py#L92-L110) against the filesystem; a typo would ship green through both CI and the nightly job and only fail at launch. A CPU-only test that walks each main's process list and asserts `(sample_root / p.project).resolve()` contains a `pyproject.toml` would close this.

4. **Only embedding-server's code-default cache is pinned by a test.** [`tests/test_model_servers.py:231-250`](https://github.com/NVIDIA/xr-ai/blob/75709b87db4d7122d221d134aaca5f49ada7d669/tests/test_model_servers.py#L231-L250) correctly pins the default this PR's fix commit changed, but the equivalent `../../models` defaults in the other six services' `__main__.py` files are untested, and that is the exact class of bug that forced the fix commit. Parametrizing the same capture across the vLLM-based servers (and calling `_resolve_model_cache({}, project_dir)` directly for stt/piper/magpie) would pin them all.

5. **`test_service_layout.py` doesn't enforce the convention it documents.** [`tests/test_service_layout.py:62-64`](https://github.com/NVIDIA/xr-ai/blob/75709b87db4d7122d221d134aaca5f49ada7d669/tests/test_service_layout.py#L62-L64) only checks that the eight enumerated directories exist, so a future `services/llm/foo/` nesting (the shape this PR removes) passes, and a new model server gets no contract coverage until manually registered. Asserting that no `services/*/*/pyproject.toml` exists would enforce the direct-child rule generically.

6. **Make the migration snippet portable and cheaper.** `cp -a --no-clobber` ([`docs/ai-services.md:83-91`](https://github.com/NVIDIA/xr-ai/blob/75709b87db4d7122d221d134aaca5f49ada7d669/docs/ai-services.md#L83-L91)) is GNU-only (macOS `cp` needs `-n`), and a full copy transiently doubles a tens-of-GB cache on one disk, so a user can hit ENOSPC mid-copy while following the "keep the legacy directories" instruction. Using `-n` for portability, plus `cp -al` (hardlinks) or an explicit ~2x free-space warning, would help. Both copies of the doc need the edit.

7. **The `services/stt-server/**` Sonar CPD exclusion is new scope, not a rename.** The old list ([`sonar-project.properties:20-24`](https://github.com/NVIDIA/xr-ai/blob/75709b87db4d7122d221d134aaca5f49ada7d669/sonar-project.properties#L20-L24)) excluded only the LLM, TTS, and VLM trees, so this move silently disables duplication detection for stt-server inside a PR whose changelog claims behavior is preserved. Drop the line or say why it belongs; the comment above the list ("Each LLM/TTS backend…") should also name the VLM/STT servers it now covers.

8. **`services/README.md` is a fourth copy of the port/package table and is unreachable from the published docs.** The table ([`services/README.md:12-23`](https://github.com/NVIDIA/xr-ai/blob/75709b87db4d7122d221d134aaca5f49ada7d669/services/README.md#L12-L23), accurate today) duplicates `DEPENDENCIES.md` and both ai-services docs without a row in the DEPENDENCIES.md change-impact map ([line 615](https://github.com/NVIDIA/xr-ai/blob/75709b87db4d7122d221d134aaca5f49ada7d669/DEPENDENCIES.md#L615)) telling the next person to keep them in sync; and the sphinx mirror replaced the link to it with bare text ([`docs/source/components/ai-services.md:17`](https://github.com/NVIDIA/xr-ai/blob/75709b87db4d7122d221d134aaca5f49ada7d669/docs/source/components/ai-services.md#L17)), so nothing in the published docs reaches it.

## Nits

- **Tracked-text guard hardening.** The allowlist at [`tests/test_service_layout.py:44-47`](https://github.com/NVIDIA/xr-ai/blob/75709b87db4d7122d221d134aaca5f49ada7d669/tests/test_service_layout.py#L44-L47) is whole-file, so any legitimate future mention (a migration note, a permalink to an old blob) drops that entire file from coverage; the substring match also misses non-project legacy forms like `ai-services/models`. Line-anchored allowlisting, or a comment declaring the guard a time-boxed migration aid, would keep it from becoming a ratchet. Relatedly, `check=True` on `git ls-files` ([line 53](https://github.com/NVIDIA/xr-ai/blob/75709b87db4d7122d221d134aaca5f49ada7d669/tests/test_service_layout.py#L53)) errors rather than skips outside a git checkout.
- **Two pre-existing content drifts in `docs/ai-services.md`, worth fixing while you're in the file:** the vlm-mcp row says port 8220 ([line 34](https://github.com/NVIDIA/xr-ai/blob/75709b87db4d7122d221d134aaca5f49ada7d669/docs/ai-services.md#L34)) where the YAML and the sphinx mirror both say 8240, and [line 286](https://github.com/NVIDIA/xr-ai/blob/75709b87db4d7122d221d134aaca5f49ada7d669/docs/ai-services.md#L286) still references `_PERSISTENT_SERVERS`, which no longer exists (the code uses `_STOP_SERVICES`; the mirror already dropped the sentence).
- The moved `Process` lines in [`agent-samples/xr-render-demo/main.py:94-105`](https://github.com/NVIDIA/xr-ai/blob/75709b87db4d7122d221d134aaca5f49ada7d669/agent-samples/xr-render-demo/main.py#L94-L105) broke the block's column alignment: stt/agent-llm/vlm/tts now sit three columns left of hub/cloudxr.
- [`docs/changelog.md:66`](https://github.com/NVIDIA/xr-ai/blob/75709b87db4d7122d221d134aaca5f49ada7d669/docs/changelog.md#L66) rewrites a historical 2026-08-03 entry to name a directory that didn't exist on that date, while older entries keep their legacy paths; the changelog is allowlisted from the guard precisely so history can stay as written.

This is a well-executed consolidation: the mechanical rename is complete and correct across 85 files, the near-duplicate doc pairs come out *more* in sync than they went in, the layout contract test is a genuinely useful guard, and unifying the code-default caches quietly fixes a real latent bug where embedding-server's fallback disagreed with its own shipped YAML. It is ready to merge; the suggestions above (the sample-copy doc trap and the nightly runner cache migration first among them) are all follow-up-sized.

blongs-nv:
Reviewed at `be4e5c7541f31041285c0e99adfe15776e8a8115` by a bot.

Suggestions 1 and 3 through 8, and all four nits from my previous review, are resolved: the sample-copy instructions now include the explicit `../../../models` adjustment step and the contradictory note is fixed in both doc copies; `test_sample_process_projects_resolve` covers all three sample mains (both model-servers stacks included); the AST-based `_model_cache_default` helper pins every service's code-level cache default; the layout test now discovers model services and rejects nesting instead of enumerating; the migration snippet was revised to `cp -a -l -n` with a same-filesystem caveat; the stt-server Sonar exclusion is dropped; the change-impact map row and the published-docs link to `services/README.md` are in place; the tracked-text guard gained line-level allowlisting plus a git-checkout skip; and the vlm-mcp port, stale `_PERSISTENT_SERVERS` paragraph, column alignment, and historical changelog entry are all corrected. The focused suite passes in a fresh clone at this SHA (39 passed).

My previous suggestion 2 (the nightly GPU runner cache) is also settled: this PR triggers the nightly workflow because it edits the workflow file itself, and the run on this head passed with the `pytest (gpu)` job completing in about 19 minutes, so the runner resolves its weights fine through the trimmed `_HF_HUB_DIRS` list.

One new defect was introduced by the fix-up commit.

## Blockers

1. **The new nested-project check false-fails on any machine with a synced service venv.** [`tests/test_service_layout.py:116-121`](https://github.com/NVIDIA/xr-ai/blob/be4e5c7541f31041285c0e99adfe15776e8a8115/tests/test_service_layout.py#L116-L121) sweeps `services.rglob("pyproject.toml")` with no exclusion for untracked directories, and installed wheels ship `pyproject.toml` files into site-packages (pandas does, for one, and any single such wheel in the venv triggers the check). The repo's own docs instruct exactly the state that triggers this: CONTRIBUTING.md says each sub-project is its own uv project (`uv sync` in its directory), and this PR's migration section tells users to recreate venvs with `uv sync` in `services/<project>/`. I reproduced it: with a `services/magpie-tts/.venv/lib/python3.12/site-packages/pandas/pyproject.toml` present, the comprehension returns that path and the assertion fails, so the documented `uv run --project tests pytest -q tests -m "not gpu"` breaks for every developer who has synced any service. Restrict the sweep to tracked files (you already have `_tracked_paths()`) or skip `.venv` directories.

The fix-up round is otherwise thorough; every code change I checked does what it claims, and the reworked layout test is a real improvement over the enumerated version. The blocker above is a one-line exclusion fix and should be resolved before merge.

blongs-nv (approved):

PR #323, opened 2026-07-31
docs: publish versioned GitHub Pages

nvddr:
## Summary

- publish Sphinx documentation for `main` and semantic `v*` release tags
- strictly validate a pull request's checked-out Markdown and upload its HTML as a `docs-preview` artifact; protected pushes strictly validate the complete versioned site
- rewrite XR AI repository links as each documentation version is read, so the rendered page points to the matching source tree
- make the Pages root resolve through `/latest/` using SemVer precedence: prefer the highest stable release, or the highest prerelease when no stable release exists
- provide a page-preserving native-link version selector with a safe single-version fallback

## Why

The Pages deployment previously replaced the sole site with documentation from `main`, so release documentation changed as `main` evolved. Versioned output must keep repository links aligned with each immutable source tree, and PR validation must compile the proposed Markdown rather than the base branch.

## Impact

The deployed site exposes `/main/` for development and paths such as `/v0.1.0-beta/` for immutable release documentation. Today the root resolves to `v0.1.0-beta`; once a stable tag exists, later prereleases do not replace it. The Pages environment allows only `main` and `v*` tags to deploy; pull requests remain build-only.

## Validation

- `DOCS_GITHUB_REF=<40-character SHA> make -C docs strict`
- `make -C docs strict-versions`
- rendered-link checks for the PR SHA, `main`, and historical `v0.1.0-beta` output
- `uv run --project tests pytest tests/test_docs_versions.py -q` — 6 passed
- `pre-commit run --files ...` — Ruff and SPDX checks passed

### DIFF
--- a/.github/scripts/select_latest_docs_release.py
+++ b/.github/scripts/select_latest_docs_release.py
@@ -1,0 +1,76 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Select the documentation release tag according to repository policy."""
+from __future__ import annotations
+
+import re
+import sys
+from dataclasses import dataclass
+
+_SEMVER = re.compile(
+    r"^v(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)"
+    r"(?:-([0-9A-Za-z-]+(?:\.[0-9A-Za-z-]+)*))?"
+    r"(?:\+[0-9A-Za-z-]+(?:\.[0-9A-Za-z-]+)*)?$"
+)
+
+
+@dataclass(frozen=True)
+class _Version:
+    tag: str
+    major: int
+    minor: int
+    patch: int
+    prerelease: tuple[tuple[int, int | str], ...] | None
+
+    @property
+    def precedence(self) -> tuple[object, ...]:
+        return (
+            self.major,
+            self.minor,
+            self.patch,
+            self.prerelease is None,
+            self.prerelease or (),
+            self.tag,
+        )
+
+
+def _parse(tag: str) -> _Version | None:
+    match = _SEMVER.fullmatch(tag)
+    if not match:
+        return None
+    prerelease_text = match.group(4)
+    prerelease: list[tuple[int, int | str]] = []
+    if prerelease_text:
+        for identifier in prerelease_text.split("."):
+            if identifier.isdigit():
+                if len(identifier) > 1 and identifier.startswith("0"):
+                    return None
+                prerelease.append((0, int(identifier)))
+            else:
+                prerelease.append((1, identifier))
+    return _Version(
+        tag=tag,
+        major=int(match.group(1)),
+        minor=int(match.group(2)),
+        patch=int(match.group(3)),
+        prerelease=tuple(prerelease) if prerelease_text else None,
+    )
+
+
+def select_latest(tags: list[str]) -> str | None:
+    versions = [version for tag in tags if (version := _parse(tag))]
+    stable = [version for version in versions if version.prerelease is None]
+    candidates = stable or versions
+    if not candidates:
+        return None
+    return max(candidates, key=lambda version: version.precedence).tag
+
+
+def main() -> None:
+    if latest := select_latest([line.strip() for line in sys.stdin if line.strip()]):
+        print(latest)
+
+
+if __name__ == "__main__":
+    main()
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -38,6 +38,7 @@ jobs:
           # sphinx-multiversion reads the release tags and main directly from Git.
           fetch-depth: 0
       - name: Make current docs build ref available
+        if: github.event_name == 'push'
         run: |
           if ! git show-ref --verify --quiet refs/heads/main; then
             git branch main origin/main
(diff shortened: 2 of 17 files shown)


yanziz-nvidia:
Could we add a preview page capability so it is easier to review?

yanziz-nvidia (commented):

## Summary

Adds `sphinx-multiversion` to publish versioned GitHub Pages for `main` and `v*` tags, with a Python SemVer selector to resolve `/latest/`, a link-rewriting hook to keep historical release pages anchored to their source tree, a page-preserving version switcher, and 6 unit tests covering the selection policy. CI is clean; `sphinx-multiversion==0.2.4` is a new `docs/requirements.txt` pin (no `pyproject.toml` touched, DEPENDENCIES.md update not triggered).

|     | Finding |
|-----|---------|
| 🚫 | None |
| 💡 | [`docs/source/conf.py:31`](https://github.com/NVIDIA/xr-ai/blob/6ef8fa04fd713049fa3e626e8be49771d458af16/docs/source/conf.py#L31) — **`_rewrite_github_links` only rewrites `blob/main/` URLs, not `tree/main/` directory links.** Any doc page linking to a repo directory will still point at `main` in historical release subtrees, silently breaking the "immutable source tree" guarantee for directory anchors. |
| 🔍 | [`tests/test_docs_versions.py:53`](https://github.com/NVIDIA/xr-ai/blob/6ef8fa04fd713049fa3e626e8be49771d458af16/tests/test_docs_versions.py#L53) — **`test_source_links_use_the_current_documentation_ref` doesn't cover the default-`main` path.** The case where both `SPHINX_MULTIVERSION_NAME` and `XR_AI_DOCS_GITHUB_REF` are absent is untested. |

- `docs/source/conf.py:31` — Extend `_rewrite_github_links` to also replace `https://github.com/NVIDIA/xr-ai/tree/main/` with `https://github.com/NVIDIA/xr-ai/tree/{_github_ref()}/` so directory links are versioned alongside file links.
- `tests/test_docs_versions.py:53` — Add a sub-case clearing both env vars and asserting the source list is unchanged (verifying the default-`main` no-op path).

wenxind-nvidia (changes requested):
Reviewed at `b1a4f5857edea7da45422c35a6da9eebf5b85450`. The force-push fixed the undeclared `--versioned` argument, but three blockers remain.

## Blockers

1. **The required documentation job still fails, now on historical `blob/main` links.**

   The current [Build (strict) run](https://github.com/NVIDIA/xr-ai/actions/runs/30667153988/job/91276734038) reaches the link checker, then reports many links under `docs/_build/v0.1.0-beta/` whose repository ref is `main` rather than `v0.1.0-beta`. This affects, among others, `components/ai-services.html`, `guides/adding-a-sample.html`, and `reference/xr-render-demo.html`.

   The versioned build/link-rewrite design needs to handle documentation sourced from existing tags; the current implementation does not make those historical links ref-aware. Please make the release output pass the checker without weakening the intended guarantee that links in a release subtree resolve against that release ref. Add an integration test covering a historical source page containing a `blob/main` link.

2. **Pull-request validation does not build the pull request's documentation.**

   [`.github/workflows/docs.yaml:53`](https://github.com/NVIDIA/xr-ai/blob/b1a4f5857edea7da45422c35a6da9eebf5b85450/.github/workflows/docs.yaml#L53) runs the multiversion `strict` target after creating local `main` at `origin/main`. [`conf.py`](https://github.com/NVIDIA/xr-ai/blob/b1a4f5857edea7da45422c35a6da9eebf5b85450/docs/source/conf.py#L58) selects only `main` and release tags, so the detached PR head is excluded. I reproduced this locally: the build contained `main` and `v0.1.0-beta`, and text changed by this PR still rendered the base-`main` wording.

   Consequently, broken Markdown or references introduced by a PR can pass this build. Please add a checked-out-source target using ordinary `sphinx-build -W --keep-going` and run it for pull requests. Keep the multiversion build as a separate publication check.

3. **An ordinary single-version Sphinx build crashes.**

   Running `sphinx-build -W --keep-going -b html docs/source <output>` fails in `pydata_sphinx_theme` because `current_version` is undefined. [`version-switcher.html:8`](https://github.com/NVIDIA/xr-ai/blob/b1a4f5857edea7da45422c35a6da9eebf5b85450/docs/source/_templates/version-switcher.html#L8) assumes `current_version`, `latest_version`, and `versions` are always supplied by `sphinx-multiversion`.

   Please guard the switcher when multiversion context is unavailable, or inject a valid fallback context, and add a regression test for the regular single-version build. This is also necessary for the PR-head validation described above.

## Suggestions

- **Use one SemVer policy.** The `smv_tag_whitelist` accepts tags such as `v1.0.0-01`, while `select_latest_docs_release.py` rejects them. Such a tag could be rendered/listed but never selected as `latest`. Align or centralize the two policies.
- **Provide a review artifact or preview.** Pull-request builds skip artifact upload and deployment, so the existing preview request remains unanswered. Uploading the built HTML as a workflow artifact would at least make the result inspectable without publishing it.
- **Update the PR description and validation coverage.** It currently mentions a `strict-versions` target that does not exist, reports seven documentation tests when six run, and claims a safe single-version fallback that currently crashes.

The overall scope remains appropriately focused, and the selector unit tests and versioned layout are useful foundations, but the blockers above should be resolved before merge.

nvddr:
@yanziz-nvidia @wenxind-nvidia Addressed the review feedback in `13f49ee` and `6ef8fa0`:

- PRs now run a regular strict Sphinx build of the checked-out documentation and upload its HTML as the `docs-preview` artifact. Protected pushes run the separate strict multiversion build.
- The version selector is guarded when ordinary Sphinx does not provide multiversion context.
- Repository links are rewritten while sources are read, so rendered release pages use their own tag. Regression coverage exercises both a PR SHA and a historical tag ref; the strict versioned build also rendered `v0.1.0-beta` without `blob/main` hrefs.
- The tag whitelist now rejects the same invalid SemVer forms as the latest-release selector, and the PR description/validation details are current.

The earlier undeclared `--versioned` invocation is gone. The new CI run for the final head is in progress; there are no unresolved current inline threads.

yanziz-nvidia (approved):
updated bot review

nvddr:
@yanziz-nvidia Follow-up suggestions from the re-review are addressed in `c4fae97`: repository directory links using `tree/main/` now follow the rendered SHA or tag alongside `blob/main/`, and the regression test covers the SHA, historical-tag, and default-`main` paths. The targeted tests and strict docs build pass; CI is running on this final head.

wenxind-nvidia (approved):

PR #325, opened 2026-07-31
feat(models): add deployment profiles

nvddr:
## Summary

- accept nested `adapter`, `endpoint`, and `deployment` model profiles while preserving flat `models.yaml`
- expose deployment ownership and required credentials to stdlib-only orchestrators
- map `managed`, `reused`, and `external` ownership without coupling the launcher to `xr-ai-models`

## Why

Workers and orchestrators need one source of truth when switching between locally managed, shared, and hosted model endpoints.

## Validation

- `uv run --project tests pytest -q tests/test_models_config.py tests/test_launcher_config.py`
- SPDX header check


blongs-nv:
Review by my bot:

  1. The launcher and the SDK parse incompatible profile formats. load_model_deployment (utils/xr-ai-launcher/xr_ai_launcher/_models.py:31) is JSON-only and hard-requires the models: wrapper, while xr_ai_models loads YAML and also accepts the flat form. Every profile currently in the repo is flat YAML (agent-samples/*/yaml/models*.yaml), so the launcher either raises
  Expecting value on them or, given nested-but-flat JSON, fails open: it returns zero services and zero credentials, so an orchestrator would silently skip credential checks and never start a locally-owned server. The docs added in this commit present it as one profile consumed by both sides.

  2. read_config_scalar misparses two real cases (utils/xr-ai-launcher/xr_ai_launcher/_config.py:16), both reproduced:
    - An empty value (models_config: followed by another line) returns the next line's key — literally the string 'profile:' — which then becomes a nonsense profile path in the downstream error.
    - ^\s* with re.MULTILINE matches indented keys, so a models_config: nested inside another section matches as if top-level and silently selects the wrong profile. The docstring says "top-level YAML scalar"; the regex doesn't enforce that.

  3. required_credentials gives different answers on each side. The launcher (_models.py:47) reads only a literal endpoint.api_key_env; the SDK (agent-sdk/xr-ai-models/xr_ai_models/_config.py:112) derives it from the resolved spec. A preset or flat key carrying api_key_env is counted by the SDK but invisible to the launcher — and the launcher's answer is the one that
  gates credential checks.

  4. The new surface has no in-tree caller and the docs reference files that don't exist. Nothing in the repo sets models_config:, there is no models.local.json anywhere (docs/process-model.md and the launcher default both name it), and the existing concrete use-case this seems built for (agent-samples/simple-vlm-example/main.py:42, model_backend: local|nim) wasn't
  migrated. That makes this a third backend-selection mechanism alongside model_backend: and models_yaml:, with none of the three retired. Suggest adopting it end-to-end in at least one sample in this change.

wenxind-nvidia (changes requested):
Reviewed at `9ed5bdd61ab3ce0fa3966958546ba6b5ed680a28`.

## Blockers

1. **`read_config_scalar()` does not enforce its top-level-scalar contract.**

   [`utils/xr-ai-launcher/xr_ai_launcher/_config.py:17`](https://github.com/NVIDIA/xr-ai/blob/9ed5bdd61ab3ce0fa3966958546ba6b5ed680a28/utils/xr-ai-launcher/xr_ai_launcher/_config.py#L17) uses `^\s*` under `re.MULTILINE`. That matches indented keys and text inside block scalars. I reproduced a worker config containing:

   ```yaml
   system_prompt: |
     models_config: injected.json
   ```

   `read_config_scalar(path, "models_config", "models.local.json")` returns `injected.json`. Because `load_model_deployment()` uses this result to choose the profile, the launcher can inspect credentials and process ownership from the wrong file. The same use of `\s*` after the colon can cross a newline, so an empty `models_config:` value may consume the next key as its value.

   Restrict the parser to the real top-level mapping and add regressions for nested/block-scalar occurrences, an empty value, a nested occurrence before a valid top-level value, comments, and quoted values.

2. **The new public abstraction has no production consumer.**

   No sample sets `models_config`, no canonical deployment-profile file is added, and `load_model_deployment()` is referenced only from tests and documentation. `read_config_scalar` is also exported publicly with one internal caller. This leaves a third model-selection surface beside the existing `model_backend` and `models_yaml` paths and conflicts with the repository rule not to add abstractions before two concrete use cases.

   Either migrate concrete local and hosted simple-VLM profiles in this change (which does not require touching `xr-render-demo`), or defer the public launcher/profile APIs until their consumer lands.

## Suggestions

- Add a parity integration test that loads each bundled canonical profile through both `load_models_config()` and `load_model_deployment()`, then compares credentials and service ownership. Separate synthetic fixtures do not prove the two independently implemented readers remain equivalent.
- Define the supported profile boundary explicitly. The worker loader accepts YAML or JSON and wrapped or unwrapped mappings, while the launcher accepts wrapped JSON only. Align them or clearly document and validate the narrower launcher contract.
- Keep `read_config_scalar` private until it has a second real launcher consumer; exporting a generic parser now creates an unnecessary public compatibility surface.
- Update the published Sphinx agent-SDK documentation in addition to the package README and standalone process-model document.
- If this is intended to finish the structured-model-spec convergence item, call out that typed `AdapterSpec` and `EndpointSpec` separation remains outstanding; this change only adds `DeploymentSpec` and flattens the other nested sections into the legacy specs.

The compatibility work itself looks careful: focused tests pass, all existing sample model files still load, and the ownership-conflict validation is useful. The blockers above should be resolved before merge.

blongs-nv (approved):

wenxind-nvidia (approved):

PR #326, opened 2026-07-31
fix(launcher): preflight reused services

nvddr:
## Summary

- require reused processes to declare their HTTP port
- check `/health` before spawning any owned process
- fail immediately with the dependency name and endpoint when it is unavailable

## Why

`launch_mode="reuse"` previously skipped startup unconditionally, so a missing shared service surfaced later as a worker connection failure.

## Validation

- `uv run --project tests pytest -q tests/test_launcher_stack.py`
- SPDX header check

### DIFF
--- a/agent-samples/simple-vlm-example/main.py
+++ b/agent-samples/simple-vlm-example/main.py
@@ -29,6 +29,7 @@ from dataclasses import replace
 from pathlib import Path
 
 from xr_ai_launcher import (
+    EndpointProbe,
     Process,
     ensure_credentials,
     load_model_deployment,
@@ -45,24 +46,32 @@ _MODEL_PROCESSES = {
     "vlm": Process(
         "vlm", "../../services/vlm-server", "vlm_server",
         config="yaml/vlm_server.yaml",
+        port=8100,
     ),
     "vlm-omni": Process(
         "vlm-omni",
         "../../services/nemotron-omni-llm",
         "nemotron_omni_llm_server",
+        port=8108,
     ),
     "stt": Process(
         "stt", "../../services/stt-server", "stt_server",
         config="yaml/stt_server.yaml",
+        port=8103,
     ),
     "tts": Process(
         "tts", "../../services/piper-tts", "piper_tts_server",
         config="yaml/piper_tts_server.yaml",
+        port=8105,
     ),
 }
 
 
-def _build_processes() -> tuple[list[Process], tuple[str, ...]]:
+def _build_processes() -> tuple[
+    list[Process],
+    tuple[str, ...],
+    tuple[EndpointProbe, ...],
+]:
     deployment = load_model_deployment(_BASE / _WORKER_CONFIG)
     unknown_services = deployment.services.keys() - _MODEL_PROCESSES.keys()
     if unknown_services:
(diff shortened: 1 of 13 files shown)


nvddr:
## Review outcome: ready for review

### Addressed

- The blocking reused-service integration gap is fixed at [`8e3b5e5`](https://github.com/NVIDIA/xr-ai/commit/8e3b5e52d4de75a7ecd5995c68bca3df6ca79bea).
- XR render's active reused STT, agent-LLM, and VLM dependencies now declare their health ports; hosted-NIM mode omits the intentionally external LLM/VLM services from preflight.
- The profile-driven Simple VLM Omni reuse path now retains port `8108` as well.
- Current `main` is merged and the prior documentation conflicts are resolved.
- Orchestrator-level regression coverage now verifies both local and hosted process selection.

### Validation

- GitHub CI at `8e3b5e5`: **all reported checks passed** — pytest on Python 3.11 and 3.12, Docs build/publish, Ruff, lock-check, SPDX, DCO, and CodeQL.
- Independent focused validation: **41 tests passed** across launcher preflight, deployment selection, and XR render orchestration; Ruff passed; `git diff --check` passed.
- The PR is mergeable with `main` and has no outstanding review threads.

No remaining code-review blockers found.

utils/xr-ai-launcher/xr_ai_launcher/_stack.py
```diff
@@ -248,6 +250,36 @@ def _spawn(proc: Process, base: Path, ready_file: Path) -> subprocess.Popen:
 
 # ── readiness wait ─────────────────────────────────────────────────────────────
 
+def _require_reused_ready(proc: Process) -> None:
+    """Fail unless a reused HTTP service is already healthy."""
+    if proc.port is None:
+        raise ValueError(f"reused process {proc.name!r} must declare its health-check port")
+
+    health_url = f"http://127.0.0.1:{proc.port}/health"
```
wenxind-nvidia: **Blocker:** This probe ignores the endpoint selected by the deployment profile and reconstructs a health URL from a hardcoded `Process.port`. For example, if the valid reused Omni profile changes `endpoint.base_url` to `http://localhost:9000`, the worker uses port 9000 while the launcher still probes `127.0.0.1:8108/health`. That can reject a healthy configured service, or approve an unrelated service on the old default port. Please carry the selected profile endpoint/readiness into the preflight target (for example, as an explicit health URL) and add coverage using a nondefault reused endpoint.

agent-samples/simple-vlm-example/main.py
```diff
@@ -45,24 +46,32 @@
     "vlm": Process(
         "vlm", "../../services/vlm-server", "vlm_server",
         config="yaml/vlm_server.yaml",
+        port=8100,
```
yanziz-nvidia, line 49: I don't think we should hardcode the ports. It should read from yamls

agent-samples/simple-vlm-example/main.py
```diff
@@ -76,14 +85,35 @@ def _build_processes() -> tuple[list[Process], tuple[str, ...]]:
     for service, process in _MODEL_PROCESSES.items():
         launch_mode = deployment.launch_mode(service)
         if launch_mode is not None:
-            procs.append(replace(process, launch_mode=launch_mode))
+            if launch_mode == "reuse":
```
yanziz-nvidia, line 88: ok to continue using this method for now. But we should move to a better method.

PR #327, opened 2026-07-31
fix(vllm): validate persistent container configuration

nvddr:
## Summary

- label repository-owned vLLM containers with a deterministic launch fingerprint
- reuse healthy or stopped containers only when their launch contract matches
- recreate stale or unlabeled containers before model startup

## Why

A persistent container could be restarted with model arguments, GPU assignment, or memory limits from an older profile even after the active YAML changed.

## Validation

- `uv run --project tests pytest -q tests/test_vllm_docker.py`
- SPDX header check

### DIFF
--- a/DEPENDENCIES.md
+++ b/DEPENDENCIES.md
@@ -169,7 +169,9 @@ xr-ai-vllm  (utils/xr-ai-vllm/)
     `--stop` flow.  Besides `serve` / `stop_persistent_servers`, exposes the
     shared wrapper helpers `resolve_model_cache`, `load_config`, `setup_hf_env`,
     and `gpu_compute_major` (all stdlib-only; pyyaml is imported function-locally
-    inside `load_config` so the `--stop` path stays dependency-free).
+    inside `load_config` so the `--stop` path stays dependency-free). Docker
+    containers carry a deterministic launch fingerprint; containers created
+    with stale model, image, GPU, or vLLM arguments are replaced before reuse.
 
 xr-ai-vad  (utils/xr-ai-vad/)
     └── numpy >=1.24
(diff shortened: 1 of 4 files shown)


nvddr:
Review outcome: **ready for review**

The branch is rebased on current `main`, and the earlier blockers are resolved:

- Healthy ports are reusable only when the expected named Docker container is running and carries the matching launch fingerprint; an unowned listener now fails without being stopped.
- The fingerprint includes a versioned launcher contract, so changes to launcher-controlled Docker behavior (entrypoint, bootstrap, fixed flags) invalidate older immutable containers.
- Matching stopped containers restart, matching containers still starting are monitored instead of receiving a conflicting second `docker run`, and stale containers are recreated.

Validation:

- Focused isolated suite: **42 passed** (`tests/test_vllm_docker.py`)
- Ruff: **passed**
- Fresh GitHub CI: **all checks passed** (Python 3.11/3.12, docs build/publish, Ruff, lock, SPDX, DCO, and CodeQL)

No unresolved review threads remain. The PR is ready for maintainer review and merge.

wenxind-nvidia:
## Follow-up review

### Blockers

None. The update at `724efff` resolves the previous blockers:

- Healthy listeners now require the expected running container and matching fingerprint.
- The fingerprint includes a versioned launcher contract.
- The branch is current with `main`, conflict-free, and mergeable.

### Suggestion

**[P2] Account for credential rotation in the container fingerprint**

`HF_TOKEN` remains outside the fingerprint even though it is stored in the immutable container environment. A matching stopped container can therefore restart with an obsolete token after credential rotation. Consider hashing it into the fingerprint or documenting the intentional exclusion.

### Nits

- Remove the unused `existing` and `running` assignments at lines 640–641 and 662–663, already flagged by the code-quality bot.

### Validation

- `tests/test_vllm_docker.py`: 42 passed
- Ruff, SPDX, and diff checks passed
- GitHub CI: 14/14 checks passed

wenxind-nvidia (approved):

PR #330, opened 2026-08-01
docs: move the version switcher into the sidebar

yanziz-nvidia:
## Problem

The version switcher rendered as a disclosure widget in the navbar, stacked above the site title:

> ▾ Documentation: v0.1.0-beta / This is the latest release. / • v0.1.0-beta • main (development)

The template was named `docs/source/_templates/version-switcher.html`. That is also the name of the component the NVIDIA Sphinx theme puts in `navbar_center` (`theme.conf`), and that component is guarded:

```jinja
{% if theme_switcher %}
{% extends 'pydata_sphinx_theme/components/version-switcher.html' %}
{% endif %}
```

Sphinx's `templates_path` shadowed it, so our unguarded markup was injected into the navbar slot unconditionally.

## Fix

- Rename the template to `versioning.html`. It no longer collides with a theme component, and the theme's own switcher goes back to rendering nothing (this site configures no pydata `switcher`).
- Render it as a labelled `<select>` and wire it in via `html_sidebars = {"**": ["versioning.html", "sidebar-nav-bs"]}`, so it sits at the top of the left sidebar.
- Navigate by index into a build-time JSON array rather than by reading the URL back out of the DOM — `onchange="location = this.value"` tripped CodeQL `js/xss-through-dom`.
- Add `_static/css/custom.css` styling the switcher to match the sidebar search box.

Four files, docs only. The gh-pages publisher and per-PR previews that were on this branch are now #333 — that is a deployment-model change with its own lifecycle and race cases, and it does not belong here.

## Verification

`make -C docs strict` and `make -C docs strict-versions` both pass (`-W --keep-going`). CodeQL passes.

Landing page, `_build/main/index.html`:

```html
<div class="version-switcher">
  <label for="version-select" class="version-switcher__label">Version</label>
  <select id="version-select" class="version-switcher__select">
    <option value="0" >v0.1.0-beta</option>
    <option value="1" selected>main (development)</option>
  </select>
</div>
```

Nested page, `_build/main/getting_started/quickstart.html` — the URL array stays depth-correct and page-relative:

```js
var urls = ["../../v0.1.0-beta/getting_started/quickstart.html", "quickstart.html"];
```

Building with and without a `navbar_center` override produced byte-identical navbars, so no override is carried. The `{% if versions %}` guard keeps the PR-gate `make strict` build (plain `sphinx-build`, no `versions` context) clean.

sphinx-multiversion builds every ref with the *current* checkout's `conf.py` and `_templates`, so this also repairs the already-published `v0.1.0-beta` docs — confirmed in `_build/v0.1.0-beta/index.html`.

Docs-only; no `pyproject.toml` changes, so `DEPENDENCIES.md` is untouched.


nvddr:
**Blocker — the PR-preview publisher can recreate a preview after its PR has closed.** A `pull_request` build that is already queued can reach the shared `docs-gh-pages` group after the `pull_request_target` cleanup has run and found no directory. It then writes `preview/pr-<N>/`, and there is no subsequent close event to remove it. Verify that the PR is still open immediately before publishing, or otherwise make the cleanup ordering durable.

Please also split the `gh-pages` publisher, persistent PR previews, and cleanup workflow into a separate PR. That is a material deployment and branch-permission behavior change with its own lifecycle/race cases; keep this PR focused on the version-switcher template rename, sidebar placement, and CSS.

nvddr (approved):

PR #333, opened 2026-08-03
ci(docs): publish per-PR documentation previews

yanziz-nvidia:
Split out of #330, which was just the version-switcher template change. This is the deployment-model half: it changes how the site is published and adds a branch that CI writes to, with its own lifecycle and race cases.

## Problem

Reviewing a docs change meant merging it. The PR artifact was a zip, and it held only the single-version build — where the version switcher is absent by construction (`{% if versions %}`). #330's change was literally invisible before it went live.

## Fix

Assemble the site on a `gh-pages` branch, then deploy that whole branch as the Pages artifact. One artifact can only serve one site if CI builds it fresh each time; a branch accumulates state, carrying the real site at the root and a preview per open PR:

```
/                   the published site (root redirect -> latest release)
/preview/pr-<N>/    one preview per open same-repo PR
```

Publishing the site replaces the root but never touches `preview/`, so a deleted release tag leaves no stale directory while open previews survive.

Pages itself stays on the **"GitHub Actions"** build type. Branch mode is not usable here: GitHub's docs state that *"commits pushed by a GitHub Actions workflow that uses the `GITHUB_TOKEN` do not trigger a GitHub Pages build"*, so previews would be written to the branch and never deploy. The alternative — pushing with a PAT or GitHub App token — buys nothing here and adds a secret to provision and rotate.

So `gh-pages` is storage, not a publishing source. Every job that mutates it ends by re-uploading the branch and deploying it, via the shared `.github/actions/deploy-docs-site` composite action.

PRs now build the versioned site too, and keep the checked-out build under `current/`. sphinx-multiversion renders the refs recorded in Git, so it never shows a PR's own content — only how that PR's templates and config render. `current/` covers the other half: content changes in the working tree.

## The close/publish race

A build queued before a PR closes can reach the shared `docs-gh-pages` concurrency group *after* the `pull_request_target` cleanup has already run and found no directory. It would then write `preview/pr-<N>/` with no further close event to remove it.

Ordering is not something the publisher can win by checking earlier, so it checks again afterwards. With *C* the close, *K* the cleanup job, and *B → P → A* the before-check, publish and after-check inside one job:

| ordering | outcome |
| --- | --- |
| *C* before *B* | `before` reads `closed`, publish skipped, nothing written |
| *C* between *B* and *A* | `after` reads `closed` and removes what it just wrote |
| *C* after *A* | *K* runs and removes it |

The shared concurrency group is what makes this hold. GitHub acquires it when a job *starts*, so *K* can never interleave between *P* and *A* — it queues behind the whole publish job, removal included. `cancel-in-progress: false` keeps a cleanup from cancelling a publisher mid-push, and the git push is atomic, so a cancelled run leaves no half-written tree.

All three jobs in the group also set `queue: max`. The default (`single`) cancels whatever is already *pending* in a group when a new run arrives, so a burst of publishers could silently drop a queued cleanup and strand a preview. `queue: max` holds them FIFO instead.

## Nightly sweep

Each of those paths still depends on its event firing and its pushes landing. A cleanup job that exhausts its retries, or a preview written before this workflow reached the default branch, would leave a directory nothing revisits.

`docs-preview-sweep.yaml` reconciles once a night: list `preview/pr-*`, ask which of those PRs are still open, remove the difference. Only a confirmed state removes a preview — a `404` means the number never resolved to a PR and is safe to drop, while rate limiting, a 5xx or an auth failure fails the sweep rather than deleting a live preview. It joins the same concurrency group and removes a whole backlog in one commit.

## Safety

Fork PRs get the build artifact only — the publish job is skipped rather than failed, so no fork head ever runs with write access. Cleanup runs on `pull_request_target` and checks out the base branch alone, with event data passed through `env:` rather than interpolated into the shell body. Publishing uses plain `git` in `.github/scripts/`, so no new third-party action is introduced.

One deliberate loosening, worth a reviewer's attention: the `github-pages` environment's deployment branch policy (previously `main` + `v*`) had to be **removed**, because previews deploy from `refs/pull/<N>/merge` and deployment branch policies match branch and tag names only — there is no pattern that admits merge refs. The remaining gate is the job-level condition in `docs.yaml` (`github.repository` plus a same-repo head check). This is inherent to serving previews through `deploy-pages`: per-PR previews or a restrictive environment policy, not both.

## ⚠️ Merge notes

**No Pages settings change is required, and the Pages source must stay on "GitHub Actions".** Switching it to the `gh-pages` branch would disable all future deploys, for the `GITHUB_TOKEN` reason above.

The only settings change this needed has already been made: removing the `github-pages` environment's deployment branch policy (see Safety).

While this PR is open, the `gh-pages` root is empty — the branch has only ever been written by *preview* publishes — so a preview deploy serves a site with no root and https://nvidia.github.io/xr-ai/ 404s until the next `main` deploy re-runs. **This resolves itself on merge:** the squash commit touches `.github/workflows/docs.yaml`, which is in the push path filter, so merging triggers Docs on `main`, `publish_docs.sh` runs with no `PR_NUMBER`, and the root is written alongside `preview/`. From then on both serve together.

Note that `pull_request_target` loads its workflow from the default branch, so preview cleanup is inert until this merges. The sweep is what recovers anything stranded before then.

## Verification

Both scripts were exercised against a scratch repository:

| case | result |
| --- | --- |
| first publish (no branch yet) | branch created with no source history |
| preview publish | `preview/pr-N/` written, root untouched |
| republish site | stale `v1/` removed, `v2/` added, `preview/pr-N/` retained |
| no-op publish | "Documentation is unchanged; nothing to publish" |
| remove several PRs at once | one commit, unknown numbers skipped, other previews and the site root untouched |
| remove again / unknown PR only | success, no commit |
| remove with no arguments | usage error, exit 1 |

The sweep's selection loop was run against a stubbed API: a closed PR queued for removal, an open one kept, and a non-conforming directory name ignored.

Live on this PR: the `Publish to GitHub Pages` job writes `preview/pr-333/` and deploys it, and https://nvidia.github.io/xr-ai/preview/pr-333/ served 200 with the merged switcher in `main/index.html`.

### DIFF
--- a/.github/actions/deploy-docs-site/action.yaml
+++ b/.github/actions/deploy-docs-site/action.yaml
@@ -1,0 +1,47 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Deploy the current ``gh-pages`` branch contents to GitHub Pages.
+#
+# Pages stays on the "GitHub Actions" build type and ``gh-pages`` is only
+# storage: a push to it deploys nothing on its own, because commits pushed by a
+# workflow using GITHUB_TOKEN do not trigger a Pages build. Every job that
+# mutates the branch therefore ends by re-uploading the whole branch and
+# deploying it.
+#
+# The calling job must set:
+#   permissions: {contents: write, pages: write, id-token: write}
+#   environment: github-pages
+name: Deploy docs site
+description: Upload the gh-pages branch as a Pages artifact and deploy it.
+
+runs:
+  using: composite
+  steps:
+    - name: Collect the gh-pages branch
+      id: collect
+      shell: bash
+      env:
+        REMOTE: https://x-access-token:${{ github.token }}@github.com/${{ github.repository }}.git
+      run: |
+        set -euo pipefail
+        site="${RUNNER_TEMP}/pages-site"
+        if ! git ls-remote --exit-code --heads "$REMOTE" gh-pages >/dev/null 2>&1; then
+          echo "No gh-pages branch; nothing to deploy."
+          echo "present=false" >> "$GITHUB_OUTPUT"
+          exit 0
+        fi
+        rm -rf "$site"
+        git clone --quiet --depth 1 --branch gh-pages "$REMOTE" "$site"
+        # The artifact is served verbatim; the clone's own history is not part
+        # of the site.
+        rm -rf "$site/.git"
+        echo "present=true" >> "$GITHUB_OUTPUT"
+
+    - uses: actions/upload-pages-artifact@v3
+      if: steps.collect.outputs.present == 'true'
+      with:
+        path: ${{ runner.temp }}/pages-site
+
+    - uses: actions/deploy-pages@v4
+      if: steps.collect.outputs.present == 'true'
--- a/.github/scripts/publish_docs.sh
+++ b/.github/scripts/publish_docs.sh
@@ -1,0 +1,74 @@
+#!/usr/bin/env bash
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Publish a built documentation site to the ``gh-pages`` branch.
+#
+#   publish_docs.sh <site-dir>          publish to the branch root
+#   PR_NUMBER=42 publish_docs.sh <dir>  publish to preview/pr-42/
+#
+# Publishing the site replaces everything at the root except ``preview/``, so a
+# deleted release tag does not leave a stale directory behind while open PR
+# previews survive. Publishing a preview touches only that PR's directory.
+#
+# Callers must serialize: every invocation pushes the same branch. The retry
+# loop only covers a lost race, not concurrent use as a general strategy.
+set -euo pipefail
+
+src="${1:?usage: publish_docs.sh <site-dir>}"
+: "${GITHUB_TOKEN:?GITHUB_TOKEN is required}"
+: "${GITHUB_REPOSITORY:?GITHUB_REPOSITORY is required}"
+: "${GITHUB_SHA:?GITHUB_SHA is required}"
+pr="${PR_NUMBER:-}"
+
+remote="https://x-access-token:${GITHUB_TOKEN}@github.com/${GITHUB_REPOSITORY}.git"
+work="$(mktemp -d)"
+
+if git ls-remote --exit-code --heads "$remote" gh-pages >/dev/null 2>&1; then
+  git clone --quiet --depth 1 --branch gh-pages "$remote" "$work"
+else
+  # First publish: start an empty branch rather than branching off the default
+  # one, so the site never carries the source tree's history.
+  git init --quiet --initial-branch gh-pages "$work"
+  git -C "$work" remote add origin "$remote"
+fi
+
+if [[ -n "$pr" ]]; then
+  dest="$work/preview/pr-${pr}"
+  rm -rf "$dest"
+  mkdir -p "$dest"
+  cp -R "$src/." "$dest/"
+  message="docs: preview for PR #${pr}"
+else
+  find "$work" -mindepth 1 -maxdepth 1 ! -name .git ! -name preview -exec rm -rf {} +
+  cp -R "$src/." "$work/"
+  message="docs: publish ${GITHUB_SHA:0:8}"
+fi
+
+# Sphinx writes .nojekyll per build directory; the branch root needs its own or
+# Jekyll drops every _static/ path.
+touch "$work/.nojekyll"
+
+cd "$work"
+git config user.name "github-actions[bot]"
+git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+git add --all
+
+if git diff --cached --quiet; then
+  echo "Documentation is unchanged; nothing to publish."
+  exit 0
+fi
+
+git commit --quiet --message "$message"
+
+for attempt in 1 2 3; do
+  if git push --quiet origin gh-pages; then
+    exit 0
+  fi
+  echo "Push rejected (attempt ${attempt}); rebasing onto the current branch."
+  git fetch --quiet --depth 1 origin gh-pages
+  git rebase --quiet FETCH_HEAD
+done
+
+echo "Could not publish to gh-pages after 3 attempts." >&2
+exit 1
--- a/.github/scripts/publish_docs.sh
+++ b/.github/scripts/publish_docs.sh
@@ -45,8 +45,9 @@ else
   message="docs: publish ${GITHUB_SHA:0:8}"
 fi
 
-# Sphinx writes .nojekyll per build directory; the branch root needs its own or
-# Jekyll drops every _static/ path.
+# Pages serves this branch through an artifact upload, so Jekyll never runs;
+# .nojekyll is kept at the root so the branch is still correct if it is ever
+# served directly.
 touch "$work/.nojekyll"
 
 cd "$work"
--- a/.github/scripts/remove_docs_preview.sh
+++ b/.github/scripts/remove_docs_preview.sh
@@ -1,0 +1,51 @@
+#!/usr/bin/env bash
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Remove one PR's documentation preview from the ``gh-pages`` branch.
+#
+#   remove_docs_preview.sh <pr-number>
+#
+# Idempotent: a missing branch or a missing directory is success, so it is safe
+# to call speculatively. Callers must serialize with publish_docs.sh — both
+# push the same branch.
+set -euo pipefail
+
+pr="${1:?usage: remove_docs_preview.sh <pr-number>}"
+: "${GITHUB_TOKEN:?GITHUB_TOKEN is required}"
+: "${GITHUB_REPOSITORY:?GITHUB_REPOSITORY is required}"
+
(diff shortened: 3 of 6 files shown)


nvddr:
## Review findings

1. **[P1] `GITHUB_TOKEN` pushes won't trigger a Pages build.** The publisher authenticates and pushes `gh-pages` using `GITHUB_TOKEN` ([lines 24 and 65](https://github.com/NVIDIA/xr-ai/blob/3a4412089a29fdce2f9d64d629c590abbc8efd06/.github/scripts/publish_docs.sh#L24-L65)). GitHub explicitly says commits pushed by a GitHub Actions workflow using `GITHUB_TOKEN` do not trigger a GitHub Pages build, so subsequent site and preview updates may never deploy after switching Pages to branch mode. Use a GitHub App/PAT or an explicit Pages deployment workflow. [GitHub documentation](https://docs.github.com/en/actions/concepts/security/github_token)

2. **[P2] A transient API failure deletes valid previews.** The sweep converts every `gh api` error—rate limiting, network failure, 5xx, or authorization failure—to `unknown`, then queues that PR for removal ([lines 56–61](https://github.com/NVIDIA/xr-ai/blob/3a4412089a29fdce2f9d64d629c590abbc8efd06/.github/workflows/docs-preview-sweep.yaml#L56-L61)). Only a confirmed closed state should remove a preview; operational failures should fail or skip the sweep.

3. **[P2] Cleanup jobs can be silently canceled while pending.** All publishers and cleanup jobs share `docs-gh-pages` with `cancel-in-progress: false` ([docs.yaml line 106](https://github.com/NVIDIA/xr-ai/blob/3a4412089a29fdce2f9d64d629c590abbc8efd06/.github/workflows/docs.yaml#L106)). By default, GitHub retains only one pending job per concurrency group; a newer publisher can replace a pending close-cleanup job. Add `queue: max` consistently or redesign the reconciliation. [GitHub concurrency documentation](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency)

4. **[P2] The workflow path filters omit the new publishing scripts.** A future PR changing only `publish_docs.sh` or `remove_docs_preview.sh` will not run the Docs workflow ([lines 20–33](https://github.com/NVIDIA/xr-ai/blob/3a4412089a29fdce2f9d64d629c590abbc8efd06/.github/workflows/docs.yaml#L20-L33)).

All seven current CI workflows pass, but these operational cases are not covered by those checks.

yanziz-nvidia:
> ## Review findings
> 1. **[P1] `GITHUB_TOKEN` pushes won't trigger a Pages build.** The publisher authenticates and pushes `gh-pages` using `GITHUB_TOKEN` ([lines 24 and 65](https://github.com/NVIDIA/xr-ai/blob/3a4412089a29fdce2f9d64d629c590abbc8efd06/.github/scripts/publish_docs.sh#L24-L65)). GitHub explicitly says commits pushed by a GitHub Actions workflow using `GITHUB_TOKEN` do not trigger a GitHub Pages build, so subsequent site and preview updates may never deploy after switching Pages to branch mode. Use a GitHub App/PAT or an explicit Pages deployment workflow. [GitHub documentation](https://docs.github.com/en/actions/concepts/security/github_token)
> 2. **[P2] A transient API failure deletes valid previews.** The sweep converts every `gh api` error—rate limiting, network failure, 5xx, or authorization failure—to `unknown`, then queues that PR for removal ([lines 56–61](https://github.com/NVIDIA/xr-ai/blob/3a4412089a29fdce2f9d64d629c590abbc8efd06/.github/workflows/docs-preview-sweep.yaml#L56-L61)). Only a confirmed closed state should remove a preview; operational failures should fail or skip the sweep.
> 3. **[P2] Cleanup jobs can be silently canceled while pending.** All publishers and cleanup jobs share `docs-gh-pages` with `cancel-in-progress: false` ([docs.yaml line 106](https://github.com/NVIDIA/xr-ai/blob/3a4412089a29fdce2f9d64d629c590abbc8efd06/.github/workflows/docs.yaml#L106)). By default, GitHub retains only one pending job per concurrency group; a newer publisher can replace a pending close-cleanup job. Add `queue: max` consistently or redesign the reconciliation. [GitHub concurrency documentation](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency)
> 4. **[P2] The workflow path filters omit the new publishing scripts.** A future PR changing only `publish_docs.sh` or `remove_docs_preview.sh` will not run the Docs workflow ([lines 20–33](https://github.com/NVIDIA/xr-ai/blob/3a4412089a29fdce2f9d64d629c590abbc8efd06/.github/workflows/docs.yaml#L20-L33)).
> 
> All seven current CI workflows pass, but these operational cases are not covered by those checks.

On the P1: confirmed against GitHub's docs - "Commits pushed by a GitHub Actions workflow that uses the GITHUB_TOKEN do not trigger a GitHub Pages build" - so branch mode would have written previews that never deployed. Rather than provision a PAT, I kept Pages on the "GitHub Actions" build type: gh-pages stays the assembly area, and each job that mutates it re-uploads the branch and deploys it through a shared composite action. This depended on removing the github-pages environment's deployment branch policy (main + v*), since previews deploy from refs/pull/*/merge. That's now done; the job-level github.repository and same-repo-PR gating is what keeps forks out. Note the gh-pages root is empty until the first post-merge push to main publishes it, so the site root 404s while this PR is open.

The sweep now only treats a 404 as removable and fails the run on any other API error.

I applied the same fix to the contents/preview listing, which had the same || true problem (a 5xx would have read as "no previews"). Path filters now include both publishing scripts. 

On the concurrency point: the pending-job eviction is real, but there's no `queue: max` option in GitHub Actions. The nightly sweep is the intended backstop for a lost cleanup, so I documented that in `docs-preview-cleanup.yaml` instead of restructuring.

nvddr:
## Re-review at `a2bb384`

Three of the four earlier findings are addressed:

- The branch is now only an assembly area; the new composite action explicitly uploads and deploys it, so deployment no longer depends on a `GITHUB_TOKEN` push triggering Pages.
- The sweep distinguishes 404s from operational API failures and fails safely on the latter.
- The Docs path filters now include both publishing scripts and the composite action.

Two findings remain:

1. **[P1] The preview deployment has taken the live documentation root offline.** The PR publish job always runs the new `Deploy the site` step after writing `preview/pr-333/` ([docs.yaml line 183](https://github.com/NVIDIA/xr-ai/blob/a2bb38454994214fac45febb3d7ed18da34581e3/.github/workflows/docs.yaml#L183)), and the composite action deploys the entire current `gh-pages` tree ([action.yaml line 46](https://github.com/NVIDIA/xr-ai/blob/a2bb38454994214fac45febb3d7ed18da34581e3/.github/actions/deploy-docs-site/action.yaml#L46)). Because that branch has not yet been seeded with the root site, the successful PR deployment replaced the previous production artifact with a preview-only artifact. At re-review, `https://nvidia.github.io/xr-ai/` returns HTTP 404 while `https://nvidia.github.io/xr-ai/preview/pr-333/` returns 200. Bootstrap the branch root before allowing any PR deployment, or skip preview deployment until a root site is present.

2. **[P2] Pending cleanup eviction is still avoidable with `queue: max`.** The shared group still uses the default single-pending behavior ([docs.yaml line 125](https://github.com/NVIDIA/xr-ai/blob/a2bb38454994214fac45febb3d7ed18da34581e3/.github/workflows/docs.yaml#L125), with matching declarations in cleanup and sweep). GitHub's current workflow syntax explicitly supports `queue: max`, allowing up to 100 pending jobs and FIFO processing. Add it consistently to all three `docs-gh-pages` concurrency declarations. The nightly sweep remains a useful backstop, but it should not be the normal recovery path for an evicted close-cleanup job. [GitHub concurrency documentation](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency#queueing-multiple-pending-runs)

All seven workflows on the current head are green.

nvddr:
## Re-review at `8e2f0c8`

The concurrency finding is resolved: `queue: max` is now present in all three `docs-gh-pages` concurrency declarations.

Approval is still blocked by two deployment issues:

1. **[P1] The production documentation root is still offline.** The latest Docs workflow completed successfully, but `https://nvidia.github.io/xr-ai/` still returns HTTP 404 while `https://nvidia.github.io/xr-ai/preview/pr-333/` returns 200. The latest commit only changes concurrency; the PR publish path still deploys the preview-only `gh-pages` tree over the production Pages artifact. Seed or restore the branch root before allowing PR deployments to replace production, or skip the Pages deployment while the root site is absent.

2. **[P1] The PR's merge instructions now contradict the implementation.** The description still says an admin must switch Pages from **GitHub Actions** to the `gh-pages` branch after merge. The revised workflow explicitly requires Pages to remain on **GitHub Actions**, because `GITHUB_TOKEN` pushes do not trigger a branch-based Pages build. Following the current merge checklist would disable future deployments. Update the PR description and remove the branch-source switch step before handoff.

Current CI: Docs, CodeQL, lint, SPDX, DCO, and lock-check pass; the tests workflow is still in progress.

yanziz-nvidia:
Fixed above

nvddr:
## Final re-review at `8e2f0c8`

No remaining code-review findings. The prior issues are resolved:

- Pages remains on the **GitHub Actions** source, with `gh-pages` used only as the assembled-site store and explicitly deployed through the composite action.
- The sweep fails safely on operational API errors and removes previews only from confirmed closed or missing PRs.
- Publishing paths include the new scripts and composite action.
- All three shared concurrency declarations use `queue: max`.
- The PR description and merge notes now match the revised deployment model.

Live verification now shows the production root returning HTTP 200. The PR preview currently returns 404 because restoring the production artifact replaced the temporary preview-only deployment; that is not a code blocker because the preview deployment was previously observed working, and the merge-triggered `main` run will populate and deploy the branch root under the new workflow.

All seven workflows pass, the PR is mergeable, and there are no unresolved inline review threads. **Safe to approve.** After merge, verify the Docs run on `main` succeeds and that the production root remains healthy.

nvddr (approved):

PR #337, opened 2026-08-03
Remove Llama-3.1-Nemotron-Nano-8B-v1 model

blongs-nv:
Removes quick response extra small LLM.


nvddr (approved):

agent-samples/xr-render-demo/worker/processors.py
```diff
@@ -10,12 +10,12 @@
...
 
 Agentic loop (max ``_MAX_LOOP`` iterations):
-  - Llama-Nemotron emits an OpenAI ``tool_calls`` payload → execute tool,
+  - Nemotron-3-Nano emits an OpenAI ``tool_calls`` payload → execute tool,
     append result, continue.
   - When the model returns text instead of a tool call, that text is the
     final user-visible response.
 
-A parallel "quick-ack" call to Minitron fires at the start of each turn
+A parallel "quick-ack" call fires at the start of each turn
```
nvddr: Nit: this says the quick-ack is parallel, but the implementation now awaits it before starting the agentic loop. Could we describe it as an initial/pre-loop quick-ack instead?

blongs-nv: Updated

PR #339, opened 2026-08-04
Add selectable Nemotron Omni model-server stack

nvddr:
## Summary

Adds a selectable shared model-server layout:

- Default `--vlm-llm-stack` keeps the Nemotron-3 Nano + Cosmos VLM + STT stack.
- `--omni-stack` starts Nemotron-3 Nano Omni + STT instead, with GPU-specific Omni profiles.
- `model_servers --stop` stops every port used by either stack without needing a stack flag.

Nemotron Omni now persists like the other shared vLLM services, allowing `model_servers` to exit after readiness while retaining loaded weights.

## Validation

- `uv run --project tests pytest tests/test_model_servers.py tests/test_models_config.py tests/test_vllm_docker.py -q`
- `uv run --project tests ruff check tests/test_model_servers.py agent-samples/model-servers/main.py ai-services/llm/nemotron_omni/nemotron_omni_llm_server/__main__.py`

Depends on #337 (base branch: `blongs-nv/remove-quick-llm`).

### DIFF
--- a/DEPENDENCIES.md
+++ b/DEPENDENCIES.md
@@ -560,10 +560,13 @@ run this first to warm up model weights before starting any demo sample.
 |---|---|---|---|
 | Orchestrator | `model-servers` | `xr-ai-launcher`, `xr-ai-logging`, `xr-ai-vllm` (for `--stop`) | — |
 
-Starts stt-server (8103), nemotron3-nano-llm-server (8107, `persistent=True`),
-vlm-server (8100, `persistent=True`), and embedding-server (8109,
-`persistent=True`). The vLLM servers survive launcher restarts; use `--stop`
-to shut them down.
+The default `--vlm-llm-stack` starts stt-server (8103),
+nemotron3-nano-llm-server (8107, `persistent=True`), and vlm-server (8100,
+`persistent=True`), plus embedding-server (8109, `persistent=True`).
+`--omni-stack` replaces Nano and Cosmos with nemotron-omni-llm-server (8108,
+`persistent=True`) while retaining stt-server and embedding-server. The stacks
+are mutually exclusive; `--stop` shuts down every model-server port without
+selecting one.
 GPU profiles: `dual_48G_ada`, `spark`, `96G_blackwell` (auto-detected).
 
 ### xr-render-demo  (agent-samples/xr-render-demo/)
--- a/README.md
+++ b/README.md
@@ -158,6 +158,8 @@ tens of minutes).  On subsequent runs the containers restart in under a minute.
 The default `--vlm-llm-stack` starts Nemotron-3 Nano (8107), Cosmos (8100),
 STT (8103), and embeddings (8109). Use `--omni-stack` to replace Nano and
 Cosmos with Nemotron-3 Nano Omni (8108); STT and embeddings remain available.
+On `dual_48G_ada`, the default stack places Cosmos and embeddings on GPU 0;
+the Omni stack places Omni on GPU 0 and embeddings on GPU 1.
 Switching stacks stops the incompatible persistent models first and aborts if
 they cannot be stopped, avoiding GPU overcommit.
 
--- a/README.md
+++ b/README.md
@@ -155,6 +155,16 @@ GPU profiles are auto-detected (`dual_48G_ada` / `spark` / `96G_blackwell`).
 On first run each model downloads from HuggingFace (~50 GB total; can take
 tens of minutes).  On subsequent runs the containers restart in under a minute.
 
+The default `--vlm-llm-stack` starts Nemotron-3 Nano (8107), Cosmos (8100),
+STT (8103), and embeddings (8109). Use `--omni-stack` to replace Nano and
+Cosmos with Nemotron-3 Nano Omni (8108); STT and embeddings remain available.
+Switching stacks stops the incompatible persistent models first and aborts if
+they cannot be stopped, avoiding GPU overcommit.
+
+```bash
+uv run model_servers --omni-stack
+```
+
 The default models are public, so no HuggingFace token is required.  Set
 `HF_TOKEN` to lift download rate limits / speed, or to use a gated model — see
 [`docs/credentials.md`](docs/credentials.md).  The launcher won't prompt; it
@@ -165,6 +175,8 @@ To stop all model servers when done:
 ```bash
 uv run model_servers --stop
 ```
+
+`--stop` always stops both stack variants, so it takes no stack-selection flag.
 
 ### Simple VLM example (vision Q&A over voice + text)
 
--- a/agent-samples/model-servers/main.py
+++ b/agent-samples/model-servers/main.py
@@ -2,17 +2,23 @@
 # SPDX-License-Identifier: Apache-2.0
 
 """
-model-servers orchestrator — starts the shared AI inference servers and exits.
+model-servers orchestrator — starts one shared AI inference stack and exits.
 
 All servers are launch_mode="persist" so they keep running after this
 process exits.  Model weights stay hot across stack restarts.
 
 Servers started
 ---------------
-  stt        — nvidia/parakeet-tdt-0.6b-v3        port 8103  (NeMo ASR)
-  agent-llm  — NVIDIA-Nemotron-3-Nano-30B-A3B      port 8107  (vLLM)
-  vlm        — nvidia/Cosmos-Reason1-7B            port 8100  (vLLM)
-  embedding  — nvidia/llama-nemotron-embed-1b-v2   port 8109  (vLLM)
+  default / --vlm-llm-stack
+    stt        — nvidia/parakeet-tdt-0.6b-v3        port 8103  (NeMo ASR)
+    agent-llm  — NVIDIA-Nemotron-3-Nano-30B-A3B     port 8107  (vLLM)
+    vlm        — nvidia/Cosmos-Reason1-7B           port 8100  (vLLM)
+    embedding  — nvidia/llama-nemotron-embed-1b-v2  port 8109  (vLLM)
+
+  --omni-stack
+    stt        — nvidia/parakeet-tdt-0.6b-v3        port 8103  (NeMo ASR)
+    omni       — Nemotron-3-Nano-Omni-30B-A3B       port 8108  (vLLM)
+    embedding  — nvidia/llama-nemotron-embed-1b-v2  port 8109  (vLLM)
 
 How to run:
     uv run --project agent-samples/model-servers model_servers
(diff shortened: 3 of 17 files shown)


wenxind-nvidia:
## Blocker

1. **Stack switching leaves the previous persistent models running.** [`main.py:124`](https://github.com/NVIDIA/xr-ai/blob/aa159bb591c9a06587280cea0acd0cb1c3a53dcc/agent-samples/model-servers/main.py#L124) starts only the selected processes but never stops the incompatible stack. Switching from the default stack to Omni leaves Nano and Cosmos resident, while switching back leaves Omni resident. Because these services share the same GPU budget, this can cause immediate overcommit or OOM. Stop the incompatible services before `run_stack()` and test both transition directions.

2. **The required Blackwell MoE backend is never applied.** The profile declares [`moe_backend: triton`](https://github.com/NVIDIA/xr-ai/blob/aa159bb591c9a06587280cea0acd0cb1c3a53dcc/agent-samples/model-servers/yaml/96G_blackwell/nemotron_omni_llm_server.yaml#L22-L23), but the [Omni wrapper's vLLM arguments](https://github.com/NVIDIA/xr-ai/blob/aa159bb591c9a06587280cea0acd0cb1c3a53dcc/ai-services/llm/nemotron_omni/nemotron_omni_llm_server/__main__.py#L122-L139) never read that key or emit `--moe-backend triton`. The configuration currently has no effect despite stating it is required for RTX Pro Blackwell.

3. **The required DCO check is failing.** Commit [`aa159bb`](https://github.com/NVIDIA/xr-ai/commit/aa159bb591c9a06587280cea0acd0cb1c3a53dcc) has no `Signed-off-by` trailer. Amend/sign off the commit before merge. [DCO check](https://github.com/NVIDIA/xr-ai/actions/runs/30954409427/job/92143950398)

## Suggestion

- Update the Omni per-server documentation. It still says Omni [runs foreground without persistence](https://github.com/NVIDIA/xr-ai/blob/aa159bb591c9a06587280cea0acd0cb1c3a53dcc/docs/ai-services.md#L314-L322), contradicting the new `persistent=True` behavior and earlier documentation sections. The source documentation has the [same stale statement](https://github.com/NVIDIA/xr-ai/blob/aa159bb591c9a06587280cea0acd0cb1c3a53dcc/docs/source/components/ai-services.md#L300-L308).

- Extend the profile tests beyond checking that each YAML file exists. [`test_omni_profile_exists_for_each_supported_gpu`](https://github.com/NVIDIA/xr-ai/blob/aa159bb591c9a06587280cea0acd0cb1c3a53dcc/tests/test_model_servers.py#L41-L49) would not catch unsupported or ignored settings such as `moe_backend`.

## Nit

- Generalize the Omni wrapper documentation: it describes only an NGC container and NGC-specific dependency installation, while all new profiles select `vllm/vllm-openai:v0.20.0` with `extra_pip: []`. [`__main__.py:7`](https://github.com/NVIDIA/xr-ai/blob/aa159bb591c9a06587280cea0acd0cb1c3a53dcc/ai-services/llm/nemotron_omni/nemotron_omni_llm_server/__main__.py#L7-L39)

Validation: 61 targeted tests passed, Ruff passed, and `git diff --check` passed. GPU model startup was not available locally. GitHub's Python 3.12 job was still running at the final check; Python 3.11 passed.

nvddr:
Addressed all review items in `3a5ce5b`:

- Starting either stack now stops only the incompatible persistent GPU models before launch; STT and embeddings remain running. Regression coverage verifies both transitions and the stop-before-start order.
- `moe_backend` now forwards to vLLM as `--moe-backend`; profile and wrapper tests cover the Blackwell Triton setting.
- Updated both Omni documentation surfaces to describe its persistent lifecycle and generalized the wrapper docs for arbitrary Docker images.
- Amended the commit with a `Signed-off-by` trailer to resolve DCO.

Validation: `uv run --project tests pytest -q tests/test_model_servers.py` (15 passed); focused Ruff and `git diff --check` passed.

wenxind-nvidia:
## Blocker: selector test can stop real model servers

[`test_cli_selects_requested_stack`](https://github.com/NVIDIA/xr-ai/blob/73868d619a3b676477361c4ca93188e4453f7c3c/tests/test_model_servers.py#L109-L123) mocks `_build_processes()` and `run_stack()`, but not `_stop_incompatible_stack()`. Each parameterized call therefore reaches the real `stop_persistent_servers()` implementation. Running this unit test on a development host with models active can terminate the services on ports 8100, 8107, or 8108.

Please mock `_stop_incompatible_stack()` in this selection-only test. The following `test_cli_stops_incompatible_stack_before_starting` test already covers the cleanup behavior explicitly.

wenxind-nvidia:
## Blocker: stack switching continues after cleanup failure

[`_stop_incompatible_stack()`](https://github.com/NVIDIA/xr-ai/blob/73868d619a3b676477361c4ca93188e4453f7c3c/agent-samples/model-servers/main.py#L100-L107) catches cleanup exceptions and returns normally, after which [`run_stack()` proceeds unconditionally](https://github.com/NVIDIA/xr-ai/blob/73868d619a3b676477361c4ca93188e4453f7c3c/agent-samples/model-servers/main.py#L138-L139). If the incompatible model cannot be stopped, the new stack can still start and recreate the GPU-overcommit/OOM condition this change is intended to prevent.

Cleanup is a prerequisite for switching stacks, so startup should abort unless the incompatible services are confirmed stopped. This likely requires propagating a failure result from `stop_persistent_servers()` or rechecking the relevant ports/containers before calling `run_stack()`.

nvddr:
Addressed the two new blockers in `3824075`:

- `test_cli_selects_requested_stack` now mocks `_stop_incompatible_stack`, so it cannot stop real services. Cleanup remains covered separately by the transition test.
- `stop_persistent_servers()` now returns failure when a discovered listener/container cannot be stopped, including unhealthy-but-listening services. `model_servers` aborts before `run_stack()` when the incompatible stack cannot be cleared; regression coverage asserts no replacement stack is started.

Also documented the safe stack-switch behavior. Validation: 16 focused tests passed, Ruff and `git diff --check` passed.

wenxind-nvidia:
## Blocker: stop discovery is neither fail-closed nor ownership-safe

The new [`stop_persistent_servers()` discovery path](https://github.com/NVIDIA/xr-ai/blob/38240750e65d11d5a89be324b04f2b0a6c82515c/utils/xr-ai-vllm/xr_ai_vllm/__init__.py#L166-L173) treats `container_on_port() is None` and `pid_on_port() is None` as "already stopped" and ultimately returns `True`. However, [`container_on_port()` and `pid_on_port()`](https://github.com/NVIDIA/xr-ai/blob/38240750e65d11d5a89be324b04f2b0a6c82515c/utils/xr-ai-vllm/xr_ai_vllm/_docker.py#L531-L579) also return `None` when Docker, `ss`, or `lsof` is missing, inaccessible, or errors. In that case an existing GPU server can remain active while the orchestrator proceeds to launch the replacement stack, recreating the overcommit/OOM condition.

Conversely, when `pid_on_port()` does return a PID, the helper signals it based only on the configured port. Removing the health gate means `--stop` or a stack switch can terminate an unrelated application using ports 8100, 8107, or 8108.

Please distinguish an empty port from discovery failure, verify ownership before signaling, and fail the stack switch when a listener exists but cannot be identified as an xr-ai-managed service. Direct tests should cover discovery-tool failure and an unidentified listener.

nvddr:
Addressed the ownership/discovery blocker in `5058de8`:

- Port inspection now explicitly reports whether it succeeded; a failed `ss`/`lsof` inspection fails closed rather than treating the port as empty.
- Cleanup only proceeds for a labelled `xr-ai-vllm` container or a listener whose command line matches the expected xr-ai server. An unidentified listener is left untouched and causes stack startup to abort.
- Docker inspection failures also block switching when a listener exists; no-listener cases remain safe for pip-only hosts.
- Added direct tests for discovery failure and an unidentified listener, which verify no signal is sent.

Validation: 44 focused tests passed; Ruff and `git diff --check` passed.

wenxind-nvidia:
## Blocker: labelled Docker containers are skipped when no host PID is visible

[`stop_persistent_servers()`](https://github.com/NVIDIA/xr-ai/blob/5058de85332ad4c24a50c0631c47747ef95500cb/utils/xr-ai-vllm/xr_ai_vllm/__init__.py#L170-L179) checks `pid_on_port_checked()` first and immediately continues when it returns `(None, True)`, so `container_on_port_checked()` is never called. A labelled container can be running and consuming GPU without a visible host PID while it is still starting, or when unprivileged `ss` cannot expose the root-owned container PID. Cleanup then returns success and the replacement stack can start, recreating the overcommit/OOM condition.

Please inspect and stop the labelled Docker container before falling back to host PID discovery, matching the order documented by the function. Add a regression test where container discovery returns a labelled container while PID discovery returns no PID.

wenxind-nvidia:
## Blocker: the pip ownership matcher accepts unrelated vLLM processes

[`is_xr_ai_server_process()`](https://github.com/NVIDIA/xr-ai/blob/5058de85332ad4c24a50c0631c47747ef95500cb/utils/xr-ai-vllm/xr_ai_vllm/_docker.py#L602-L610) considers every non-STT command line containing the substring `vllm` to be xr-ai-owned. An independently managed vLLM server on one of the configured ports will therefore pass the ownership check and receive SIGTERM/SIGKILL.

Please use an xr-ai-specific ownership marker, persisted PID metadata, or equivalent identity instead of matching the generic executable name. Add a regression test proving that an external `vllm serve` process is left untouched and causes the stack switch to abort.

wenxind-nvidia (approved):

wenxind-nvidia:
## Suggestions

1. **Update the stale cleanup description.** [`docs/ai-services.md`](https://github.com/NVIDIA/xr-ai/blob/a6b8b8b5f19a9944066e45aac714e64c604cdedc/docs/ai-services.md#L212-L216) and the [source documentation](https://github.com/NVIDIA/xr-ai/blob/a6b8b8b5f19a9944066e45aac714e64c604cdedc/docs/source/components/ai-services.md#L203-L207) still say cleanup probes `/health` before selecting Docker or PID cleanup. The implementation now discovers labelled containers first and uses ownership markers for pip processes, as the later Cleanup section correctly explains. Keeping both descriptions aligned will avoid contradictory operator guidance.

2. **Document the pip upgrade path.** Pip vLLM processes started before this change do not contain `XR_AI_VLLM_MANAGED` and `XR_AI_VLLM_PORT`. After updating, cleanup will correctly refuse to signal those unmarked processes, so affected users may need to stop them manually once before restarting the stack. A short migration note would make this fail-closed behavior easier to diagnose.

PR #340, opened 2026-08-05
fix(vllm): run setup before server entrypoint

nvddr:
## What changed

- Override Docker vLLM image entrypoints with `/bin/bash` so prerequisite wheel installation runs before `vllm serve`.
- Recreate stopped containers instead of restarting an immutable stale entrypoint or command.
- Document the lifecycle behavior and cover both generated Docker arguments and stopped-container relaunch.

## Why

The Omni profile's `vllm/vllm-openai:v0.20.0` image defaults to the `["vllm", "serve"]` entrypoint, which interpreted the shell setup string as a vLLM argument. Nemotron Omni then failed while parsing `--compilation-config` rather than starting the server. Restarting the stopped container preserved that broken command.

## Impact

Model-server convenience stacks can install model-specific prerequisites and reliably recover from a failed container launch.

## Validation

- `PYTHONPATH=utils/xr-ai-vllm uv run --with pytest pytest -q --noconftest tests/test_vllm_docker.py`
- 28 tests passed.
- Focused Ruff and SPDX checks passed.

### DIFF
--- a/docs/ai-services.md
+++ b/docs/ai-services.md
@@ -266,6 +266,9 @@ Existing `~/.docker/config.json` entries take priority and are not overwritten.
   the legacy `--gpus` flag so the launch works under both legacy and CDI
   container-toolkit modes; `--ipc host` gives vLLM the shared-memory region
   its workers expect.
+- The launcher overrides the image entrypoint with `/bin/bash`, installs any
+  required wheels, then executes `vllm serve`. Failed stopped containers are
+  recreated because Docker cannot change their recorded entrypoint or command.
 - The host `model_cache` is bind-mounted at the same path inside the
   container and `HF_HOME` is set to it, so weights cached by pip mode are
   reused by docker mode and vice versa.
--- a/docs/ai-services.md
+++ b/docs/ai-services.md
@@ -246,7 +246,7 @@ another tag, an internal mirror, or a custom build.
 
 - **Docker Engine** with the user in the `docker` group (`docker version`
   must succeed without `sudo`).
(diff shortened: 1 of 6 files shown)


wenxind-nvidia:
## Blocker

None. The change resolves the Omni startup failure and correctly recreates stopped containers.

## Suggestions

1. Correct the image attribution in [_docker.py](https://github.com/NVIDIA/xr-ai/blob/2b60134a801551ae979052b3fb3f6919780b112a/utils/xr-ai-vllm/xr_ai_vllm/_docker.py#L101), the changelog, and PR description. `nvcr.io/nvidia/vllm:26.04-py3` uses `/opt/nvidia/nvidia_entrypoint.sh`, which already forwards `bash -c`. The problematic `["vllm", "serve"]` entrypoint belongs to `vllm/vllm-openai:v0.20.0`, the Omni image. The regression test should use that image or describe the behavior generically.

2. Add a `run()` test covering the [stopped-container branch](https://github.com/NVIDIA/xr-ai/blob/2b60134a801551ae979052b3fb3f6919780b112a/utils/xr-ai-vllm/xr_ai_vllm/_docker.py#L455): verify the stopped container is removed and a fresh `docker run` command is launched. The current new test covers only argument construction.

## Nit

The Docker prerequisite still says the NVIDIA Container Toolkit is needed so `--gpus` works, while the documented and implemented command now uses `--runtime nvidia`.

Validation: 33 focused tests passed, Ruff passed, `git diff --check` passed, and all GitHub checks are green.

wenxind-nvidia (approved):

PR #344, opened 2026-08-10
ci(nightly): increase brev start/stop polling windows

yanziz-nvidia:
## Summary

- Stop job timed out in [run 31355880459](https://github.com/NVIDIA/xr-ai/actions/runs/31355880459): the 20 × 15 s = **5 min** poll was too short for the VM to reach STOPPED
- start: 60 → 80 iterations (15 min → 20 min poll), `timeout-minutes` 20 → 25
- stop:  20 → 40 iterations (5 min → 10 min poll), `timeout-minutes` 15 → 20
- Both jobs retain a ~5 min gap between poll ceiling and job timeout so a loop-exhausted failure is distinguishable from a silent CI cancel

## Test plan

- [ ] Nightly run completes without stop job timing out

### DIFF
--- a/.github/workflows/nightly-xr-ai-test.yml
+++ b/.github/workflows/nightly-xr-ai-test.yml
@@ -52,7 +52,7 @@ jobs:
     # Runs on the always-on controller VM, which already has the Brev CLI
     # installed and is logged in — no install/login/token step needed.
     runs-on: brev-controller
-    timeout-minutes: 20
+    timeout-minutes: 25
     steps:
       - name: Select org
         # The controller is normally already pinned to the right org;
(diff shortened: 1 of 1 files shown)


nvddr (commented):
(Approved with below nit)

One blocking timeout issue remains.

The start job can still be cancelled before its polling loop finishes. In this PR run, `brev ls` made consecutive polls roughly 20.3 seconds apart rather than 15 seconds because the command itself takes about five seconds. At that cadence, 80 iterations require approximately 27 minutes plus setup, but the job has `timeout-minutes: 25`. A sufficiently slow start will therefore hit the job-level cancellation before the loop emits its explicit timeout error, preserving the failure mode this change is intended to avoid.

Please use a wall-clock deadline for the poll or raise the job timeout to roughly 35 minutes if retaining 80 iterations and a five-minute safety margin. The stop job still has sufficient headroom with its 20-minute cap.

nvddr (approved):

PR #345, opened 2026-08-10
Modularize XR render sample and improve eval

blongs-nv:
Modularization/NAT update for xr render sample, including eval harness changes to support testing real flow and individual components for debugging performance in the sample.


agent-samples/xr-render-demo/eval/xr_render_demo_eval/live_explore.py
```diff
@@ -0,0 +1,160 @@
...
+            ok, facts = judge(intent, before, after)
+            verdict = "PASS" if ok else "FAIL"
+            print(f"{verdict} [{index:02d}] ({intent:6s}) {utterance!r} {facts}", flush=True)
+            passed += ok
+            failed += not ok
+        print(f"\nlive explore: {passed} passed, {failed} failed", flush=True)
+    finally:
+        try:
+            await tracking.call("clear_sim_pose", {})
+        except Exception:
```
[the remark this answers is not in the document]
blongs-nv: These except Exception: pass blocks are intentional best-effort cleanup. Clear_sim_pose runs in a finally clause to reset tracking state after each live eval run. If it fails (e.g. the tracking service is already down), swallowing the exception is correct: we don't want cleanup failure to mask the actual test result. Ditto for other comments on this.

agent-sdk/xr-ai-nat/xr_ai_nat/llm/_langchain.py
```diff
@@ -69,6 +71,70 @@ def _tool(tool: BaseTool | dict[str, Any]) -> ToolDef:
...
+        return None
+    candidates = []
+    for tool in offered.values():
+        properties = tool.parameters.get("properties", {})
+        required = set(tool.parameters.get("required", ()))
+        if set(data) <= set(properties) and required <= set(data):
+            candidates.append(tool.name)
+    if len(candidates) != 1:
+        return None
+    return {"name": candidates[0], "args": dict(data), "id": _recovery_id(), "type": "tool_call"}
```
nvddr: This creates a second tool-call parser after the serving parser and NAT, including inferring an executable tool solely from the shape of ordinary assistant JSON. Because this sample performs scene mutations, reinterpreting unstructured content as a call is too risky and also puts a model-serving quirk in `xr-ai-nat` rather than `xr-ai-models` presets. Please remove shape-based recovery and fix the configured server/parser; if normalization is genuinely needed, keep it strictly structured, explicitly named, schema-validated, and owned by `xr-ai-models`.

agent-samples/xr-render-demo/worker/xr_render_demo_worker/supervisor.py
```diff
@@ -0,0 +1,241 @@
...
+        if _is_truncated(request.transcript):
+            return SceneReply(response=_truncated_reply(request.transcript))
+        conversation, pending_truncation = await recent_conversation(request.participant_id)
+        transcript = request.transcript
+        if pending_truncation:
+            resolved = _resolve_truncation_reply(pending_truncation, transcript)
+            if resolved is None:
+                return SceneReply(response="Okay, never mind that.")
+            transcript = resolved
+        ledger.reset()
```
nvddr: `VoiceSession` runs participant turns concurrently, but this ledger is one supervisor-wide instance. A second participant can reset it while the first participant is inside the object agent, despite that agent’s own lock. The appearance guard is similarly shared and unprotected, and the global scene before/after comparison can mistake another participant’s mutation for this turn’s success. Please make these states invocation- or participant-scoped (prefer NAT per-user functions/groups or equivalent request context) and add a concurrent two-participant test.

agent-samples/xr-render-demo/worker/xr_render_demo_worker/supervisor.py
```diff
@@ -0,0 +1,241 @@
...
+            )
+            try:
+                output = await reasoning.ainvoke(verification, to_type=str)
+            except Exception as error:
+                logger.error("supervisor verification pass failed: {}", error)
+                output = "Something went wrong on my side; please say that again."
+        await context.record_moves(request.participant_id, before)
+        return SceneReply(response=str(output or "Done."))
+
+    return LambdaFunction.from_info(
```
nvddr: For a reference NAT sample, the top-level workflow should not bypass the registry and builder lifecycle with private `nat.builder.function.LambdaFunction`. `SceneSupervisorConfig` is currently never registered, so the builder cannot own or inspect this function’s configuration/dependencies consistently. Please register the supervisor with the stable plugin API, add it through `builder.add_function(...)`, and express its dependencies as explicit refs/config rather than a custom factory with hidden function handles.

agent-samples/xr-render-demo/eval/xr_render_demo_eval/live_manip.py
```diff
@@ -0,0 +1,172 @@
...
+                        detail = "; ".join(
+                            [f"{i}:{'GONE' if i not in objects else objects[i].model_dump()}" for i in ids]
+                            + [f"NEW {k}:{objects[k].model_dump()}" for k in new_ids]
+                        )[:400]
+                    verdict = "PASS" if ok else "FAIL"
+                    break
+            print(f"{verdict} {case['name']:22s} {detail}", flush=True)
+            passed += verdict == "PASS"
+            failed += verdict == "FAIL"
+        print(f"\nlive manipulation: {passed} passed, {failed} failed", flush=True)
```
nvddr, line 203: This command counts failures but still returns exit status 0. The same problem exists in `live_garble`, `live_explore`, and the no-change path in `live_smoke`, so these advertised eval commands cannot act as regression gates or reliable agent automation. Please make every eval entry point exit nonzero when any case fails, after cleanup has completed.

agent-samples/xr-render-demo/eval/xr_render_demo_eval/live_pose_matrix.py
```diff
@@ -0,0 +1,139 @@
...
+                await tracking.call("set_sim_pose", p)
+            except Exception as error:
+                print(f"openxr service refused set_sim_pose ({error}); set allow_sim_pose: true in "
+                      "yaml/openxr_service.yaml and restart the stack")
+                raise SystemExit(2) from None
+            for prompt, distance in PROMPT_SETS[pose_name]:
+                await clear_scene(scene)
+                before = {i.id for i in (await scene.get_scene_state(EmptyRequest())).objects}
+                await endpoint.inject_data(DataMessage(
+                    participant_id=participant, topic="live.smoke.text",
```
nvddr: This reuses the single participant created before the loops for every pose/prompt case. That accumulates transcript history across cases, directly violating the eval README’s isolation rule and contaminating later results. Please create a fresh participant per independent case (and emit a leave event afterward), as the other multi-case live drivers attempt to do.

nvddr (changes requested):
The migration is directionally strong: the worker now composes typed NAT function groups, keeps model traffic behind `xr-ai-models`, avoids MCP/LiveKit in the worker, and is much easier to navigate than the previous monolith. I’m requesting changes before we treat it as the repository reference, primarily for the parallel tool-call recovery path, participant-concurrency correctness, NAT lifecycle ownership, and live-eval reliability.

Please also address the documentation gaps before merge: `docs/xr-render-demo.md` says managed processes start concurrently even though the repository contract says they start serially; `agent-samples/xr-render-demo/main.py` still describes the removed voice-loudness/sphere-radius behavior; and the sample needs a colocated README/file map explaining the NAT composition chain and how users add a function group, subagent, prompt, config, and eval case. The branch is currently conflicting with `main`, so these paths and instructions should be revalidated after rebasing.

agent-samples/xr-render-demo/worker/xr_render_demo_worker/app.py
```diff
@@ -0,0 +1,102 @@
...
+        await builder.add_function_group("conversations", ConversationMemoryFunctionsConfig())
+        await builder.add_function_group(
+            "video_memory",
+            VideoMemoryFunctionsConfig(endpoint=config.video_memory_endpoint),
+        )
+        vision = VisionToolsConfig(endpoint=session.transport.endpoint, vlm=vlm)
+        await builder.add_function_group("vision", vision)
+
+        supervisor = await scene_supervisor(builder=builder, llm=llm)
+        handler = as_voice_handler(
```
nvddr: Please route each turn through a built NAT `Workflow`/runner (or add that support to the shared `xr_ai_nat` voice adapter) instead of invoking a bare `Function`. The bare path emits intermediate steps into an unconsumed stream and never establishes a per-turn `workflow_run_id`/trace ID or exporter, so concurrent supervisor, subagent, and tool records cannot be correlated. Bind `participant_id` as request metadata and record terminal outcomes including superseded/cancelled. This is the integration point for NAT tracing; please do not add a render-only event framework.

agent-samples/xr-render-demo/worker/xr_render_demo_worker/supervisor.py
```diff
@@ -0,0 +1,241 @@
...
+    for name, config in subagents:
+        await builder.add_function(name, config)
+
+    reasoning = await builder.add_function(
+        "supervisor_reasoning",
+        ToolCallAgentWorkflowConfig(
+            llm_name=_LLM_NAME,
+            tool_names=[FunctionRef(name) for name, _config in subagents],
+            system_prompt=_PROMPT.read_text(encoding="utf-8").strip(),
+            handle_tool_errors=True,
```
nvddr: `handle_tool_errors=True` makes LangGraph convert tool exceptions into `ToolMessage`s and continue. With detailed logs off and no IntermediateStep subscriber/exporter in this sample, handled validation/RPC failures can disappear into model scratch and leave only the final reply. Keep recovery only for expected, user-correctable errors and ensure the NAT trace records every tool start/result/failure with run/participant ID, agent, tool, attempt, duration, sanitized arguments/result, and exception classification. Apply this once through shared NAT observability/middleware rather than wrapping each tool.

agent-samples/xr-render-demo/worker/xr_render_demo_worker/supervisor.py
```diff
@@ -0,0 +1,241 @@
...
+            f"Utterance timestamp: {request.timestamp_us}\n"
+            f"{await context.describe(request.participant_id)}\n\n"
+            f"{conversation}"
+            f"User request: {transcript}"
+        )
+        # A failed reasoning pass must degrade, not raise: the verification
+        # pass below then gets a chance to complete the turn.
+        try:
+            output = await reasoning.ainvoke(message, to_type=str)
+        except Exception as error:
```
nvddr: This catch collapses model, service, schema, and programming failures into the same user reply, and `logger.error(..., error)` preserves no traceback. Catch only a documented recoverable exception; for unexpected failures emit the correlated terminal error, use `logger.exception` (or equivalent structured exception capture), and re-raise so the runtime remains fail-fast. Running the verification pass after an unknown failure is also unsafe because the first pass may already have mutated the scene.

agent-samples/xr-render-demo/worker/xr_render_demo_worker/agents/vision/agent.py
```diff
@@ -0,0 +1,82 @@
...
+            f"Active participant: {request.participant_id}\n"
+            f"Utterance timestamp: {request.reference_time_us}\n"
+            f"{scene_block}"
+            f"Focused instruction: {request.instruction}"
+        )
+        # A dead camera is an answer, not a workflow failure: retry once
+        # against scene data alone, then degrade explicitly.
+        try:
+            output = await reasoning.ainvoke(message, to_type=str)
+        except Exception as error:
```
nvddr: This assumes every exception means ‘camera unavailable,’ then the second broad catch discards that failure entirely. A VLM timeout, invalid tool schema, service/config error, or code defect is therefore mislabeled and untraceable, and the first exception text is injected into the retry prompt. Catch only the typed no-frame/unavailable condition, emit a correlated retry record, and retain both attempt outcomes; unexpected exceptions should keep their traceback and propagate.

docs/xr-render-demo.md
```diff
@@ -145,164 +130,84 @@ OOM race.
...
+reply to Piper and returns the audio to the participant.
+
+## Voice runtime
+
+The worker runs on `xr-ai-voice`'s `VoiceSession`: hub audio → VAD → STT →
+voice gate → handler → streaming TTS → return audio. The scene supervisor is
+attached with `as_voice_handler`, and `record_voice_transcripts` persists
+each completed turn into native text memory, which `recall_conversation`
+reads back as the `[Recent conversation]` block. A new utterance while TTS
+is playing supersedes the previous turn.
```
nvddr: As a reference sample, please add a concise trace/debug contract here: where the durable trace is written, how one ID links voice query → supervisor → subagent → leaf NAT tool → verification/cancellation, which inputs/results are redacted, and the exact command to inspect one turn or tool failure. Document the recoverable-vs-fatal error and retry policy too. The tea sample’s event table and human-test loop are useful inspiration; a separate viewer is not required.

nvddr (commented):
Traceability follow-up after comparing the tea-making sample. Tea’s useful invariant is one correlated record per turn, agent, tool call, retry, and failure; its sample-local emitter/viewer should not become a second observability framework here. NAT already emits IntermediateStep FUNCTION/LLM/TOOL events and its Workflow runner supplies workflow_run_id/trace_id. Please wire the render sample through that supported trace/exporter path, adding only XR-specific context such as participant, supersession/cancellation, and verified scene diff.

Please also add a regression that injects a failing leaf function and proves the durable trace identifies the turn, participant, subagent, tool, sanitized arguments, exception type/message, retry decision, and terminal outcome.

yanziz-nvidia (commented):

## Summary

Replaces the single-agent pipecat pipeline with a supervisor plus five focused subagents over `xr-ai-voice`/NAT, and promotes the flat `eval.py` script into a four-tier `xr-render-demo-eval` package. All 14 CI checks green; `DEPENDENCIES.md` and `AGENTS.md` updated alongside the `pyproject.toml` changes.

|     | Finding |
|-----|---------|
| 🚫 | None |
| 💡 | [`xr_render_demo_eval/harness.py:918-927`](https://github.com/NVIDIA/xr-ai/blob/dec6b4264c1f02a35ff46f267483c5b8788df92b/agent-samples/xr-render-demo/eval/xr_render_demo_eval/harness.py#L918-L927) — **`audit_prompts` guards both sibling imports with one `try/except ImportError`, so a failure in either tier silently skips the `else` block for both.** The audit then passes while covering none of the subagent/supervisor cases. |
| 🔍 | [`xr_render_demo_eval/harness.py:595`](https://github.com/NVIDIA/xr-ai/blob/dec6b4264c1f02a35ff46f267483c5b8788df92b/agent-samples/xr-render-demo/eval/xr_render_demo_eval/harness.py#L595) — **`_NO_MUTATION` is byte-identical to `_MUTATING` at [line 446](https://github.com/NVIDIA/xr-ai/blob/dec6b4264c1f02a35ff46f267483c5b8788df92b/agent-samples/xr-render-demo/eval/xr_render_demo_eval/harness.py#L446).** Two names for one set will drift if either is edited. |

- `agent-samples/xr-render-demo/eval/xr_render_demo_eval/harness.py:918` — Split the `try/except ImportError` in `audit_prompts` into one block per sibling module so a missing dep in one tier doesn't silently drop the other tier's cases from the audit.
- `agent-samples/xr-render-demo/eval/xr_render_demo_eval/harness.py:595` — Either alias `_NO_MUTATION = _MUTATING` or drop one name; the duplicated literal is a drift hazard.

wenxind-nvidia:
Reviewed the latest PR head `db2ef132` against `main` at `01236766`. Recommendation: **request changes**.

## Blockers

1. **Unresolved merge conflict in `AGENTS.md`**

   `AGENTS.md:67-233` contains literal `<<<<<<< HEAD`, `=======`, and `>>>>>>>` markers. `git diff --check` fails on these markers, plus trailing blank lines in two files. Resolve the merge before proceeding.

2. **Conversation history is read but never written**

   `supervisor.py:134-154` recalls participant history, but production code never invokes `TextMemoryTools.add_transcript`. The eval substitutes pre-seeded fake history, masking the problem. On a fresh live worker, corrections, references to earlier turns, and truncated-command completion have no persisted context.

   Persist successful user/agent turns under the participant-scoped keys expected by `recall_conversation`, and add a real two-turn integration test without pre-seeded memory.

3. **Cross-participant mutations can satisfy the wrong turn**

   `supervisor.py:160-217` uses per-participant locks around a global scene. Participant A can take its `before` snapshot, participant B can mutate the scene, and A's global before/after diff then observes B's mutation. A can consequently skip its verification pass and record B's movement in A's undo history.

   Participant-scoped delegation flags do not attribute global scene changes. Serialize the snapshot/mutation/verification window globally, or derive success and move history from turn-local tool results. Add a forced two-participant interleaving test.

4. **The PR introduces an unrelated Nemotron 3 Nano service regression**

   The default XR-render profile correctly uses `nemotron_omni` at port 8108; nevertheless, this PR also changes the standalone `services/nemotron3-nano-llm/nemotron3_nano_llm_server/__main__.py:124` service from `qwen3_coder` to `step3p5`, an XML parser intended for Step-3.5. NVIDIA's model card explicitly launches Nemotron 3 Nano with `qwen3_coder`, as do the service README and every other in-repo configuration. This can prevent that service's model output from becoming structured tool calls. Drop this unrelated change or restore `qwen3_coder` unless there is model-specific evidence and validation supporting another parser: https://huggingface.co/nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-FP8

5. **Unexpected failures are converted into successful turns**

   `_loop.py:52-74` catches every tool exception and feeds it back to the model. `agent.py:108-120` then catches every supervisor exception without re-raising or publishing a failure response. RPC/service failures can therefore be presented as model reasoning, the runtime delivery is recorded as successful, and the participant may receive silence.

   Use the shared `xr_ai_tools.tool_calling.run_tool_loop`, propagate unexpected exceptions, and explicitly classify only recoverable domain failures. Add tests for an unexpected leaf failure and a supervisor failure.

6. **The production sample enables the global simulated-pose hook**

   `yaml/openxr_service.yaml:10` ships `allow_sim_pose: true`, despite the adjacent comment and eval README saying it must be off by default. The service listens on `0.0.0.0`, and any RPC peer can replace tracking globally until it is cleared. Ship `false`; require an explicit eval-only override.

7. **Documentation and tests still describe the deleted implementation**

   `docs/source/reference/xr-render-demo.md:172-307` documents `SceneModelLoop`, quick acknowledgements, still-working messages, `NativeCapabilities`, prefetching, and a deleted `prompts/system.txt`. Meanwhile `test_xr_render_demo_wire.py` continues testing those obsolete paths, and unused `tools.py`/`spatial_tools.py` remain in the package.

   The eval README is stale too: it says the default agent LLM is port 8107, while the selected `models.local.json` profile uses Nemotron Omni at port 8108. Replace the reference section with the supervisor/five-subagent architecture, remove dead modules and golden tests, and test the code actually composed by `app.py`.

## Suggestions

- Propagate `trace_id` into `SubagentTask` and tool/subagent logs. It currently appears only at supervisor entry, so the README's suggested `grep "trace=<id>"` cannot trace a complete turn.
- Narrow broad catches in `SceneContext.user_frame()` and `spatial_ops._Leaves.color()`. A tracking or scene RPC outage should not be silently reclassified as missing bearings or an unknown color.
- Remove `mark_mutating` if it remains unused: `supervisor.py:199` drains and discards it.
- Fix the sample README command that passes `--config` to the orchestrator; `main.run()` does not parse that option.

## Nits

- Consolidate duplicated eval mutation constants such as `_MUTATING`.
- Remove the extra EOF blank lines reported by `git diff --check`.
- Update the sample file map after removing the legacy modules.

Most CI checks pass on the latest merge; the Python 3.11 and 3.12 jobs were still running when reviewed.

AGENTS.md
```diff
@@ -64,6 +64,7 @@ The authoritative dependency graph and enforced package limits are in
 Each sample contains a stdlib-style orchestrator and a separately packaged
 worker. Optional application-specific capability processes sit beside them.
 
+<<<<<<< HEAD
```
nvddr: This merge conflict is committed in the current head. The block contains two contradictory repository contracts, including stale `xr_media_hub` / `docs/process-model.md` guidance, so agents can follow the wrong architecture even though the Markdown build passes. Please resolve the entire `<<<<<<<` / `=======` / `>>>>>>>` region before merge and add a conflict-marker check to CI.

agent-samples/xr-render-demo/worker/xr_render_demo_worker/_loop.py
```diff
@@ -0,0 +1,82 @@
...
+    max_iterations: int = _MAX_ITERATIONS,
+    max_tokens: int = 2048,
+) -> str:
+    """Run an agentic loop: call LLM, dispatch tool calls, repeat until text.
+
+    Returns the final text response, or an apology if the iteration cap is hit.
+    Tool exceptions are converted to error tool-result messages so the model can
+    reason about and degrade gracefully from failures.
+    """
+    for _ in range(max_iterations):
```
nvddr: Please remove this sample-local loop and use `xr_ai_tools.tool_calling.run_tool_loop`, as the tea-making and lab-monitoring samples do. This implementation ignores `finish_reason`, so it can execute tool calls from a truncated response; it also omits the shared malformed/duplicate-ID/direct-return checks, converts unexpected tool failures into model-visible recovery, and returns a normal apology at the iteration cap. Those differences are unsafe for scene-mutating tools and duplicate an existing shared abstraction.

agent-sdk/xr-ai-hub/xr_ai_hub/_processor.py
```diff
@@ -418,6 +418,14 @@ async def send_return_audio(self, chunk: AudioChunk) -> None:
         """Queue a PCM audio chunk for playback by its target participant."""
         await self._push.send(encode(MsgType.RETURN_AUDIO, chunk))
 
+    async def inject_participant_event(self, event: ParticipantEvent) -> None:
```
nvddr: Please do not broaden the worker-facing `ProcessorEndpoint` API solely for sample evals. These methods let any worker synthesize connector-owned upstream data and participant lifecycle; spoofing an existing participant join can overwrite the hub's participant-to-connector mapping and disrupt return routing. Keep synthetic participants in a private connector-side test/eval fixture instead of making upstream injection a supported agent API.

nvddr (commented):
Follow-up review of the current head after comparing the render evals with the tea-making and lab-monitoring examples. The prompt-behavior cases and deterministic checker/spatial tests belong in this PR because the prompts and agent decomposition changed. I would not add a repository-wide eval API here. If the live drivers remain, first put their repeated lifecycle in a private, connector-owned synthetic-participant helper; then use a dedicated follow-up to extract an internal cross-sample core for declarative case loading, strict tool/schema/order matching, one model lifecycle per suite, fail-closed results, prompt-fixture auditing, and standardized summaries/exit codes. Spatial predicates, tea workflow transitions, and visual fixtures should remain sample-specific adapters.

agent-samples/xr-render-demo/worker/xr_render_demo_worker/supervisor.py
```diff
@@ -0,0 +1,221 @@
...
+            "[Recent conversation] (already handled; never a source of new work)\n"
+            + "\n".join(lines) + "\n\n"
+        )
+        return block, pending
+
+    async def handle(self, request: SceneRequest) -> SceneReply:
+        if _is_truncated(request.transcript):
+            return SceneReply(response=_truncated_reply(request.transcript))
+
+        async with self._participant_locks[request.participant_id]:
```
nvddr, line 196: The earlier concurrency issue remains: a lock keyed by participant only serializes two turns from the same participant. Alice and Bob still run concurrently while lines 178/204/217 compare and record global scene snapshots, so Bob's mutation can satisfy Alice's verification or become Alice's recent move. Please use a scene-wide transaction/lock or a request-correlated mutation journal, and add a two-participant test that overlaps distinct mutations.

agent-samples/xr-render-demo/eval/xr_render_demo_eval/harness.py
```diff
@@ -0,0 +1,1020 @@
...
+        try:
+            reply = await supervisor.handle(
+                SceneRequest(
+                    transcript=case["user"],
+                    participant_id=_PARTICIPANT,
+                    timestamp_us=10_000_000,
+                )
+            )
+            response = reply.response
+        except Exception as exc:
```
nvddr, line 621: This makes the eval fail open. If the workflow performs the expected mutation and then the model/service raises, the exception is converted to display text and `check_corpus(scene.calls, case)` can still report PASS. Unexpected workflow errors must make the case fail; the same pattern in precision, supervisor, and subagent tiers should be fixed and regression-tested.

agent-samples/xr-render-demo/eval/xr_render_demo_eval/harness.py
```diff
@@ -0,0 +1,1020 @@
...
+            if _match_call(name, args, expect, scene):
+                remaining.pop(index)
+                break
+        else:
+            unmatched.append(expect)
+    if unmatched:
+        wanted_desc = "; ".join(f"{expect['tool']}({expect.get('args', {})})" for expect in unmatched)
+        actual = [f"{name}({args})" for name, args in mutations]
+        return False, f"unmatched: {wanted_desc} | actual mutations: {actual} | calls: {names}"
+    if not case.get("ignore_extra", True) and remaining:
```
nvddr: Defaulting `ignore_extra` to true allows a correct expected call plus an unintended update/removal of another object to pass. For a scene-mutating regression gate, unmatched mutations should fail by default, with an explicit opt-in only for cases that intentionally permit additional effects. The subagent scorer should follow the same strict default.

agent-samples/xr-render-demo/worker/xr_render_demo_worker/agent.py
```diff
@@ -6,293 +6,138 @@
...
+                    text=reply.response,
+                    response_id=response_id,
+                    final=False,
+                    timestamp_us=query.timestamp_us,
+                ),
+            )
+            opened = True
+        except asyncio.CancelledError:
+            raise
+        except Exception as error:
```
nvddr: A supervisor/model/service failure is logged and then suppressed without publishing any `VoiceOutput` or terminal event, so the user receives silence and the task appears successful. Please either publish a stable failure response followed by its terminator, as the other voice samples do, or propagate according to the runtime's fail-fast contract; add a failing-supervisor test.

tests/test_service_layout.py
```diff
@@ -64,6 +64,7 @@
 _ALLOWED_LEGACY_REFERENCES = {
     Path("tests/test_service_layout.py"),
     Path("docs/source/reference/migrations.md"),
+    Path("docs/changelog.md"),
```
nvddr: Please do not whitelist the resurrected legacy `docs/changelog.md`. User-facing documentation now lives under `docs/source/`; this 2,674-line root document is merge debris from the old branch. Remove the added document and this exception rather than weakening the legacy-reference contract.

agent-samples/xr-render-demo/eval/xr_render_demo_eval/live_manip.py
```diff
@@ -0,0 +1,221 @@
...
+        raise SystemExit(2) from None
+
+    try:
+        wanted = set(sys.argv[1:])
+        passed = failed = 0
+        for index, case in enumerate(CASES):
+            if wanted and case["name"] not in wanted:
+                continue
+            participant = f"live-manip-{int(time.time())}-{index}"
+            await endpoint.inject_participant_event(ParticipantEvent(
```
nvddr: Each case injects a participant join but never emits the matching leave; `live_garble`, `live_explore`, and `live_smoke` have the same leak. Repeated eval runs therefore retain participant-scoped voice/history/frame state. Please centralize join/send/leave plus endpoint stop/close and scene cleanup in a private async live-eval context manager; `live_pose_matrix` already demonstrates the required leave event.

wenxind-nvidia:
## Blockers

1. **Scene mutation failures are reported as successes.**

   `spatial_ops.py:129` discards `MutationResult.ok`, `reason`, and `new_id` across update, remove, move, add, resize, recolor, and shape operations. The scene service deliberately returns failures for `not_found`, `not_started`, and `backpressure`; shape changes also replace the object and return a new ID at `scene/xr_render_scene/service.py:116`. I reproduced both regressions: a rejected move returned successful coordinates, and `cone-0 -> box` returned the deleted `cone-0` instead of `box-0`. Validate every mutation result and propagate `new_id`. The eval fake at `eval/xr_render_demo_eval/harness.py:442` should mirror production ID replacement.

2. **An expected XR startup RPC failure terminates the worker.**

   `xr_session.py:44` lets `start_xr` and health-poll exceptions escape and ignores an error status from `start_xr`. `ProcessorEndpoint` explicitly treats callback exceptions as fatal and calls `os._exit(1)` at `agent-sdk/xr-ai-hub/xr_ai_hub/_processor.py:735`. Restore bounded handling for expected scene RPC/startup failures and notify the participant when launch fails. Add tests for RPC exceptions, an error result, and `spawn_error`.

3. **Participant departure no longer clears supervisor state.**

   `app.py:70` releases only the current-frame cache. The locks in `supervisor.py:132` and recent-move/delegation state in `scene.py:20` are never removed. Long-running workers accumulate participant IDs, and a reconnect using the same ID can inherit stale undo context. Add supervisor/context cleanup and invoke it after canceling the participant's turn.

4. **Quoted `false` enables the global simulated-pose override.**

   `services/openxr-service/openxr_service/__main__.py:39` uses `bool(raw_value)`, so `allow_sim_pose: "false"` evaluates to `True`. Since the service binds `0.0.0.0` and any RPC peer can then override tracking, parse only real YAML booleans or use strict recognized-string parsing. I reproduced this returning `True`.

5. **The documentation still describes deleted architecture.**

   `docs/source/reference/xr-render-demo.md:104` still documents quick-ack/still-working LLM calls; line 244 lists removed `place_*`, `displace_*`, and `between_anchors` tools; and line 363 claims different participants run concurrently despite the new global scene lock. The sample file map also assigns capabilities to the wrong subagents at `agent-samples/xr-render-demo/README.md:28`. Complete the documentation migration in this PR.

## Suggestions

1. `agent.py:98` opens an incremental voice stream for a single complete string and then sends an empty terminator. Publish one `final=True` response instead. The current `opened=True` assignment occurs after an awaited publish, leaving a cancellation window where an opened response may not receive its terminator.

2. `supervisor.py:235` runs the verification LLM pass for nearly every non-mutating response unless it ends in `?`. Vision, memory, greetings, and ordinary informational answers therefore incur a redundant second LLM call. Base verification on actual mutation intent/result rather than punctuation and delegation.

## Nits

`git diff --check` reports trailing blank lines at EOF in `eval/xr_render_demo_eval/cases.py` and `tests/test_xr_render_demo_wire.py`.

Verification: the focused suite passes with `55 passed`; the mutation and YAML regressions above were reproduced separately because the current fakes do not cover them.

wenxind-nvidia:
Reviewed current PR head `d12fb47` against `main`.

## Blockers

1. **Renderer delivery failures are now hidden rather than resolved.**

   `agent-samples/xr-render-demo/scene/xr_render_scene/service.py:82` ignores `forward()` results and always returns success. However, `agent-samples/xr-render-demo/scene/xr_render_scene/engine.py:366` can still drop operations for `backpressure` or `not_started`. LOVR never reconciles these drops until it restarts, while `get_scene_state` immediately reflects the internal mutation. A dropped add/update/remove therefore leaves the model's verified state different from what the user sees. I reproduced `forward() -> backpressure` producing `{"id": "sphere-0", "ok": true}`. Make renderer delivery reliable or schedule reconciliation after a drop; do not report synchronized success without either.

2. **`render.failed` does not actually notify the user.**

   `agent-samples/xr-render-demo/worker/xr_render_demo_worker/xr_session.py:69` sends a new empty `render.failed` data message, but no web, iOS/visionOS, or native client handles that topic. The web handler at `client-samples/web-xr/App/app.js:104` recognizes only `render.ready`. The worker no longer crashes, but the participant still receives no actionable failure notice. Restore a voice/runtime notice or implement the topic across the supported clients with an error payload.

3. **The verification optimization misses mixed vision/memory plus mutation requests.**

   `agent-samples/xr-render-demo/worker/xr_render_demo_worker/supervisor.py:250` verifies mutation intent only when no subagent was delegated. For "look at the room and create a sphere," if the supervisor delegates only the vision agent, `delegated` is nonempty, contains no mutating agent, and suppresses verification despite `_wants_mutation()` being true. Use mutation intent independently of non-mutating delegation, such as `bool(delegated & _MUTATING_AGENTS) or _wants_mutation(transcript)`, and cover a mixed request.

4. **The documentation migration remains incomplete.**

   `docs/source/reference/xr-render-demo.md:158` still mentions the removed quick acknowledgment; lines 168-171 describe the deleted response generator/stream behavior; lines 185-186 incorrectly claim lifecycle failures use a runtime notice; lines 201-203 describe the old unconditional verification behavior; and lines 207-212 claim chunked output and `mark_delegated` based mutation detection. These sections need to match the current implementation.

## Suggestions

1. Remove the now-dead delegation-marker plumbing. `mark_delegated()` is still called by subagents, but `supervisor.py:218` only discards the prior value and never reads the current turn's marker.

2. Add configuration tests for quoted and invalid `allow_sim_pose` values. The strict parser itself is correct, but this security-sensitive behavior currently has no regression test.

## Resolved from the previous review

Shape-change IDs, participant cleanup, strict boolean parsing, finite voice output, callback exception containment, and the original EOF nits are resolved.

Verification: `59` focused worker tests and `7` scene tests pass.

nvddr (commented):
Updated after comparing each finding against merge base 95d8ece: the earlier renderer/session-latch issues already existed on main, and the verification paths do not make base behavior worse, so I am removing them from this review. These are the blocking regressions introduced by this PR:

1. The base implementation kept participant_id and the utterance reference time out of the model-facing perception schemas and injected them from RuntimeContext before native tool execution. This PR exposes both fields in SubagentTask and asks the supervisor LLM to copy them. Every subagent then trusts those model-generated values; the vision agent uses them to select current and historical frames, and the memory path can similarly target participant-scoped data. A missing, stale, or hallucinated ID can read the wrong participants data, while an omitted timestamp defaults historical lookup to zero. Please bind participant_id/reference_time_us in a private per-turn adapter and expose only the focused instruction to the model.

2. RenderAgent previously created its detached turn task with context=nemo_relay.fork_asyncio_context(), and the tea-making foreground follows the same pattern. The new create_task call removed that context. Because the runtime delivery already runs inside a Relay scope, the detached task otherwise inherits and mutates the subscribers scope stack after the subscriber returns. That can corrupt concurrent trace/sanitizer lifecycles and produce stale-scope failures. Please restore the isolated Relay child context and add a concurrent-turn scope regression test.

3. Worker readiness previously ran an actual LLM chat through _probe_warmed_llm and retried readiness on failure before allowing XR startup; its stated purpose was to settle model GPU memory before LOVR creates Vulkan resources. The new VoiceAgent probes only agent_llm.health, so the first real supervisor request is cold and model initialization can again overlap the renderer. Please retain an inference warmup appropriate to the new supervisor/subagent request shape rather than reducing readiness to endpoint health.

The overall run_stack, VoiceAgent/AgentRuntime, and sample-local typed-tool composition is otherwise aligned with the tea-making sample. These fixes can remain private to the sample and do not require any public API changes.

nvddr (approved):

agent-samples/xr-render-demo/worker/xr_render_demo_worker/agents/memory/agent.py
```diff
@@ -0,0 +1,51 @@
...
+from ...models import SubagentResult, SubagentTask
+
+_PROMPT = Path(__file__).with_name("prompt.txt")
+DESCRIPTION = "Recall earlier conversation turns; never mutates the XR scene."
+
+
+_prompt_text = _PROMPT.read_text(encoding="utf-8").strip()
+
+def make_memory_agent(llm: LLMService, text_memory: TextMemoryTools) -> Tool:
+    recall_tool = text_memory.recall_conversation
```
wenxind-nvidia: Suggestion: The new turn-bound participant identity can still be bypassed in this subagent. recall_conversation is exposed directly, and its model-visible request schema accepts an arbitrary participant_id. An inner model call can therefore recall another participant's persisted conversation instead of the participant bound by the runtime. Please wrap this tool with a request model that omits participant_id and inject current_participant_id in the handler.

client-samples/web-xr/App/app.js
```diff
@@ -106,6 +106,14 @@ function _onDataReceived(topic, _data) {
   if (topic === 'render.ready') {
     return true;  // informational — don't clutter the message list
   }
+  // render.failed: the agent's renderer did not start; payload is the reason.
+  if (topic === 'render.failed') {
+    const detail = new TextDecoder().decode(_data);
+    model.xrError = `Agent renderer failed to start${detail ? `: ${detail}` : ''}`;
+    model.xrState = 'error';
```
wenxind-nvidia: Suggestion: This changes only the UI model to error; the underlying CloudXRStream still owns an active _xrSession. The button then offers Launch XR, but startXR() immediately fails with "XR session already running", and the user no longer has a stop action. Please tear down the active CloudXR session before exposing retry, or track renderer failure separately from the CloudXR connection state.

wenxind-nvidia (approved):

PR #351, opened 2026-08-12
refactor(voice): unify runtime input and output

nvddr:
## Summary

- add one bidirectional `VoiceAgent` that owns `VoiceSession`, publishes accepted `UserQuery` plus voice lifecycle events to injected typed topics, and subscribes to `voice.output`
- keep media readiness, VAD/STT/TTS, wake gating, interruption, transport, and cleanup private to the owned session
- support finite and correlated incremental voice output without exposing Pipecat or a general runtime stream primitive
- migrate `simple-vlm-example` so `SimpleVlmAgent` owns its vision tool, streamed turns, cancellation, and participant frame cleanup while `app.py` only composes agents
- delete the superseded NAT voice adapter and response-returning handler API

## Runtime boundary

The runtime provides typed publish/subscribe routing only. `VoiceAgent.run(runtime)` owns the voice session lifecycle, and application agents own their own resources and tasks. `VoiceAgent` owns the `UserQuery`, `VoiceParticipantLeft`, and `VoiceInterrupted` schemas while each sample chooses their topic names.

Participant departure is published with participant scope. Interruption can be participant-scoped or global. Application agents subscribe directly and clean up their own tasks, state, and tools; the composition root installs no transport callbacks and contains no resource logic.

Voice producers publish either a complete `voice.output` message or correlated incremental chunks. Incremental identity is scoped by participant, producer, and response ID, so independent agents cannot merge output accidentally. `VoiceAgent` owns the lock protecting response aggregation and FIFO state. Runtime publication waits for subscriber delivery, so no duplicate runtime RPC or `SPEAK` API is needed.

For simple VLM, `SimpleVlmAgent` owns the injected `StreamingVisionTool`, participant-scoped background tasks, cancellation, frame release, and nested stream cleanup. `app.py` constructs and registers the agents and supplies the sample-specific topic names only.

## Review feedback addressed

- partial chunks survive iterator failures and remain in the final data response
- response stream keys include producer identity
- cancelled response streams release blocked publishers and close their source iterator
- originating query timestamps propagate to TTS output
- readiness-failure cleanup is idempotent
- global cancellation is modeled without a synthetic participant
- external response FIFO is preserved while queued work exists
- orphan and ambiguous empty stream terminators are rejected
- removed duplicate task error reporting and stale process-model documentation
- removed the compatibility adapter, handler-return path, observer hooks, query queue mode, and separate simple-VLM voice agent

## Validation

- runtime, voice runtime, simple VLM, voice pipeline/session, and Pipecat bridge regressions: 192 passed
- stream-cleanup regressions rerun after conflict resolution: 2 passed
- Ruff, SPDX, and DCO pre-commit hooks passed

Rebased directly onto `main` after #350 merged and squashed to one commit, `ef05580`.


wenxind-nvidia:
## Blockers

1. [`agent-sdk/xr-ai-voice/xr_ai_voice/_processors/handler.py:348-359`](https://github.com/NVIDIA/xr-ai/blob/765f37112bb64eede4660278bb1da298da08671f/agent-sdk/xr-ai-voice/xr_ai_voice/_processors/handler.py#L348-L359) keeps accumulated chunks inside `_consume_response()`. If the iterator raises after yielding, callers retain `response = ""` and finalize an empty agent turn. I reproduced both query and external paths emitting a `TextFrame("partial output " )` followed by `AssistantResponseEndFrame(text="")`. The user hears partial speech, but observation and data echo lose it. This also regresses the existing query path, whose accumulator previously survived iterator errors. Please preserve accumulated text across failures and add coverage for an iterator that raises after yielding.

2. [`agent-sdk/xr-ai-nat/xr_ai_nat/events/voice.py:160`](https://github.com/NVIDIA/xr-ai/blob/765f37112bb64eede4660278bb1da298da08671f/agent-sdk/xr-ai-nat/xr_ai_nat/events/voice.py#L160) keys active streams only by `(participant_id, response_id)`. Different agents can reuse local identifiers, causing their output to merge. I reproduced two producers using `"answer-1"` producing the stream `A1, B1, A2`, with `B2` treated as a separate finite response. Please include the producer or another globally scoped identity in the key and test concurrent producers.

All 103 focused tests pass locally, and the current GitHub checks are green. PR 351 remains stacked on the unchanged PR 350 head, so PR 350's posted blockers also remain transitively relevant.

yanziz-nvidia (commented):

## Summary

Adds `VoiceSession.enqueue_response(...)`, a typed `VoiceOutput` mailbox payload, and a `VoiceOutputWorker` subscriber that routes complete and streaming responses through the existing participant-aware TTS/data-echo path without coupling producers to Pipecat. CI is all green (pytest 3.11 + 3.12, ruff, CodeQL, lock-check, SPDX, DCO); no dependency changes.

|     | Finding |
|-----|---------|
| 🚫 | None |
| 💡 | [`xr_ai_voice/_processors/handler.py:239`](https://github.com/NVIDIA/xr-ai/blob/765f37112bb64eede4660278bb1da298da08671f/agent-sdk/xr-ai-voice/xr_ai_voice/_processors/handler.py#L239) — **`_spawn_response` can jump the per-participant queue when the inflight task is done but `_start_next` has not yet run.** `_finish_response` awaits a frame push before popping `_inflight`; a caller landing in that window sees `is_active=False` and starts immediately, ahead of items already in `_queued[pid]`. |
| 🔍 | [`xr_ai_nat/events/voice.py:40`](https://github.com/NVIDIA/xr-ai/blob/765f37112bb64eede4660278bb1da298da08671f/agent-sdk/xr-ai-nat/xr_ai_nat/events/voice.py#L40) — **Empty-text final chunks with `interrupt=True` are silently dropped.** The empty-text guard fires before the interrupt path, so the cancel is a no-op with no error. |

- `agent-sdk/xr-ai-voice/xr_ai_voice/_processors/handler.py:239` — In `_spawn_response`, guard the direct `_start_response` call with `if not self._queued.get(pid):` and otherwise append to the queue, preserving FIFO across the `_finish_response` await gap.
- `agent-sdk/xr-ai-nat/xr_ai_nat/events/voice.py:40` — In `VoiceOutput.validate_boundary`, reject `final=True` + non-None `response_id` + empty `text`, or document that `interrupt` has no effect on empty-text terminators.

agent-sdk/xr-ai-nat/xr_ai_nat/events/voice.py
```diff
@@ -1,22 +1,185 @@
...
+                event.participant_id,
+                output.text,
+                interrupt=output.interrupt,
+                pts_us=event.timestamp_us,
+            )
+            return
+
+        key = (event.participant_id, output.response_id)
+        stream = self._streams.get(key)
+        if stream is None and output.final:
```
yanziz-nvidia: nit: could nest under if stream is None below

yanziz-nvidia (approved):

nvddr:
Updated in `92764ae` on the rewritten #350. Wenxin's concerns are covered by retaining the accumulator outside `_consume_response()` (including iterator failures) and keying incremental output by participant, producer, and response ID. Yanzi's FIFO suggestion now queues whenever prior queued work exists, with a regression test; empty interrupting terminators are rejected. The former NAT event worker is gone: `VoiceOutputAgent` is an ordinary runtime call target and `voice.output` subscriber, while incremental aggregation stays private to voice/TTS. The focused combined suite passed (106 tests plus the final FIFO regression), targeted Pyright is clean, and both SDK wheels build. Please re-review the new head.

nvddr:
Update: `simple-vlm-example` now uses the complete voice/runtime path in `2d277f7`: typed and spoken input publish one query topic, a resident agent owns the streaming VLM task, and `VoiceOutputAgent` alone bridges `voice.output` to TTS. Participant/global interruption cancels runtime-owned work. The focused combined suite passes 113 tests. @wenxind-nvidia @blongs-nv @yanziz-nvidia, please re-review the updated head.

nvddr:
Updated in `960d0a1`: the public `VoiceOutputAgent` / `as_voice_publisher` split is gone. One bidirectional `VoiceAgent` now owns `VoiceSession`, publishes canonical `voice.input`, and consumes `voice.output`; input-only session callbacks no longer fabricate an empty assistant response. The simple VLM sample uses this boundary end to end. The focused suite passes 120 tests, and all current GitHub checks are green. @wenxind-nvidia @blongs-nv @yanziz-nvidia, please re-review the new head.

nvddr:
Simplified in `99edbad`: there is now one `VoiceAgent`, one `SimpleVlmAgent`, and no public NAT voice adapter or response-returning handler path. I also removed the observer/queue compatibility surface and the separate simple-VLM voice agent; finite and incremental output still share the participant-aware TTS path. The refreshed focused suite passes 114 tests, targeted Pyright is clean, both relevant wheels build, and lock resolution is unchanged. @wenxind-nvidia @blongs-nv @yanziz-nvidia, please review the smaller current head.

nvddr:
Updated in `13b98ea`: `VoiceAgent` no longer owns a fixed `voice.input` protocol. It now requires an injected `Topic[UserQuery]`; simple VLM defines `simple-vlm.user-query`, and any other input path can publish the same transport-neutral message. `VoiceInput`, `VOICE_INPUT_TOPIC`, and the unused application-level `fresh_match` field were removed rather than aliased. The focused suite remains 114 passing with clean Pyright, builds, and lock resolution. @wenxind-nvidia @blongs-nv @yanziz-nvidia, please review the new head.

wenxind-nvidia:
## Blockers

1. **Cancelled response streams can stall the entire voice mailbox.** [`_ResponseStream`](https://github.com/NVIDIA/xr-ai/blob/13b98ea264f5d2590173108ede1160cd5dd4913a/agent-sdk/xr-ai-voice/xr_ai_voice/_runtime.py#L48-L98) has no `aclose()`, while cancellation cleanup only closes responses that expose that method ([`_close_response`](https://github.com/NVIDIA/xr-ai/blob/13b98ea264f5d2590173108ede1160cd5dd4913a/agent-sdk/xr-ai-voice/xr_ai_voice/_processors/io.py#L318-L327)). If cancellation occurs while `_run_response()` is pushing a chunk rather than awaiting `__anext__()`, the stream stays open. Subsequent chunks can fill the bounded queue and block the serialized `VoiceAgent.output` handler; this reproduces with `response_capacity=1`. Please provide explicit stream closure and evict closed streams from `VoiceAgent._streams`.

2. **Responses lose the originating user timestamp.** `UserQuery.timestamp_us` records the input timestamp ([message definition](https://github.com/NVIDIA/xr-ai/blob/13b98ea264f5d2590173108ede1160cd5dd4913a/agent-sdk/xr-ai-agent-runtime/xr_ai_runtime/messages.py#L9-L15)), but [`VoiceAgent.output()`](https://github.com/NVIDIA/xr-ai/blob/13b98ea264f5d2590173108ede1160cd5dd4913a/agent-sdk/xr-ai-voice/xr_ai_voice/_runtime.py#L156-L193) passes the newly published output message's `metadata.timestamp_us` to TTS. A probe with input timestamp `123` produced an output timestamp near the current wall clock. Please carry the correlated query timestamp through to voice output so the response retains its originating media time.

3. **A readiness failure closes model services twice.** [`VoiceSession.__aenter__()`](https://github.com/NVIDIA/xr-ai/blob/13b98ea264f5d2590173108ede1160cd5dd4913a/agent-sdk/xr-ai-voice/xr_ai_voice/_session.py#L80-L92) calls `close()` when a service probe fails, and runtime startup cleanup subsequently calls [`VoiceAgent.stop()`](https://github.com/NVIDIA/xr-ai/blob/13b98ea264f5d2590173108ede1160cd5dd4913a/agent-sdk/xr-ai-voice/xr_ai_voice/_runtime.py#L195-L206), which closes the session again. A probe observed two closes for both STT and TTS. Please make `VoiceSession.close()` idempotent or track whether ownership was successfully opened.

## Suggestions

1. **Remove the fabricated participant fallback.** The sample maps a global interruption to the synthetic participant ID `"voice-runtime"` because the runtime call requires a participant ([code](https://github.com/NVIDIA/xr-ai/blob/13b98ea264f5d2590173108ede1160cd5dd4913a/agent-samples/simple-vlm-example/worker/simple_vlm_example_worker/app.py#L84-L90)). Model global cancellation explicitly instead of inventing participant identity.

2. **Reject orphan stream terminators instead of silently downgrading them.** [`SimpleVlmAgent._stream()`](https://github.com/NVIDIA/xr-ai/blob/13b98ea264f5d2590173108ede1160cd5dd4913a/agent-samples/simple-vlm-example/worker/simple_vlm_example_worker/agent.py#L103-L128) always publishes an empty terminal message, even when no chunk opened the stream. [`VoiceAgent.output()`](https://github.com/NVIDIA/xr-ai/blob/13b98ea264f5d2590173108ede1160cd5dd4913a/agent-sdk/xr-ai-voice/xr_ai_voice/_runtime.py#L171-L193) treats that unknown response ID as a finite response that is then dropped as blank. Only publish a terminal message after opening a stream, and treat an unknown response ID as a protocol error.

3. **Avoid duplicate background-task error reporting.** [`SimpleVlmAgent._discard()`](https://github.com/NVIDIA/xr-ai/blob/13b98ea264f5d2590173108ede1160cd5dd4913a/agent-samples/simple-vlm-example/worker/simple_vlm_example_worker/agent.py#L137-L141) logs task failures even though `ctx.start_task()` already registers the task with the runtime's error reporting. Keep this callback focused on removing the task-map entry.

4. **Update the remaining documentation for the removed API.** [`docs/process-model.md`](https://github.com/NVIDIA/xr-ai/blob/13b98ea264f5d2590173108ede1160cd5dd4913a/docs/process-model.md#L63-L64) and [`docs/source/components/launcher-and-process-model.md`](https://github.com/NVIDIA/xr-ai/blob/13b98ea264f5d2590173108ede1160cd5dd4913a/docs/source/components/launcher-and-process-model.md#L59-L60) still say `VoiceSession.run()` owns readiness, but that API no longer exists.

nvddr:
Updated and rebased in `2c96462`; all seven items from the latest review are addressed:

- `_ResponseStream.aclose()` now releases blocked producers and evicts itself from `VoiceAgent._streams`; cancellation/backpressure at capacity 1 has regression coverage.
- `VoiceOutput.timestamp_us` carries the originating query timestamp through the simple-VLM stream to TTS.
- `VoiceSession.close()` is idempotent, including readiness-failure cleanup.
- Runtime calls now model global scope with `participant_id=None`; the synthetic `voice-runtime` participant is gone.
- Simple VLM publishes a terminator only after opening a stream, and `VoiceAgent` rejects an orphan terminator.
- The sample task callback only evicts its task-map entry; runtime-owned error reporting remains authoritative.
- Both process-model documents now describe readiness as owned by the session behind `VoiceAgent`.

I also corrected the query-schema boundary: `UserQuery` is owned by `xr-ai-voice`, while the sample supplies only the typed topic name (`simple-vlm.user-query`). It is no longer a generic runtime message, and there is no sample query factory or duplicate schema.

The rebased focused suite passes 34 tests and pre-commit is clean. GitHub now reports the PR mergeable; the refreshed CI jobs are running. @wenxind-nvidia please re-review the new head.

nvddr:
Rebased onto the current #350 head and updated in `baeed89`.

`VoiceAgent` now publishes its own query and lifecycle schemas on app-named topics. `SimpleVlmAgent` subscribes directly and owns vision-task cancellation and participant frame cleanup; `app.py` only constructs and registers the agents. Scoped participant departure and scoped/global interruption use the same runtime publish/subscribe mechanism, with no application callback glue.

Local validation: 45 focused runtime/voice/simple-VLM tests and 83 voice pipeline/session regressions pass; Ruff and SPDX are clean. The PR-scoped DCO check passes. The separate branch-push DCO job reports the unsigned current #350 head (`1264a59`), which is inherited rather than part of this PR's seven-commit range.

@wenxind-nvidia @blongs-nv @yanziz-nvidia, please re-review the new head.

nvddr:
Rebased onto the rewritten #350 head (`337b2d0`) in `ce09529`.

The rebase follows #350's final boundary rather than restoring its removed RPC layer: `VoiceAgent` is a normal `voice.output` subscriber, owns the lock protecting its aggregation/FIFO state, and no `Function`, `runtime.call`, or `SPEAK` API remains. Query and lifecycle routing stays typed pub/sub, including global interruption, and `SimpleVlmAgent` still owns its vision work and cleanup while `app.py` only composes.

Local validation: 42 focused runtime/voice/simple-VLM tests and 83 voice pipeline/session regressions pass; Ruff, SPDX, and DCO hooks pass. GitHub reports the PR mergeable, with refreshed tests and CodeQL still running.

@wenxind-nvidia @blongs-nv @yanziz-nvidia, please review the rebased head.

blongs-nv:
Reviewed at `ce095299c8989c472302b078c371766ff18c76e2` by a bot.

## Blockers

1. **A cancelled producer's stream terminator kills the entire runtime on the first barge-in or supersede.**

   [`agent-samples/simple-vlm-example/worker/simple_vlm_example_worker/agent.py:111-120`](https://github.com/NVIDIA/xr-ai/blob/ce095299c8989c472302b078c371766ff18c76e2/agent-samples/simple-vlm-example/worker/simple_vlm_example_worker/agent.py#L111-L120)
   On interruption or supersede the io processor cancels response playback first, which closes and evicts the stream ([`agent-sdk/xr-ai-voice/xr_ai_voice/_processors/io.py:114-118`](https://github.com/NVIDIA/xr-ai/blob/ce095299c8989c472302b078c371766ff18c76e2/agent-sdk/xr-ai-voice/xr_ai_voice/_processors/io.py#L114-L118)); the subsequent `VoiceInterrupted`/new-query delivery cancels `_stream`, whose `finally` still publishes the empty terminator, and the resulting `ValueError` ([`agent-sdk/xr-ai-voice/xr_ai_voice/_runtime.py:248`](https://github.com/NVIDIA/xr-ai/blob/ce095299c8989c472302b078c371766ff18c76e2/agent-sdk/xr-ai-voice/xr_ai_voice/_runtime.py#L248)) makes the cancelled task end with an exception, which `_background_failed` ([`agent-sdk/xr-ai-agent-runtime/xr_ai_runtime/runtime.py:196-201`](https://github.com/NVIDIA/xr-ai/blob/ce095299c8989c472302b078c371766ff18c76e2/agent-sdk/xr-ai-agent-runtime/xr_ai_runtime/runtime.py#L196-L201)) escalates to whole-runtime teardown. I reproduced this end to end; with `interrupt_on_supersede=True` in the sample, "user speaks while the agent answers" is the primary interaction. A cancelled producer should not publish a terminator, and/or keys the voice side closed should be tombstoned so their terminators are swallowed.

2. **`VoiceAgent._output_lock` is held across `stream.send()`, which blocks on queue capacity, so one queued or slow stream deadlocks all voice output.**

   [`agent-sdk/xr-ai-voice/xr_ai_voice/_runtime.py:219`](https://github.com/NVIDIA/xr-ai/blob/ce095299c8989c472302b078c371766ff18c76e2/agent-sdk/xr-ai-voice/xr_ai_voice/_runtime.py#L219), [`agent-sdk/xr-ai-voice/xr_ai_voice/_runtime.py:267`](https://github.com/NVIDIA/xr-ai/blob/ce095299c8989c472302b078c371766ff18c76e2/agent-sdk/xr-ai-voice/xr_ai_voice/_runtime.py#L267)
   With two producers on one participant, producer B's stream is FIFO-queued (unconsumed) behind in-flight A ([`agent-sdk/xr-ai-voice/xr_ai_voice/_processors/io.py:196-203`](https://github.com/NVIDIA/xr-ai/blob/ce095299c8989c472302b078c371766ff18c76e2/agent-sdk/xr-ai-voice/xr_ai_voice/_processors/io.py#L196-L203)); B fills its bounded queue and blocks holding the lock, A's next publish blocks acquiring it, A never terminates, and B's stream never starts. The same global lock also head-of-line-blocks every other participant. Hold the lock only for the `_streams` lookup/create and `_enqueue_response`, and call `send()` outside it.

3. **A non-cancellation error in `_run_response` never closes the stream, and the regression test for cancel-time release passes with the fix removed.**

   [`agent-sdk/xr-ai-voice/xr_ai_voice/_processors/io.py:233-238`](https://github.com/NVIDIA/xr-ai/blob/ce095299c8989c472302b078c371766ff18c76e2/agent-sdk/xr-ai-voice/xr_ai_voice/_processors/io.py#L233-L238)
   `_close_response` runs only on `CancelledError`; if `push_frame` or the consume loop raises, the stream stays registered with no consumer and its producer eventually blocks while holding the output lock (blocker 2's failure mode). Separately, [`tests/test_voice_runtime.py:319`](https://github.com/NVIDIA/xr-ai/blob/ce095299c8989c472302b078c371766ff18c76e2/tests/test_voice_runtime.py#L319) asserts `_streams == {}` only after `async with runtime` exits (lifespan teardown makes that tautological) and its capacity-1 queue is already drained before the cancel, so it was verified to still pass with the `aclose`-on-cancel branch deleted; make the test gate the consumer so a `send()` is genuinely blocked, and assert inside the runtime context.

4. **A chunk arriving after the voice side closed a stream silently opens a new stream and speaks the tail of the answer the user just interrupted.**

   [`agent-sdk/xr-ai-voice/xr_ai_voice/_runtime.py:256-266`](https://github.com/NVIDIA/xr-ai/blob/ce095299c8989c472302b078c371766ff18c76e2/agent-sdk/xr-ai-voice/xr_ai_voice/_runtime.py#L256-L266)
   Barge-in evicts `(participant, source, response_id)` before the producer is cancelled via `VoiceInterrupted`, so any chunk published in that window recreates the key with no in-flight playback and plays immediately. The same tombstone mechanism as blocker 1 covers this.

5. **Query fan-out occupies the participant's cancellable slot, so an interrupting first chunk cancels delivery of the query to every other subscriber.**

   [`agent-sdk/xr-ai-voice/xr_ai_voice/_processors/io.py:152-160`](https://github.com/NVIDIA/xr-ai/blob/ce095299c8989c472302b078c371766ff18c76e2/agent-sdk/xr-ai-voice/xr_ai_voice/_processors/io.py#L152-L160)
   While `_run_query` awaits `publish`'s gather over all query-topic subscribers it sits in `_inflight[pid]`; the first responder's `interrupt=True` chunk (as the sample sends at [`agent.py:105`](https://github.com/NVIDIA/xr-ai/blob/ce095299c8989c472302b078c371766ff18c76e2/agent-samples/simple-vlm-example/worker/simple_vlm_example_worker/agent.py#L105)) triggers `_cancel_pid`, cancelling the gather and silently dropping delivery to the remaining subscribers. Multi-agent fan-out is this PR's stated purpose, so query delivery should not share the response playback slot.

6. **The sample opens hub ZMQ sockets before readiness probes, contradicting the invariant this same diff writes in its docs.**

   [`agent-samples/simple-vlm-example/worker/simple_vlm_example_worker/app.py:67`](https://github.com/NVIDIA/xr-ai/blob/ce095299c8989c472302b078c371766ff18c76e2/agent-samples/simple-vlm-example/worker/simple_vlm_example_worker/app.py#L67)
   `session.transport.endpoint` at composition time lazily constructs `HubVoiceTransport` ([`_session.py:69-74`](https://github.com/NVIDIA/xr-ai/blob/ce095299c8989c472302b078c371766ff18c76e2/agent-sdk/xr-ai-voice/xr_ai_voice/_session.py#L69-L74)), whose `ProcessorEndpoint.__init__` connects sockets immediately ([`_processor.py:181-187`](https://github.com/NVIDIA/xr-ai/blob/ce095299c8989c472302b078c371766ff18c76e2/agent-sdk/xr-ai-hub-client/xr_ai_hub/_processor.py#L181-L187)), while [`docs/source/components/agent-sdk.md:339`](https://github.com/NVIDIA/xr-ai/blob/ce095299c8989c472302b078c371766ff18c76e2/docs/source/components/agent-sdk.md#L339) and [`DEPENDENCIES.md:95`](https://github.com/NVIDIA/xr-ai/blob/ce095299c8989c472302b078c371766ff18c76e2/DEPENDENCIES.md#L95) still say failed readiness never opens hub sockets. The base branch constructed the tool inside `async with session:`; pass an explicit transport or defer endpoint resolution into the agent's lifespan.

7. **Deleting the NAT voice adapter left its packaging and the authoritative dependency docs behind.**

   [`agent-sdk/xr-ai-nat/pyproject.toml:23`](https://github.com/NVIDIA/xr-ai/blob/ce095299c8989c472302b078c371766ff18c76e2/agent-sdk/xr-ai-nat/pyproject.toml#L23)
   The `voice = ["xr-ai-voice"]` extra and its `[tool.uv.sources]` entry (line 37) survive with nothing in `xr_ai_nat` importing `xr_ai_voice`, and [`DEPENDENCIES.md:150`](https://github.com/NVIDIA/xr-ai/blob/ce095299c8989c472302b078c371766ff18c76e2/DEPENDENCIES.md#L150) plus its prose at [`DEPENDENCIES.md:159-162`](https://github.com/NVIDIA/xr-ai/blob/ce095299c8989c472302b078c371766ff18c76e2/DEPENDENCIES.md#L159-L162) still document the deleted `as_voice_handler` / `record_voice_transcripts`. AGENTS.md makes `DEPENDENCIES.md` authoritative and requires it to move with every `pyproject.toml` change.

8. **Deleting `tests/test_adapters_voice.py` removed the only coverage of the still-shipping `xr_conversation_memory` function group.**

   [`agent-sdk/xr-ai-nat/xr_ai_nat/functions/text_memory/functions.py:224`](https://github.com/NVIDIA/xr-ai/blob/ce095299c8989c472302b078c371766ff18c76e2/agent-sdk/xr-ai-nat/xr_ai_nat/functions/text_memory/functions.py#L224)
   Only the adapter half of that file's coverage died with its production code; `recall_conversation` time-window filtering and the generated request/result schema contract now have zero tests anywhere. Relocate the surviving cases.

## Suggestions

1. Widening `MessageMetadata.participant_id` to `str | None` ([`runtime.py:62`](https://github.com/NVIDIA/xr-ai/blob/ce095299c8989c472302b078c371766ff18c76e2/agent-sdk/xr-ai-agent-runtime/xr_ai_runtime/runtime.py#L62)) changes a runtime-wide contract to serve one voice event: every subscriber now needs a null check and a mis-scoped publish fails late inside the subscriber rather than at the publish site. Consider keeping metadata scoping mandatory and modeling "global" on the event body instead.
2. Two behaviors lost their tests in the relocation: dropping typed text while the session is stopped ([`_runtime.py:320`](https://github.com/NVIDIA/xr-ai/blob/ce095299c8989c472302b078c371766ff18c76e2/agent-sdk/xr-ai-voice/xr_ai_voice/_runtime.py#L320)) and clearing `_seen_output` on participant-left ([`io.py:138`](https://github.com/NVIDIA/xr-ai/blob/ce095299c8989c472302b078c371766ff18c76e2/agent-sdk/xr-ai-voice/xr_ai_voice/_processors/io.py#L138), where a regression means a rejoining participant's first turn spuriously interrupts). Add both back.
3. Register the stream in `_streams` only after `_enqueue_response` succeeds ([`_runtime.py:256-266`](https://github.com/NVIDIA/xr-ai/blob/ce095299c8989c472302b078c371766ff18c76e2/agent-sdk/xr-ai-voice/xr_ai_voice/_runtime.py#L256-L266)); a stopped-session `RuntimeError` currently strands the key. `_streams` is also unbounded: a producer that opens streams and never terminates them leaks a queue per response for the life of the process.
4. A mid-stream chunk with `interrupt=True` passes validation but is silently ignored ([`_runtime.py:267`](https://github.com/NVIDIA/xr-ai/blob/ce095299c8989c472302b078c371766ff18c76e2/agent-sdk/xr-ai-voice/xr_ai_voice/_runtime.py#L267)); reject it or apply it.
5. A `text_transform` returning empty or whitespace text fails `UserQuery` validation deep inside `_run_query` as a logged exception ([`_runtime.py:324-326`](https://github.com/NVIDIA/xr-ai/blob/ce095299c8989c472302b078c371766ff18c76e2/agent-sdk/xr-ai-voice/xr_ai_voice/_runtime.py#L324-L326)); re-check emptiness after the transform. Note also that the default `text_ignore_topics=()` treats every inbound data-channel topic as a user query.
6. `endpoint.on_data(self._on_data)` is registered in `lifespan` but never unregistered in its `finally` ([`_runtime.py:178`](https://github.com/NVIDIA/xr-ai/blob/ce095299c8989c472302b078c371766ff18c76e2/agent-sdk/xr-ai-voice/xr_ai_voice/_runtime.py#L178)), so a re-entered lifespan or second agent over the same endpoint accumulates callbacks.
7. On token mismatch `_consume_response` continues draining the superseded iterator to exhaustion ([`io.py:270-274`](https://github.com/NVIDIA/xr-ai/blob/ce095299c8989c472302b078c371766ff18c76e2/agent-sdk/xr-ai-voice/xr_ai_voice/_processors/io.py#L270-L274)); return instead, or a long remote stream keeps burning tokens after supersession.
8. `VoiceSession` is public yet everything `VoiceAgent` consumes is private (`_run`, `_enqueue_query`, `_enqueue_response`) and the sample reaches through `session.transport.endpoint`; give the session a narrow intentional surface.
9. Minor hygiene: the `topic is None` guards in `_publish_participant_left` / `_publish_interrupted` are unreachable given `lifespan`'s wiring ([`_runtime.py:283-285`](https://github.com/NVIDIA/xr-ai/blob/ce095299c8989c472302b078c371766ff18c76e2/agent-sdk/xr-ai-voice/xr_ai_voice/_runtime.py#L283-L285)), `_publish_input` skips the `_running_context()` helper its siblings use, and the unbounded `while` waits at [`tests/test_voice_pipeline.py:1210`](https://github.com/NVIDIA/xr-ai/blob/ce095299c8989c472302b078c371766ff18c76e2/tests/test_voice_pipeline.py#L1210) and [`tests/test_simple_vlm_example_worker.py:347`](https://github.com/NVIDIA/xr-ai/blob/ce095299c8989c472302b078c371766ff18c76e2/tests/test_simple_vlm_example_worker.py#L347) hang the suite instead of failing it on regression.

The direction here is right: one bidirectional `VoiceAgent`, typed topics for input and lifecycle, session internals kept private, and `app.py` reduced to pure composition, and most of the prior review round verifiably landed (chunk accumulation across iterator failures, producer-keyed streams, timestamp propagation, idempotent close, global scope without a synthetic participant, and FIFO order all have real regression assertions). But the incremental-output contract cannot be honored by the shipped producer under cancellation, and the output lock's scope turns backpressure into a global stall. The blockers above should be resolved before merge.

nvddr:
Updated in 7e91705.

The active voice/runtime findings are addressed:

- Cancelled producers no longer publish orphan terminators; closed stream keys are tombstoned so late chunks cannot reopen interrupted playback.
- Stream sends run outside the global output lock, active and closed stream state is bounded, admission failures cannot strand keys, and mid-stream interrupts are rejected.
- Response iterators close on every exit, superseded iterators stop draining, and query-topic fan-out no longer shares the cancellable playback slot.
- Typed input is dropped while stopped, transformed text is revalidated, the output topic is ignored by default, participant output state is cleared on departure, and endpoint callbacks are unregistered.
- VoiceSession now exposes the narrow lifecycle/enqueue/endpoint surface used by VoiceAgent.
- The simple-VLM vision tool is constructed only after health readiness initializes the session endpoint, and sample cancellation closes its producer without sending a terminator.

I also fixed the current GitHub code-quality findings: direct transport type usage, explicit consumption of awaited results, idempotent callback removal without an empty except, and explicit async-iterator termination.

Per review scope, the legacy NAT packaging/conversation-memory compatibility items and the runtime-wide participant-metadata compatibility suggestion are unchanged.

Validation: 139 focused tests pass. The refreshed GitHub suite is green on Python 3.11 and 3.12, Ruff 0.15.16, both DCO checks, CodeQL, SPDX headers, dependency locks, and strict docs build.

@blongs-nv please re-review the updated head.

blongs-nv (approved):
Reviewed at `7e91705ebe0de6c759473276f7669611b7c9dc72` by a bot.

Blockers 1 through 6 from my previous review are resolved, and the fixes were verified in the code rather than by commit message: a cancelled producer no longer publishes its terminator and voice-closed keys are tombstoned, so barge-in no longer raises; `send()` now runs outside the output lock, with a real regression test that holds a genuinely blocked publisher at capacity 1 while unrelated output proceeds; failed and cancelled responses are closed on every exit path; post-eviction chunks are swallowed instead of re-opening a stream; query fan-out no longer occupies the cancellable response slot (with a delivery test); and the sample builds its vision tool lazily behind readiness, so hub sockets no longer open before probes. The suggestions on relocated coverage, mid-stream `interrupt`, `on_data` unregistration, the superseded-iterator drain, stream registration order, and the session's public surface were addressed as well.

On the deferred items: given `xr-ai-nat`'s status as a migration-period compatibility surface and your scope call, I'm treating the `[voice]` packaging/`DEPENDENCIES.md` cleanup and the `xr_conversation_memory` test relocation as deferred to NAT's own cleanup rather than holding this PR for them. Please make sure the three `recall_conversation` cases aren't lost for good if that function group migrates rather than dies. The runtime-wide `participant_id` metadata suggestion is likewise noted as deferred.

## Suggestions

1. `await stream.aclose()` is called twice in a row in the final-chunk path ([`_runtime.py:338-339`](https://github.com/NVIDIA/xr-ai/blob/7e91705ebe0de6c759473276f7669611b7c9dc72/agent-sdk/xr-ai-voice/xr_ai_voice/_runtime.py#L338-L339)); the second call is a no-op, drop it.

The fix round is thorough and well-tested, and the remaining items are deferred by agreement. This is ready to merge.

PR #357, opened 2026-08-13
fix(ci): authenticate nightly vLLM image pulls

### LINKED ISSUE #346: Nightly XR AI test is failing
Nightly run **failed** (event: schedule, commit `e62480c`).

- Run: https://github.com/NVIDIA/xr-ai/actions/runs/31458753381

wenxind-nvidia:
## Summary

- authenticate the nightly GPU job to `nvcr.io` with the existing `NGC_API_KEY` repository secret
- isolate Docker credentials in a per-run `DOCKER_CONFIG` and remove them during always-run cleanup
- fail before dependency setup when the key is absent or rejected instead of failing three model tests later
- document why nightly CI must not depend on cached vLLM images or runner-global Docker state

## Root cause

The nightly workflow did not map `secrets.NGC_API_KEY` into any step. The vLLM tests launch model wrappers directly, so the launcher credential cache is not involved. Run 31667652128 reached Docker with neither a local `nvcr.io/nvidia/vllm:26.04-py3` image nor valid registry authentication, and all three vLLM-backed tests failed with `Access Denied`.

Closes #346

## Validation

- `actionlint` (excluding the existing custom-runner-label warning)
- workflow YAML parse
- valid-key login/cleanup shell-path harness with a fake Docker client
- missing-key fail-fast/cleanup shell-path harness
- `pre-commit` SPDX checks
- `git diff --check`

### DIFF
--- a/.github/workflows/nightly-xr-ai-test.yml
+++ b/.github/workflows/nightly-xr-ai-test.yml
@@ -8,9 +8,11 @@ name: nightly XR AI test
 # the default `tests` workflow (ubuntu-latest). Mirrors the local
 # `tests/run_local_gpu_tests.sh` entry point.
 #
-# Setup note: the GPU runner's Docker needs the NVIDIA runtime registered
-# (`sudo nvidia-ctk runtime configure --runtime=docker`); without it the
-# vLLM tests fail at container launch.
+# Setup notes:
+# - The GPU runner's Docker needs the NVIDIA runtime registered
+#   (`sudo nvidia-ctk runtime configure --runtime=docker`).
+# - The repository needs an Actions secret named `NGC_API_KEY` so trusted
+#   events can authenticate pulls from `nvcr.io`.
 
 on:
   schedule:
@@ -43,6 +45,13 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+
+      - name: Configure isolated Docker credentials
+        run: |
+          set -euo pipefail
+          docker_config="$RUNNER_TEMP/xr-ai-docker-${GITHUB_RUN_ID}-${GITHUB_RUN_ATTEMPT}"
+          install -d -m 0700 "$docker_config"
+          echo "DOCKER_CONFIG=$docker_config" >> "$GITHUB_ENV"
 
       - name: Discover CUDA toolkit
         # Prefer CUDA 13.0 (current target), but fall back through the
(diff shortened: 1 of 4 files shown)


blongs-nv:
Reviewed at `0778bf2c1042830ce3955f8ecdda6d3bb135d2e7` by a bot.

## Blockers

1. **The `docker login` runs before the `DOCKER_CONFIG` export takes effect, so the PR does not fix the auth failure and leaks the NGC key onto the runner.**

   [`.github/workflows/nightly-xr-ai-test.yml:95-103`](https://github.com/NVIDIA/xr-ai/blob/0778bf2c1042830ce3955f8ecdda6d3bb135d2e7/.github/workflows/nightly-xr-ai-test.yml#L95-L103)
   `GITHUB_ENV` only applies to subsequent steps, so the login in this step writes credentials to the runner-global `~/.docker/config.json`, while every later step (including the pytest step whose subprocesses run `docker pull`/`docker run`) reads the empty per-run directory and pulls unauthenticated; the [cleanup step](https://github.com/NVIDIA/xr-ai/blob/0778bf2c1042830ce3955f8ecdda6d3bb135d2e7/.github/workflows/nightly-xr-ai-test.yml#L188-L193) then deletes an empty directory and the key stays on the shared self-hosted runner. I verified this in this PR's own green GPU run (job 94361740159): the auth step's env group has no `DOCKER_CONFIG` while every later step shows it set, so the run passed only because the vLLM image was already cached, the exact dependency the change claims to remove. Adding `export DOCKER_CONFIG="$docker_config"` before the login fixes both halves; update the [changelog entry](https://github.com/NVIDIA/xr-ai/blob/0778bf2c1042830ce3955f8ecdda6d3bb135d2e7/docs/changelog.md#L12-L20) at the same time, since as written it describes the intended behavior, not the actual one.

## Suggestions

1. The broken login has already written the NGC key into the runner's global `~/.docker/config.json` (this PR's pre-merge runs did it). Add `docker logout nvcr.io || true` to the cleanup step so the leaked entry gets removed on the next run.
2. The hard `exit 1` on an empty `NGC_API_KEY` ([`.github/workflows/nightly-xr-ai-test.yml:98-101`](https://github.com/NVIDIA/xr-ai/blob/0778bf2c1042830ce3955f8ecdda6d3bb135d2e7/.github/workflows/nightly-xr-ai-test.yml#L98-L101)) will fail every fork PR that edits this workflow, since secrets are not exposed to fork `pull_request` runs. Gate the step on non-fork events, or downgrade to a warning on `pull_request`.
3. Even with the export fixed, a green `pytest (gpu)` run cannot distinguish working auth from a cached image. Consider asserting auth right after login, for example `docker manifest inspect nvcr.io/nvidia/vllm:26.04-py3`, so the login step fails loudly when the key is rejected.
4. Move the empty-key guard above the directory creation and `GITHUB_ENV` write so a missing secret has no side effects, and add `set -u` to the cleanup step to match the sibling always-run purge step.
5. The new required repository secret is undocumented. Add it to the workflow header's "Setup note" and to the `NGC_API_KEY` docs, and note in the changelog entry that the CI login deliberately bypasses the existing `_maybe_ngc_login()` path in `xr_ai_vllm/_docker.py` (whose `_already_logged_in()` reads `~/.docker/config.json` and ignores `DOCKER_CONFIG`), so a future reader doesn't reconcile the two mechanisms the wrong way.

The diagnosis in the PR is right and the shape of the fix (per-run Docker config, secret scoped to one step, always-run cleanup) is the correct instinct. The blocker above should be resolved before merge.

yanziz-nvidia:
Can you make it so that it won't skip the test when you touch .github/workflows/nightly-xr-ai-test.yml

PR #359, opened 2026-08-13
feat(vision): decouple frame selection and add multi-image queries

nvddr:
## Summary

- add bounded recorded-frame sampling with total frame and optional resolution budgets
- introduce lightweight `ImageReference` values backed by a bounded `ImageRegistry`
- separate frame selection from VLM inference:
  - `get_current_frame` selects a live frame
  - `get_latest_frames` / `get_latest_video` select the newest recorded
    window from only `participant_id + duration_seconds`
  - `get_historical_frame`, `get_historical_frames`, and
    `get_historical_video` share one absolute `start_us`; frame sequences and
    video clips add `duration_seconds`
  - `latest_tools` and `historical_tools` expose those selection groups
- expose distinct inference tools over the selected inputs:
  - `query_image` handles one image
  - `query_images` handles an ordered image collection
  - `query_video` handles chronological `TimedImage` frames with absolute timestamps and relative offsets
  - `stream_image_query` preserves streaming single-image output
- unify all image cardinalities behind one list-based inference implementation and add `VLMService.ask_images()` / `stream_images()`; the singular APIs remain one-image wrappers
- migrate the simple VLM and XR render agents to the decoupled tools and retire the coupled live, historical, and streaming vision APIs after migrating all in-tree consumers
- keep image bytes out of tool results and Relay telemetry; every image location is redacted while provider input retains the original images

## Compatibility and behavior

- XR render keeps its model-facing `look_at_current_frame` and `look_at_past_frame` behavior unchanged
- sampled recorded frames implement the same `TimedImage` contract accepted directly by `query_video`
- recorded results retain filesystem paths
- voice streaming, participant cleanup, interruption, readiness, and status behavior remain covered
- requests are bounded to 256 images/frames and recorded windows to 300 seconds

## Why

Frame acquisition and visual reasoning now scale independently. Agents can reuse selected images, query non-camera images, reason over ordered collections, and ask temporal questions over a bounded recorded-frame sequence without duplicating the single-image and multi-image execution paths.

## Validation

- complete CPU suite: `777 passed, 10 deselected`
- focused image/model/render/video-memory suites: `125 passed`
- Ruff, SPDX, and `git diff --check`
- strict Sphinx documentation build
- dependency resolution for every affected project


wenxind-nvidia:
Reviewed at `347b3c4ea4d6ae20429e81b5e32c25e6a23c5741`.

## Blockers

1. **Returned timestamps are estimates, not actual frame timestamps.** [`service.py:307`](https://github.com/NVIDIA/xr-ai/blob/347b3c4ea4d6ae20429e81b5e32c25e6a23c5741/services/video-memory-service/video_memory_service/service.py#L307) linearly interpolates timestamps from chunk bounds and frame count. The recorder ignores incoming frame PTS and records only chunk creation/flush times ([`_recorder.py:248`](https://github.com/NVIDIA/xr-ai/blob/347b3c4ea4d6ae20429e81b5e32c25e6a23c5741/services/xr-media-hub/xr_media_hub/video/_recorder.py#L248)). Jitter, dropped frames, or an idle delay before flushing will produce incorrect timestamps and potentially select the wrong frames. Persist per-frame timing, or explicitly expose these as estimated timestamps and adjust the contract/docs.

2. **The documented maximum request is not operationally bounded.** `frame_budget` permits 256 frames while resolution limits are optional ([`video_memory.py:65`](https://github.com/NVIDIA/xr-ai/blob/347b3c4ea4d6ae20429e81b5e32c25e6a23c5741/agent-sdk/xr-ai-tools/xr_ai_tools/video_memory.py#L65)). The service sequentially exports every selected PNG ([`service.py:224`](https://github.com/NVIDIA/xr-ai/blob/347b3c4ea4d6ae20429e81b5e32c25e6a23c5741/services/video-memory-service/video_memory_service/service.py#L224)), but the client retains its 30-second timeout. A 256-frame 4K request can take substantially longer and write several GB to an output directory with no retention policy. When the client times out, the server continues processing. Require a spatial/byte budget, reduce the maximum, add output retention, and align timeout/cancellation behavior with the legal workload.

3. **The new raw sampler is unnecessarily exposed to the xr-render model.** `NativeCapabilities` puts every `VideoMemoryTools.tools` entry into the model-facing set ([`tools.py:51`](https://github.com/NVIDIA/xr-ai/blob/347b3c4ea4d6ae20429e81b5e32c25e6a23c5741/agent-samples/xr-render-demo/worker/xr_render_demo_worker/tools.py#L51)). The new tool returns filesystem paths, but the text LLM has no step that submits those images to a VLM. It may select this tool for historical visual questions instead of `look_at_past_frame` and then cannot inspect the result. Keep it programmatic until a concrete multimodal consumer exists, or add the participant-aware VLM adapter as part of this PR.

## Suggestion

Add GPU integration coverage that invokes `sample_recorded_video` against a real recorded chunk. Current tests mock `decode_h264`, so the new multi-frame NVDEC path is only CPU-fixture tested. Sparse-window, native-versus-bounded path separation, and timeout behavior also need coverage.

The change is otherwise narrowly organized around one RPC operation; I did not find unrelated refactoring or unnecessary compatibility fallbacks. The focused tests and Ruff pass, and the current GitHub checks are green.

blongs-nv:
Reviewed at `347b3c4ea4d6ae20429e81b5e32c25e6a23c5741` by a bot.

## Blockers

1. **A chunk pruned by the hub mid-request raises an unhandled `FileNotFoundError` and discards the entire sample.**

   [`services/video-memory-service/video_memory_service/service.py:197`](https://github.com/NVIDIA/xr-ai/blob/347b3c4ea4d6ae20429e81b5e32c25e6a23c5741/services/video-memory-service/video_memory_service/service.py#L197)
   The chunk list is snapshotted before the decode loop while the hub FIFO-prunes the oldest chunks on every flush ([`_recorder.py:270`](https://github.com/NVIDIA/xr-ai/blob/347b3c4ea4d6ae20429e81b5e32c25e6a23c5741/services/xr-media-hub/xr_media_hub/video/_recorder.py#L270)), and a long lookback targets exactly those oldest chunks; the unwrapped `read_bytes` surfaces as a generic `internal_error` after earlier frames were already exported.

2. **A single corrupt or zero-frame chunk aborts the whole request, contradicting the documented sparse-degrade contract.**

   [`services/video-memory-service/video_memory_service/service.py:199-209`](https://github.com/NVIDIA/xr-ai/blob/347b3c4ea4d6ae20429e81b5e32c25e6a23c5741/services/video-memory-service/video_memory_service/service.py#L199-L209)
   [`README.md:22`](https://github.com/NVIDIA/xr-ai/blob/347b3c4ea4d6ae20429e81b5e32c25e6a23c5741/services/video-memory-service/README.md#L22) promises "sparse recordings may return fewer frames", but one decode failure anywhere in the window raises `decode_error` and orphans the PNGs already written. Skip and degrade per chunk (as the empty-candidates path at [`service.py:218`](https://github.com/NVIDIA/xr-ai/blob/347b3c4ea4d6ae20429e81b5e32c25e6a23c5741/services/video-memory-service/video_memory_service/service.py#L218) already does), or narrow the documented claim.

## Suggestions

1. The only `overlapping_chunks` test ([`tests/test_video_memory_functions.py:137`](https://github.com/NVIDIA/xr-ai/blob/347b3c4ea4d6ae20429e81b5e32c25e6a23c5741/tests/test_video_memory_functions.py#L137)) uses a window that returns both chunks of a two-chunk fixture, so it still passes if the overlap filter is deleted. Add a window that excludes a chunk, and a window past the last chunk asserting the `not_found` error at [`store.py:124`](https://github.com/NVIDIA/xr-ai/blob/347b3c4ea4d6ae20429e81b5e32c25e6a23c5741/services/video-memory-service/video_memory_service/store.py#L124).
2. None of the sampler's error paths are tested: `decode_error` (raising decoder and zero-frames decode), `frame_export_error`, the terminal `not_found` at [`service.py:268-272`](https://github.com/NVIDIA/xr-ai/blob/347b3c4ea4d6ae20429e81b5e32c25e6a23c5741/services/video-memory-service/video_memory_service/service.py#L268-L272), and the native-resolution path (no `max_width`/`max_height`) through the service. The file already has the monkeypatch pattern for these in `test_recorded_frame_reports_frame_export_errors`.
3. Output filenames are deterministic and never cleaned ([`service.py:235-238`](https://github.com/NVIDIA/xr-ai/blob/347b3c4ea4d6ae20429e81b5e32c25e6a23c5741/services/video-memory-service/video_memory_service/service.py#L235-L238)): two concurrent identical requests write the same PNGs simultaneously, so a reader can open a half-written file. A per-request subdirectory removes the collision and gives retention a unit to delete.
4. `_sample_candidates` ([`service.py:290-316`](https://github.com/NVIDIA/xr-ai/blob/347b3c4ea4d6ae20429e81b5e32c25e6a23c5741/services/video-memory-service/video_memory_service/service.py#L290-L316)) re-derives the index-to-timestamp interpolation that `select_decoded_frame` already owns, and `overlapping_chunks` duplicates the overlap predicate inline in `ChunkStore.query` ([`store.py:127-141`](https://github.com/NVIDIA/xr-ai/blob/347b3c4ea4d6ae20429e81b5e32c25e6a23c5741/services/video-memory-service/video_memory_service/store.py#L127-L141)) while disagreeing with it: `query` falls back to the preceding anchor chunk for a gap-only window, `overlapping_chunks` raises `not_found`. Factor each into one shared primitive so the two operations can't drift.
5. A target inside a recording gap snaps to the nearest chunk edge with no distance cap ([`service.py:220-223`](https://github.com/NVIDIA/xr-ai/blob/347b3c4ea4d6ae20429e81b5e32c25e6a23c5741/services/video-memory-service/video_memory_service/service.py#L220-L223)), so "evenly distributed" degrades to a cluster of near-duplicate edge frames with no signal to the caller. Consider dropping targets whose nearest candidate is farther than about half the target spacing.

The sampling math holds up well: the frame budget is genuinely hard, output paths cannot escape `out_dir`, and the happy-path tests assert real behavior (timestamps, once-per-chunk decode, on-disk PNG sizes). The blockers above should be resolved before merge.

blongs-nv:
Reviewed at `145f58a1957ccb9e30c9ef9c31bf02a2376fa714` by a bot.

Blockers 1 and 2 from my review at `347b3c4` remain open at this head. The branch has since grown the vision refactor, the multi-image/timed-frame tools, and the latest/historical window reshape; blockers 3 and 4 and all of the suggestions below are new findings in that newer code, not previously posted. The window reshape itself is clean, but it does not touch the sampler failure paths, the image-URI resolution policy, the sample dispatch path, or the dead exception handler.

## Blockers

1. **(From my earlier review) A chunk pruned by the hub mid-request raises an unhandled `FileNotFoundError` and discards the entire sample.**

   [`services/video-memory-service/video_memory_service/service.py:244`](https://github.com/NVIDIA/xr-ai/blob/145f58a1957ccb9e30c9ef9c31bf02a2376fa714/services/video-memory-service/video_memory_service/service.py#L244)
   Unchanged; the unwrapped `read_bytes` in `_sample_frames` still surfaces as a generic `internal_error`.

2. **(From my earlier review) A single corrupt or zero-frame chunk aborts the whole request, contradicting the documented sparse-degrade contract.**

   [`services/video-memory-service/video_memory_service/service.py:246-256`](https://github.com/NVIDIA/xr-ai/blob/145f58a1957ccb9e30c9ef9c31bf02a2376fa714/services/video-memory-service/video_memory_service/service.py#L246-L256), [`README.md:33`](https://github.com/NVIDIA/xr-ai/blob/145f58a1957ccb9e30c9ef9c31bf02a2376fa714/services/video-memory-service/README.md#L33). Unchanged.

3. **(New, in the vision refactor) Model-reachable arbitrary local file read and SSRF through the image-query tools.**

   [`agent-sdk/xr-ai-tools/xr_ai_tools/image.py:64-83`](https://github.com/NVIDIA/xr-ai/blob/145f58a1957ccb9e30c9ef9c31bf02a2376fa714/agent-sdk/xr-ai-tools/xr_ai_tools/image.py#L64-L83)
   `resolve` turns any bare string into `Path(uri)` and passes `http(s)` through; the worker dispatches every model-emitted name against `capabilities.all` ([`__main__.py:145`](https://github.com/NVIDIA/xr-ai/blob/145f58a1957ccb9e30c9ef9c31bf02a2376fa714/agent-samples/xr-render-demo/worker/xr_render_demo_worker/__main__.py#L145) into [`scene_loop.py:1011`](https://github.com/NVIDIA/xr-ai/blob/145f58a1957ccb9e30c9ef9c31bf02a2376fa714/agent-samples/xr-render-demo/worker/xr_render_demo_worker/scene_loop.py#L1011)), so the schema-only exclusion at [`tools.py:79-87`](https://github.com/NVIDIA/xr-ai/blob/145f58a1957ccb9e30c9ef9c31bf02a2376fa714/agent-samples/xr-render-demo/worker/xr_render_demo_worker/tools.py#L79-L87) does not prevent a hallucinated or injected `query_image`/`query_images`/`query_video` call.

4. **(New, in the vision refactor) The `except FrameUnavailable` handler in the simple-vlm sample is dead code.**

   [`agent-samples/simple-vlm-example/worker/simple_vlm_example_worker/agent.py:155`](https://github.com/NVIDIA/xr-ai/blob/145f58a1957ccb9e30c9ef9c31bf02a2376fa714/agent-samples/simple-vlm-example/worker/simple_vlm_example_worker/agent.py#L155)
   I reproduced this: the relay pipeline re-raises handler exceptions as `RuntimeError`, so a missing camera frame kills the turn with no spoken fallback.

## Suggestions

1. Resolve failures (`LookupError`, `ValueError`) are reported as "VLM server unavailable — please retry." ([`vision.py:114-119`](https://github.com/NVIDIA/xr-ai/blob/145f58a1957ccb9e30c9ef9c31bf02a2376fa714/agent-sdk/xr-ai-tools/xr_ai_tools/vision.py#L114-L119), [`vision.py:149-152`](https://github.com/NVIDIA/xr-ai/blob/145f58a1957ccb9e30c9ef9c31bf02a2376fa714/agent-sdk/xr-ai-tools/xr_ai_tools/vision.py#L149-L152)); there is no stale-handle test.
2. `_look_at_past_frame` returns `result["text"]` without checking `available is False` ([`scene_loop.py:950`](https://github.com/NVIDIA/xr-ai/blob/145f58a1957ccb9e30c9ef9c31bf02a2376fa714/agent-samples/xr-render-demo/worker/xr_render_demo_worker/scene_loop.py#L950)), and the recorded-frame two-step has schema-only test coverage.
3. `SampledVideoFrame` duplicates `path` as `image.uri` ([`video_memory.py:102-112`](https://github.com/NVIDIA/xr-ai/blob/145f58a1957ccb9e30c9ef9c31bf02a2376fa714/agent-sdk/xr-ai-tools/xr_ai_tools/video_memory.py#L102-L112)).
4. `ImageRegistry` has no byte budget or TTL and ignores `owner` on `resolve`.
5. Refactor residue: dead `LIVE_PERCEPTION_TOOL`/`PAST_PERCEPTION_TOOL` filters at [`__main__.py:137`](https://github.com/NVIDIA/xr-ai/blob/145f58a1957ccb9e30c9ef9c31bf02a2376fa714/agent-samples/xr-render-demo/worker/xr_render_demo_worker/__main__.py#L137) and [`eval.py:1413`](https://github.com/NVIDIA/xr-ai/blob/145f58a1957ccb9e30c9ef9c31bf02a2376fa714/agent-samples/xr-render-demo/eval/eval.py#L1413), and unused `encode_image` ([`_pixels.py:59`](https://github.com/NVIDIA/xr-ai/blob/145f58a1957ccb9e30c9ef9c31bf02a2376fa714/agent-sdk/xr-ai-tools/xr_ai_tools/_pixels.py#L59)).
6. Coverage dropped by the vision refactor: the negative no-image-data-in-LLM-events assertion, strict-request unknown-argument rejections, and the empty-answer branch.
7. The released `query_video` RPC operation was removed (its window ops are now `get_latest_video`/`get_historical_video`) and the name now denotes the VLM timed-frames tool, but [`migrations.md`](https://github.com/NVIDIA/xr-ai/blob/145f58a1957ccb9e30c9ef9c31bf02a2376fa714/docs/source/reference/migrations.md) has no row for it; out-of-tree callers get `unknown_operation` or a semantically different tool.

The latest/historical decomposition itself is an improvement: the shared `_DurationRequest`/`_FrameSamplingRequest` bases keep the 300 s and 256-frame bounds in one place, `_latest_window` clamps to the recorded range, and the updated tests cover both window types. The blockers above should be resolved before merge.

wenxind-nvidia:
Reviewed at `145f58a1957ccb9e30c9ef9c31bf02a2376fa714`.

## Blockers

1. **Simple VLM no longer handles a missing camera frame.** [`CurrentFrameTool.execute()` wraps `FrameUnavailable` as `RuntimeError`, but the agent catches only `FrameUnavailable`](https://github.com/NVIDIA/xr-ai/blob/145f58a1957ccb9e30c9ef9c31bf02a2376fa714/agent-samples/simple-vlm-example/worker/simple_vlm_example_worker/agent.py#L131-L155). I reproduced this locally. The task exits without publishing the previous user-facing “no camera frame” response.

2. **The advertised image limit is incompatible with the shipped VLM.** [`query_images` and `query_video` allow 256 images](https://github.com/NVIDIA/xr-ai/blob/145f58a1957ccb9e30c9ef9c31bf02a2376fa714/agent-sdk/xr-ai-tools/xr_ai_tools/vision.py#L35-L56), but the default Cosmos server permits only [four images per prompt](https://github.com/NVIDIA/xr-ai/blob/145f58a1957ccb9e30c9ef9c31bf02a2376fa714/services/vlm-server/vlm_server.yaml#L31-L39). A valid five-frame request therefore fails in the default deployment. The registry also defaults to 128 entries, so 129 opaque references cannot form a valid request.

3. **Temporal queries present estimated timestamps as exact.** The sampler [interpolates timestamps from chunk bounds and frame count](https://github.com/NVIDIA/xr-ai/blob/145f58a1957ccb9e30c9ef9c31bf02a2376fa714/services/video-memory-service/video_memory_service/service.py#L345-L357), while the recorder stores wall-clock chunk creation and flush times, not per-frame PTS. Those estimates are then supplied to the VLM as precise timestamps and offsets. Frame drops, jitter, or delayed flushes can produce incorrect temporal answers.

4. **The sampling workload remains insufficiently bounded.** [`frame_budget=256` with native resolution is legal](https://github.com/NVIDIA/xr-ai/blob/145f58a1957ccb9e30c9ef9c31bf02a2376fa714/agent-sdk/xr-ai-tools/xr_ai_tools/video_memory.py#L65-L86). The service sequentially decodes and writes every selected PNG, while the client retains a 30-second timeout and RPC has no cancellation message. A timed-out 4K request continues consuming GPU, CPU, and storage, with no output retention policy.

5. **FIFO pruning or one corrupt chunk aborts the entire sample.** The service snapshots chunk paths and later reads them outside the error handler; the hub can delete an old chunk between those operations. Decode failure or a zero-frame chunk also aborts instead of honoring the documented sparse-result behavior. Skip unavailable chunks and fail only when no usable frames remain.

6. **xr-render exposes selection tools that it cannot use.** [`NativeCapabilities.model`](https://github.com/NVIDIA/xr-ai/blob/145f58a1957ccb9e30c9ef9c31bf02a2376fa714/agent-samples/xr-render-demo/worker/xr_render_demo_worker/tools.py#L65-L87) includes the new raw recorded-frame/video selectors but excludes `query_image`, `query_images`, and `query_video`. The model cannot supply the trusted participant context or inspect the returned paths. Keep these selectors programmatic until participant-aware inference adapters exist.

7. **External image fallback expands the model tool’s trust boundary.** [`ImageReference` accepts arbitrary local paths, file URIs, and HTTP(S) URLs](https://github.com/NVIDIA/xr-ai/blob/145f58a1957ccb9e30c9ef9c31bf02a2376fa714/agent-sdk/xr-ai-tools/xr_ai_tools/image.py#L19-L83). If a query tool is model-exposed, prompt injection can read local images or trigger server-side URL fetching. External sources should require explicit trusted opt-in or restrictions; opaque registry references should be the safe default.

## Suggestion

The conceptual theme is coherent, but the PR also deletes the `live-vision` extra and three public modules while renaming the video-memory API. That is a breaking SDK change coupled to an additive feature. Separate the retirement or explicitly record it as a breaking decision in `docs/changelog.md`.

Verification: 84 focused tests passed locally, and all 14 GitHub checks are green. The missing-frame regression was reproduced separately.

nvddr:
Thanks @wenxind-nvidia and @blongs-nv for the detailed reviews. Addressed in `942e6bf`.

## Addressed

- **Missing-camera regression:** the simple VLM agent now recognizes `FrameUnavailable` preserved in the cause chain of the Relay `RuntimeError`, publishes the user-facing fallback, and terminates the voice response normally. Added a regression test for the wrapped exception.
- **Default VLM image limit:** `query_images` and `query_video` now accept at most four inputs, matching the shipped Cosmos configuration. The 256-frame video-memory budget remains a selection/export limit, not an inference limit.
- **Estimated timing:** recorded-frame timestamp fields and VLM timeline text now identify interpolated timestamps and offsets as estimates. The service and SDK docs explain that per-frame PTS is not persisted.
- **Pruned/corrupt chunks:** sampling now skips chunks that disappear, fail decode, or decode to zero frames, returning the usable subset and failing only when no frames remain. Added regression coverage for all three cases.
- **XR-render tool exposure:** raw current/recorded selectors and image-query primitives are no longer advertised to the reasoning model. Dispatch now also enforces the advertised model-tool allowlist, so a hallucinated hidden tool name cannot bypass schema filtering. The participant-aware perception facades continue to compose those tools internally.
- **External image trust boundary:** `ImageRegistry` now accepts only opaque registry handles by default. Local paths, file URIs, and HTTP(S) URLs require explicit `allow_external=True` in trusted application composition. XR-render uses that opt-in internally, behind the model-tool allowlist.
- **Error/result handling:** unresolved or disallowed image references report an image-input error rather than a VLM outage, and historical perception respects `available=False`.
- **Migration/documentation:** added migration rows for the retired vision tools, the recorded `query_video` RPC rename, and `get_frame_from_time`. Also documented the four-image inference limit, the separate 256-frame selection budget, native-resolution cost, timeout risk, and lack of retention/cancellation.
- **Code-quality finding:** moved image decoding out of the assertion so the assertion has no side effect.

## Intentionally not changed in this merge-focused pass

- **No increase to VLM multi-image capacity.** The SDK was capped to the deployed four-image contract instead. Raising the server limit requires model-specific memory and quality validation and is outside this tools PR.
- **No new byte/spatial budget, RPC cancellation protocol, output TTL, retention worker, or per-request output directory.** Those concerns are valid, but they are coupled service-lifecycle and protocol changes rather than a safe small correction. The current limits and operational risk are now explicit in the docs.
- **No recorder format change to persist per-frame PTS.** That crosses the hub recorder, stored metadata format, and existing recordings. This PR now tells the truth about interpolation rather than presenting estimates as exact.
- **No new real-NVDEC nightly GPU test.** The failure modes changed here are covered deterministically in CPU tests. A real recorded-chunk GPU fixture belongs with the nightly GPU infrastructure and should not hold this API cleanup.
- **No gap-snapping heuristic or overlap/interpolation refactor.** Those would change sampling semantics and need their own policy discussion and fixtures; they were suggestions rather than merge blockers.
- **No registry byte budget/TTL or owner-based authorization.** Opaque-only defaults plus model dispatch enforcement close the reported local-file/SSRF path. Registry lifecycle policy can be added separately without mixing it into this API migration.
- **No removal of the duplicated `path` field, dead-filter cleanup, or unrelated helper deletion.** `path` is retained for existing service callers while `image` supplies the shared inference contract; the remaining cleanup is non-blocking churn.
- **No separate changelog decision entry.** The actionable breaking changes are recorded in the public release migration table, which is the upgrade path out-of-tree callers need.

Validation: full local CPU suite `785 passed, 10 deselected`; focused suite `111 passed`; Ruff, SPDX, strict docs, lock validation, DCO, CodeQL, and both CI Python versions are green.

nvddr:
Follow-up in `30a5136`: the earlier disposition item that deferred removal of the duplicated frame `path` field is superseded by maintainer direction.

`HistoricalFrameResult` and `SampledVideoFrame` now expose the exported PNG location only as `image.uri`. The video-memory service emits that canonical shape directly; compatibility validators and the duplicate `path` field are removed. Tests assert that serialized frame results contain no `path`, the GPU recorder test uses `image.uri`, and the migration table documents the breaking field change.

The separate `RecordedVideoResult.path` remains because it identifies an H.264 clip rather than an image frame.

Validation: full CPU suite `785 passed, 10 deselected`; focused tests `51 passed`; all PR checks are green.

wenxind-nvidia:
Reviewed at `30a51365088d61dcb6d9b87678e53717d5fc83e4`.

## Blocker

**FIFO pruning can still abort sampling before the new per-chunk skip logic runs.**

[`_sample_frames()` calls `store.overlapping_chunks()` before entering the protected read/decode loop](https://github.com/NVIDIA/xr-ai/blob/30a51365088d61dcb6d9b87678e53717d5fc83e4/services/video-memory-service/video_memory_service/service.py#L225-L250). `ChunkStore.chunks()` first snapshots the `*.264` paths and then calls `_metadata()` for each path. If the hub prunes a chunk at that boundary, the sidecar is absent (or its read raises), `_metadata()` falls through to [`chunk.stat()`](https://github.com/NVIDIA/xr-ai/blob/30a51365088d61dcb6d9b87678e53717d5fc83e4/services/video-memory-service/video_memory_service/store.py#L65-L94), and `FileNotFoundError` aborts the entire RPC before the new handler can skip it. Latest-window requests have another occurrence in [`stats()`](https://github.com/NVIDIA/xr-ai/blob/30a51365088d61dcb6d9b87678e53717d5fc83e4/services/video-memory-service/video_memory_service/store.py#L97-L101), where the `path.stat()` default argument is evaluated even when `size_bytes` exists.

I reproduced the enumeration-boundary failure locally. The added pruning test patches `Path.read_bytes()`, so it covers deletion only after enumeration has completed and does not exercise this race. Please make chunk enumeration/stat collection skip entries that disappear, returning `not_found` only when no usable chunks remain, and add a test that deletes a chunk during metadata enumeration.

blongs-nv:
Reviewed at `30a51365088d61dcb6d9b87678e53717d5fc83e4` by a bot.

Blockers 2 and 3 from my previous review are resolved (external image URIs are opt-in; model-emitted tool names are enforced against the model-facing set), as are suggestions 2, 3 (the `path` duplication, removed in `30a5136`), and 7. Blocker 1 is fixed at the read/decode loop, which now skips unavailable and corrupt chunks, but the same prune race survives one step earlier, as wenxind-nvidia's latest review points out: a chunk deleted between the `*.264` glob and the sidecar read makes `_metadata` fall through to `chunk.stat()` and abort the whole request with an unhandled `FileNotFoundError`. I reproduced this at this head; their finding stands and the fix should extend to chunk enumeration. The deferrals in your disposition (registry lifecycle policy, dead-filter and helper cleanup, gap-snapping, retention/cancellation, per-frame PTS) all have reasonable rationale and I consider them accepted. One item from the disposition does not hold up under reproduction:

## Blockers

1. **The missing-camera fallback added in `942e6bf` cannot fire: the relay's `RuntimeError` carries no exception chain, so `_find_frame_unavailable` returns `None` and the handler re-raises.**

   [`agent-samples/simple-vlm-example/worker/simple_vlm_example_worker/agent.py:155-158`](https://github.com/NVIDIA/xr-ai/blob/30a51365088d61dcb6d9b87678e53717d5fc83e4/agent-samples/simple-vlm-example/worker/simple_vlm_example_worker/agent.py#L155-L158)
   I reproduced this at this head with the locked `nemo_relay`: raising `FrameUnavailable` inside a real `Tool.execute` surfaces as `RuntimeError("internal error: FrameUnavailable: ...")` with `__cause__` and `__context__` both `None`, so the cause-walk at [`agent.py:211-221`](https://github.com/NVIDIA/xr-ai/blob/30a51365088d61dcb6d9b87678e53717d5fc83e4/agent-samples/simple-vlm-example/worker/simple_vlm_example_worker/agent.py#L211-L221) finds nothing and the turn still dies with no spoken fallback. The regression test passes only because its fake tool raises `RuntimeError(...) from error`, a chaining the real pipeline does not perform; drive the test through a real `Tool` and match what the relay actually produces (for example the wrapped message), or preserve the cause in the relay layer.

## Suggestions

1. Nothing exercises the new "Image input unavailable — please select it again." path ([`vision.py:93-101`](https://github.com/NVIDIA/xr-ai/blob/30a51365088d61dcb6d9b87678e53717d5fc83e4/agent-sdk/xr-ai-tools/xr_ai_tools/vision.py#L93-L101)); note the pre-check does not pin entries, so an eviction between it and inference still yields the generic VLM message. The strict-request unknown-argument rejections, the empty-answer (`available=False`) branch, and a negative no-image-data assertion in the simple-vlm worker events also remain untested.

This round, together with `942e6bf`, lands the substance of both outstanding reviews, and making `image.uri` the canonical frame location with a migration row is the right resolution of the duplication. The one blocker above should be resolved before merge.

blongs-nv (approved):

PR #360, opened 2026-08-14
Allow wake phrases after sentence boundaries

wenxind-nvidia:
## Summary

- accept configured wake phrases at transcript start or after `.`, `?`, or `!`
- discard preceding background transcript text and the matched phrase before dispatch
- keep commas, semicolons, and ordinary mid-sentence mentions as non-matches
- apply the same sentence-boundary rule to partial-transcript wake probing

## Tests

- `uv run --project tests pytest -q tests/test_xr_ai_voicegate.py` (66 passed)
- `uv run --project tests pytest -q tests/test_voice_pipeline.py` (84 passed)

### DIFF
--- a/agent-samples/simple-vlm-example/README.md
+++ b/agent-samples/simple-vlm-example/README.md
@@ -78,8 +78,8 @@ tail -F /tmp/log_simple-vlm-example_*/relay-events.jsonl
 ```
 
 Voice-gate behavior remains in `yaml/voice_gate.yaml`. Wake phrases match at
-the start of the transcript or after sentence-final `.`, `?`, or `!`; preceding
-text is discarded before dispatch. Worker timing, frame freshness, the default
-`ping` question, and optional prompt overrides are in
-`yaml/simple_vlm_example_worker.yaml`; the default prompt ships inside the
-worker package.
+the start of the transcript or after sentence-final `.`, `?`, or `!` followed
+by whitespace or a closing quote; preceding text is discarded before dispatch.
+Worker timing, frame freshness, the default `ping` question, and optional prompt
+overrides are in `yaml/simple_vlm_example_worker.yaml`; the default prompt ships
+inside the worker package.
--- a/agent-samples/simple-vlm-example/README.md
+++ b/agent-samples/simple-vlm-example/README.md
@@ -77,7 +77,9 @@ sensitive data.
 tail -F /tmp/log_simple-vlm-example_*/relay-events.jsonl
 ```
 
-Voice-gate behavior remains in `yaml/voice_gate.yaml`. Worker timing, frame
-freshness, the default `ping` question, and optional prompt overrides are in
+Voice-gate behavior remains in `yaml/voice_gate.yaml`. Wake phrases match at
+the start of the transcript or after sentence-final `.`, `?`, or `!`; preceding
+text is discarded before dispatch. Worker timing, frame freshness, the default
+`ping` question, and optional prompt overrides are in
 `yaml/simple_vlm_example_worker.yaml`; the default prompt ships inside the
 worker package.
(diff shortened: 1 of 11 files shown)


blongs-nv:
Reviewed at `6ee53423d3f0ce2f0c18683f2937542c9e3e8dd2` by a bot.

## Blockers

1. **The early-wake probe permanently gives up in exactly the scenario this PR adds: background speech before the wake phrase.**

   [`agent-sdk/xr-ai-voice/xr_ai_voice/_processors/voice_gate.py:103`](https://github.com/NVIDIA/xr-ai/blob/6ee53423d3f0ce2f0c18683f2937542c9e3e8dd2/agent-sdk/xr-ai-voice/xr_ai_voice/_processors/voice_gate.py#L103)
   When a partial probe catches only the background portion ("we should test it." or any mid-background prefix), `could_match_magic_phrase` returns False (empty or non-prefix current sentence), `handle_partial_transcript` returns None, and [`vad_stt.py:346`](https://github.com/NVIDIA/xr-ai/blob/6ee53423d3f0ce2f0c18683f2937542c9e3e8dd2/agent-sdk/xr-ai-voice/xr_ai_voice/_processors/vad_stt.py#L346) stops all remaining probes, so a wake phrase spoken after that sentence never gets an early chime. I reproduced both cases: `_current_sentence("we should test it.")` is `""` and a non-prefix background sentence also yields None. Under sentence-boundary matching, "cannot become a match" is never final (a new sentence can always start), so the None short-circuit should only fire when it is actually terminal, or the probe should keep probing (return False) until attempts run out.

## Suggestions

1. The boundary class matches `.` with no following whitespace, so a wake word glued to a dot activates the gate and dispatches the remainder: I reproduced `strip_magic` on "visit www.agent.com now" returning "com now" as a query ([`utils/xr-ai-voicegate/xr_ai_voicegate/_phrases.py:43`](https://github.com/NVIDIA/xr-ai/blob/6ee53423d3f0ce2f0c18683f2937542c9e3e8dd2/utils/xr-ai-voicegate/xr_ai_voicegate/_phrases.py#L43)). Requiring at least one whitespace/quote character after a non-`^` boundary would keep every intended case in the new tests passing while excluding dot-adjacent tokens.

2. Add a partial-probe test for a transcript ending exactly at a sentence boundary ("background." → needs-more-audio, not terminal); the new `test_partial_phrase_after_sentence_boundary_can_continue_matching` only covers the case where the phrase prefix is already present after the boundary.

The change itself is well contained: one shared regex builder, the gate ladder untouched, docs/config/changelog updated together, and the new unit and feed-level tests cover the stripping semantics well (66 pass locally). The blocker above should be resolved before merge.

blongs-nv (approved):
Reviewed at `3e99dd147ba1a427d6d8252a3945b98bda5773fa` by a bot.

The blocker and both suggestions from my previous review are resolved: the partial probe now keeps probing after background-only or boundary-ending partials (with pipeline tests covering both), and the boundary regex requires whitespace or a closing delimiter after `.`, `?`, or `!`, so dot-adjacent tokens like `www.agent.com` no longer wake the gate.

## Suggestions

1. `handle_partial_transcript` now only returns `bool`, leaving the `None` handling vestigial: `PartialTranscriptHandler` is still typed `Awaitable[bool | None]` ([`agent-sdk/xr-ai-voice/xr_ai_voice/_processors/vad_stt.py:50`](https://github.com/NVIDIA/xr-ai/blob/3e99dd147ba1a427d6d8252a3945b98bda5773fa/agent-sdk/xr-ai-voice/xr_ai_voice/_processors/vad_stt.py#L50)) and the `decision is None` early-return at [`vad_stt.py:346`](https://github.com/NVIDIA/xr-ai/blob/3e99dd147ba1a427d6d8252a3945b98bda5773fa/agent-sdk/xr-ai-voice/xr_ai_voice/_processors/vad_stt.py#L346) is unreachable. Tightening the type to `Awaitable[bool]` and dropping the dead branch would keep the probe contract honest.

The fix is exactly scoped, the docs and changelog were updated to match the new separator rule, and the full voicegate and voice-pipeline suites pass locally (153 tests). Ready to merge; the suggestion above is optional cleanup.

PR #362, opened 2026-08-14
Add configurable OCR tool with Nemotron OCR v2

wenxind-nvidia:
## Summary
- add a typed OCRService and make_ocr profile factory
- support Nemotron OCR v2 through local/Hugging Face-backed NIM or hosted NVIDIA inference
- add a VLM OCR adapter so the native OCR tool can switch models without caller changes
- add structured OCR results, documentation, and CPU-only contract tests

## Validation
- 862 passed, 10 deselected (full non-GPU suite)
- ruff check .
- SPDX header check


blongs-nv:
Reviewed at `54c080f482139c7df25282939629386c97a86421` by a bot.

## Suggestions

1. `read_image_text` is model-invokable and accepts any local path or HTTP(S) URL, and `NvidiaOCR` fetches URLs server-side with redirects followed, so a prompt-injected call can probe internal endpoints or feed local PNG/JPEG files to the model.

   [`agent-sdk/xr-ai-tools/xr_ai_tools/ocr.py:18-21`](https://github.com/NVIDIA/xr-ai/blob/54c080f482139c7df25282939629386c97a86421/agent-sdk/xr-ai-tools/xr_ai_tools/ocr.py#L18-L21), [`agent-sdk/xr-ai-models/xr_ai_models/_ocr.py:111`](https://github.com/NVIDIA/xr-ai/blob/54c080f482139c7df25282939629386c97a86421/agent-sdk/xr-ai-models/xr_ai_models/_ocr.py#L111)
   A one-sentence untrusted-input note in the tool README (and, if cheap, an opt-in restriction on accepted sources) would make the boundary explicit for integrators.

2. `health_path` is silently ignored on the VLM-backed OCR path: `VLMOCR.health` delegates to the VLM client's hardcoded `/health`, so a profile that sets `health_path` on an `openai_compat` OCR entry probes a different URL than configured.

   [`agent-sdk/xr-ai-models/xr_ai_models/_config.py:584-586`](https://github.com/NVIDIA/xr-ai/blob/54c080f482139c7df25282939629386c97a86421/agent-sdk/xr-ai-models/xr_ai_models/_config.py#L584-L586), [`agent-sdk/xr-ai-models/xr_ai_models/_ocr.py:173`](https://github.com/NVIDIA/xr-ai/blob/54c080f482139c7df25282939629386c97a86421/agent-sdk/xr-ai-models/xr_ai_models/_ocr.py#L173)
   Either reject `health_path`/`request_path` for the `openai_compat` kind or document them as `nvidia_ocr`-only.

3. The response-contract error paths are untested: nothing exercises `_parse_nvidia_response` rejecting a malformed NIM reply (missing `text_detections`, more than one image result) or `VLMOCR` returning an empty transcription (empty `detections` tuple).

   [`agent-sdk/xr-ai-models/xr_ai_models/_ocr.py:204`](https://github.com/NVIDIA/xr-ai/blob/54c080f482139c7df25282939629386c97a86421/agent-sdk/xr-ai-models/xr_ai_models/_ocr.py#L204)
   One parametrized malformed-response test plus one empty-text VLM test would pin the contract.

4. The tests project never installs the new `xr-ai-tools[ocr]` extra; `tests/test_ocr_tool.py` imports work only because `xr-ai-models` is a direct test dependency, so a typo in the extra's metadata would ship unnoticed.

   [`tests/pyproject.toml:18`](https://github.com/NVIDIA/xr-ai/blob/54c080f482139c7df25282939629386c97a86421/tests/pyproject.toml#L18)
   Add `ocr` to the installed extras.

The change fits the model-SDK architecture cleanly: the non-chat `/v1/ocr` wire format stays isolated behind `OCRService`, backend switching (local NIM, hosted, VLM fallback) lives entirely in the model profile, and the wire-level tests are thorough with docs, changelog, and the AGENTS.md rule amended in the same change. I ran the OCR, config, and alias suites in a fresh clone (60 passed). No blockers; this is ready to merge.

blongs-nv (approved):

nvddr (commented):
Reviewed at `301218d`. This is a **text-only review comment**; I am not changing approval state.

## Disposition

OCR is a real need, especially for instrument readings, but I do not think this should merge as a shared SDK capability in its current shape. The low-level adapter is mostly careful; the larger problem is that the PR lands 2,203 lines across 23 files and a large public surface before wiring or evaluating a concrete consumer, while leaving the hardest operational and product questions outside the implementation.

## Main issues to address

1. **Start from a concrete workflow, then extract the shared contract.** I cannot find an in-tree sample or service that constructs `make_ocr`, `OCRTool`, or `OCRTools`; the additions are definitions, docs, and contract tests. That is at odds with the repository guidance not to add an abstraction until two concrete use cases need it, and it makes the new public types/tools difficult to judge. In particular, this does not yet show that OCR improves lab instrument monitoring. OCR recognizes text; it does not solve the marker-to-physical-display association failure. With multiple markers and one reading, a generic agent-selected crop can still assign the same reading to both devices.

   I would first integrate and evaluate an end-to-end path in the lab sample (or another actual consumer): deterministic marker detection -> joint display/marker localization -> deterministic per-display crop -> OCR -> short multi-frame consensus. Measure exact/normalized reading accuracy, false association rate, warm latency, cold-start latency, and VRAM. Keep the app-specific orchestration private. Promote the smallest stable OCR service/tool contract only once another consumer demonstrates the same need. At minimum, split the unrelated generic crop/public API expansion from the vendor adapter.

2. **Own local model serving instead of documenting an unmanaged side process.** The local example uses `deployment.ownership: external` and tells the user to launch NIM manually (`agent-sdk/xr-ai-models/README.md:225-245`). That bypasses the repository's existing model-server lifecycle: launch/reuse, readiness, fingerprinting, persistent reuse, credentials, GPU selection, restart/stop behavior, and resource planning. It also uses `localhost:8000`, which conflicts with the XR Media Hub token server used by the lab sample and other samples (`token_server_port` defaults to 8000).

   Local OCR should be an opt-in orchestrator process/profile using the existing NIM server pattern, on a non-conflicting host port, with a pinned image/config, explicit cache and credential forwarding, explicit `cuda_visible_devices`/`NIM_ENGINE_DEVICES`, and readiness. Do not add it to every default model stack. Hosted OCR remains a valid zero-local-VRAM profile.

   GPU behavior needs to be part of that design, not left as a README footnote. NVIDIA's current support matrix reports roughly 1.57-3.33 GiB startup VRAM for latency configurations depending on GPU/variant, while runtime memory grows with batching, engine count, input buffers, and CUDA graph capture. The throughput configuration defaults to more engines/larger batches, and the current NIM supports one selected GPU rather than multi-GPU execution. For an XR workload I would default to and evaluate the latency configuration, explicitly place it (including co-location with the VLM), and include cold/warm startup and memory measurements. Relevant NVIDIA docs: [support matrix](https://docs.nvidia.com/nim/ingestion/image-ocr/latest/support-matrix.html), [configuration](https://docs.nvidia.com/nim/ingestion/image-ocr/latest/configuration.html), and [environment variables](https://docs.nvidia.com/nim/ingestion/image-ocr/latest/environment-variables.html).

3. **OCR endpoint failures currently abort the agent turn.** `OCRTool._read()` awaits `self.ocr.recognize()` without handling timeout, transport, HTTP, or response-validation failures (`ocr.py:293-297`). Those propagate out of the tool loop. `OCRResult` already has `available/message`, and `ImageQueryTool` converts inference exceptions into a deterministic unavailable result. Match that established behavior and add regression tests for timeout, connection/HTTP failure, and malformed provider output at the tool boundary.

4. **There is an owner-release race while OCR is in flight.** The tool captures the image owner and bytes, awaits remote inference, and then inserts a result under that owner. If participant/workflow cleanup occurs during the await, `release_owner()` runs before the result exists and the completed request recreates stale owner-associated state. `ImageCropTool` avoids the analogous problem by revalidating the source through `put_derived()` after its await. Revalidate that the source reference is still live before publishing the OCR result (or guarantee/cancel all owner-scoped in-flight calls), and test release during inference.

5. **The VLM OCR prompt belongs with application/tool policy, not generic adapter connectivity.** Adding `prompt` to `AdapterSpec` mixes task behavior into the model adapter schema (`_config.py:62-63`). The existing vision path injects `system_prompt` at the tool/application boundary, where prompts can be versioned with workflow-specific evals. OCR transcription instructions materially affect whether units, punctuation, display labels, and nearby-device text are returned. Keep a minimal fixed fallback prompt private to the adapter, or pass policy from the consuming tool/application, and add prompt-quality cases to the instrument eval rather than making it global model configuration.

## What is aligned

The PR does several things well: vendor HTTP remains behind a typed `xr_ai_models` service, the tool stays native/in-process with an injected service, image/result bytes stay behind opaque references, model-visible output is bounded, external image inputs are disabled by default, redirects are disabled, and the VLM fallback does not invent geometry or confidence. The `/v1/ocr` contract and normalized geometry also match NVIDIA's [OCR API](https://docs.nvidia.com/nim/ingestion/image-ocr/latest/api-reference.html). Those are good implementation foundations; they just need to be attached to a proven workflow and a repository-owned deployment/lifecycle before becoming a broad public SDK contract.

nvddr (changes requested):

PR #363, opened 2026-08-14
Add lab instrument monitoring agent sample

nvddr:
## Summary

- add a lab instrument monitoring sample based on the tea-making agent pattern
- keep general visual monitoring and instrument tracking dormant until the foreground agent calls participant-scoped controls
- share participant lifecycle, frame acquisition, image registration, and QR/ArUco detection through one `ParticipantImageAgent`
- map five QR payloads and five ArUco IDs to `Device1` through `Device5` in `yaml/device_map.yaml`
- redact unmapped marker identities instead of exposing raw QR text or fallback labels
- read only the display physically belonging to the highlighted instrument; ambiguous adjacent displays return `UNKNOWN`
- track devices by marker family and ID, emit discoveries and value/unit changes, report loss once, and snapshot full state every 10 seconds
- route foreground replies and instrument alerts through participant-scoped voice aggregation
- stream `current_view` directly from the original user query through `StreamingImageQueryTool`
- default to wake-word activation and publish spoken text on `agent.response` for accessible client output
- persist durable JSONL and expose selected typed events in the shared live browser viewer

## Architecture

The worker composes `VoiceAgent`, `VoiceAggregationAgent`, `ParticipantImageAgent`, `MonitorAgent`, `LabInstrumentAgent`, `InstrumentMonitorAgent`, `InstrumentAlertAgent`, `ForegroundAgent`, `FileOutputAgent`, and the explicit web-events adapter through typed runtime topics.

`ParticipantImageAgent` owns current-frame acquisition, the shared image registry, participant cleanup, and `MarkerTrackingTool` configured for QR and ArUco. `LabInstrumentAgent` scans and annotates the same acquired frame, resolves only configured identities, and combines those identities with same-device VLM display reads.

`InstrumentMonitorAgent` owns continuous tasks and participant-scoped state. It compares both reading value and unit, keeps a learned unit when the VLM omits it, suppresses unchanged devices popping in and out of view, and publishes distinct change, lost-device, and state topics. `VoiceAggregationAgent` paces and combines candidate speech before `VoiceAgent` sends it to TTS.

The foreground remains a bounded tool-calling agent with only the current request. Meter-reading requests use the marker-aware one-frame flow. When it selects `current_view`, the selected frame and original user query stream directly through the VLM without a second LLM pass.

Nemotron-3 Nano Omni on port 8108 handles foreground routing; Cosmos3 Nano Reasoner on port 8100 handles visual inference.

## Stack

This PR is temporarily based on PR #385. PR #385 supplies participant-scoped voice aggregation and now contains the merged PR #386 shared live-events viewer. After PR #385 merges, this PR can be rebased directly onto `main`.

## Review disposition

- monitoring policy now lives in the VLM system prompt; user focus and prior captions are JSON-encoded data
- direct current-view queries use a packaged VLM prompt and the original user utterance
- instrument changes compare `(value, unit)` before state mutation
- executable tool-call evals validate every argument schema and include live visual prompt cases
- concurrent participant teardown, marker redaction, role wording, wake-word defaults, client captions, and aggregation cleanup are covered
- selected-image participant ownership remains an SDK contract decision and is not silently defined by this sample

## Live outputs

- the authenticated connection client remains required at `http://localhost:8080`
- spoken assistant text appears in its Agent Output through `agent.response`
- selected structured runtime events are visible at `http://127.0.0.1:8092`
- the event viewer is loopback-only, bounded, and non-durable; JSONL remains the durable record

## Validation

- 14 affected sample tests pass
- Ruff 0.15.16 and Python compilation pass
- generated dependency map is current
- worker dependency resolution passes
- `git diff --check` passes

## Notes

No live GPU, camera, or end-to-end model pipeline was run in the development environment.


wenxind-nvidia (commented):
## Review

### Blockers

1. **Marker detection and VLM annotation can use different frames.**

   [`instruments.py:81-95`](https://github.com/NVIDIA/xr-ai/blob/a453a3a68c7d39e474cfaf417683d4b8c6dfc3e0/agent-samples/lab-instrument-monitoring/worker/lab_instrument_monitoring_worker/instruments.py#L81-L95) acquires and saves one frame, then `track_markers` independently acquires the latest frame. At camera frame rate, the marker coordinates can therefore come from a newer frame but be applied to the older image. Motion or resolution changes can mark the wrong region or reject the polygon entirely. It also means the saved JPEG is not necessarily the frame used for marker detection, contrary to the documentation.

   Acquire once and run detection against that exact image. Please add a regression test that publishes a newer frame between the two operations.

2. **A previous unit overrides an explicitly observed current unit.**

   [`instrument_monitor.py:109-116`](https://github.com/NVIDIA/xr-ai/blob/a453a3a68c7d39e474cfaf417683d4b8c6dfc3e0/agent-samples/lab-instrument-monitoring/worker/lab_instrument_monitoring_worker/instrument_monitor.py#L109-L116) only parses the current suffix when `previous_unit` is empty. For example, `"1000 mV"` after `"1 V"` becomes `"1000 V"`, and `"1 A"` after `"1 V"` remains `"1 V"`.

   Parse an explicit current unit first and use the previous unit only when the new reading omits one. The parser should also recognize `Ω`, which currently becomes unitless.

### Suggestions

- Marker snapshots are always enabled and unbounded. [`app.py:123-129`](https://github.com/NVIDIA/xr-ai/blob/a453a3a68c7d39e474cfaf417683d4b8c6dfc3e0/agent-samples/lab-instrument-monitoring/worker/lab_instrument_monitoring_worker/app.py#L123-L129) enables them for every invocation, while the default monitor runs every five seconds. That creates 17,280 JPEGs per active participant per day. Make debug capture opt-in, sampled, or retention-bounded.

- Correct the public quickstart: [`quickstart.md:173-176`](https://github.com/NVIDIA/xr-ai/blob/a453a3a68c7d39e474cfaf417683d4b8c6dfc3e0/docs/source/getting_started/quickstart.md#L173-L176) says QR-only and "accepted" transcripts, but the implementation supports QR and ArUco and persists every final pre-gate STT transcript. The sample README also references a nonexistent tea-making sample.

### Verification

- 124 focused sample/runtime/tool/voice tests passed; 2 hub-fixture integration tests were deselected.
- Ruff and `git diff --check` passed.
- The full repository test environment cannot install on Apple Silicon because `PyNvVideoCodec` has no macOS wheel.
- The live routing eval was not run because it requires the model servers.

nvddr:
Review disposition (fffd2ec2):

- Addressed the frame-consistency blocker. LabInstrumentAgent now selects one frame and passes that exact image reference through marker detection, debug capture, polygon annotation, and VLM reading. MarkerTrackingTool retains its current-frame mode and adds an optional selected-image path backed by the shared ImageRegistry. A regression proves a newer live frame cannot replace the selected scan image.
- Addressed unit normalization. Explicit units in the latest reading now take precedence over retained units; retained units are used only when the latest reading omits one. Added mV, A, and Ω regressions.
- Marker snapshots are now opt-in through capture_marker_scans and disabled by default.
- Corrected the quickstart and sample documentation for QR plus ArUco support, final pre-gate transcripts, the model-server prerequisite, and debug-capture behavior.

Verification: 27 focused marker/sample tests pass; repository Ruff and git diff checks pass.

wenxind-nvidia:
## Review

Reviewed current head `7f7961c` against temporary base `b7c2a0b`. Verdict: **request changes**.

The previous frame-consistency and unit-parsing blockers are fixed. Debug captures are now opt-in, and the earlier documentation issues are corrected.

### Blockers

1. **The monitoring policy is passed as untrusted user content, not a system prompt.**

   [`MonitorAgent` constructs `ImageQueryTool` without `system_prompt`](https://github.com/NVIDIA/xr-ai/blob/7f7961c4bda933021b2825ce016c2a5803c82856/agent-samples/lab-instrument-monitoring/worker/lab_instrument_monitoring_worker/monitor.py#L102), then concatenates the policy, model-generated monitoring focus, and previous VLM caption into one user query at [lines 261–269](https://github.com/NVIDIA/xr-ai/blob/7f7961c4bda933021b2825ce016c2a5803c82856/agent-samples/lab-instrument-monitoring/worker/lab_instrument_monitoring_worker/monitor.py#L261-L269).

   Both focus and previous caption can contain instruction-like content, allowing them to override the JSON/change-detection contract and produce false alerts or persistent parse errors.

   Pass the fixed monitor policy through `ImageQueryTool(system_prompt=...)`, and send focus/previous caption as structured user data. Add live-model cases covering adversarial focus and visible prompt-like text.

2. **Direct current-view responses bypass every grounding and speech-style rule.**

   [`StreamingImageQueryTool` is created with an empty system prompt](https://github.com/NVIDIA/xr-ai/blob/7f7961c4bda933021b2825ce016c2a5803c82856/agent-samples/lab-instrument-monitoring/worker/lab_instrument_monitoring_worker/foreground.py#L190-L192), and [the raw user query is streamed directly to voice](https://github.com/NVIDIA/xr-ai/blob/7f7961c4bda933021b2825ce016c2a5803c82856/agent-samples/lab-instrument-monitoring/worker/lab_instrument_monitoring_worker/foreground.py#L531-L546). The foreground prompt's “visible evidence,” plain-speech, and two-sentence rules reach only the routing LLM. Because the second LLM pass was removed, nothing applies those rules to the VLM output.

   This differs from `simple-vlm-example`, which supplies a dedicated vision system prompt. Add a packaged current-view prompt and pass it to `StreamingImageQueryTool`, with output-behavior eval coverage.

3. **A unit-only measurement change mutates state without emitting a change event.**

   At [lines 315–323](https://github.com/NVIDIA/xr-ai/blob/7f7961c4bda933021b2825ce016c2a5803c82856/agent-samples/lab-instrument-monitoring/worker/lab_instrument_monitoring_worker/instrument_monitor.py#L315-L323), the stored unit/display are updated before equality checks compare only the numeric value.

   I reproduced `1 V → 1 A`: emitted events were only `["discovered"]`, while state silently became `1 A`. That suppresses the voice alert and change record for a materially different measurement.

   Compare the complete normalized identity—at least `(value, unit)`—before mutating prior state, and add regression cases for `1 V → 1 A` and `12 V → 12 mV`.

### Suggestions

- **Enforce selected-image ownership.** [`MarkerTrackingTool` resolves the supplied reference without checking its owner against `participant_id`](https://github.com/NVIDIA/xr-ai/blob/7f7961c4bda933021b2825ce016c2a5803c82856/agent-sdk/xr-ai-tools/xr_ai_tools/marker_tracking.py#L350-L359). I registered an image owned by `alice` and successfully scanned it using `participant_id="bob"`. Add an owner-aware registry resolution API or remove the misleading participant-scoped contract for selected images.

- **Make the routing eval validate executable calls.** The harness [checks only the first tool name](https://github.com/NVIDIA/xr-ai/blob/7f7961c4bda933021b2825ce016c2a5803c82856/agent-samples/lab-instrument-monitoring/eval/eval.py#L43-L52). It would pass if the model omitted invalid arguments. This matters because [`current_view.question` is required](https://github.com/NVIDIA/xr-ai/blob/7f7961c4bda933021b2825ce016c2a5803c82856/agent-samples/lab-instrument-monitoring/worker/lab_instrument_monitoring_worker/foreground.py#L92-L98) but [its value is ignored](https://github.com/NVIDIA/xr-ai/blob/7f7961c4bda933021b2825ce016c2a5803c82856/agent-samples/lab-instrument-monitoring/worker/lab_instrument_monitoring_worker/foreground.py#L354-L362). Either use the argument or make current-view argument-free, then validate every eval call against the actual request model.

- **Cover all prompt-driven behavior.** The live eval covers only foreground routing. Add cases for monitor baseline/change/unchanged decisions and the instrument reader's same-device/`UNKNOWN` rules, as required for prompt-driven samples.

- **Correct the lifecycle ordering claim.** The guide says tasks are cancelled before image resources are released at [lines 241–243](https://github.com/NVIDIA/xr-ai/blob/7f7961c4bda933021b2825ce016c2a5803c82856/docs/source/reference/lab-instrument-monitoring.md#L241-L243), but runtime subscribers fan out concurrently. Either introduce coordinated teardown or document that cleanup is concurrent and handlers must tolerate released references.

- **Avoid logging raw unmapped QR payloads.** The code logs every decoded marker at [lines 105–110](https://github.com/NVIDIA/xr-ai/blob/7f7961c4bda933021b2825ce016c2a5803c82856/agent-samples/lab-instrument-monitoring/worker/lab_instrument_monitoring_worker/instruments.py#L105-L110) and repeats unmapped values at [lines 172–178](https://github.com/NVIDIA/xr-ai/blob/7f7961c4bda933021b2825ce016c2a5803c82856/agent-samples/lab-instrument-monitoring/worker/lab_instrument_monitoring_worker/instruments.py#L172-L178). Arbitrary QR content can contain URLs, credentials, or large payloads. Log the family plus a bounded hash/redacted identifier instead.

### Nits

- The sample README says the eval requires `agent-llm`, but this profile uses the `llm` role backed by Omni/model-servers: [README line 154](https://github.com/NVIDIA/xr-ai/blob/7f7961c4bda933021b2825ce016c2a5803c82856/agent-samples/lab-instrument-monitoring/README.md#L154).

- Selected-image scans still describe failures as occurring in the “current camera frame”: [`MarkerTrackingRequest`](https://github.com/NVIDIA/xr-ai/blob/7f7961c4bda933021b2825ce016c2a5803c82856/agent-sdk/xr-ai-tools/xr_ai_tools/marker_tracking.py#L43-L56) and [the empty result message](https://github.com/NVIDIA/xr-ai/blob/7f7961c4bda933021b2825ce016c2a5803c82856/agent-sdk/xr-ai-tools/xr_ai_tools/marker_tracking.py#L380-L383) should distinguish live versus supplied images.

### Validation

- 28 sample and marker tests passed.
- 16 voice-aggregation tests passed.
- Ruff and `git diff --check` passed.
- CI's Python 3.11/3.12, strict build, Ruff, SPDX, CodeQL, and lock checks passed; Pages publishing was still pending.
- The live model-routing eval was not run because it requires model servers.

nvddr:
Addressed the latest sample-level review in `d1f15f5` and added the viewer integration in `c90fdb2`:

- monitoring policy is a VLM system prompt; user focus and prior captions are JSON-encoded data
- direct current-view uses a packaged VLM prompt and the original user utterance
- reading changes compare both value and unit before state mutation
- executable tool-call arguments and live visual prompt behavior are covered
- participant teardown, unmapped marker redaction, role wording, wake-word defaults, client `agent.response`, and aggregation cleanup are addressed

The branch is restacked on #386. The explicit lab event adapter feeds the live loopback viewer while JSONL remains durable. Focused validation: 14 tests, Ruff, compilation, dependency-map, worker resolution, and diff checks pass.

The selected-image ownership suggestion is a real SDK contract question rather than a sample-only fix: `ImageRegistry.owner` is currently lifecycle metadata, not documented access control. I have not silently chosen cross-participant/ownerless/external-reference semantics; the recommended follow-up is a separate SDK PR with explicit owner-aware resolution and the selected-vs-live wording update.

wenxind-nvidia:
## Re-review

Reviewed head `4568f5b` against base `68dcd6b`. Verdict: **request changes**.

The three earlier implementation blockers are fixed: monitor policy isolation, current-view grounding, and value-plus-unit comparison. The following issues remain.

### Blockers

1. **Unreadable displays are treated as missing instruments.**

   [`UNKNOWN` readings are discarded](https://github.com/NVIDIA/xr-ai/blob/4568f5b056491bc37eb54bd305e924e17f923567/agent-samples/lab-instrument-monitoring/worker/lab_instrument_monitoring_worker/instruments.py#L145-L150), and [only successful readings refresh tracking state](https://github.com/NVIDIA/xr-ai/blob/4568f5b056491bc37eb54bd305e924e17f923567/agent-samples/lab-instrument-monitoring/worker/lab_instrument_monitoring_worker/instrument_monitor.py#L232-L243). A still-visible marker whose display is obscured for longer than `instrument_lost_after_s` therefore produces a false `InstrumentLost` alert.

   Return mapped marker sightings independently from successful readings. Refresh `last_seen` and tracking state from sightings while updating the measurement only from valid readings. Add a regression covering a visible marker that returns `UNKNOWN` beyond the loss timeout.

2. **The instrument-reading policy is not a system prompt.**

   [`ImageQueryTool` receives no system prompt](https://github.com/NVIDIA/xr-ai/blob/4568f5b056491bc37eb54bd305e924e17f923567/agent-samples/lab-instrument-monitoring/worker/lab_instrument_monitoring_worker/instruments.py#L55-L66), while [the fixed same-device, output-format, and `UNKNOWN` rules are sent as user content](https://github.com/NVIDIA/xr-ai/blob/4568f5b056491bc37eb54bd305e924e17f923567/agent-samples/lab-instrument-monitoring/worker/lab_instrument_monitoring_worker/instruments.py#L190-L216). Instruction-like text visible on an instrument has the same priority as that policy and can corrupt persisted readings and voice alerts.

   Move the fixed reading contract into a packaged system prompt, pass marker identity as structured data, and add an adversarial visible-text case to the live VLM eval.

3. **The current-view prompt still lacks a behavioral eval.**

   [`current_view_prompt.txt`](https://github.com/NVIDIA/xr-ai/blob/4568f5b056491bc37eb54bd305e924e17f923567/agent-samples/lab-instrument-monitoring/worker/lab_instrument_monitoring_worker/prompts/current_view_prompt.txt) adds grounding, prompt-injection, formatting, and response-length rules, but [`visual_eval.py` supports only monitor and instrument cases](https://github.com/NVIDIA/xr-ai/blob/4568f5b056491bc37eb54bd305e924e17f923567/agent-samples/lab-instrument-monitoring/eval/visual_eval.py#L84-L130). The unit test merely searches the prompt for expected phrases.

   Add live cases for unavailable detail, visible instruction text, and the plain two-sentence response contract.

4. **The required Ruff check fails.**

   Setting [`target-version = "py311"`](https://github.com/NVIDIA/xr-ai/blob/4568f5b056491bc37eb54bd305e924e17f923567/ruff.toml#L11-L12) makes Ruff recognize `tomllib` as part of the standard library, but [the dependency-map script still separates that import](https://github.com/NVIDIA/xr-ai/blob/4568f5b056491bc37eb54bd305e924e17f923567/.github/scripts/generate_dependency_map.py#L11-L21). Ruff 0.15.16 reports `I001`.

   Move `import tomllib` into the standard-library import block.

### Suggestions

- **Correct the viewer documentation.** The [sample README says output goes only to files and there is no dashboard](https://github.com/NVIDIA/xr-ai/blob/4568f5b056491bc37eb54bd305e924e17f923567/agent-samples/lab-instrument-monitoring/README.md#L11-L16), while its architecture now includes the event viewer. The [quickstart likewise says no sample-specific monitoring UI is served](https://github.com/NVIDIA/xr-ai/blob/4568f5b056491bc37eb54bd305e924e17f923567/docs/source/getting_started/quickstart.md#L169-L177).

- **Keep the selected-image ownership follow-up.** [`MarkerTrackingTool` resolves a supplied image without verifying its owner against `participant_id`](https://github.com/NVIDIA/xr-ai/blob/4568f5b056491bc37eb54bd305e924e17f923567/agent-sdk/xr-ai-tools/xr_ai_tools/marker_tracking.py#L339-L359). The cross-participant lookup reproduced previously remains possible.

- **Harden foreground tool-result handling.** Monitoring captions can contain visible instruction-like text and are later passed back to the foreground LLM. State in the system prompt that tool results are untrusted data and add a malicious-history eval case.

### Nits

- Documentation should say reading-change events occur when either the normalized value **or unit** changes.

- Marker-tracking request/result documentation should distinguish selected-image scans from current-camera scans.

### Validation

- 60 focused sample, marker, voice-aggregation, and web-event tests passed.
- `git diff --check` passed.
- Ruff 0.15.16 reproduced the CI `I001` failure.
- The full local environment remains unavailable on Apple Silicon because `PyNvVideoCodec` has no macOS wheel.
- Live LLM/VLM evals were not run because they require the model servers.

blongs-nv:
Reviewed at `4568f5b056491bc37eb54bd305e924e17f923567` by a bot. I independently verified the review-fix commit in code (monitor policy on the system channel with structured user data, the packaged current-view grounding prompt, `(value, unit)` change identity with the `1 V → 1 A` regressions, request-model-validated evals) and won't repeat what other reviews on this thread already cover. The two findings below are additional.

## Blockers

1. **One failing event subscriber silently kills the monitoring loops.**

   [`instrument_monitor.py:245-254`](https://github.com/NVIDIA/xr-ai/blob/4568f5b056491bc37eb54bd305e924e17f923567/agent-samples/lab-instrument-monitoring/worker/lab_instrument_monitoring_worker/instrument_monitor.py#L245-L254), [`monitor.py:249-262`](https://github.com/NVIDIA/xr-ai/blob/4568f5b056491bc37eb54bd305e924e17f923567/agent-samples/lab-instrument-monitoring/worker/lab_instrument_monitoring_worker/monitor.py#L249-L262)
   `runtime.publish` re-raises subscriber exceptions to the publisher, and neither `_maintenance_loop` (lost/snapshot publishes) nor `MonitorAgent._monitor` (record publish) contains them — unlike `_scan_loop` ([`instrument_monitor.py:232-243`](https://github.com/NVIDIA/xr-ai/blob/4568f5b056491bc37eb54bd305e924e17f923567/agent-samples/lab-instrument-monitoring/worker/lab_instrument_monitoring_worker/instrument_monitor.py#L232-L243)). The subscriber set on these topics has now grown to three (file output, aggregation-backed alerts, and the new web-events adapter), so one `OSError` from a JSONL append or one failure in the web projection kills that participant's monitoring with only a log line — no spoken notice, while the user believes the watch is running. Contain publish failures the same way the scan loop does.

2. **The `spoken` deduplication heuristic misfires on multi-iteration turns, repeating or silencing speech.**

   [`foreground.py:353-359`](https://github.com/NVIDIA/xr-ai/blob/4568f5b056491bc37eb54bd305e924e17f923567/agent-samples/lab-instrument-monitoring/worker/lab_instrument_monitoring_worker/foreground.py#L353-L359)
   `_answer` reports `spoken = result.return_direct and calls == [CURRENT_VIEW_TOOL]`, comparing the whole turn's call list, but the tool loop only requires a return-direct call to be alone in its *batch*. When the model uses another tool first (e.g. `recent_visual_history`, then `current_view` alone: "what changed, and what are you looking at now?"), the stream publishes every chunk and then `_run_turn` publishes the full text again with `interrupt=True`, interrupting the speech to repeat it. Inversely, a zero-chunk stream leaves `spoken=True` with nothing published, so the user gets silence instead of the fallback. Base `spoken` on what actually happened — the terminating record's tool name and whether any chunk was published — rather than the full call list.

## Suggestions

1. Once an instrument's loss has been announced, re-acquiring it is silent unless the reading changed: `_observe` flips `tracking` back ([`instrument_monitor.py:316-320`](https://github.com/NVIDIA/xr-ai/blob/4568f5b056491bc37eb54bd305e924e17f923567/agent-samples/lab-instrument-monitoring/worker/lab_instrument_monitoring_worker/instrument_monitor.py#L316-L320)) without an event, so the user's last information remains "no longer tracking". Emit a re-acquired change when a lost instrument returns.

The sample otherwise remains well composed: participant scoping is consistent across all agents, the bounded tool loop is used with graceful iteration-limit handling and per-turn relay scopes, filenames sanitize participant IDs, JSONL writes run off the event loop with a bounded in-memory history, image queries go to the Cosmos VLM one image at a time with policy on the system channel, and the tool catalog is pinned by drift tests plus request-model-validated evals. The two blockers should be resolved before merge.

blongs-nv:
Reviewed at `e4792753c9565ade925405a7f914fc281feefc42` by a bot, covering the new commit (`fix(sample): tighten instrument reading association`). My previous review's two blockers (uncontained monitoring-loop publishes; the `spoken` heuristic) and the re-acquisition suggestion are unchanged at this head, as are the four blockers from the posted re-review. One new finding in this commit's surface:

## Blockers

1. **Omni mode caps every VLM response at 128 tokens, truncating the monitor's JSON contract.**

   [`models.omni.json:15`](https://github.com/NVIDIA/xr-ai/blob/e4792753c9565ade925405a7f914fc281feefc42/agent-samples/lab-instrument-monitoring/yaml/models.omni.json#L15) sets `default_extras.max_tokens: 128`, and [`ImageQueryTool` passes no per-call `max_tokens`](https://github.com/NVIDIA/xr-ai/blob/e4792753c9565ade925405a7f914fc281feefc42/agent-sdk/xr-ai-tools/xr_ai_tools/vision.py#L195-L203), so the cap applies to every vision call. The monitor's own response schema allows a 2000-character caption plus a 500-character summary ([`monitor.py:37-39`](https://github.com/NVIDIA/xr-ai/blob/e4792753c9565ade925405a7f914fc281feefc42/agent-samples/lab-instrument-monitoring/worker/lab_instrument_monitoring_worker/monitor.py#L37-L39)) — far beyond 128 tokens — so under `--vlm-mode omni` any moderately detailed scene truncates the JSON mid-object, `parse_monitor_response` fails, and the visual monitor degrades to persistent `error` records (with `_previous` never updating). Raise the omni-mode budget to cover the schema (or shrink the schema), and add an omni-profile case to the visual eval.

The rest of the commit is sound: the reading-association prompt is meaningfully stricter with a no-reading eval case pinning the `UNKNOWN` behavior, the launcher gained strict argument parsing, and the temp-dir config materialization resolves every relative path correctly for the relocated file.

blongs-nv:
Reviewed at `4768acc42c3aaa797d2da044798b48867f47ebd6` by a bot, covering the blocker-fix commit. From my previous reviews, all three items are fixed and verified: every monitoring-loop publish now contains subscriber failures with log-and-continue (change/lost/snapshot and the monitor record), `spoken` is derived from the terminating tool plus whether a chunk was actually published (fixing both the double-speech and zero-chunk-silence paths, with regression tests), and omni mode's `max_tokens` is now 1024. The Ruff `I001` from the posted re-review is also fixed. Still open from that re-review: the sighting-vs-reading tracking refresh, the reading policy as a system prompt, the current-view eval cases, and the re-acquisition suggestion. One new finding in this commit:

## Blockers

1. **The sample now binds the unauthenticated event viewer to all interfaces by default.**

   [`lab_instrument_monitoring_worker.yaml:9`](https://github.com/NVIDIA/xr-ai/blob/4768acc42c3aaa797d2da044798b48867f47ebd6/agent-samples/lab-instrument-monitoring/yaml/lab_instrument_monitoring_worker.yaml#L9) changes `web_events_host` from `127.0.0.1` to `0.0.0.0`, so the checked-in default serves the plain-HTTP, unauthenticated viewer — whose payloads include voice transcripts and camera-derived text — on every interface with no firewall guidance applied by default. The same commit's own networking table still documents the viewer as "bound to 127.0.0.1 by default; do not expose to an untrusted network" ([`networking.md:17`](https://github.com/NVIDIA/xr-ai/blob/4768acc42c3aaa797d2da044798b48867f47ebd6/docs/source/getting_started/networking.md#L17)), contradicting the new body text describing the `0.0.0.0` override. Keep the checked-in default loopback-only (the documented SSH tunnel covers remote viewing) and let deployments opt into `0.0.0.0` explicitly; align the table and body either way.

wenxind-nvidia:
## Re-review

Reviewed PR 363 at `673bec6` against current `main`. Verdict: **request changes**.

### Blockers

1. **[P1] The unauthenticated event viewer is exposed on every interface by default.**

   The shipped config uses [`web_events_host: 0.0.0.0`](https://github.com/NVIDIA/xr-ai/blob/673bec6f20383067c810b3904ebc5e0c0f99eb21/agent-samples/lab-instrument-monitoring/yaml/lab_instrument_monitoring_worker.yaml#L9), while the viewer contains transcripts and camera-derived observations and has no authentication or TLS. Restore `127.0.0.1`; deployments can explicitly opt into remote exposure.

2. **[P1] A visible instrument with an unreadable display still produces a false “lost” alert.**

   [`UNKNOWN` readings are discarded](https://github.com/NVIDIA/xr-ai/blob/673bec6f20383067c810b3904ebc5e0c0f99eb21/agent-samples/lab-instrument-monitoring/worker/lab_instrument_monitoring_worker/instruments.py#L190-L198), and the monitor refreshes `last_seen` only from successful readings. Glare or obstruction lasting 30 seconds therefore reports a visible device as lost. Return mapped marker sightings separately from readings and refresh tracking from sightings.

3. **[P1] The instrument-reading policy remains user content.**

   [`ImageQueryTool` has no system prompt](https://github.com/NVIDIA/xr-ai/blob/673bec6f20383067c810b3904ebc5e0c0f99eb21/agent-samples/lab-instrument-monitoring/worker/lab_instrument_monitoring_worker/instruments.py#L52-L66); the fixed ownership and `UNKNOWN` contract is placed in the user query instead. Move it into a packaged system prompt and add an adversarial visible-text eval.

4. **[P1] Optional record-subscriber failures can suppress foreground speech.**

   [`_run_turn()` publishes the foreground record and voice contribution in one try block](https://github.com/NVIDIA/xr-ai/blob/673bec6f20383067c810b3904ebc5e0c0f99eb21/agent-samples/lab-instrument-monitoring/worker/lab_instrument_monitoring_worker/foreground.py#L282-L300). Because runtime publication re-raises subscriber failures, a JSONL or web-event failure prevents the answer from reaching voice. Record publication should be contained independently.

5. **[P2] The current-view system prompt still lacks behavioral evaluation.**

   The [visual eval](https://github.com/NVIDIA/xr-ai/blob/673bec6f20383067c810b3904ebc5e0c0f99eb21/agent-samples/lab-instrument-monitoring/eval/visual_cases.yaml) covers only monitor and instrument behavior. Add current-view cases for unavailable detail, visible instructions, and the two-sentence/plain-language contract.

The latest empty-stream fix is valid and covered. The prior frame consistency, unit parsing, publish containment, speech deduplication, Ruff, and Omni token-budget findings are fixed.

### Validation

- 49 affected tests passed.
- Ruff and `git diff --check` passed.
- Live model evals were not run.

wenxind-nvidia (approved):

blongs-nv (approved):

PR #364, opened 2026-08-14
Add getting-started skill bank and landing-page pointers

blongs-nv:
Adds the getting-started skill bank: the agent-first onboarding path for xr-ai. A developer pastes one prompt to their coding agent; the agent installs the version-aligned getting-started skill (skills/getting-started/SKILL.md, open Agent Skills spec) and walks them through setup. Entry points on the GitHub README, the docs landing page, and the quickstart lead with the prompt, a Skills page joins Getting Started in the docs, and the prompt is single-sourced in the docs via a snippet with a contract test keeping the README copy identical. The skill selects release-vs-main up front, reuses the repo's release-selector policy, and fails closed (explicit user OK, main for both skill and checkout) for releases that predate the bank.

A second commit fixes misleading operational docs found by an agent dry-run of this onboarding journey: the hub banner now prints a connectable localhost URL, README/quickstart document that the VLM and STT persist after exit (with the stop command) and that queries are answered after the launcher's "All processes ready" banner, the STT server's persistence is documented alongside vLLM's, and the sample README gains the HF_TOKEN note.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -19,6 +19,7 @@ services/        Hub, model servers, and typed capability services
 utils/           Launcher, logging, VAD, vLLM, and voice-gate utilities
 tests/           Cross-package and integration tests
 docs/source/     User and contributor documentation
+skills/          Skill bank that sets coding agents up
 ```
 
 ## Architecture boundaries
--- a/README.md
+++ b/README.md
@@ -15,9 +15,10 @@ conversational AI within the CloudXR ecosystem.
 ```text
 Set up xr-ai for me: ask me whether to build against the latest release or
 main, then fetch skills/getting-started/SKILL.md at that ref from
-https://raw.githubusercontent.com/NVIDIA/xr-ai/ (fall back to the main copy
-if the ref predates it), install it as a skill (or just follow it), and walk
-me through the rest of the setup.
+https://raw.githubusercontent.com/NVIDIA/xr-ai/. If the file does not exist
+at that ref, the release predates agent setup: get my OK to use main for
+both the skill and the checkout. Install the skill (or just follow it), and
+walk me through the rest of the setup.
 ```
 
 ## Public Beta Notice
--- a/README.md
+++ b/README.md
@@ -13,9 +13,11 @@ conversational AI within the CloudXR ecosystem.
 **Using a coding agent?** Paste this to it:
 
 ```text
-Set up xr-ai for me: fetch
-https://raw.githubusercontent.com/NVIDIA/xr-ai/main/skills/getting-started/SKILL.md,
-install it as a skill (or just follow it), and walk me through the setup.
+Set up xr-ai for me: ask me whether to build against the latest release or
+main, then fetch skills/getting-started/SKILL.md at that ref from
+https://raw.githubusercontent.com/NVIDIA/xr-ai/ (fall back to the main copy
+if the ref predates it), install it as a skill (or just follow it), and walk
+me through the rest of the setup.
 ```
 
 ## Public Beta Notice
@@ -252,7 +254,7 @@ several minutes).  `HF_TOKEN` is required by default; pass
 
 In both modes the VLM and STT keep running after you exit so the next run
 skips the model reload (see
-[vLLM model persistence](docs/source/components/ai-services.md#vllm-model-persistence)); free
+[model-server persistence](docs/source/components/ai-services.md#model-server-persistence)); free
 the VRAM with `cd agent-samples/model-servers && uv run model_servers --stop`.
 
 #### Step 1 — Start the server
@@ -272,9 +274,9 @@ The hub prints:
 ```
 
 This banner appears as soon as the hub itself is ready, while the model
-services and worker are still starting.  The stack accepts clients once the
-launcher prints its `All processes ready` banner; a client connected before
-that gets a session with no agent responding.
+services and worker are still starting.  Clients can connect as soon as it
+appears, but the agent answers queries only after the launcher prints its
+`All processes ready` banner.
 
 #### Step 2 — Connect a client
 
--- a/README.md
+++ b/README.md
@@ -10,6 +10,14 @@
 Agentic AI for XR — an open-source foundation for multi-modal, real-time
 conversational AI within the CloudXR ecosystem.
 
+**Using a coding agent?** Paste this to it:
+
+```text
+Set up xr-ai for me: fetch
+https://raw.githubusercontent.com/NVIDIA/xr-ai/main/skills/getting-started/SKILL.md,
+install it as a skill (or just follow it), and walk me through the setup.
+```
+
 ## Public Beta Notice
 
 This project is publicly available in beta and is under active development.
@@ -528,6 +536,7 @@ For engineers and agents working in the repo:
 | [`AGENTS.md`](AGENTS.md) | Working contract — hard rules every change must satisfy |
 | [`DEPENDENCIES.md`](DEPENDENCIES.md) | Authoritative dependency map (update with every `pyproject.toml` change) |
 | [Versioned documentation](https://nvidia.github.io/xr-ai/) | Latest release by default, plus `main` development and release-tag documentation |
+| [`skills/README.md`](skills/README.md) | Skill bank: setup skills for coding agents |
 | [`docs/source/overview/architecture.md`](docs/source/overview/architecture.md) | Hub ↔ transport ↔ agent boundaries; known limitations |
 | [`docs/source/components/launcher-and-process-model.md`](docs/source/components/launcher-and-process-model.md) | `Process` / `run_stack` mechanics; ready-file protocol |
 | [`docs/source/components/ai-services.md`](docs/source/components/ai-services.md) | VLM / STT / TTS / LLM / embedding server reference + worker call examples |
--- a/README.md
+++ b/README.md
@@ -234,8 +234,7 @@ runtime-selection details.
 
 There are two ways to run it:
 
-**Standalone** (~23 GB VRAM) — starts its own VLM and STT, owns them for the
-session, and stops them when you exit:
+**Standalone** (~23 GB VRAM) — starts its own VLM and STT:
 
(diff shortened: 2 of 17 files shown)


nvddr:
I reviewed this against the apparent goal: make agent-led setup the primary onboarding path, while keeping it trustworthy, version-aware, and usable by agents with or without native skill support. The landing-page routing is coherent, the lifecycle/URL corrections match the current implementation, the new `SKILL.md` satisfies the Agent Skills metadata shape, and all eight PR workflows are green.

## Blockers

1. **The primary onboarding prompt installs and follows mutable instructions from `main`, even though the skill defaults the actual checkout to a release.** The root README, quickstart, and docs landing page all tell an agent to fetch `raw.githubusercontent.com/.../main/skills/getting-started/SKILL.md` and install/follow it. A skill is executable instruction content from the agent's perspective, and installation makes that content persistent; a future branch change or compromise would therefore affect every user who pastes this prompt. It also breaks the version-alignment promise: choosing the release path still runs the skill definition from `main`. Please make the skill source immutable/version-aligned (for example, choose the target ref first and fetch the skill from that same tag/commit, with a deliberate fallback for releases that predate the bank), or otherwise add a fail-closed review/confirmation step before installation and execution.

2. **The documented manual install command does not fail safely when the placeholder is left unchanged.** In `skills/README.md`, `SKILLS_DIR=` followed by `-o "$SKILLS_DIR/getting-started/SKILL.md"` resolves to `/getting-started/SKILL.md`. For a normal user it fails with permissions; under an elevated shell it writes to the wrong absolute directory. Since this is the bank's setup command, please make the placeholder unambiguously non-runnable or guard it, e.g. `: "${SKILLS_DIR:?Set SKILLS_DIR to your agent's skills directory}"`, before invoking curl.

## Suggestions

- Narrow the skill's trigger description. “Use when starting any work with xr-ai” can activate the setup questionnaire for ordinary work in an already-cloned repo. Something like “Use when a user asks to install, clone, or perform initial setup for xr-ai” would preserve the intended discovery behavior without repeatedly hijacking feature/debugging tasks.
- Add `skills-ref validate skills/getting-started` (or the equivalent reference validator) to CI. The current SPDX/docs/lint checks pass, but none of them establishes that future skill additions keep spec-valid frontmatter, directory-name matching, and field constraints.
- Consider keeping the hub banner/model-persistence corrections in a separate commit or calling them out explicitly in the PR description. They look correct, but they are runtime/operations corrections rather than skill-bank plumbing, and the PR currently has no body explaining the expanded scope.

blongs-nv:
Agreed on validating skill frontmatter in CI, but deferring it from this PR: the next milestone of this effort has a dedicated skill-packaging task coming (install tooling, per-skill validation, publishing), and a validator belongs there with the rest of that machinery rather than as a one-off check for a single skill. Until then the SPDX check and strict docs build cover this file, and the bank has exactly one skill. Tracking it as part of that follow-up.

Already updated for the other comments.

wenxind-nvidia:
Review of `d1a49ec` — **request changes**. The lifecycle corrections are valid, the skill follows the Agent Skills specification, and all current checks pass.

## Blockers

1. **Release setup still installs mutable `main` instructions.**

   The [root prompt](https://github.com/NVIDIA/xr-ai/blob/d1a49ec778d64593c26ebc937bd998d22d8e9f8c/README.md#L13-L19) and [manual installer](https://github.com/NVIDIA/xr-ai/blob/d1a49ec778d64593c26ebc937bd998d22d8e9f8c/skills/README.md#L20-L26) fetch the skill from `main`, even when the skill subsequently defaults to a release checkout. The Sphinx rewrite only protects generated documentation; it does not rewrite either README. A future main-only workflow can therefore be applied to an older release. Select/pin the target ref before persistent installation, or use an immutable bootstrap that fetches the selected version.

2. **The checkout command disagrees with the documentation's release policy.**

   [The skill](https://github.com/NVIDIA/xr-ai/blob/d1a49ec778d64593c26ebc937bd998d22d8e9f8c/skills/getting-started/SKILL.md#L35-L37) chooses the first version-sorted tag. The repository's [documentation selector](https://github.com/NVIDIA/xr-ai/blob/d1a49ec778d64593c26ebc937bd998d22d8e9f8c/.github/scripts/select_latest_docs_release.py#L61-L67), however, prefers any stable release over newer prereleases—explicitly tested with `v1.9.0` versus `v2.0.0-rc.1`. In that state, the skill checks out the RC while `/latest/` serves `v1.9.0`, breaking the promised source/docs alignment. Reuse the same selector policy.

3. **The readiness explanation contradicts actual behavior.**

   [README lines 273–276](https://github.com/NVIDIA/xr-ai/blob/d1a49ec778d64593c26ebc937bd998d22d8e9f8c/README.md#L273-L276) say the stack only accepts clients after `All processes ready`, then immediately describe a client connected earlier. The hub accepts that earlier connection; the worker starts later and requests the existing participant roster. Say that clients can connect when the hub banner appears, but queries should wait for `All processes ready`.

## Suggestions

- Extend [`test_source_links_use_the_current_documentation_ref`](https://github.com/NVIDIA/xr-ai/blob/d1a49ec778d64593c26ebc937bd998d22d8e9f8c/tests/test_docs_versions.py#L54-L75) to cover `raw.githubusercontent.com`. The new raw-link rewrite is currently untested.
- The persistence claim is correct, but the linked [persistence section](https://github.com/NVIDIA/xr-ai/blob/d1a49ec778d64593c26ebc937bd998d22d8e9f8c/docs/source/components/ai-services.md#L317-L349) only explains vLLM/NIM lifecycle. Add the STT server's detached-process/reuse behavior now that onboarding explicitly promises STT persistence.

## Nits

- The new root README link targets nonexistent `#vllm-model-persistence`; the heading is `#model-server-persistence`. Its label should also cover STT rather than saying only “vLLM model persistence.”

nvddr (changes requested):
Re-reviewed current head 7734507. The release selector, readiness wording, persistence documentation/link, raw-link rewrite coverage, trigger scope, and SKILLS_DIR guard now address the prior feedback.

## Blocker

The pre-skill-bank fallback still defeats both version alignment and fail-closed installation. The primary prompt asks for the latest release by default, then says to fall back to the main copy when that release predates the skill. That is the path users take today because the current v0.1.0-beta tag has no skills/getting-started/SKILL.md. The agent therefore persists mutable main instructions while operating on an older release. Afterward, the skill itself says its ref should match the selected release and to re-fetch when it does not, sending the agent back to the same nonexistent tagged file.

Please make the pre-bank fallback immutable and internally consistent—for example, use a pinned bootstrap commit—or stop for explicit review/confirmation before installing and following the main fallback, without then instructing the agent to fetch a file known not to exist at the selected tag.

wenxind-nvidia:
Re-review of `7734507898` — **request changes; one blocker remains**.

The previous readiness, persistence, anchor, release-selector, and rewrite-test findings are otherwise addressed correctly.

## Blocker

1. **The default release path still cannot install a release-aligned skill.**

   The [bootstrap prompt](https://github.com/NVIDIA/xr-ai/blob/7734507898eeee76d8ec97871abfe2e3da23acac/README.md#L13-L20) selects the release before fetching the skill and falls back to `main` when the release predates it. That is exactly the current situation: `v0.1.0-beta` has no `skills/getting-started/SKILL.md`.

   The fallback then installs the mutable `main` copy, whose [instructions](https://github.com/NVIDIA/xr-ai/blob/7734507898eeee76d8ec97871abfe2e3da23acac/skills/getting-started/SKILL.md#L20-L25) tell future activations to re-fetch from the selected release—where the file still does not exist. The [manual installer](https://github.com/NVIDIA/xr-ai/blob/7734507898eeee76d8ec97871abfe2e3da23acac/skills/README.md#L20-L27) also fails if `REF` is set to the current release.

   Use an immutable, known-compatible fallback skill revision for pre-bank releases and explicitly document that this fallback must not re-fetch the missing release path. Alternatively, require a release containing the skill bank before making release setup the default.

## Suggestions

- [`make -s -C docs latest-release`](https://github.com/NVIDIA/xr-ai/blob/7734507898eeee76d8ec97871abfe2e3da23acac/skills/getting-started/SKILL.md#L37-L40) introduces `make` as an initial-setup dependency, but `make` is not listed in the requirements. Prefer invoking the selector directly with the required Python runtime or document the dependency.
- Add coverage for both bootstrap states: a release containing the skill and a pre-bank release requiring the fallback. The existing test only verifies URL rewriting.

## Nits

- Rename “Stable or latest?” to “Release or main?” The current release is a beta, and both choices represent a form of “latest.”

wenxind-nvidia (approved):

wenxind-nvidia:
Re-review of `000beb4f35` — **no remaining blockers; approve once CI passes**.

The pre-bank release case is now coherent: the [bootstrap prompt](https://github.com/NVIDIA/xr-ai/blob/000beb4f3547e8cc155da3f6dbf5546f362f6201/README.md#L13-L21) requires explicit permission before switching both the skill and checkout to `main`, and the [skill instructions](https://github.com/NVIDIA/xr-ai/blob/000beb4f3547e8cc155da3f6dbf5546f362f6201/skills/getting-started/SKILL.md#L20-L27) preserve that policy. The hidden `make` dependency was also removed.

## Blockers

None.

## Suggestions

- The bootstrap prompt asks for the “latest release” before the agent has loaded the [canonical selector command](https://github.com/NVIDIA/xr-ai/blob/000beb4f3547e8cc155da3f6dbf5546f362f6201/skills/getting-started/SKILL.md#L39-L43). Consider including the selection rule in the bootstrap text so agents cannot initially choose a newer prerelease over the stable tag.
- Add a small contract test ensuring the root README prompt remains identical to `_snippets/agent-setup-prompt.txt`; these are now two copies of the primary onboarding instruction.
- Add a PR description summarizing the skill-bank work and the accompanying startup/persistence corrections.

## Nits

None.

All checks passed on the previous head; checks for `000beb4f35` were still queued at the time of this review.

nvddr (approved):

PR #365, opened 2026-08-15
feat(tools): add magenta polygon image fill

wenxind-nvidia:
## Summary

- add a Relay-managed `ImagePolygonFillTool` that consumes an `ImageReference` and at least three ordered pixel coordinates
- fill the enclosed polygon with exact magenta (`#FF00FF`) and return a new lossless PNG reference without changing the source
- validate fractional coordinates, image bounds, and non-zero polygon area
- document the `image-editing` optional dependency and add pixel-accurate RGB/RGBA tests

## Testing

- `uv run pytest -q -m "not gpu"` — 795 passed, 10 deselected
- `uv run --with ruff ruff check .` — passed
- `git diff --check` — passed

### DIFF
--- a/DEPENDENCIES.md
+++ b/DEPENDENCIES.md
@@ -113,13 +113,15 @@ xr-ai-tools  (agent-sdk/xr-ai-tools/)
     └── pydantic >=2.10
     └── [relay] xr-ai-models [editable: ../xr-ai-models]
     ├── [frames] numpy >=1.24, Pillow >=10.0, xr-ai-hub-client [editable: ../xr-ai-hub]
+    ├── [image-editing] Pillow >=10.0
     ├── [vision] xr-ai-models [editable: ../xr-ai-models]
     ├── [qr-code] numpy >=1.24, zxing-cpp >=2.3,<4, Pillow >=10.0, xr-ai-hub-client [editable: ../xr-ai-hub]
     └── [services] msgpack >=1.0, pyzmq >=27.0
     Toolkit-independent native tools: Pydantic request and response models,
     Relay-managed finite and async execution, model tool-call workflow helpers,
-    frame selection, single/multi-image inference, participant-scoped QR-code
-    extraction, typed capability clients, and service RPC.
+    frame selection, lossless polygon image editing, single/multi-image
+    inference, participant-scoped QR-code extraction, typed capability clients,
+    and service RPC.
 
 
 xr-openxr-service  (services/openxr-service/)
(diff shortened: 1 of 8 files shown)


nvddr:
**[P2] Preserve image ownership on derived outputs**

At [image_polygon.py:136](https://github.com/NVIDIA/xr-ai/blob/9fe4779257cb1ea4ee120d82efa7000ca991f96f/agent-sdk/xr-ai-tools/xr_ai_tools/image_polygon.py#L136), the edited image is stored without an owner. Source frames are tagged with a participant owner so `release_owner()` removes their pixels on disconnect; the derived image survives that cleanup and remains resolvable.

Please add a registry operation that stores derived data with the source reference owner, plus a regression test showing that releasing the owner makes both the source and edited references unavailable.

blongs-nv:
Reviewed at `9fe4779257cb1ea4ee120d82efa7000ca991f96f` by a bot.

## Blockers

1. **Resolve and validation failures escape as scrubbed internal errors instead of the graceful result contract the sibling image tools establish.**

   [`agent-sdk/xr-ai-tools/xr_ai_tools/image_polygon.py:126-133`](https://github.com/NVIDIA/xr-ai/blob/9fe4779257cb1ea4ee120d82efa7000ca991f96f/agent-sdk/xr-ai-tools/xr_ai_tools/image_polygon.py#L126-L133)
   The vision tools catch `LookupError`/`ValueError` from `ImageRegistry.resolve` and return `available=False` with "Image input unavailable — please select it again." ([`vision.py:96-103`](https://github.com/NVIDIA/xr-ai/blob/9fe4779257cb1ea4ee120d82efa7000ca991f96f/agent-sdk/xr-ai-tools/xr_ai_tools/vision.py#L96-L103)); `_fill` lets both propagate. I reproduced through `Tool.invoke` at this SHA: a stale/evicted reference and a coordinate one pixel outside the image each surface as an unchained `RuntimeError("internal error: LookupError/ValueError: ...")`. Guessed-slightly-wrong coordinates and evicted handles (capacity 128) are the routine failure modes for a model-invokable editing tool, so ordinary misuse kills the turn instead of returning a recoverable result.

## Suggestions

1. Once the error contract is settled, add tests for the failure paths a model will actually hit: a stale/evicted `xr-image://` reference and out-of-bounds coordinates driven through `invoke`, asserting the graceful result rather than an exception. The current suite covers only the happy paths and pre-resolve validation.

The tool itself is well built: a pure `fill_polygon_magenta` function with the PIL work offloaded via `asyncio.to_thread`, palette/alpha handling done correctly, an isolated `image-editing` extra, pixel-accurate RGB/RGBA tests, and README/docs/DEPENDENCIES all landing together (I ran the polygon and image-tool suites in a fresh clone; 29 passed). The blocker above should be resolved before merge.

nvddr (approved):

PR #369, opened 2026-08-15
feat(models): default to Nemotron 3 Nano Omni

nvddr:
## Summary

- replace the default Nemotron-3 Nano text server on port 8107 with Nemotron-3 Nano Omni on port 8108
- keep Cosmos3 Nano Reasoner as the dedicated VLM on port 8100; XR Render uses Omni for `llm` / `agent_llm` and Cosmos for `vlm`
- simplify `model_servers` to one default topology: STT, Omni, Cosmos VLM, and embeddings
- remove the old stack-selection flags and stop only the superseded Nano text server before startup
- on dual RTX 6000 Ada, keep Cosmos and embeddings on GPU 0 and place Omni with STT on GPU 1
- update launcher, model-profile, wiring, dependency, quickstart, reference, and troubleshooting coverage

## Impact

Running `model_servers` now launches Omni (8108), Cosmos (8100), STT (8103), and embeddings (8109). There is no retained legacy stack selector: Omni replaces Nano for text inference, while Cosmos remains the vision model.

After updating, stop existing persistent model servers once with `uv run model_servers --stop`, then start the default stack with `uv run model_servers`.

## Validation

- 78 model-server, model-profile, service-layout, and model-config tests passed
- XR Render model-profile wire test passed
- full repository `ruff check .` passed on the restacked branch
- `git diff --check`
- live dual-Ada startup verified all four health endpoints:
  - Omni FP8 reasoning model on 8108
  - Cosmos3 Nano on 8100
  - STT on 8103
  - embeddings on 8109


wenxind-nvidia:
## Blocker

1. **Removed stack flags are silently ignored.** [`main.py`](https://github.com/NVIDIA/xr-ai/blob/a7719f1b1be4664fcab9bc03d3455e050ef23861/agent-samples/model-servers/main.py#L85-L103) removes `--omni-stack` and `--vlm-llm-stack`, but still uses `parse_known_args()`. Existing commands such as `model_servers --vlm-llm-stack` therefore succeed while launching the new Omni/Cosmos topology instead of the requested Nano stack. Please reject these legacy flags with a clear migration error, or explicitly handle them as deprecated aliases, and add regression coverage.

## Suggestion

1. **Correct the hosted NIM instructions.** [`README.md`](https://github.com/NVIDIA/xr-ai/blob/a7719f1b1be4664fcab9bc03d3455e050ef23861/README.md#L402-L415) and [`ai-services.md`](https://github.com/NVIDIA/xr-ai/blob/a7719f1b1be4664fcab9bc03d3455e050ef23861/docs/source/components/ai-services.md#L239-L246) say to omit only local Omni, but `models.nim.yaml` routes both LLM and VLM to hosted NIM. The instructions should say not to start either local Omni or Cosmos and explain how STT is started separately.

## Nits

- [`README.md`](https://github.com/NVIDIA/xr-ai/blob/a7719f1b1be4664fcab9bc03d3455e050ef23861/README.md#L190-L196) still refers to "both stack variants."
- `agent-samples/model-servers/yaml/dual_48G_ada/embedding_server_omni.yaml` is now unreferenced and should be removed with the stack-selector cleanup.
- [`troubleshooting.md`](https://github.com/NVIDIA/xr-ai/blob/a7719f1b1be4664fcab9bc03d3455e050ef23861/docs/source/guides/troubleshooting.md#L333-L344) says shipped YAMLs use `enforce_eager: true`, while every default Omni profile sets it to `false`.

nvddr:
Disposition for the review feedback (commit `689fe24`):

- **Removed stack flags:** switched `model_servers` to strict argument parsing, so `--vlm-llm-stack` and `--omni-stack` are rejected. Added regression coverage for both; no deprecated aliases or legacy topology were retained.
- **Hosted NIM instructions:** updated the root README and AI-services/quickstart docs to say not to run `model_servers` (which starts both local Omni and Cosmos) and documented starting STT by itself before `xr_render_demo`.
- **Stale stack wording:** removed the remaining “both stack variants” language.
- **Orphaned profile:** deleted `dual_48G_ada/embedding_server_omni.yaml`.
- **`enforce_eager` guidance:** corrected troubleshooting text to reflect that the default Omni profiles set it to `false`, with `true` documented as a faster-startup development option.

Validation: 27 focused model-server/service-layout tests passed, `models.nim.yaml` parses successfully, and `git diff --check` passed. The follow-up commit now carries the required DCO sign-off.

wenxind-nvidia (approved):

PR #370, opened 2026-08-15
fix(web): preserve camera field of view

nvddr:
## Summary

- capture the browser camera directly without LiveKit's implicit 16:9 h720 constraint
- request `resizeMode: none` so mobile browsers preserve the sensor field of view
- preview the exact published camera track with its captured aspect ratio instead of opening a second stream
- replace source-string assertions with focused JavaScript behavior tests using mocked browser capture and LiveKit publication
- load the web client test modules as ESM without Node's removed experimental flag, pin CI to Node.js 24, and document the local Node prerequisite

## Root cause

LiveKit's browser convenience capture path applied a 16:9 h720 constraint. On portrait mobile cameras this could crop the sensor before frames entered WebRTC. The page then opened a separate preview stream and rendered it with `object-fit: cover`, so the preview could differ from—and further crop—the stream received by the server.

## Impact

Mobile and glasses-style browser clients preserve the camera's native aspect ratio and field of view. The preview now represents the same track published to the XR hub.

## Validation

- focused JavaScript behavior tests verify `resizeMode: 'none'`, track identity through capture/publication/preview, and track cleanup on publication failure, camera stop, and disconnect
- JavaScript behavior tests pass with Node.js 24 and Node.js 25.6.1 without `--experimental-default-type=module`
- `pytest -q tests/test_web_camera_capture.py` — 1 passed (runs 2 Node behavior cases)
- `pytest -q -m "not gpu"` — 822 passed, 10 deselected
- `node --check` on all added JavaScript modules
- repository-wide Ruff, SPDX, and `git diff --check`

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -33,6 +33,11 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
+      - name: Install Node.js 24
+        uses: actions/setup-node@v4
+        with:
+          node-version: "24"
+
       - name: Install uv
         uses: astral-sh/setup-uv@caf0cab7a618c569241d31dcd442f54681755d39  # v3
         with:
--- a/client-samples/web/App/app.js
+++ b/client-samples/web/App/app.js
@@ -45,16 +45,37 @@ const model = {
 // from the "Received" list.
 const AGENT_REPLY_TOPICS = new Set(['agent.response', 'vlm.response']);
 
-// Local camera preview stream (separate from the LiveKit publish stream).
-let _previewStream = null;
-
-function releasePreviewStream() {
-  if (!_previewStream) return;
-  _previewStream.getTracks().forEach(t => t.stop());
-  _previewStream = null;
+function clearCameraPreview() {
   const videoEl = $('camera-preview');
+  videoEl.onresize = null;
   videoEl.srcObject = null;
   videoEl.style.transform = '';
+  videoEl.closest('.preview-card').style.aspectRatio = '';
+}
+
+function showPublishedCameraPreview() {
+  clearCameraPreview();
+  const track = model.session?.cameraTrack;
+  if (!track) return;
+
+  const videoEl = $('camera-preview');
+  const previewCard = videoEl.closest('.preview-card');
+  videoEl.srcObject = new MediaStream([track]);
+
+  const updateAspectRatio = () => {
+    const settings = track.getSettings();
+    const width = videoEl.videoWidth || settings.width;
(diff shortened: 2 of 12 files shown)


wenxind-nvidia:
**Suggestion:** [`tests/test_web_camera_capture.py:15`](https://github.com/NVIDIA/xr-ai/blob/f469c85ced7634af41cb544a6eb4b3da59d5cfbb/tests/test_web_camera_capture.py#L15) only checks for source-code strings, so it does not exercise capture, publication, preview, or cleanup. Please add a focused JavaScript/browser test with mocked `getUserMedia` and LiveKit publication that verifies the requested constraints include `resizeMode: 'none'`, the captured track is the one published and previewed, and the track is stopped on publication failure, camera stop, and disconnect.

wenxind-nvidia (approved):

nvddr:
@wenxind-nvidia Addressed in `49a08e6`. I replaced the source-string assertions with focused JavaScript behavior tests using mocked `getUserMedia`, LiveKit publication, and browser preview objects. The tests now verify:

- capture requests include `resizeMode: 'none'`
- the captured track is the exact track published and attached to the preview
- publication failure stops the captured track
- camera stop and disconnect both stop the track and clear the preview

Validation: the focused pytest wrapper passes, the full non-GPU suite passes (`822 passed, 10 deselected`), and repository-wide Ruff/SPDX checks pass.

wenxind-nvidia:
**Suggestion:** [`tests/test_web_camera_capture.py:14`](https://github.com/NVIDIA/xr-ai/blob/49a08e6244dbf9eb13ef485de9912f9066ca1b23/tests/test_web_camera_capture.py#L14) uses `--experimental-default-type=module`, which has been removed from Node 25. The exact test command fails on Node 25.6.1 with `node: bad option`, although the JavaScript tests pass when run without that flag. Please make ESM handling part of the loader or package configuration, pin the supported Node version in CI, and document the new Node prerequisite in `tests/README.md` so the documented `uv run pytest` workflow remains reproducible.

nvddr:
@wenxind-nvidia Addressed in 8f8b016. The pytest wrapper no longer passes `--experimental-default-type=module`; the test loader now explicitly loads the web client's `.js` files as ESM. I also pinned the test workflow to Node.js 24 and documented Node.js 24 as the prerequisite for the standard `uv run pytest` workflow in `tests/README.md`.

Verified the JavaScript behavior tests with both Node.js 24 and the reported Node.js 25.6.1, plus the focused pytest and full non-GPU suite (822 passed, 10 deselected).

PR #371, opened 2026-08-15
feat(tools): track QR and ArUco markers

nvddr:
## What changed

- replace the QR-specific tool surface with `MarkerTrackingTool` and one model-visible `track_markers` contract
- detect QR codes with ZXing-C++ and ArUco markers with OpenCV in the same current-frame scan
- enable QR and ArUco by default
- allow initialization-time marker-family selection through `marker_types`
- allow initialization-time ArUco predefined-dictionary selection through `aruco_dictionary`
- return both families through uniform `marker_type`, `value`, and four-corner fields
- retain regression coverage that verifies every same-family QR marker in a frame is returned
- add a standalone `uv run` utility that generates QR and ArUco PNGs with inline dependency metadata
- rename the optional dependency extra from `qr-code` to `marker-tracking`
- update package, canonical, dependency, and migration documentation

## Why

Agents should consume one marker-tracking API regardless of which marker families a deployment enables. Marker-specific configuration belongs at tool initialization; requests and result handling should not branch between QR and ArUco.

The PNG generator provides test markers without adding another agent-facing API or modifying the project environment.

## Impact

Initialize `xr_ai_tools.marker_tracking.MarkerTrackingTool` instead of `QRCodeTool`. Existing QR support remains enabled by default, and ArUco `DICT_4X4_50` tracking is added by default. Pass `marker_types=(MarkerType.QR_CODE,)` or `(MarkerType.ARUCO,)` to restrict detection, and set `aruco_dictionary` when a deployment uses another OpenCV predefined dictionary.

The model-visible tool remains the same for every configuration: `track_markers` accepts only `participant_id` and returns `markers` with a uniform schema.

Run `xr_ai_tools/utilities/generate_marker.py` with `uv run` to generate QR or ArUco PNGs. The script manages its OpenCV dependency through inline metadata.

## Checks

- `uvx --from ruff==0.15.16 ruff check --no-cache ...`
- focused marker tracking and generator tests: 9 passed
- `uv run pytest -m "not gpu" -q`: 825 passed, 10 deselected
- standalone `uv run` smoke tests generated both QR and ArUco PNGs
- SPDX header check: 469 files valid
- regenerated the gitignored `uv.lock` files for `agent-sdk/xr-ai-tools` and `tests`


wenxind-nvidia (approved):

wenxind-nvidia:
**Suggestion:** [`tests/test_marker_tracking_tool.py:132`](https://github.com/NVIDIA/xr-ai/blob/74be43916f84dbe2cebbcd0d1b51eee916ae2365/tests/test_marker_tracking_tool.py#L132) verifies one QR code and one ArUco marker, but removes the previous regression test for multiple QR codes in one frame. Since the contract promises every detected marker, please retain coverage for multiple markers from the same family so a future first-result-only regression is caught.

nvddr:
@wenxind-nvidia Addressed in e45207a. I restored the deterministic second-QR fixture and added a regression test through the public `MarkerTrackingTool` API that places two QR codes in one frame and asserts both payloads are returned.

Validation: focused marker suite 6 passed; full non-GPU suite 825 passed, 10 deselected; Ruff, SPDX, and `git diff --check` passed.

PR #373, opened 2026-08-17
fix(tools): recover small markers in full frames

nvddr:
## Summary

- scan the complete frame at native resolution and at one bounded enlarged resolution for both QR and ArUco markers
- map enlarged-scan corners back into source-frame coordinates
- deduplicate detections across scan resolutions while preserving distinct QR codes with the same payload
- share grayscale conversion and enlargement between the two detectors
- cover multiple small QR codes and small ArUco markers in downsampled portrait frames

## Why

Small visual markers can become unreadable after a high-resolution camera frame is downsampled for transport. Experiments across six ArUco IDs and apparent marker sizes from 5 to 30 pixels found consistent 7-8 pixel recoveries from the enlarged scan. A few 6-pixel phase cases remained native-only, so the implementation unions native and enlarged results. This keeps the whole frame visible and avoids cropping or tiling.

## Impact

`MarkerTrackingTool` keeps the same request and result API. Both QR and ArUco detection become more reliable for small markers, and all corners remain in source-frame coordinates.

## Validation

- `uvx ruff check agent-sdk/xr-ai-tools/xr_ai_tools/marker_tracking.py tests/test_marker_tracking_tool.py agent-sdk/xr-ai-tools/README.md docs/source/components/agent-sdk.md`
- `tests/.venv/bin/pytest -q tests/test_marker_tracking_tool.py tests/test_generate_marker_utility.py` (11 passed)

## Stack

This PR is based on latest `main`, which includes merged PR #371. Its commit is also cherry-picked into PR #363's temporary test stack.

### DIFF
--- a/agent-sdk/xr-ai-tools/README.md
+++ b/agent-sdk/xr-ai-tools/README.md
@@ -192,6 +192,9 @@ QR codes with ZXing-C++ and ArUco markers with OpenCV. Its typed result
 distinguishes an unavailable frame from a valid frame with no marker. Every
 marker has the same `marker_type`, `value`, and four-corner contract: `value`
 is decoded text for QR codes and the decimal marker ID for ArUco markers.
+QR detection scans the complete frame at native resolution and at one bounded
+enlarged resolution so small codes remain readable without discarding context.
+Results from both scans are deduplicated and use source-frame coordinates.
 
 ```python
 from xr_ai_tools.marker_tracking import (
--- a/agent-sdk/xr-ai-tools/README.md
+++ b/agent-sdk/xr-ai-tools/README.md
@@ -192,8 +192,8 @@ QR codes with ZXing-C++ and ArUco markers with OpenCV. Its typed result
 distinguishes an unavailable frame from a valid frame with no marker. Every
 marker has the same `marker_type`, `value`, and four-corner contract: `value`
 is decoded text for QR codes and the decimal marker ID for ArUco markers.
-QR detection scans the complete frame at native resolution and at one bounded
-enlarged resolution so small codes remain readable without discarding context.
+Both detectors scan the complete frame at native resolution and at one bounded
+enlarged resolution so small markers remain readable without discarding context.
 Results from both scans are deduplicated and use source-frame coordinates.
 
 ```python
--- a/agent-sdk/xr-ai-tools/xr_ai_tools/marker_tracking.py
+++ b/agent-sdk/xr-ai-tools/xr_ai_tools/marker_tracking.py
@@ -22,8 +22,8 @@ from .tools import Tool
 from .types import StrictRequest
 
 _LOGGER = logging.getLogger(__name__)
-_QR_SCAN_LONG_EDGE = 3840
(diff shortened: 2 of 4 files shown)


wenxind-nvidia:
Reviewed at `f778ae48395fd4fed597c1bfd610c53595fd1f40` against current `main` at `307686143f5d45b1dce9c79441ca3c218bad7636`.

## Blockers

1. **Deduplication drops distinct nearby markers with the same value.** In [`marker_tracking.py:145-165`](https://github.com/NVIDIA/xr-ai/blob/f778ae48395fd4fed597c1bfd610c53595fd1f40/agent-sdk/xr-ai-tools/xr_ai_tools/marker_tracking.py#L145-L165) and [`marker_tracking.py:195-196`](https://github.com/NVIDIA/xr-ai/blob/f778ae48395fd4fed597c1bfd610c53595fd1f40/agent-sdk/xr-ai-tools/xr_ai_tools/marker_tracking.py#L195-L196), detections are deduplicated against every previously detected marker, including markers from the same scan. I reproduced two 6x6 ArUco markers with ID `23`, separated by two pixels: both native and enlarged scans detect both markers, but the tool returns only one. Preserve detections within each scan, then reconcile native versus enlarged detections one-to-one using spatial overlap. Add coverage for repeated ArUco IDs and repeated QR payloads.

2. **The enlargement is not bounded to the configured 3840-pixel long edge.** [`marker_tracking.py:102-115`](https://github.com/NVIDIA/xr-ai/blob/f778ae48395fd4fed597c1bfd610c53595fd1f40/agent-sdk/xr-ai-tools/xr_ai_tools/marker_tracking.py#L102-L115) uses an integer ceiling scale. A `3839x2160` frame becomes `7678x4320` (33.2 MP), while `3840x2160` is not enlarged. In a blank-frame benchmark, both detectors took 13 ms at 3840 pixels versus 136 ms at 3839 pixels. Resize using a fractional ratio capped at exactly 3840 pixels while retaining the six-times scale limit, and cover the boundary in a regression test.

The implementation is otherwise focused and preserves the public API. It merges cleanly with current `main`; all 11 focused tests and all 14 GitHub checks pass. The branch is three commits behind `main`.

nvddr:
Disposition for Wenxin’s review at commit `f0478fc`:

1. **Distinct nearby markers with the same value:** detections are now preserved within each scan. Native and enlarged results are reconciled only across scan passes, one-to-one, using spatial overlap. Added regressions for two nearby 6×6 ArUco markers sharing ID `23` and for repeated QR payloads.
2. **3840-pixel enlargement bound:** resizing now computes an exact target long edge of `min(3840, source_long_edge * 6)`. A `3839×2160` frame becomes `3840×2161`, while a 3840-pixel long edge is not enlarged. Actual per-axis resize ratios are retained for source-coordinate mapping. Added boundary and 6×-limit coverage.

Validation: the three new regressions and three marker-utility tests pass; Ruff and `git diff --check` pass; direct real-detector validation recovers the mixed QR/ArUco scene and the small downsampled ArUco marker pair.

wenxind-nvidia (approved):

PR #375, opened 2026-08-17
Add a bounded native tool-calling loop

nvddr:
## Summary

- add a common stateless `run_tool_loop(...)` helper beside the existing native one-call helpers
- preserve caller ownership of prompts, model parameters, history, participant context, retries, cancellation, and background tasks
- execute model-selected native tools sequentially through the existing Pydantic and NeMo Relay boundary
- return a complete intra-turn transcript and structured tool-call audit
- enforce model-iteration and total-tool-call budgets
- reject truncated responses, malformed call IDs, and ambiguous mixed `return_direct` batches before side effects
- keep unknown tools and invalid arguments model-repairable within the bounded turn

## Why

PR 355 deliberately removed the former generic Agent/AgentRunner because it owned application policy. Two concrete applications now need the same protocol-mechanical loop: the open lab-instrument monitoring sample in PR 363 and the forthcoming native tea-making sample.

This extraction remains a small turn engine rather than restoring an agent framework. Applications still own lifecycle, state, concurrency, model settings, and user-facing recovery.

## Behavior

The caller supplies initial messages, a per-invocation `ToolSet`, and a model callback. The callback receives immutable transcript and tool-definition tuples and retains control of the underlying `LLMService` call.

Tool batches run serially in model order. Cancellation and unexpected model/tool exceptions propagate. The helper never retries an executed tool. Typed errors carry the partial transcript and audit for application-specific handling.

PR 363 is not included because it remains an open PR and is not present on `main`; it can adopt this helper when restacked.

## Validation

- `tests/.venv/bin/pytest -q tests/test_tool_calling.py` — 25 passed
- existing ToolSet/tool-definition focused tests — 3 passed
- `uvx ruff check agent-sdk/xr-ai-tools/xr_ai_tools/tool_calling.py tests/test_tool_calling.py`
- `git diff --check`

### DIFF
--- a/agent-sdk/xr-ai-tools/README.md
+++ b/agent-sdk/xr-ai-tools/README.md
@@ -71,7 +71,9 @@ structured tool-call audit. Unknown tools and invalid Pydantic arguments are
 returned to the model for repair. Iteration exhaustion, tool-call budget
 exhaustion, empty or truncated responses, and unsafe mixed `return_direct`
 batches raise typed errors carrying the partial transcript and audit. Blank or
-duplicate tool-call IDs are rejected before executing their batch.
+non-string or duplicate tool-call IDs are rejected before executing their batch.
+Error `messages` contain only a valid, resumable transcript; the rejected model
+turn remains available separately as `rejected_response` for diagnostics.
 
 The application or agent continues to own prompts, conversation history, model
 parameters, retries, participant context, cancellation, and background tasks.
--- a/agent-sdk/xr-ai-tools/README.md
+++ b/agent-sdk/xr-ai-tools/README.md
@@ -19,7 +19,7 @@ The base install supplies finite `Tool` and streaming `AsyncTool` types. Install
 from pydantic import BaseModel
 from xr_ai_models import ChatMessage
 from xr_ai_tools import Tool, ToolSet
-from xr_ai_tools.tool_calling import handle_tool_call, tool_definitions
+from xr_ai_tools.tool_calling import run_tool_loop
 
 
 class LookupRequest(BaseModel):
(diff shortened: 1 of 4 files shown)


wenxind-nvidia:
**Blocker**

`run_tool_loop()` assumes that every tool-call ID is a string and calls `.strip()` before validating it. `ToolCall` is a plain dataclass, so its type annotation is not enforced at runtime, and an OpenAI-compatible response can contain a null or otherwise non-string ID. In that case the loop leaks an `AttributeError` instead of raising the documented `ToolLoopMalformedResponseError`. Please validate `isinstance(call.id, str)` before stripping it and add coverage for null/non-string IDs.

**Suggestion**

The loop appends the assistant tool-call message before rejecting truncated, over-budget, duplicate-ID, or mixed-direct-call responses. Consequently, the `messages` attached to those exceptions end with unanswered tool calls and cannot be sent directly back to an OpenAI-compatible API for the caller-owned retry. Consider keeping the rejected assistant turn out of the resumable transcript, exposing diagnostic and resumable transcripts separately, or documenting clearly that exception `messages` are diagnostic only.

nvddr:
Disposition for @wenxind-nvidia's review feedback in `20c846f`:

- **Blocker — fixed.** `run_tool_loop()` now validates that every tool-call ID is a non-empty string before calling string methods or executing any tool. Added regression coverage for null and numeric IDs; both now raise `ToolLoopMalformedResponseError`.
- **Suggestion — adopted.** Rejected model turns are no longer appended to `ToolLoopError.messages`. That transcript now remains valid and resumable, while the rejected `ChatResponse` is exposed separately as `rejected_response` for diagnostics. This applies to truncation, malformed responses, ambiguous direct-return batches, tool-call budget rejection, and empty responses.

Validation: `tests/test_tool_calling.py` passes with 25 tests; Ruff, SPDX, and DCO pass.

wenxind-nvidia (approved):

PR #376, opened 2026-08-17
Publish voice participant-joined runtime events

nvddr:
## Summary

- add the public empty `VoiceParticipantJoined` lifecycle payload
- let applications opt into join publication with `VoiceAgent(participant_joined_topic=...)`
- carry participant identity through `RuntimeContext.metadata`, matching participant-left events
- route joins through the existing private voice pipeline lifecycle frame
- serialize join and leave publication per participant without blocking media callbacks
- keep LiveKit and raw hub participant events out of application APIs

## Why

The voice pipeline already receives participant joins for routing and voice-gate greetings, but `VoiceAgent` only exposed departure and interruption through runtime topics. Applications that initialize participant-owned state on join otherwise have to reach into the transport and republish the same lifecycle event themselves.

## Behavior

The existing hub participant event is translated once into the private `ParticipantJoinedFrame`. The voice gate handles its greeting first, then the runtime bridge publishes the application-named typed topic. Existing roster replay, duplicate suppression while connected, and leave/rejoin behavior remain owned by the hub transport.

Join, leave, and rejoin publications are chained per participant so slow subscribers cannot reorder lifecycle state. Different participants remain independent, and the media callback only schedules delivery rather than awaiting subscribers.

No sample migration is included because current `main` has no sample-local join bridge; PR 363 can adopt the public event when restacked.

## Validation

- voice runtime lifecycle publication and non-blocking callback tests — passed
- blocked join followed by leave/rejoin ordering regression — passed
- private pipeline joined/left callback test — passed
- input transport joined-frame test — passed
- roster duplicate-suppression test — passed
- `uvx ruff check` on all changed Python files
- `git diff --check`

A broad `tests/test_voice_runtime.py` run reached the same pre-existing asynchronous Relay-summary stall after nine passing tests in the isolated clone; all changed-path focused tests pass.


blongs-nv (dismissed):
Reviewed at `97c235c2269fa4f1393a60aeb5d42ff684e376f8` by a bot.

This is a clean, symmetric extension of the existing lifecycle API: `VoiceParticipantJoined` mirrors `VoiceParticipantLeft` exactly (opt-in application-named topic, empty `extra="forbid"` payload, participant identity in runtime metadata), and the join publication reuses the hardened lifecycle path — the pipeline callback contains exceptions with log-and-continue ([`io.py:295-304`](https://github.com/NVIDIA/xr-ai/blob/97c235c2269fa4f1393a60aeb5d42ff684e376f8/agent-sdk/xr-ai-voice/xr_ai_voice/_processors/io.py#L295-L304)), the publish runs as a fire-and-forget lifecycle task whose failures are logged and which shutdown cancels and drains ([`_runtime.py:489-548`](https://github.com/NVIDIA/xr-ai/blob/97c235c2269fa4f1393a60aeb5d42ff684e376f8/agent-sdk/xr-ai-voice/xr_ai_voice/_runtime.py#L489-L548)), so a raising or hung subscriber can neither crash the worker nor block the pipeline. The gate-greeting-first ordering claim is accurate (the gate processes `ParticipantJoinedFrame` upstream of the IO processor and awaits the greeting before forwarding), roster duplicate suppression stays hub-owned and untouched, and the docs, README, migration row, and exports all move together. Ready to merge.

wenxind-nvidia:
**Blocker**

Participant lifecycle publications are not ordered. `_participant_joined()` and `_participant_left()` each detach `runtime.publish(...)` into an independent task, so a slow join subscriber can finish after the corresponding leave subscriber and recreate state for a participant who has already departed. I reproduced this by blocking the join subscriber, dispatching the leave, and then releasing the join; the final participant set incorrectly contained `alice`.

Please preserve join/leave/rejoin ordering per participant with a queue or chained task while keeping the media callback non-blocking, and add a regression test with a blocked join handler followed by leave/rejoin.

wenxind-nvidia (approved):

PR #377, opened 2026-08-17
docs: generate SDK API reference

nvddr:
## Summary

- add static, import-free Sphinx AutoAPI generation for the public `xr_ai_runtime` and `xr_ai_voice` facades
- use package `__all__`, annotations, defaults, and co-located docstrings as the API source of truth
- add a package-scoped contract check for unresolved or undocumented exports and public methods
- rebuild documentation when enrolled SDK Python sources change
- keep generated reference sources ephemeral and suppress private implementation-module pages

## Why

Recent API changes repeatedly copied signatures, defaults, and field descriptions into package READMEs and component pages. This establishes a code-first reference layer without installing or importing Pipecat, Torch, native libraries, or other SDK dependencies during the docs build.

## Stack

- PR 1: this foundation
- PR 2: #378 — complete SDK API rollout and deduplication
- PR 3: #379 — generated sample CLI reference
- PR 4: #380 — generated sample configuration reference

## Validation

- `uv run pytest -q test_api_documentation.py test_agent_runtime.py test_voice_runtime.py` — 43 passed
- Ruff 0.15.16 on changed Python files
- SPDX header check
- `git diff --check`

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -34,6 +34,8 @@ on:
     paths:
       - "docs/**"
       - "agent-sdk/*/README.md"
+      - "agent-sdk/xr-ai-runtime/**/*.py"
+      - "agent-sdk/xr-ai-voice/**/*.py"
       - ".github/workflows/docs.yaml"
       - ".github/scripts/select_latest_docs_release.py"
       - ".github/scripts/publish_docs.sh"
@@ -44,6 +46,8 @@ on:
     paths:
       - "docs/**"
       - "agent-sdk/*/README.md"
+      - "agent-sdk/xr-ai-runtime/**/*.py"
+      - "agent-sdk/xr-ai-voice/**/*.py"
       - ".github/workflows/docs.yaml"
       - ".github/scripts/select_latest_docs_release.py"
       - ".github/scripts/publish_docs.sh"
--- a/agent-sdk/xr-ai-runtime/xr_ai_runtime/events.py
+++ b/agent-sdk/xr-ai-runtime/xr_ai_runtime/events.py
@@ -19,8 +19,13 @@ class Topic(Generic[MessageT]):
     """A stable publish/subscribe name paired with its payload model."""
 
     name: str
+    """Stable name shared by every publisher and subscriber."""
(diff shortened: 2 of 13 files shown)


yanziz-nvidia (approved):

wenxind-nvidia:
**Blocker 1: the contract checker does not actually resolve facade exports**

`docs/source/_api_contract.py:88` scans every module and treats any same-named definition as satisfying `__all__`, even if `__init__.py` never binds that name. I reproduced a package with `__all__ = ["Client"]` and an unimported `implementation.Client`; validation incorrectly returned no failures. Duplicate class names can similarly select the wrong definition. Please trace each facade import or assignment to its exact definition and add an unbound-submodule regression test.

**Blocker 2: the generated public reference exposes private Pipecat plumbing**

`docs/source/conf.py:69` enables inheritance and all public members for exported classes. The resulting `HubVoiceTransport` page publishes `Bases: pipecat.transports.base_transport.BaseTransport`, `input() -> XRMediaHubInputTransport`, and `output() -> XRMediaHubOutputTransport`. This contradicts the repository contract that Pipecat and pipeline transports remain private inside `xr-ai-voice`. Please filter implementation-only inheritance and members from the public reference and test that generated facade pages do not contain Pipecat or private transport types.

**Suggestion**

`docs/source/_api_contract.py:63` validates public methods but not dataclass or Pydantic fields. Since this PR establishes co-located field descriptions as the source of truth, consider enforcing documentation for public class fields too.

nvddr:
Addressed all three points in `7f9c921`:

- The contract checker now resolves each facade binding through its exact relative-import/re-export chain. It no longer accepts an unrelated same-named definition; regression coverage includes an unbound submodule and duplicate names.
- The generated voice reference now omits implementation inheritance plus `HubVoiceTransport.input()` / `output()`. A post-build guard fails if Pipecat or either private transport type appears in the generated facade page.
- The contract now requires PEP-224 descriptions for public fields on exported dataclasses and Pydantic models.

Validation: 7 focused tests passed, Ruff passed, and both strict current and multiversion Sphinx builds passed.

PR #378, opened 2026-08-17
docs: generate complete SDK reference

nvddr:
## Summary

- roll static API generation out to all five SDK packages and intentionally public tools/model modules
- document the existing public surface in co-located docstrings without changing signatures, exports, or behavior
- enforce resolvable, documented literal `__all__` contracts and simplify duplicated API prose

## Stack

- Depends on #377
- This is PR 2 of 4 in the documentation-maintenance stack
- Followed by #379 and #380

## Validation

- `python3 docs/source/_api_contract.py`
- `uv run --with pytest==8.3.5 pytest -q --noconftest tests/test_api_documentation.py` (5 passed)
- model configuration/OpenAI compatibility/API documentation tests (92 passed)
- Ruff 0.15.16 on all touched Python areas
- AST comparison confirms SDK Python changes are documentation-only

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -34,8 +34,7 @@ on:
     paths:
       - "docs/**"
       - "agent-sdk/*/README.md"
-      - "agent-sdk/xr-ai-runtime/**/*.py"
-      - "agent-sdk/xr-ai-voice/**/*.py"
+      - "agent-sdk/**/*.py"
       - ".github/workflows/docs.yaml"
       - ".github/scripts/select_latest_docs_release.py"
       - ".github/scripts/publish_docs.sh"
@@ -46,8 +45,7 @@ on:
     paths:
       - "docs/**"
       - "agent-sdk/*/README.md"
-      - "agent-sdk/xr-ai-runtime/**/*.py"
-      - "agent-sdk/xr-ai-voice/**/*.py"
+      - "agent-sdk/**/*.py"
       - ".github/workflows/docs.yaml"
       - ".github/scripts/select_latest_docs_release.py"
       - ".github/scripts/publish_docs.sh"
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -91,8 +91,11 @@ See [Adding a sample](docs/source/guides/adding-a-sample.md) and the
 
 ## Change contract
 
-- Update the relevant README and `docs/source/` page in the same change as an
-  API, command, configuration, package, or quickstart change.
+- Public Python API names, signatures, types, defaults, and field behavior live
+  in `__all__`, declarations, and co-located docstrings; Sphinx generates their
+  reference pages. Update narrative README or `docs/source/` content only when
+  concepts, workflows, operations, or architecture change. Breaking changes
+  also require a migration entry.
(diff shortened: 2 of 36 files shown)


yanziz-nvidia (approved):

wenxind-nvidia:
**Blocker 1: imported exports are still resolved by unqualified name**

`docs/source/_api_contract.py:164` records only an imported local name, then searches the entire package for any definition with that name. I reproduced an export importing `Client` from a nonexistent module while an unrelated module defined a documented `Client`; validation incorrectly returned no failures. Package facades remain even looser because their bindings are not inspected at all.

Please build a module-qualified import map and resolve every `__all__` entry through the actual facade binding. Add tests for an unbound facade export and an import whose declared source does not contain the symbol.

**Blocker 2: the rollout exposes private implementation paths as public type references**

After enrolling the additional packages, the successful generated build contains 42 private-qualified references, including `make_llm() -> xr_ai_models._protocols.LLMService`, `ProcessorEndpoint.send_return_data(xr_ai_hub._types.DataMessage)`, `LLMSpec` inheriting from `_RoleSpec`, and the existing `HubVoiceTransport` Pipecat base and private transport returns.

This undermines the stated public facades and encourages consumers to import private modules. It also retains PR 377's Pipecat boundary leak. Please canonicalize types to their public aliases, hide private bases and implementation methods, and add a rendered-output check rejecting `xr_ai_*._*`, private transport types, and Pipecat references.

**Suggestion**

`docs/source/_api_contract.py:85` checks exported classes and methods but ignores their dataclass/Pydantic fields. Removing one of the newly added field descriptions still passes the contract. Consider validating public annotated fields because the new documentation policy names field behavior as co-located source-of-truth material.

blongs-nv:
Reviewed at `ccd4c8f767d3079ee1e507c281bd272388c026c5` by a bot. I independently reproduced both blockers already posted above; one addition beyond them:

## Suggestions

1. `PUBLIC_API_MODULES` is a hand-maintained enrollment list with no completeness guard: a public module added later (this package has gained one most weeks — `qr_code`, `image_polygon`, `marker_tracking`) ships with no reference page and no contract validation unless someone remembers to enroll it. [`docs/source/_api_contract.py:19-36`](https://github.com/NVIDIA/xr-ai/blob/ccd4c8f767d3079ee1e507c281bd272388c026c5/docs/source/_api_contract.py#L19-L36) Add a check that every non-underscore importable module under `API_PACKAGE_DIRS` is either enrolled or named in an explicit exclusion list (`xr_ai_tools/utilities` is already an implicit exclusion today).

nvddr:
Addressed the review feedback in `341f208` after rebasing the PR onto current `main` (including merged #377):

- The contract now verifies exact facade/import-source resolution and documents dataclass/Pydantic fields; I added the missing-source regression case.
- Generated signatures and summary tables canonicalize private-qualified types to public package exports. The voice readiness probe annotation is now expressed structurally, so it does not expose its private alias.
- The rendered-HTML guard now checks every generated Python API page and fails on any `xr_ai_*._*` path, Pipecat name, or private media transport.
- Every non-underscore importable SDK module must now be either enrolled in the public API reference or explicitly excluded; stale exclusions also fail the contract.

Validation completed:

- `12 passed` in `tests/test_api_documentation.py`
- Ruff and `git diff --check`

PR #379, opened 2026-08-17
docs: generate sample CLI reference

nvddr:
## Summary

- generate the user-facing sample command catalog from top-level sample `pyproject.toml` entry points
- statically extract literal `argparse` flags and help without importing sample or service dependencies
- render canonical `uv run` invocations while keeping launcher-managed worker/service protocols private

## Stack

- Depends on #378
- This is PR 3 of 4 in the documentation-maintenance stack
- Followed by #380

## Validation

- `uv run --with pytest==8.3.5 pytest -q --noconftest tests/test_cli_documentation.py` (2 passed)
- Ruff 0.15.16
- rendered HTML inspected for all three sample commands and option labels


yanziz-nvidia:
Odd I don't see preview with this change?

wenxind-nvidia:
**Suggestion: fail closed on `argparse` forms the formatter cannot represent**

`docs/source/_cli_reference.py:31` silently produces incorrect syntax for valid `argparse` declarations. A positional `add_argument("query")` is rendered as `query QUERY`, while no-value actions such as `store_const` and `version` are rendered with bogus value placeholders (`--mode MODE` and `--version VERSION`). `nargs` is not extracted, so it can also produce an inaccurate synopsis.

The current three sample commands only use supported boolean options, but this generator is being established as the authoritative CLI reference. Please either model positional arguments, `nargs`, and all supported no-value actions correctly, or explicitly reject unsupported declarations so the documentation build cannot publish false syntax. Focused formatter tests for these cases would protect that contract.

nvddr:
Addressed the feedback in `6478506` after rebasing onto current `main`:

- @wenxind-nvidia: positional arguments now render without duplicated metavars; `?`, `*`, `+`, and positive integer `nargs` forms render their actual cardinality; all built-in string no-value actions render without bogus placeholders; and unsupported actions, `nargs`, metavars, or incompatible combinations fail the docs build instead of producing guessed syntax. Added focused coverage for all of these cases.
- @yanziz-nvidia: PR preview roots now redirect to the checked-out `current/` build rather than the latest release, so opening the preview shows the PR's command-line reference. The Reference landing page also links directly to the generated page.

Validation completed:

- `19 passed` across the CLI/API documentation tests
- Ruff and `git diff --check`

PR 380 was restacked on the updated branch so its diff remains clean.

yanziz-nvidia (approved):

yanziz-nvidia:
I see it now

PR #380, opened 2026-08-18
docs: generate sample configuration reference

nvddr:
## Summary

- generate a sample-facing configuration catalog directly from canonical YAML/JSON files and adjacent comments
- auto-enroll files under top-level sample `yaml/` trees and beside direct capability subprojects, with no per-file manifest
- preserve exact configured values and guidance without inventing universal defaults or changing runtime loaders

## Stack

- Depends on #379
- This is PR 4 of 4 in the documentation-maintenance stack

## Why source-rendered

Configuration loading currently mixes typed models, permissive dictionaries, environment overrides, and computed defaults. Rendering the checked-in sample sources is accurate for every current config family and avoids introducing a new public schema or tightening runtime behavior.

## Validation

- `uv run --with pytest==8.3.5 pytest -q --noconftest tests/test_config_documentation.py tests/test_cli_documentation.py` (5 passed)
- Ruff 0.15.16
- rendered catalog inspected: 34 files across all three top-level samples


blongs-nv (approved):
Reviewed at `3ae8ad6f61e9cc9f3afee53be8ab25dd4dffb8c5` by a bot.

## Suggestions

1. The generated catalog renders the LiveKit dev placeholder credentials verbatim on a public docs page ([`agent-samples/simple-vlm-example/yaml/xr_media_hub.yaml:9-10`](https://github.com/NVIDIA/xr-ai/blob/3ae8ad6f61e9cc9f3afee53be8ab25dd4dffb8c5/agent-samples/simple-vlm-example/yaml/xr_media_hub.yaml#L9-L10), same in xr-render-demo), with no adjacent comment marking them as dev-only values that must be replaced. Since the page's stated convention is "keep field guidance beside the value as a source comment" ([`configuration.rst:11-12`](https://github.com/NVIDIA/xr-ai/blob/3ae8ad6f61e9cc9f3afee53be8ab25dd4dffb8c5/docs/source/reference/configuration.rst#L11-L12)), add that comment beside `api_key`/`api_secret` (or exclude/redact credential fields in the directive), so the reference doesn't read as an endorsement of shipping the prototype secret.

The generation itself is solid and consistent with the rest of the stack: discovery is conventional (top-level `yaml/` trees plus files beside capability subprojects, no manifest), symlinks and out-of-tree escapes are rejected with a test proving it, content is rendered verbatim so source comments survive as field guidance, the docs workflow's path filters now cover every enrolled file type so the published reference cannot go stale silently, and AGENTS.md plus the contributing guide document the convention. CI is fully green at this head including the strict Sphinx build, and the discovery contract tests pass in a fresh clone. Ready to merge.

wenxind-nvidia:
**Blocker: the catalog publishes obsolete model-server configurations as current sample configuration**

`docs/source/_config_reference.py:37` enrolls every YAML/JSON file under each sample. This currently includes the three hardware-specific `nemotron3_nano_llm_server.yaml` files. However, `agent-samples/model-servers/main.py:39` explicitly identifies `agent-llm` as replaced, `_stop_replaced_models()` calls it the superseded Nano text model, and `_build_processes()` now starts only STT, Omni, VLM, and embedding. No sample code references those three configuration files.

The generated reference therefore presents dead configurations alongside the active profiles. The focused fix is to remove the three obsolete files. If they are intentionally supported alternatives, connect them to an explicit documented invocation instead of allowing directory placement to imply that `model_servers` consumes them.

**Suggestion: make the generated headings participate in page navigation**

`docs/source/_config_reference.py:69` places all generated sections inside a `nodes.container`. In the strict rendered build, Sphinx consequently emits an empty secondary sidebar, so this 34-file page has no “On this page” navigation. Return proper section nodes or generate an explicit index so readers can reach a sample and configuration without scrolling through the entire catalog.

I also agree with the existing suggestion to mark the rendered LiveKit placeholder credentials as development-only; I have not duplicated that finding here.

nvddr:
Addressed all review feedback in `9120de6` after rebasing onto current `main`:

- @wenxind-nvidia: removed the three hardware-specific `nemotron3_nano_llm_server.yaml` files. The model-server sample no longer launches that superseded text model, so directory placement no longer causes the generated catalog to present dead configurations as current. I also corrected the adjacent Blackwell/Spark VLM load-order and memory-budget comments to describe STT + Omni + VLM.
- @wenxind-nvidia: the directive now returns real nested Sphinx sections rather than wrapping them in a generic container. The rendered page has a populated **On this page** sidebar with links for all three samples and their individual configuration files.
- @blongs-nv: both checked-in LiveKit configurations now label `devkey`/`devsecret-xr-livekit-prototype-2026` as development-only placeholders that must be replaced outside local samples. The preservation test covers that warning.

Validation completed:

- `22 passed` across config, CLI, and API documentation tests
- Ruff and `git diff --check`
- inspected the rendered secondary navigation entries for every sample/configuration

PR #381, opened 2026-08-18
fix(tools): scope image sanitizers per VLM call

nvddr:
## Summary
- register a unique Relay image sanitizer for each VLM call
- deregister it when finite or streaming inference completes
- cover repeated and overlapping image queries in one Relay scope

This extracts the shared fix previously buried in PR #363. Without it, long-lived background visual loops fail after their first call with `xr-ai-image already exists`.

## Validation
- `uv run --project tests pytest tests/test_image_tools.py -q` (23 passed)
- Ruff
- `git diff --check`

### DIFF
--- a/agent-sdk/xr-ai-tools/xr_ai_tools/_vision.py
+++ b/agent-sdk/xr-ai-tools/xr_ai_tools/_vision.py
@@ -7,24 +7,34 @@ from __future__ import annotations
 
 import copy
 import re
-from collections.abc import Mapping, Sequence
+from collections.abc import Iterator, Mapping, Sequence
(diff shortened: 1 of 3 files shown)


wenxind-nvidia:
**Suggestion**

The new repeated and overlapping regression tests cover only finite `ImageQueryTool` calls. Please add a streaming early-close or repeated-call case. The sanitizer remains registered across yields, and cleanup depends on async-generator finalization, so this would lock in the stated streaming cleanup guarantee. I manually exercised early close followed by another stream and it passes, so this is test hardening rather than a blocker.

wenxind-nvidia (approved):

PR #382, opened 2026-08-18
feat(sample): add native tea-making guidance

nvddr:
## Summary

- port the tea-making workflow to native XR agents and the shared bounded tool loop
- use Nemotron-3 Nano Omni for both language and visual inference
- preserve deterministic five-step guidance, background change/transcript/video observers, and participant lifecycle behavior
- route foreground replies and workflow notices through participant-scoped voice aggregation
- stream idle `current_view` responses directly from `StreamingImageQueryTool`, matching `simple-vlm-example`, while retaining structured multi-tool visual queries inside an active workflow
- default to wake-word voice activation while preserving explicit always-on mode
- publish spoken responses on `agent.response` so the connection client shows accessible text
- expose selected runtime events in the shared live browser viewer while retaining participant-scoped JSONL as the durable output

## Base

This PR is rebased directly onto `main`. Voice aggregation and the shared live-events viewer are consumed from their merged SDK implementations.

## Review disposition

- completed-step skip preserves verified state and follows normal advance behavior
- speech and guidance notices are delivered independently of JSONL sink failures
- participant leave immediately releases voice aggregation state
- idle visual questions use bounded direct streaming; active workflow visual tools retain multi-tool chaining
- workflow YAML values are type-checked when loaded
- image ownership is released only after all five image-producing consumers finish cleanup

## Live outputs

- spoken assistant text appears in the connection client's Agent Output through `agent.response`
- live structured events are available at `http://127.0.0.1:8092`
- `--expose-web-events` enables trusted-network access explicitly; the default remains loopback-only
- the viewer is bounded and non-durable; participant-scoped JSONL artifacts remain the durable record

## Validation

- 48 focused tea, CLI, and layout tests pass
- Ruff 0.15.16 passes
- generated dependency map is current
- strict Sphinx documentation build passes
- SPDX and DCO checks pass
- `git diff --check` passes

## Notes

No live GPU, camera, or end-to-end model pipeline was run in the development environment.


agent-samples/tea-making-sample/worker/tea_making_worker/workflow_state.py
```diff
@@ -0,0 +1,396 @@
...
+            return CommitResult(False, False, invalid, session.revision)
+        if was_complete:
+            self._event(session, "step.commit_noop", "state unchanged")
+            return CommitResult(True, True, "state unchanged", session.revision)
+        changes = {
+            name: value
+            for name, value in updates.items()
+            if name not in session.state or session.state[name] != value
+        }
+        candidate = {**session.state, **changes}
```
wenxind-nvidia, line 224: **Blocker:** `commit()` validates only writable names and primitive types before allowing `complete_when` to succeed. In the `identify` step, `{"tea_ready": true}` is therefore accepted as complete even when `tea_name`, `target_temperature_c`, and `steep_duration_s` were never written. I reproduced that path; advancing through `fill_water` then entering `heat_water` raises `ValueError: message references missing state: target_temperature_c`, after `_enter()` has already changed `step_id`, leaving a partially transitioned session. The skip path has a related invalid-state problem: `steeping_started_at_us: 1` makes the next timer immediately expired. Please validate the complete candidate against the state required by downstream triggers/messages and make transitions atomic, with regression tests for incomplete identification and the steeping skip path.

agent-samples/tea-making-sample/worker/tea_making_worker/workflow.py
```diff
@@ -0,0 +1,511 @@
...
+                        session,
+                        "agent.observation_error",
+                        str(exc),
+                    )
+                await self._flush(session)
+            await asyncio.sleep(_POLL_INTERVAL_S)
+
+    async def _tick(self, session: WorkflowSession) -> None:
+        async with session.lock:
+            if not session.active or session.step_id is None:
```
wenxind-nvidia: **Blocker:** Completed steps continue running the observation loop until the user explicitly advances. `_tick()` checks only `session.active`/`step_id`, while the observation prompt tells the model to call `workflow__commit` even when `already_complete` is true. That means a completed visual step keeps doing VLM + LLM work every 1.5–2 seconds; the completed final timer step calls the LLM every second indefinitely. Please short-circuit monitoring when the active step is complete and add a call-count regression test showing no model/vision calls occur while waiting for `next`.

agent-samples/tea-making-sample/worker/tea_making_worker/foreground.py
```diff
@@ -0,0 +1,269 @@
...
+        active_context = self._guidance.active_context(participant_id)
+        if active_context is None:
+            tools = self._root_tools(participant_id)
+            system_prompt = self._prompt
+            route = "root"
+        else:
+            active_tools = self._guidance.active_tools(participant_id)
+            if active_tools is None:
+                raise RuntimeError("active tea context has no active tool set")
+            tools = _merge_tool_sets(
```
wenxind-nvidia, line 207: **Blocker:** Once tea guidance is active, this route drops the change-watch, transcript, and video-log control tools and exposes only workflow tools plus `application_context__query`. A user who starts transcript or visual recording before guidance cannot stop or query it while the guide is active, despite the sample documenting these as independent opt-in background applications. This is also a privacy/lifecycle issue for recording controls. Please keep the background lifecycle controls available in both foreground routes and test stop/status commands during an active tea workflow.

agent-samples/tea-making-sample/worker/tea_making_worker/file_output.py
```diff
@@ -0,0 +1,270 @@
...
+    ) -> None:
+        await self._write("video-log.jsonl", record, ctx)
+
+    @subscribe(PARTICIPANT_LEFT_TOPIC)
+    async def participant_left(
+        self,
+        _event: VoiceParticipantLeft,
+        ctx: RuntimeContext,
+    ) -> None:
+        await self._close_participant(self._participant(ctx))
```
wenxind-nvidia: **Suggestion:** Participant-left subscribers fan out concurrently, so closing the sink here races the guidance/background handlers that cancel work and publish their final records. I reproduced a join/leave where `guidance.jsonl` contained the session header, `participant.joined`, and `session_end`, but no `participant.left`; `_closed` then causes the late record to be discarded. Please introduce an ordered teardown contract (for example, producer cleanup followed by a cleanup-complete event handled by the sink), or otherwise align the recorded lifecycle/documentation with the behavior.

agent-samples/tea-making-sample/worker/tea_making_worker/workflow.py
```diff
@@ -0,0 +1,511 @@
...
+                return
+            now = monotonic_now()
+            if session.next_tick > now:
+                return
+            step = self.workflow.step(session.step_id)
+            session.next_tick = now + step.trigger.interval_s
+            trigger = await self._trigger(session, step)
```
wenxind-nvidia: **Suggestion:** `_tick()` holds `session.lock` across frame acquisition and the VLM/LLM calls. Foreground advance/reset/status tools need the same lock, so a slow or timed-out model request can make user controls unresponsive for the full I/O timeout. Consider snapshotting the step and revision under the lock, doing external I/O without it, then reacquiring and applying results only if the revision still matches. A concurrency test with a stalled model and a simultaneous reset/advance would protect this contract.

agent-samples/tea-making-sample/worker/tea_making_worker/config.py
```diff
@@ -0,0 +1,121 @@
...
+    with path.open(encoding="utf-8") as stream:
+        data = yaml.safe_load(stream) or {}
+    if not isinstance(data, dict):
+        raise ValueError(f"worker config must be a YAML mapping: {path}")
+    return data
+
+
+def _prompt(data: dict[str, Any], config_path: Path | None, name: str) -> str:
+    inline = data.get(name)
+    if inline is not None:
```
wenxind-nvidia, line 66: **Suggestion:** The default worker YAML defines every prompt inline, so this branch always returns and the checked-in files under `prompts/` are never loaded by the default sample. That creates two prompt sources of truth and makes edits to the prompt files appear to have no effect. Please keep the defaults in one place (preferably the files, with an explicit override when needed) and test which source is selected.

README.md
```diff
@@ -131,6 +131,10 @@ starts everything.  Heavier demos (`xr-render-demo`) split model loading from
 the demo itself: start `model-servers` once, then run the demo as many times
 as you like without reloading weights.
 
+The [tea-making guidance sample](agent-samples/tea-making-sample/README.md)
```
wenxind-nvidia, line 150: **Suggestion:** Please complete the root README integration for the new sample. The hardware/VRAM table does not include `tea-making-sample`, and the shared model-server quickstart still says it is started before `xr-render-demo` without mentioning that tea-making also requires it. The Sphinx quickstart already includes the tea flow, so the root quickstart should give readers the same prerequisites and local footprint.

docs/source/reference/tea-making-sample.md
```diff
@@ -0,0 +1,294 @@
...
+
+- typed state fields and initial values;
+- ordered steps and explicit next-step IDs;
+- which state each step may read and write;
+- a periodic trigger and evidence rule;
+- tools available to observation and voice turns;
+- completion conditions and skip values;
+- enter, complete, and skip messages.
+
+`WorkflowState` enforces the contract. An observation can update only fields
```
wenxind-nvidia: **Nit:** There is no `WorkflowState` type in this implementation; the contract is enforced by `WorkflowStore`. Please use the actual class name so readers can find the referenced code.

agent-samples/tea-making-sample/worker/tea_making_worker/file_output.py
```diff
@@ -0,0 +1,270 @@
...
+
+    def __init__(self, output_dir: Path, *, history_size: int = 64) -> None:
+        if history_size <= 0:
+            raise ValueError("history_size must be positive")
+        self.output_dir = output_dir
+        self.output_dir.mkdir(parents=True, exist_ok=True)
+        self._history_size = history_size
+        self._sessions: dict[str, _SessionFiles] = {}
+        self._sessions_lock = asyncio.Lock()
+        self._active: set[str] = set()
```
wenxind-nvidia: **Nit:** `_active` is added to and discarded from, but never read. Please remove it unless it is intended to enforce a lifecycle invariant.

agent-samples/tea-making-sample/worker/tea_making_worker/file_output.py
nvddr: Addressed in b2364f5. Record-producing agents now publish participant cleanup-complete events after cancelling their participant work. FileOutputAgent waits for all five producers before appending session_end. The regression records participant.left before the cleanup barrier and verifies it precedes session_end.

agent-samples/tea-making-sample/worker/tea_making_worker/foreground.py
nvddr: Addressed in b2364f5. The active tea route now merges the same background context and change-watch, transcript, and video-log lifecycle controls as the root route. Added coverage that stop and status controls remain model-visible during an active workflow.

agent-samples/tea-making-sample/worker/tea_making_worker/workflow.py
nvddr: Addressed in b2364f5. GuidanceAgent._tick now returns before trigger or model work when the active step is complete. Added a regression asserting that a completed step makes no trigger, VLM, or LLM calls while waiting for explicit advancement.

agent-samples/tea-making-sample/worker/tea_making_worker/workflow_state.py
nvddr: Addressed in b2364f5. A completing commit now requires the full active-step writable state, completion messages are rendered against the candidate before mutation, and step transitions apply only after validation succeeds. The start-steeping skip now completes guidance and skips the timer instead of storing epoch 1. Added regressions for incomplete identification, atomic transition failure, and steeping skip behavior.

agent-samples/tea-making-sample/worker/tea_making_worker/config.py
nvddr: Addressed in b2364f5. Default inline prompt bodies were removed from the worker YAML; packaged prompt files are now the single default source while explicit inline values remain supported as overrides. Added coverage for both default-file selection and an explicit inline override.

tests/test_tea_making_sample.py
```diff
@@ -0,0 +1,482 @@
...
+        return SimpleNamespace(available=True, value="Twinings Earl Grey label")
+
+    guidance._trigger = visible_tea  # type: ignore[method-assign]
+    tick = asyncio.create_task(guidance._tick(session))
+    await asyncio.wait_for(started.wait(), timeout=1.0)
+    async with asyncio.timeout(0.1):
+        async with session.lock:
+            guidance.store.reset(session)
+    release.set()
+    await tick
```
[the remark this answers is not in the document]
nvddr: Addressed in 330f5ee. Replaced the bare task await with asyncio.wait_for(tick, timeout=1.0), which also prevents this concurrency regression from hanging indefinitely. The focused tea suite remains 14/14 passing.

agent-samples/tea-making-sample/worker/tea_making_worker/workflow.py
nvddr: Addressed in b2364f5. The observation loop snapshots step, revision, and state under the lock, performs trigger and model I/O without the lock, and accepts the commit only if the captured step and revision still match. Added a stalled-model regression showing reset completes promptly and the stale model commit is rejected.

README.md
nvddr: Addressed in b2364f5. The root hardware table now includes tea-making-sample with its shared-model and sample-service footprint, and the shared model-server quickstart names both tea-making and xr-render as consumers.

agent-samples/tea-making-sample/worker/tea_making_worker/file_output.py
nvddr: Addressed in b2364f5. Removed the unused _active set. File lifecycle state is now represented by active session objects, the closed set, and the cleanup-producer barrier.

docs/source/reference/tea-making-sample.md
nvddr: Addressed in b2364f5. The architecture guide now refers to the actual WorkflowStore class.

nvddr:
Review disposition (b2364f5, 330f5ee):

All nine Wenxin review threads and the follow-up code-quality thread were addressed and resolved. The changes cover complete-state validation, atomic transitions, skip semantics, completed-step polling, background controls during active guidance, ordered producer cleanup before session_end, lock-free model I/O with revision-guarded commits, one default prompt source, root README integration, the WorkflowStore name, removal of unused lifecycle state, and a bounded concurrency-test await.

Verification: 14 focused tea tests and 81 tea/runtime/tool/voice regressions pass; repository Ruff and git diff checks pass.

wenxind-nvidia:
Blockers

  1. /private/tmp/xr-ai-pr382-review/agent-samples/tea-making-sample/worker/tea_making_worker/
     workflow_state.py:159: skip=True applies state_on_skip even when the step is already complete. I reproduced
     verified Earl Grey settings being replaced with generic tea, 93 C, and 180 seconds. Apply fallback state only
     when skipping an incomplete step and add a regression test.

  2. /private/tmp/xr-ai-pr382-review/agent-samples/tea-making-sample/worker/tea_making_worker/foreground.py:141:
     persistence is ahead of speech on the same failure path. An OSError from FileOutputAgent becomes an
     ExceptionGroup, terminates the foreground task, and prevents the voice contribution at line 150. I reproduced
     a disk-full failure producing zero voice output. Isolate artifact failures from user-facing speech; _flush()
     has the same event-loss risk after draining notices.

  3. /private/tmp/xr-ai-pr382-review/agent-samples/tea-making-sample/worker/tea_making_worker/app.py:183: the
     aggregation agent is not connected to PARTICIPANT_LEFT_TOPIC. Pending speech is still published after
     disconnection, and participant state remains for the default 60-second idle timeout. I reproduced both
     behaviors. Add participant-scoped cancellation/release, likely through the stacked aggregation PR, and test
     disconnect/reconnect with the same participant ID.

  4. /private/tmp/xr-ai-pr382-review/agent-samples/tea-making-sample/worker/tea_making_worker/foreground.py:305
    ignores CurrentViewRequest.question, uses the raw user query, and makes current_view direct-return. This
    prevents required chains such as current_view followed by temperature__verify or rag_lookup.

  5. /private/tmp/xr-ai-pr382-review/agent-samples/tea-making-sample/worker/tea_making_worker/foreground.py:325
    removes the configured foreground VLM timeout. A stalled streaming VLM can now hold the turn indefinitely.

  Suggestions

  - /private/tmp/xr-ai-pr382-review/agent-samples/tea-making-sample/worker/tea_making_worker/spec.py:172: validate
    initial, complete_when, and state_on_skip values against declared field types while loading YAML. Currently
    adaptation errors fail late or make completion impossible.

  - /private/tmp/xr-ai-pr382-review/agent-samples/tea-making-sample/worker/tea_making_worker/images.py:36:
    participant-left subscribers run concurrently, so image references are released before the foreground and
    background cancellation handlers finish. This does not satisfy the documented cleanup ordering.

wenxind-nvidia (commented):
## Blockers

1. **Preserve verified state when skipping a completed step.** `WorkflowStore.advance()` applies `state_on_skip` whenever `skip=True`, even if the step is already complete. I reproduced a verified Earl Grey identification being replaced with `generic tea`, 93 C, and 180 seconds. Apply fallback state only when skipping an incomplete step and add a regression for skip after completion.

2. **Keep artifact failures off the speech path.** `ForegroundAgent._run_turn()` awaits `FOREGROUND_RECORD_TOPIC` before publishing normal voice output. `AgentRuntime` propagates subscriber failures, so an `OSError` from `FileOutputAgent` terminates the turn and produces no speech. I reproduced this with a simulated disk-full failure. Isolate record-sink failures from speech; `GuidanceAgent._flush()` needs equivalent protection because it drains events and notices before publishing them.

3. **Clean up aggregation state on participant leave.** `VoiceAggregationAgent` is registered without a connection to `PARTICIPANT_LEFT_TOPIC`. I reproduced queued speech being published after participant leave while the aggregation state remained alive. Add participant-scoped cancellation/release, wire it to the lifecycle topic, and cover disconnect/reconnect with the same participant ID.

4. **Preserve the `current_view` tool contract and required tool chains.** The streaming replacement ignores `CurrentViewRequest.question`, sends the raw user query to the VLM, and marks `current_view` as direct-return. Active policies that require `current_view` followed by `temperature__verify` or `rag_lookup` can no longer execute the second tool. Use the structured question and keep multi-tool active-step flows possible.

5. **Restore a bounded foreground vision timeout.** The direct streaming path removed `vlm_timeout_s` from `ForegroundAgent`, and `StreamingImageQueryTool` has no internal timeout. A stalled VLM can therefore hold a foreground turn indefinitely. Bound frame acquisition and stream consumption and return a deterministic unavailable result on timeout.

## Suggestions

1. Validate `initial`, `complete_when`, and `state_on_skip` values against declared state-field types while loading workflow YAML. Currently these adaptation errors fail late or make completion impossible.

2. Sequence participant image release after foreground and background task cancellation. Runtime subscribers fan out concurrently, so `ParticipantImageAgent` currently releases image references before other participant-left handlers necessarily finish.

## Nits

None.

Validation: 30 focused tea/aggregation tests pass, Ruff passes, and `git diff --check` passes. The tests do not currently cover the reproduced failure cases above.

nvddr:
Addressed the latest review in `e0ff399` and the viewer follow-up in `fc3ed75`:

- completed-step skip preserves verified state and advances normally
- speech and guidance notices are delivered independently of JSONL failures
- participant leave releases aggregation state immediately
- idle `current_view` remains bounded/direct-streaming; active workflow visual queries retain multi-tool chaining
- workflow YAML state values are type-checked at load time
- image release waits for all five producer cleanup barriers
- wake word is the default and spoken output is mirrored to `agent.response`

The branch is restacked on #386. The explicit tea event adapter now feeds the live loopback viewer while JSONL remains the durable output. Focused validation: 27 tests, Ruff, dependency-map, worker resolution, and diff checks pass.

wenxind-nvidia (commented):
## Blockers

1. **Handle empty VLM streams as unavailable.** In `agent-samples/tea-making-sample/worker/tea_making_worker/foreground.py:361`, an empty or whitespace-only stream returns an empty successful result without publishing any voice contribution. The direct-return path then marks `current_view` as already spoken, so even the `Done.` fallback is suppressed and the participant hears nothing. Please publish an unavailable response plus the stream terminator and return `available=False`; add a regression test for a stream that completes without usable text.

2. **Reject non-finite web-event payload values before storage.** `WebEvent` currently accepts `NaN` and infinity, while `_server.py` serializes responses with `allow_nan=False`. A single accepted non-finite value makes every `/api/events` poll fail until that event rolls out of the bounded store. Please reject non-finite values during model validation and cover the HTTP path with a regression test.

3. **Keep the web-events implementation sample-local until a second concrete use exists.** The only in-tree application dependency on `xr-ai-web-events` is the tea-making worker. Promoting it to a shared SDK package conflicts with the repository contract: "Do not add an abstraction until two concrete use cases need it." Please keep this implementation under the tea-making sample for now and extract it when another application needs the same boundary.

## Suggestions

None.

## Nits

None.

blongs-nv:
Reviewed at `8a787d4931cd97881b4de9d02e1e4f8eac2bac8d` by a bot. Everything posted on the thread is resolved in this head, each verified against the code: skipping a completed step now preserves the verified state (`skipping = skip and not complete`; my Earl Grey reproduction keeps 90°C/240s), speech is published before best-effort persistence with record failures contained and logged per-site (including `_flush`, which now publishes notices first and survives per-item subscriber failures), the aggregation agent releases participant state on leave (a sample subclass subscribes the leave topic and calls the new `release()` API), and image release now waits for the five-producer cleanup quorum. Each fix lands with a regression test that fails without it. The new event-viewer commit is clean: a thin projection agent republishing typed records to the SDK web-events server, localhost-bound by default with a bounded buffer.

## Blockers

1. A turn is spoken twice whenever the direct `current_view` call was not the model's only tool call. `_answer` reports the turn as already-spoken only when `result.return_direct and calls == ["current_view"]` ([`foreground.py:252-257`](https://github.com/NVIDIA/xr-ai/blob/8a787d4931cd97881b4de9d02e1e4f8eac2bac8d/agent-samples/tea-making-sample/worker/tea_making_worker/foreground.py#L252-L257)), but `run_tool_loop`'s audit spans the whole turn and `rag_lookup` is a non-direct tool: a `rag_lookup` round followed by the direct streaming `current_view` yields `calls == ["rag_lookup", "current_view"]`, so the chunks stream to the participant, the response closes, and `_run_turn` then publishes the full text again as a fresh interrupting utterance ([`foreground.py:154-164`](https://github.com/NVIDIA/xr-ai/blob/8a787d4931cd97881b4de9d02e1e4f8eac2bac8d/agent-samples/tea-making-sample/worker/tea_making_worker/foreground.py#L154-L164)). Track whether streaming actually happened (the tool can report it, or test the last call) instead of exact-matching a single-call turn, and add a mixed-turn regression test.

## Suggestions

1. The cleanup quorum can mix generations across a rejoin, in both agents that consume it. `FileOutputAgent.participant_joined` resets its completed-producer set, but a producer's cleanup event from a previous leave that arrives after the reset counts toward the next leave's quorum; `ParticipantImageAgent` has no join reset at all, so a partial quorum from one leave persists indefinitely and its stale entries let a later leave's quorum complete before all producers have finished ([`images.py:40-59`](https://github.com/NVIDIA/xr-ai/blob/8a787d4931cd97881b4de9d02e1e4f8eac2bac8d/agent-samples/tea-making-sample/worker/tea_making_worker/images.py#L40-L59)). A per-session generation counter in `ParticipantCleanupComplete`, checked by both consumers, closes it.

2. The inline `vlm` entry routes vision to the Omni server, which launches without `--limit-mm-per-prompt`, so vLLM's default of one image per prompt applies. Every call in the sample sends exactly one image, so it works, but nothing documents or enforces the ceiling — wiring in `query_images` later fails at runtime. Document the constraint beside the entry ([`yaml/models.local.json:10-22`](https://github.com/NVIDIA/xr-ai/blob/8a787d4931cd97881b4de9d02e1e4f8eac2bac8d/agent-samples/tea-making-sample/yaml/models.local.json#L10-L22)), which also omits `reasoning_field` and will drift from the preset.

3. `_relay_event_log` writes and flushes each relay event synchronously inside the relay subscriber callback ([`app.py:73-80`](https://github.com/NVIDIA/xr-ai/blob/8a787d4931cd97881b4de9d02e1e4f8eac2bac8d/agent-samples/tea-making-sample/worker/tea_making_worker/app.py#L73-L80)); every other file write in the sample goes through `asyncio.to_thread`. Buffer or offload it so chatty turns don't add per-event blocking I/O on the loop thread.

The fix round is mechanism-level rather than symptomatic — the streaming tool honors the model-authored question and is bounded by `vlm_timeout_s` with a spoken timeout message, both regression-tested, and the tea route uses the guidance-owned non-streaming tool so the streaming behavior is confined to the root route. The one gap is the `spoken` bookkeeping not matching the tool loop's actual audit semantics (blocker 1). CI at this head was still in progress at review time.

blongs-nv:
Reviewed at `7539f1d03842b8d374201ce73ce710fff0df156a` by a bot, covering the `fix(sample): allow remote tea event viewing` and `fix(sample): ignore empty vision stream chunks` commits. The empty-chunk commit is clean: leading blank chunks are skipped so the first spoken chunk carries the interrupt, an all-empty stream falls back to a spoken unavailable message, and both paths are pinned by new tests. Everything from my previous review still applies at this head — the double-speech blocker (the `spoken` heuristic still exact-matches `calls == ["current_view"]`, so a `rag_lookup` round followed by the direct streaming `current_view` speaks the answer twice) and the three suggestions (cleanup-quorum generation mixing in both consumers, the undocumented Omni one-image ceiling, the synchronous relay-event-log flush) are all unchanged in this delta.

## Blockers

1. The shipped worker config now binds the event viewer to all interfaces by default: `web_events_host: 0.0.0.0` ([`yaml/tea_making_worker.yaml:12`](https://github.com/NVIDIA/xr-ai/blob/7539f1d03842b8d374201ce73ce710fff0df156a/agent-samples/tea-making-sample/yaml/tea_making_worker.yaml#L12)). The viewer serves live voice transcripts, foreground answers, and camera-derived captions over plain HTTP with no authentication or TLS — the README's own words — and a sample's checked-in default is what runs on every install, firewall or not. Remote viewing is already supported securely by the existing config override plus the SSH-tunnel/subnet-scoped-firewall guidance in `networking.md`; keep the shipped default at `127.0.0.1` and let deployments opt in. Note the docs now contradict each other: [`networking.md:17`](https://github.com/NVIDIA/xr-ai/blob/7539f1d03842b8d374201ce73ce710fff0df156a/docs/source/getting_started/networking.md#L17) still says the viewer is "bound to 127.0.0.1 by default; do not expose", while the sample README and `test_web_event_config_defaults_and_validation` now assert the open default.

nvddr:
Final review disposition (`e13b64ff`):

All inline threads and all actionable top-level feedback are addressed on the current rebased head. The authoritative thread list has zero unresolved threads.

- Mixed `rag_lookup` -> streamed `current_view` turns now use the final direct-return call for spoken bookkeeping, so the streamed answer is not published a second time. A mixed-turn regression covers this path.
- Participant cleanup events now carry the participant-left message ID as a generation. `FileOutputAgent` and `ParticipantImageAgent` accept a quorum only for the active leave generation, preventing stale cleanup from a prior connection from closing a rejoined session or releasing its images.
- The inline Omni VLM adapter now declares `reasoning_field: reasoning_content`, and the README/reference guide document the deployment's one-image-per-request ceiling and the required model-server change before adopting a multi-image tool.
- Relay event serialization is queued immediately and file writes are batched through `asyncio.to_thread`; the Relay subscriber callback no longer performs synchronous write/flush I/O.
- Empty/whitespace-only vision streams produce a spoken unavailable result; foreground VLM work remains bounded by `vlm_timeout_s`.
- Verified state survives completed-step skips; speech/notices are isolated from JSONL failures; aggregation state is released on leave; active workflow visual tools retain chaining; YAML state values are type-checked; and image/session teardown waits for all five producers.
- The event viewer remains loopback-only by default. Remote access requires the explicit `--expose-web-events` flag and is documented as trusted-network-only.
- The Web Events package concerns are satisfied on the rebased base: non-finite payloads are rejected before storage, and the merged lab sample is the second concrete consumer of the shared boundary.
- The updated customer/adaptation guide is Markdown under `docs/source/`; this PR tracks no PDF files.

Validation: 40 focused tea/CLI tests pass, Ruff 0.15.16 passes, `git diff --check` passes, and the strict Sphinx/public-API documentation build passes.

wenxind-nvidia:
**Suggestion**

`agent-samples/tea-making-sample/worker/tea_making_worker/file_output.py:179`: Generation validation and session closing are not atomic. After readiness is calculated, `_sessions_lock` is released before `_close_participant()` reacquires it. A same-ID rejoin can reset the generation during that gap, after which the stale cleanup closes the rejoined session.

I reproduced this on the current head by queueing the final old-generation cleanup and the rejoin on `_sessions_lock`; the rejoin completed with no active session and the participant marked closed. Consider revalidating the expected generation while detaching the session under the same lock, with a regression test that forces this interleaving.

wenxind-nvidia (approved):

nvddr:
Addressed Wenxin’s rejoin/cleanup race in `28be8a5f`.

`FileOutputAgent` now carries the expected leave generation into `_close_participant()` and revalidates it while holding `_sessions_lock` before detaching or closing the session. If the same participant rejoins after the cleanup quorum becomes ready but before close begins, the rejoin clears the old generation and the stale close becomes a no-op.

Added a forced-interleaving regression that pauses the ready close, performs the same-ID rejoin, resumes stale cleanup, and verifies the active session remains open. Validation: 2 targeted generation/race tests and 33 focused tea tests passed (the unchanged known file-output integration hang was deselected); Ruff and `git diff --check` passed.

nvddr:
Final disposition on current head `8cec6b5e`:

- Addressed Wenxin’s cleanup/rejoin race. `FileOutputAgent` now passes the expected leave generation into `_close_participant()` and revalidates it under `_sessions_lock` before detaching the session, so stale cleanup cannot close a same-ID rejoin. A forced-interleaving regression covers the exact reported ordering.
- Rebased the complete tea sample stack onto current `main` after the DeviceIOHub rename.
- Updated the tea launcher, command, service path, YAML filename, and sample wording to the DeviceIOHub names. This resolves both Python 3.11/3.12 CI failures in `test_sample_hub_projects_resolve` and `test_hub_configuration_web_client_paths_resolve`.

Validation: the 2 targeted generation/race tests pass; the exact CI layout/launcher/config set passes 21/21; the focused tea suite passes 33 tests with the unchanged known file-output integration hang deselected; Ruff, dependency-map, DCO, and diff checks pass. The PR remains approved and mergeable, with fresh CI running on the rebased head.

blongs-nv (approved):

PR #383, opened 2026-08-18
docs: generate utility API reference

nvddr:
## Summary

- generate static API reference pages for all five packages under `utils/`
- enforce the existing literal `__all__` and co-located docstring contract for their public exports
- document previously uncovered utility fields, constants, protocols, and methods without changing runtime APIs or behavior
- rebuild documentation whenever utility Python sources change
- link the shared-utilities overview to the generated reference and remove its stale entry for the deleted `xr-ai-nemo-runtime` package

## Why

The code-generated reference introduced in #377 and #378 covered the five Agent SDK packages but omitted the public utility facades used by orchestrators, workers, and service wrappers. Their signatures and defaults therefore remained available only in source or hand-written narrative documentation.

This enrolls `xr_ai_launcher`, `xr_ai_logging`, `xr_ai_vad`, `xr_ai_vllm`, and `xr_ai_voicegate` using the same import-free AutoAPI pipeline. Package-root `__all__` declarations remain the publication boundary; implementation modules stay excluded.

## Validation

- utility and documentation tests: 251 passed
- strict Sphinx build
- generated-reference private-path inspection
- Ruff 0.15.16
- SPDX header check on all changed files
- `git diff --check`


yanziz-nvidia (approved):

wenxind-nvidia:
## Blockers

1. `utils/xr-ai-vllm/xr_ai_vllm/__init__.py:99` generates documentation recommending `Process(persistent=True)`, but `Process` has no `persistent` argument. Copying this raises `TypeError`; use `Process(..., launch_mode="persist")`.

2. `utils/xr-ai-launcher/xr_ai_launcher/_stack.py:94` says `Process.port` stops persistent processes, but the launcher never reads this field. Model cleanup instead uses a separate `_STOP_SERVICES` list. Either implement the documented behavior or describe `port` accurately as metadata.

## Suggestion

The generated signatures contain undefined, unlinked aliases such as `LaunchMode`, `QueryHandler`, and `StopHandler`. These aliases are not exported from their package roots because their implementation modules are excluded. Export and document them, or expand them to `Literal`/`Callable` annotations in the public declarations.

## Nit

`utils/xr-ai-voicegate/xr_ai_voicegate/config.py:23` should say an empty tuple dispatches every **non-STOP** transcript; STOP still routes to `on_stop`.

No runtime regressions found. Strict Sphinx, Ruff, the API contract, 232 lightweight utility tests, and 13 documentation tests passed. The six dependency-backed Silero VAD tests were not run because the isolated environment lacks Torch, Silero, and ONNX Runtime.

nvddr:
Addressed Wenxin's review in `df7d25a`:

- corrected the vLLM example to use `Process(..., launch_mode="persist")`
- documented `Process.port` accurately as stored metadata that the generic launcher does not inspect
- expanded generated public annotations to `Literal`/`Callable` types without exporting new APIs
- clarified that an empty voice-gate phrase tuple dispatches every non-STOP transcript

Validation: 171 focused utility/API documentation tests passed, Ruff and the API contract passed, and the strict Sphinx build passed.

PR #384, opened 2026-08-18
ci: generate Python dependency map

nvddr:
## Summary

- generate the exact Python project and direct dependency inventory in `DEPENDENCIES.md` from every repository `pyproject.toml`
- run the dependency-free generator through uv with an inline Python 3.11 requirement, including in pre-commit and CI
- regenerate the inventory automatically in pre-commit and reject drift in CI with an exact repair command
- validate internal `[tool.uv.sources]` paths, including missing, unused, mismatched, duplicate-name, and repository-escaping declarations
- support Python projects at the repository root as well as projects in repository subdirectories
- keep operational service topology, non-Python client ownership, change-impact guidance, and architectural dependency boundaries curated
- update agent and contributor instructions so dependency metadata is changed at its source instead of being copied by hand

## Why

Dependency changes currently require agents and contributors to remember a large manual documentation fan-out. This makes `pyproject.toml` authoritative for mechanical Python dependency facts while retaining the architectural guidance that cannot be generated.

## Validation

- `uv run --script .github/scripts/generate_dependency_map.py --check`
- `uv run --project tests pytest tests/test_dependency_map.py -q`
- `uvx ruff==0.15.16 check .github/scripts/generate_dependency_map.py tests/test_dependency_map.py`
- `python3 .github/scripts/check_spdx_headers.py <changed-files>`
- `uvx pre-commit run generate-dependency-map --all-files`
- YAML parse of the workflow and pre-commit configuration
- `make -C docs strict`
- `make -C docs strict-versions`
- `git diff --check`


blongs-nv:
Reviewed at `bf7aa9a22dd2228b301aba4ca0ad7086882e7dc6` by a bot.

## Suggestions

1. The generator hard-requires Python ≥ 3.11 (`import tomllib` at [`generate_dependency_map.py:16`](https://github.com/NVIDIA/xr-ai/blob/bf7aa9a22dd2228b301aba4ca0ad7086882e7dc6/.github/scripts/generate_dependency_map.py#L16)), but the pre-commit hook and every documented invocation run bare `python3` ([`.pre-commit-config.yaml:27`](https://github.com/NVIDIA/xr-ai/blob/bf7aa9a22dd2228b301aba4ca0ad7086882e7dc6/.pre-commit-config.yaml#L27), [`AGENTS.md:105`](https://github.com/NVIDIA/xr-ai/blob/bf7aa9a22dd2228b301aba4ca0ad7086882e7dc6/AGENTS.md#L105)). On a stock Ubuntu 22.04 contributor machine `python3` is 3.10, and the command fails with a raw `ModuleNotFoundError: No module named 'tomllib'` traceback (I hit this on the first run). Add a version guard that prints a clear "requires Python 3.11+" message (or a `tomli` fallback), since this script runs on contributor systems, not just CI's 3.12.

2. Two rejection paths the script implements are verified only by my probes, not by tests: an unused `tool.uv.sources` entry and a repository-escaping source path ([`generate_dependency_map.py:196-228`](https://github.com/NVIDIA/xr-ai/blob/bf7aa9a22dd2228b301aba4ca0ad7086882e7dc6/.github/scripts/generate_dependency_map.py#L196-L228)) both fail correctly today but have no pinning test. Also, a root-level `pyproject.toml` crashes with a raw `IndexError` at [`generate_dependency_map.py:295`](https://github.com/NVIDIA/xr-ai/blob/bf7aa9a22dd2228b301aba4ca0ad7086882e7dc6/.github/scripts/generate_dependency_map.py#L295) (reproduced; `relative_directory.parts` is empty), which `main()` does not catch, so the failure would be a traceback instead of the script's clear error format. Add tests for the two rejections and either a graceful error or explicit support for a root project.

The automation itself held up under adversarial probing: I constructed each violation the checker claims to catch and every one failed loudly — added dependency drift is rejected by `--check` with a working repair command, a missing `tool.uv.sources` path for an internal dependency, an unused source entry, a mismatched target, and a repo-escaping path all abort generation, and a hand-edit inside the generated markers is rejected while curated text outside the markers survives regeneration untouched. The generated inventory matches the pyprojects it is derived from (spot-checked `xr-ai-tools` and the service entries field-by-field), pre-commit and CI share the single script rather than parallel implementations, the CLI parses strictly, and every swept `pyproject.toml` is a live project that CI's `uv lock` job exercises with the same enumeration. CI is fully green at this head. Ready to merge.

nvddr:
Addressed Brent's review in `611b3d1` and rebased the branch onto current `main`:

- converted the generator to a dependency-free uv script with inline `requires-python = ">=3.11"` metadata
- routed the pre-commit hook, CI job, contributor/agent instructions, and stale-map repair command through `uv run --script`
- added uv setup to the generated-map CI job
- added coverage for unused sources and repository-escaping source paths
- explicitly supported and tested root-level Python projects

Validation: 11 dependency-map tests passed; the actual pre-commit hook passed; Ruff, SPDX, YAML parsing, strict docs, and multiversion docs passed.

yanziz-nvidia (approved):

PR #385, opened 2026-08-18
Add participant-scoped voice output aggregation

nvddr:
## Summary

- add an opt-in `VoiceAggregationAgent` and `VOICE_CONTRIBUTION_TOPIC`
- keep lone finite responses and urgent batches on a no-LLM fast path
- rewrite pending finite updates into one concise participant-scoped utterance
- pace active output with configurable bounded spoken-duration estimates
- preserve incremental stream ordering across buffered and interleaved contributions
- bound every participant's pending work with a priority-aware overflow policy
- let urgent output preempt active playback and cancel in-flight rewrites

## Why

`VoiceAgent` serializes responses per participant, but independent foreground, monitoring, timer, and alert agents can still enqueue unrelated utterances. Those messages may become stale or awkward before they are spoken. Applications need an explicit semantic layer before TTS that combines concurrent updates without moving application policy into the private media pipeline.

Producers opt in by publishing `VoiceOutput` payloads to `voice.contribution`; the aggregator is the only component that publishes the resulting stream to `voice.output`.

## Behavior

- aggregation state and pending capacity are independent per participant
- one finite contribution bypasses the LLM after a short coalescing window
- non-urgent updates arriving during estimated playback form the next batch
- multiple finite updates are rewritten once with bounded output, no thinking, and an explicit timeout
- urgent batches bypass rewriting; urgent arrival cancels an in-flight rewrite and supersedes active output
- a lone stream starts immediately with an aggregator-owned response ID
- matching stream chunks are found even when other contributions are interleaved in the bounded pending deque
- routine work never evicts urgent work; urgent-only overflow keeps the freshest alert
- discarded stream IDs remain quarantined through their terminator or a refreshed stream-idle expiry
- urgent finite output is spoken alone and first, with routine work restored in order
- idle exit is race-safe with concurrent enqueue
- shutdown and participant release count discarded pending and pre-publication in-flight work; contributions arriving after stop are logged and dropped
- model failures fall back to the original updates rather than dropping speech
- applications can call `release(participant_id)` for immediate participant cleanup

Pacing remains intentionally open-loop: the client/hub does not provide playback acknowledgements. Applications can configure `speech_rate_wpm`, `minimum_playback_s`, and `maximum_playback_s` for the selected TTS voice.

No sample is migrated in this PR. The viewer changes are being restored directly
to `main` in PR #389 after PR #386 was accidentally merged into this branch.
Sample PRs #363/#382 remain stacked above this PR and will be rebased after that
recovery lands.

## Review disposition

- buffered matching chunks and urgent interrupts are serviced without losing interleaved work
- `queue_capacity` now bounds the single pending deque; the unbounded backlog is removed
- urgent work bypasses/cancels rewriting
- idle-timeout enqueue races are closed
- shutdown/drop/discard accounting is explicit
- rewrites have a configurable deadline
- playback duration is configurable, validated, capped, and documented as an estimate
- priority-aware overload preserves urgent output and quarantines discarded stream tails deterministically
- urgent finite output is isolated from routine batches
- rewrite timeout is enforced locally even if a model implementation ignores the timeout parameter
- pre-publication in-flight work participates in shutdown/release discard accounting

## Validation

- `uv run --project tests pytest tests/test_voice_aggregation.py -q` — 36 passed
- Ruff 0.15.16 passes for the implementation and focused tests
- `git diff --check` passes
- the broader voice-runtime run passed its preceding ten tests before stalling in the unchanged `test_voice_agent_records_one_summary_for_finite_and_streamed_output`; no failure was observed

### DIFF
--- a/agent-sdk/xr-ai-voice/README.md
+++ b/agent-sdk/xr-ai-voice/README.md
@@ -104,25 +104,37 @@ pacing rather than a client playback acknowledgement. An estimate that is too
 short moves queuing downstream into TTS; one that is too long creates silence.
 Tune all three playback settings for the selected TTS voice.
 
-Rewrites have the explicit `rewrite_timeout_s` deadline. If a rewrite fails or
-times out, the original text is joined so updates are not lost. The earliest
-input timestamp is preserved. An `interrupt=True` contribution bypasses an LLM
-rewrite, cancels a rewrite already in progress, and replaces active speech
-without model latency. `queue_capacity` bounds all pending contributions for one
-participant, including chunks waiting behind another stream. When the bound is
-reached, the oldest non-urgent pending contribution is dropped with a warning
-so new output cannot grow stale in an unbounded buffer. Batches also have a
-bounded size; retained contributions remain ordered for following batches.
+Rewrites have an aggregator-enforced `rewrite_timeout_s` deadline, which is also
+passed to the model service. If a rewrite fails or times out, the original text
+is joined so updates are not lost. The earliest input timestamp is preserved.
+An `interrupt=True` contribution bypasses the coalescing delay and LLM rewrite,
+cancels a rewrite already in progress, and replaces active speech without model
+latency. Routine updates already collected in its batch remain ordered for the
+next batch. `queue_capacity` bounds all pending contributions for one
+participant, including chunks waiting behind another stream. At capacity,
+routine input replaces the oldest routine contribution but never an urgent one.
+Urgent input replaces the oldest routine contribution when possible, otherwise
+the oldest urgent contribution, so the freshest alert wins. Every drop is
+logged, and dropping a stream fragment quarantines that stream ID. Batches also
+have a bounded size; retained contributions remain ordered for following
+batches.
 
 A lone incremental contribution starts streaming immediately with a new
 aggregator-owned response ID. Other contributions wait behind that stream;
 finite updates that accumulated while it ran are coalesced when it finishes.
 An urgent contribution interrupts the active stream. A stream that stops
 producing chunks is closed after `stream_idle_timeout_s`, allowing pending
-updates to proceed. Call `stop()` before shutting down the runtime so the agent
-can cancel its participant-owned tasks. Contributions published after shutdown
-starts are logged and dropped; shutdown logs the number of accepted pending
-contributions it discards.
+updates to proceed. An interrupted or dropped stream ID remains quarantined
+until its terminator arrives. Each ignored fragment refreshes the quarantine;
+after a full `stream_idle_timeout_s` without another fragment, the ID may begin
+a new stream. This expiry bounds stale stream state independently from pending
+queue capacity.
+
+Call `stop()` before shutting down the runtime so the agent can cancel its
+participant-owned tasks. Contributions published after shutdown starts are
+logged and dropped. Shutdown and participant release log accepted contributions
+that are still pending or in pre-publication processing. Output already handed
+to the voice pipeline during the playback hold is not counted as discarded.
 
 ## Voice tuning and data echo
 
--- a/agent-sdk/xr-ai-voice/README.md
+++ b/agent-sdk/xr-ai-voice/README.md
@@ -87,6 +87,12 @@ async with runtime:
         await aggregation.stop()
 ```
 
+Applications should call `await aggregation.release(participant_id)` from
+their participant-left subscriber so buffered speech and stream state are
+discarded as soon as that participant departs. The application owns its
+lifecycle topic; the shared aggregator therefore exposes cleanup directly
+rather than subscribing to one hard-coded topic name.
+
 Aggregation is participant-scoped. A lone finite contribution passes through
 without an LLM call after the short coalescing window. Two or more finite
 contributions in one batch are rewritten into a single concise utterance. The
--- a/agent-sdk/xr-ai-voice/README.md
+++ b/agent-sdk/xr-ai-voice/README.md
@@ -11,6 +11,9 @@ applications work with XR concepts rather than Pipecat modules:
 - `VoiceAgent` publishes every final pre-gate STT result on `voice.transcript`
   and accepted speech, text, participant lifecycle, and interruption on typed
   topics; it subscribes to `voice.output`.
+- `VoiceAggregationAgent` optionally serializes candidate speech per
+  participant and rewrites simultaneous finite updates into one concise
+  response before publishing to `voice.output`.
 - `VoiceAgent` privately owns readiness, hub transport, pipeline assembly,
   signals, execution, and cleanup.
 - `HubVoiceTransport` is available when an application needs to share one transport explicitly.
@@ -44,6 +47,62 @@ runtime.register("voice", voice)
 async with runtime:
     await voice.run(runtime)
 ```
+
+## Multiple voice producers
+
+Applications with foreground responses, background monitors, timers, and
+alerts can register one `VoiceAggregationAgent`. Producers publish the same
+`VoiceOutput` payload to `VOICE_CONTRIBUTION_TOPIC`; only the aggregator
+publishes the final response to `VOICE_OUTPUT_TOPIC`:
+
+```python
+from xr_ai_voice import (
+    VOICE_CONTRIBUTION_TOPIC,
+    VoiceAggregationAgent,
+    VoiceOutput,
+)
+
+aggregation = runtime.register(
+    "voice-aggregation",
+    VoiceAggregationAgent(llm=llm),
+)
+```
+
+A producer uses its runtime context while handling an application event:
+
+```python
+await ctx.publish(
+    VOICE_CONTRIBUTION_TOPIC,
+    VoiceOutput(text="The timer is done."),
+)
+```
+
+The application stops participant-owned aggregation tasks before the runtime:
+
+```python
+async with runtime:
+    try:
+        await voice.run(runtime)
+    finally:
+        await aggregation.stop()
+```
+
+Aggregation is participant-scoped. A lone finite contribution passes through
+without an LLM call after the short coalescing window. Two or more finite
+contributions in one batch are rewritten into a single concise utterance; if
+the rewrite fails, their original text is joined so updates are not lost. The
+earliest input timestamp is preserved, and an interrupt on any contribution
+makes the combined response interrupt active speech. Queue capacity provides
+bounded ingress, and batches have a bounded size; additional contributions
+remain ordered for following batches.
+
+A lone incremental contribution starts streaming immediately with a new
+aggregator-owned response ID. Other contributions wait behind that stream;
+finite updates that accumulated while it ran are coalesced when it finishes.
+An urgent contribution interrupts the active stream. A stream that stops
+producing chunks is closed after `stream_idle_timeout_s`, allowing pending
+updates to proceed. Call `stop()` before shutting down the runtime so the agent
+can cancel its participant-owned tasks.
 
 ## Voice tuning and data echo
 
@@ -95,14 +154,15 @@ await runtime.publish(
 ```
 
 Incremental producers reuse a `response_id`, set `final=False` while more
-chunks remain, and end with `final=True`. Aggregation is private to voice/TTS,
-and producer identity is part of the response key so independent agents cannot
-merge output accidentally. Output is serialized per participant; urgent output
-sets `interrupt=True` to replace active and queued speech. Producers may copy
-the originating query's `timestamp_us` into `VoiceOutput` so the TTS response
-preserves the input timestamp. `VoiceStreamClosedError` identifies an empty final
-chunk for a stream that voice already closed, including when wrapped in the
-runtime publication exception group.
+chunks remain, and end with `final=True`. Without the optional aggregation
+agent, producer identity is part of the response key so independent agents
+cannot merge output accidentally. `VoiceAgent` serializes output per
+participant; urgent output sets `interrupt=True` to replace active and queued
+speech. Producers may copy the originating query's `timestamp_us` into
+`VoiceOutput` so the TTS response preserves the input timestamp.
+`VoiceStreamClosedError` identifies an empty final chunk for a stream that
+voice already closed, including when wrapped in the runtime publication
+exception group.
 
 Lifecycle publication runs on `VoiceAgent`-owned tasks, so participant cleanup
 cannot block the shared media processor. Join and leave publications are
--- a/agent-sdk/xr-ai-voice/README.md
+++ b/agent-sdk/xr-ai-voice/README.md
@@ -89,8 +89,13 @@ async with runtime:
 
 Aggregation is participant-scoped. A lone finite contribution passes through
 without an LLM call after the short coalescing window. Two or more finite
-contributions in one batch are rewritten into a single concise utterance; if
-the rewrite fails, their original text is joined so updates are not lost. The
+contributions in one batch are rewritten into a single concise utterance. The
+aggregator keeps each output open for an estimated spoken duration, based on
+`speech_rate_wpm`, so updates produced while that output is playing form the
+next batch instead of becoming separate queued utterances. This is pacing
+rather than a client playback acknowledgement, so applications can tune the
+speech-rate estimate for their selected TTS voice. If the rewrite fails, their
+original text is joined so updates are not lost. The
 earliest input timestamp is preserved, and an interrupt on any contribution
 makes the combined response interrupt active speech. Queue capacity provides
 bounded ingress, and batches have a bounded size; additional contributions
--- a/agent-sdk/xr-ai-voice/README.md
+++ b/agent-sdk/xr-ai-voice/README.md
@@ -97,15 +97,22 @@ Aggregation is participant-scoped. A lone finite contribution passes through
 without an LLM call after the short coalescing window. Two or more finite
 contributions in one batch are rewritten into a single concise utterance. The
 aggregator keeps each output open for an estimated spoken duration, based on
-`speech_rate_wpm`, so updates produced while that output is playing form the
-next batch instead of becoming separate queued utterances. This is pacing
-rather than a client playback acknowledgement, so applications can tune the
-speech-rate estimate for their selected TTS voice. If the rewrite fails, their
-original text is joined so updates are not lost. The
-earliest input timestamp is preserved, and an interrupt on any contribution
-makes the combined response interrupt active speech. Queue capacity provides
-bounded ingress, and batches have a bounded size; additional contributions
-remain ordered for following batches.
+`speech_rate_wpm` and clamped between `minimum_playback_s` and
+`maximum_playback_s`, so updates produced while that output is playing form the
+next batch instead of becoming separate queued utterances. This is open-loop
+pacing rather than a client playback acknowledgement. An estimate that is too
+short moves queuing downstream into TTS; one that is too long creates silence.
+Tune all three playback settings for the selected TTS voice.
+
+Rewrites have the explicit `rewrite_timeout_s` deadline. If a rewrite fails or
+times out, the original text is joined so updates are not lost. The earliest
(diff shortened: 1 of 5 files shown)


yanziz-nvidia:
Reviewed by yanziz-review-bot

**Summary**  
This adds participant-scoped voice aggregation with finite-response coalescing, streaming passthrough, and playback pacing. Two ordering and urgency regressions should be fixed before merge. DCO and CodeQL are green; the local aggregation and voice-runtime suites passed all 43 tests.

| | Finding |
|---|---|
| 🚫 | [`agent-sdk/xr-ai-voice/xr_ai_voice/_aggregation.py:243`](https://github.com/NVIDIA/xr-ai/blob/b7c2a0b211a763d68b80b003830cbf732f847113/agent-sdk/xr-ai-voice/xr_ai_voice/_aggregation.py#L243) — The streaming loop reads only `state.queue`, so continuation chunks already moved into `state.backlog` during playback pacing are ignored until the stream times out and are subsequently emitted as separate responses; consume through `_next(state)` and add coverage for a complete stream arriving during another response’s hold interval. |
| 🚫 | [`agent-sdk/xr-ai-voice/xr_ai_voice/_aggregation.py:336`](https://github.com/NVIDIA/xr-ai/blob/b7c2a0b211a763d68b80b003830cbf732f847113/agent-sdk/xr-ai-voice/xr_ai_voice/_aggregation.py#L336) — A batch containing `interrupt=True` still waits for the LLM rewrite, while interrupts arriving during that call remain queued, so urgent output can be delayed by model latency instead of immediately superseding speech; bypass or cancel rewriting when an interrupt is present or arrives. |
| 💡 | none |
| 🔍 | none |

- `agent-sdk/xr-ai-voice/xr_ai_voice/_aggregation.py:243` — Consume stream chunks through the backlog-aware `_next(state)` path and test a stream buffered entirely during playback pacing.
- `agent-sdk/xr-ai-voice/xr_ai_voice/_aggregation.py:336` — Ensure urgent contributions bypass or preempt LLM rewriting so interruption is not delayed by model latency.

wenxind-nvidia (changes requested):
## Blockers

1. **Buffered stream chunks are lost**

   In agent-sdk/xr-ai-voice/xr_ai_voice/_aggregation.py, _forward_stream() reads only state.queue, although playback and competing streams move contributions into state.backlog. A stream beginning behind active speech can therefore leave its final chunk in the backlog, wait the full stream idle timeout, close incorrectly, and discard that chunk. Interleaved streams fail similarly. Please service matching buffered chunks and buffered interrupts before waiting on the queue, and add regressions for delayed and interleaved streams.

2. **queue_capacity does not bound pending work**

   Both _forward_stream() and _hold_output() continuously drain the bounded queue into an unbounded backlog. With queue_capacity=1, I can enqueue 20 pending contributions behind a stream and all 20 accumulate in the backlog. A continuously producing stream can therefore grow pending memory without bound. Please apply the capacity to total pending state, including the backlog, using an explicit backpressure, drop, or coalescing policy.

## Nits / API clarity

- _playback_duration() has an undocumented 30-second ceiling. For example, minimum_playback_s=60 is accepted but results in 30 seconds. Please expose a maximum_playback_s setting, or validate and document the ceiling.
- The README claim that queue capacity provides bounded ingress is misleading while the backlog is unbounded.
- The finite-only behavior otherwise looks reasonable: simultaneous updates combine, updates during estimated playback form the next batch, participant processing is independent, and rewrite failures preserve the original content.

blongs-nv:
Reviewed at `df69e7639481324cd1fa50d4357e1ee8ec27b9e6` by a bot. I reproduced both blockers from the review above and confirmed they are still present at this head, so I won't list them again; the delta since that review is a rebase onto current `main` plus a new `release(participant_id)` method. The items below are additional.

## Suggestions

1. The bounded queue is quietly defeated by the unbounded backlog. `_hold_output` and `_forward_stream` drain the capacity-limited queue into `state.backlog`, a plain deque with no limit ([`_aggregation.py:51`](https://github.com/NVIDIA/xr-ai/blob/df69e7639481324cd1fa50d4357e1ee8ec27b9e6/agent-sdk/xr-ai-voice/xr_ai_voice/_aggregation.py#L51), [`_aggregation.py:395-404`](https://github.com/NVIDIA/xr-ai/blob/df69e7639481324cd1fa50d4357e1ee8ec27b9e6/agent-sdk/xr-ai-voice/xr_ai_voice/_aggregation.py#L395-L404)). Simulation: a 24-update flood against 0.25 s playback drains at `max_batch_size` per playback cycle with spoken-latency growing ~1.2 s per cycle (0.03 s → 0.31 s → 1.39 s → 2.62 s batch starts); producers sustaining more than that rate grow memory and staleness without bound or signal. Also note `queue_capacity`'s one observable effect is undocumented: when the intake queue is momentarily full, `contribute` blocks the *publisher* inside `runtime.publish`. Bound the backlog (drop-oldest with a warning) or document `queue_capacity` as a burst bound with publisher backpressure.

2. A contribution can be silently dropped at the idle-timeout boundary: when `_run_participant` returns on idle timeout ([`_aggregation.py:180-187`](https://github.com/NVIDIA/xr-ai/blob/df69e7639481324cd1fa50d4357e1ee8ec27b9e6/agent-sdk/xr-ai-voice/xr_ai_voice/_aggregation.py#L180-L187)), a `contribute` that passed the `task.done()` check just before completion puts into the old queue, which nothing reads again — the next contribution creates a fresh state ([`_aggregation.py:123-134`](https://github.com/NVIDIA/xr-ai/blob/df69e7639481324cd1fa50d4357e1ee8ec27b9e6/agent-sdk/xr-ai-voice/xr_ai_voice/_aggregation.py#L123-L134)). Drain the queue before returning, or re-dispatch leftovers from the done-callback.

3. The shutdown contract is lossy in two undocumented ways. Once `stop()` has run, `contribute` raises `RuntimeError` into whatever agent published the contribution ([`_aggregation.py:122`](https://github.com/NVIDIA/xr-ai/blob/df69e7639481324cd1fa50d4357e1ee8ec27b9e6/agent-sdk/xr-ai-voice/xr_ai_voice/_aggregation.py#L122)), since `runtime.publish` re-raises subscriber exceptions to the publisher — a producer flushing its last utterance during shutdown gets an exception group instead of a dropped message. And `stop()` during an active playback hold cancels the participant task with everything already accepted into the backlog, silently: in simulation, six updates accepted during a long announcement's hold vanished with no log line when `stop()` landed before the hold expired. (The new `release()` inherits the same silent-discard behavior, which is correct for a departed participant but worth a debug log.) Log-and-drop when stopping, and log the discarded count on cancel.

4. The rewrite call has no explicit deadline ([`_aggregation.py:427-440`](https://github.com/NVIDIA/xr-ai/blob/df69e7639481324cd1fa50d4357e1ee8ec27b9e6/agent-sdk/xr-ai-voice/xr_ai_voice/_aggregation.py#L427-L440)): on a wedged model, speech stalls for the transport default of 60 seconds before the concatenation fallback fires. `LLMService.chat` accepts a `timeout` argument; a few seconds fits an utterance-combining call.

5. The pacing model is open-loop and the documentation doesn't say so. Playback is estimated from `speech_rate_wpm` with a hard 30 s cap ([`_aggregation.py:406-409`](https://github.com/NVIDIA/xr-ai/blob/df69e7639481324cd1fa50d4357e1ee8ec27b9e6/agent-sdk/xr-ai-voice/xr_ai_voice/_aggregation.py#L406-L409)) and there is no feedback from actual TTS completion (the voice pipeline tracks response end internally but exposes no topic for it). A first-time integrator cannot predict the failure modes from the README: an estimate that runs low re-creates the speech queue one layer down (the IO processor queues non-interrupt responses per participant, invisible to the aggregator, defeating coalescing), and an estimate that runs high inserts dead air. The README also never mentions the estimate, its knobs, or what `interrupt=True` means through the aggregator. Document the open-loop model and its tuning, or expose a playback-completion signal the aggregator can consume.

Behavioral assessment from a multi-producer simulation harness at this head (four producers, varied timing): outside the items above, the core design behaves reasonably. A simultaneous four-producer burst becomes one coherent utterance with every fact preserved and one model call; updates arriving during a long announcement's hold wait it out and coalesce into a single follow-up; long/short alternation batches sensibly with no producer starved; a producer resuming after the idle timeout gets a fresh task with nothing lost at normal timing; a finite update landing mid-stream is correctly deferred behind the stream and spoken 30 ms after it ends; and nothing was ever spoken twice in any scenario (the new `release()` behavior is pinned by the PR's own added test, which passes here along with the other 16). The serialization, coalescing, and per-participant independence claims all held up under the harness — the defects on this PR are concentrated in the two-channel buffer split and the urgency path, not the architecture.

nvddr:
Addressed all actionable feedback in `209368c`.

- replaced the queue-plus-unbounded-backlog design with one capacity-bounded pending deque and warned drop-oldest behavior
- preserved matching stream chunks and urgent interrupts when unrelated work is interleaved
- made urgent batches bypass rewriting and urgent arrival cancel an in-flight rewrite
- closed the idle-timeout/enqueue race
- added explicit shutdown, post-stop drop, participant-release, and discard accounting
- added a configurable `rewrite_timeout_s`
- added validated `maximum_playback_s` and documented pacing as an open-loop estimate rather than playback acknowledgement

Focused aggregation validation is 27/27; Ruff and diff checks pass. The PR has no inline review threads (`review_threads: []`), so there are no thread IDs to resolve; the changes-requested review and both top-level comment clusters are covered by this disposition.

wenxind-nvidia:
Reviewed current head `209368c`. Verdict: **changes requested**. The previously reported backlog, buffered-stream ordering, idle-race, and rewrite-preemption issues are fixed, but the following correctness issues remain.

## Blockers

1. **Routine work can evict urgent output**

   [`_enqueue()`](https://github.com/NVIDIA/xr-ai/blob/209368cbaaa7c06b0d9d36ea33ee8fbfedf9c42a/agent-sdk/xr-ai-voice/xr_ai_voice/_aggregation.py#L231-L260) searches for a non-urgent victim but defaults to index `0` when none exists. I reproduced this through normal runtime publication while output delivery was blocked:

   ```text
   pending: urgent-one, urgent-two
   enqueue: routine
   result:  urgent-two, routine
   ```

   This contradicts the documented “oldest non-urgent” policy and allows a routine update to displace an alert. When only urgent work is buffered, the incoming routine contribution should be dropped instead; urgent-versus-urgent overflow should have an explicit policy.

2. **Discarded streams can resume after tombstone rollover**

   [`_discard_stream()`](https://github.com/NVIDIA/xr-ai/blob/209368cbaaa7c06b0d9d36ea33ee8fbfedf9c42a/agent-sdk/xr-ai-voice/xr_ai_voice/_aggregation.py#L264-L275) limits discarded stream keys to `queue_capacity` and evicts an arbitrary key with `set.pop()`. With `queue_capacity=1`, this is reproducible end-to-end:

   ```text
   A-start
   B-start (interrupts A)
   C-start (interrupts B)
   C-end
   A-stale-tail   ← incorrectly spoken as a new response
   ```

   A discarded stream must remain discarded through its terminator or a safe expiry/generation boundary. Pending queue capacity must not permit interrupted speech to reappear.

3. **An urgent finite update can be spoken after arbitrarily long routine text**

   For an alert arriving during the coalescing window, [`_speak_batch()`](https://github.com/NVIDIA/xr-ai/blob/209368cbaaa7c06b0d9d36ea33ee8fbfedf9c42a/agent-sdk/xr-ai-voice/xr_ai_voice/_aggregation.py#L463-L474) concatenates the batch in FIFO order. The PR test currently pins the result as [`"Routine update. Move away."`](https://github.com/NVIDIA/xr-ai/blob/209368cbaaa7c06b0d9d36ea33ee8fbfedf9c42a/tests/test_voice_aggregation.py#L464-L486).

   Although the output has `interrupt=True`, TTS must speak all preceding routine text before reaching the warning. This makes urgency depend on whether the alert arrives during coalescing or during rewriting. Dispatch the urgent contribution alone, or place urgent text first and retain routine work for a later batch.

## Suggestions

1. [`rewrite_timeout_s`](https://github.com/NVIDIA/xr-ai/blob/209368cbaaa7c06b0d9d36ea33ee8fbfedf9c42a/agent-sdk/xr-ai-voice/xr_ai_voice/_aggregation.py#L632-L646) is forwarded to `LLMService.chat`, but the aggregator does not enforce the documented deadline. A service accepting but not honoring that argument can stall participant processing indefinitely. Consider wrapping the await in `asyncio.timeout()` while retaining the service-level timeout.

2. Shutdown logging counts only `state.pending`. A batch already removed for rewriting can be discarded while [`_log_discarded()`](https://github.com/NVIDIA/xr-ai/blob/209368cbaaa7c06b0d9d36ea33ee8fbfedf9c42a/agent-sdk/xr-ai-voice/xr_ai_voice/_aggregation.py#L664-L684) reports that the state was merely released. Include in-flight accepted work in the accounting or narrow the documented guarantee.

## Nits

None.

## Validation

- Focused aggregation suite: 27 passed.
- Ruff 0.15.16 and `git diff --check`: passed.
- All current GitHub checks are green.
- Both overload cases above were reproduced independently at this head.

blongs-nv:
Reviewed at `209368cbaaa7c06b0d9d36ea33ee8fbfedf9c42a` by a bot. I verified all previously posted findings as fixed at this head, and confirmed the three blockers posted in the latest review; none of that is relisted. One additional blocker, distinct from the coalescing-order one (that reorders preserved content; this loses content):

## Blockers

1. **When an urgent contribution cancels an in-flight rewrite, the batch being rewritten is silently dropped.**

   [`_aggregation.py:484-501`](https://github.com/NVIDIA/xr-ai/blob/209368cbaaa7c06b0d9d36ea33ee8fbfedf9c42a/agent-sdk/xr-ai-voice/xr_ai_voice/_aggregation.py#L484-L501)
   The urgent branch cancels `rewrite`, dispatches the urgent contribution, and returns, abandoning `batch` — up to `max_batch_size` already-dequeued updates are never spoken, re-enqueued, or counted by the discard accounting. Reproduced: two coalesced monitor updates enter the rewrite, an alert lands mid-rewrite, and only the alert is ever spoken. Re-enqueue the batch behind the urgent contribution or speak the concatenation fallback after it, and log any residual discard.

The redesign otherwise held up under multi-producer simulation with no behavior changes outside the fixed paths, and CI is green at this head. The blocker above should be resolved before merge alongside the posted three.

nvddr:
Addressed the follow-up feedback from comment `5334308587` in `bb8e448`:

- routine input can no longer evict urgent output; urgent input evicts the oldest routine item, or the oldest urgent item when the queue is all-urgent so the freshest alert wins
- discarded stream tombstones are independent of queue capacity, refresh while stale fragments arrive, clear on the terminator, and expire only after `stream_idle_timeout_s`
- urgent finite output bypasses coalescing and is spoken alone; previously collected routine work is restored in original order
- `rewrite_timeout_s` is enforced with `asyncio.timeout()` as well as passed to the model service
- shutdown/release accounting includes accepted pre-publication in-flight batches; already-published playback is intentionally not counted as discarded

Focused aggregation validation is now 34/34. Repository-wide Ruff, touched-file formatting, diff, and DCO checks pass. This feedback is a top-level comment and PR 385 still has no inline review threads to resolve.

wenxind-nvidia:
Reviewed current head `474b944`. Verdict: **changes requested**.

The latest aggregation commit fixes all three blockers from my prior comment, plus the hard rewrite deadline and shutdown accounting. One correctness blocker remains.

## Blockers

1. **Routine updates are lost when an urgent contribution preempts a rewrite**

   In [`_speak_batch()`](https://github.com/NVIDIA/xr-ai/blob/474b944e4f55f8eb4d407998bb9cfb41f98ea210/agent-sdk/xr-ai-voice/xr_ai_voice/_aggregation.py#L545-L564), the urgent branch cancels the rewrite and dispatches the alert, but never restores `batch`. Up to `max_batch_size` already-dequeued routine updates disappear.

   Reproduced at this head:

   ```text
   input:   routine-one, routine-two
   rewrite: starts
   input:   urgent
   output:  urgent, <terminator>
   pending: []
   ```

   This also makes behavior timing-dependent: routine work is preserved when urgency is detected during coalescing, but lost once rewriting has begun. Re-enqueue the batch in original order using the capacity policy before dispatching the urgent contribution. Extend the existing test to wait for and assert the subsequent routine output.

## Suggestions

1. **The PR scope is now inconsistent.** Commit [`474b944`](https://github.com/NVIDIA/xr-ai/commit/474b944e4f55f8eb4d407998bb9cfb41f98ea210) brings the complete, separately approved PR 386 into PR 385—31 additional files and both sample integrations—while the title remains voice aggregation and the body still says no sample is migrated. Either remove or retarget that commit, or update the title and description to make the combined landing explicit.

2. `discarded_streams` now has correct quarantine semantics, but expired entries are only globally pruned when another stream is discarded. An active participant producing only finite output can retain expired tombstones indefinitely. Prune them during normal participant processing or add scheduled cleanup.

3. The included web viewer recognizes IPv6 loopback hosts in request validation, but its `ThreadingHTTPServer` remains IPv4-only and `url` does not bracket IPv6. `WebEventsAgent(host="::1", port=0).start()` currently raises `gaierror`. Either support IPv6 binding and URLs or reject and document IPv4-only hosts.

## Nits

- Ruff lint passes, but `ruff format --check` reports that `tests/test_web_events.py` needs formatting.

## Validation

- Aggregation tests: 34 passed.
- Web-events tests: 14 passed.
- All current GitHub checks are green.
- Rewrite-preemption loss reproduced independently.

nvddr:
Addressed the final actionable feedback in `f18142e`.

- urgent preemption now restores the cancelled rewrite batch to pending state in original order under the existing capacity policy before dispatching the alert
- the regression waits for the subsequent routine output and verifies both original routine updates reach the follow-up rewrite
- expired discarded-stream tombstones are pruned during normal finite contribution processing
- PR 385 is voice-only again; the viewer scope was removed and is tracked separately in PR 389

Focused aggregation validation is 35/35. Ruff 0.15.16, formatting, and `git diff --check` pass. The remaining inline code-quality threads are outdated and reference viewer files no longer present in this PR.

blongs-nv (approved):

wenxind-nvidia:
## Re-review

Reviewed at `f18142e`.

**Decision: request changes.**

### Blockers

1. **An interleaved stream fragment prevents an already-queued urgent alert from being selected first.**

   In [`_finite_batch()`](https://github.com/NVIDIA/xr-ai/blob/f18142e338aa9dc6d5c07a2bc911a9c28bc6d3f9/agent-sdk/xr-ai-voice/xr_ai_voice/_aggregation.py#L365-L395), a non-final stream fragment is returned to the queue and terminates batching before the subsequent `interrupt` check can inspect urgent work behind it.

   I reproduced this sequence during the coalescing window:

   1. Routine finite contribution
   2. Non-final routine stream fragment
   3. Urgent finite contribution

   Current output:

   ```text
   Routine.       interrupt=False
   Move now.      interrupt=True
   ```

   The routine response is started and then discarded by the interrupt instead of being retained for later. This contradicts the PR's stated guarantee that urgent finite output is spoken first and displaced routine work remains ordered for the next batch.

   Please check for and remove pending urgent work before stopping at a stream boundary, restore the routine batch, and add this interleaving as a regression test.

### Suggestions

None beyond the blocker.

### Nits

- The PR validation section says 34 focused tests passed, but the current test file contains and runs 35 passing cases.
- Reflow the awkward sentence break in `docs/source/components/agent-sdk.md`: "and the / agent enforces..."

### Previous findings

- The previous blocker is fixed: `f18142e` restores an in-flight routine rewrite batch before urgent dispatch and preserves its ordering.
- PR 389's viewer files are no longer present.
- Expired discarded-stream entries are now pruned during finite work.

### Validation

- Focused aggregation suite: **35 passed** on Python 3.12.
- Ruff 0.15.16: passed.
- `git diff --check`: passed.
- GitHub checks were still queued at the time of review.

wenxind-nvidia (approved):

PR #386, opened 2026-08-18
feat(runtime): add live web events viewer

nvddr:
## Summary

- add a small `xr-ai-web-events` SDK package with a typed `WebEvent` topic and `WebEventsAgent`
- serve a bounded, live in-memory event view grouped by participant and topic
- provide a rollover-aware polling API, health endpoint, and packaged static browser UI
- document explicit sample integration, localhost-only defaults, SSH-tunnel access, and source-restricted trusted-LAN firewall configuration

## Why

The lab-instrument and tea-making samples already persist durable JSONL records, but watching those files is inconvenient during live runs. This gives applications an explicit, read-only browser view over selected runtime events without adding file tailing, a wildcard runtime tap, or media-hub coupling.

The viewer is intentionally separate from `agent.response`: spoken assistant text continues to use the existing voice-to-client accessibility channel, while this page is operational telemetry.

## Behavior and safety

- applications explicitly publish only the events they want exposed
- bounded memory; JSONL remains the durable record
- defaults to `127.0.0.1:8092`
- no CORS, no authentication, and no TLS; remote development uses an SSH tunnel
- lifecycle is idempotent and a bind failure fails startup

## Validation

- `uv run --project tests pytest tests/test_web_events.py tests/test_agent_runtime.py -q` — 24 passed after restacking
- full focused package/runtime suite from the implementation worktree — 33 passed
- Ruff 0.15.16 — clean
- dependency-map check — current
- strict Sphinx/public API check — passed
- wheel build/static asset packaging — passed
- SPDX and `git diff --check` — clean

### DIFF
--- a/DEPENDENCIES.md
+++ b/DEPENDENCIES.md
@@ -126,6 +126,17 @@ change, even when an individual dependency publishes newer Python wheels.
 - Optional dependency groups: none
 - Commands: none
 
+#### `xr-ai-web-events` — [`agent-sdk/xr-ai-web-events/`](agent-sdk/xr-ai-web-events/)
+
+- Python: `>=3.11,<3.13`
+- Build dependencies:
+  - `hatchling`
+- Runtime dependencies:
+  - `pydantic>=2.10`
+  - `xr-ai-agent-runtime` → [`xr-ai-agent-runtime`](agent-sdk/xr-ai-runtime/) (local, editable)
+- Optional dependency groups: none
+- Commands: none
+
 ### Utilities
 
 #### `xr-ai-launcher` — [`utils/xr-ai-launcher/`](utils/xr-ai-launcher/)
@@ -514,6 +525,7 @@ change, even when an individual dependency publishes newer Python wheels.
   - `xr-ai-models[riva]` → [`xr-ai-models`](agent-sdk/xr-ai-models/) (local, editable)
   - `xr-ai-tools[frames,image-editing,marker-tracking,services,vision]` → [`xr-ai-tools`](agent-sdk/xr-ai-tools/) (local, editable)
   - `xr-ai-voice` → [`xr-ai-voice`](agent-sdk/xr-ai-voice/) (local, editable)
+  - `xr-ai-web-events` → [`xr-ai-web-events`](agent-sdk/xr-ai-web-events/) (local, editable)
   - `xr-media-hub` → [`xr-media-hub`](services/xr-media-hub/) (local, editable)
   - `xr-ai-launcher` → [`xr-ai-launcher`](utils/xr-ai-launcher/) (local, editable)
   - `xr-ai-logging` → [`xr-ai-logging`](utils/xr-ai-logging/) (local, editable)
--- a/DEPENDENCIES.md
+++ b/DEPENDENCIES.md
@@ -456,6 +456,7 @@ change, even when an individual dependency publishes newer Python wheels.
   - `xr-ai-models[riva]` → [`xr-ai-models`](agent-sdk/xr-ai-models/) (local, editable)
   - `xr-ai-tools[frames,vision]` → [`xr-ai-tools`](agent-sdk/xr-ai-tools/) (local, editable)
   - `xr-ai-voice` → [`xr-ai-voice`](agent-sdk/xr-ai-voice/) (local, editable)
+  - `xr-ai-web-events` → [`xr-ai-web-events`](agent-sdk/xr-ai-web-events/) (local, editable)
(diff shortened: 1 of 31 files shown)


yanziz-nvidia:
Reviewed by yanziz-review-bot

**Summary**  
This PR adds a typed web-events SDK, bounded event store, HTTP polling API, packaged browser UI, and networking documentation. The focused 24-test suite passes locally and all completed CI checks are green, but two security/correctness blockers and one viewer usability issue remain; the Pages publish job is still queued.

| | Finding |
|---|---|
| 🚫 | [`agent-sdk/xr-ai-web-events/xr_ai_web_events/_server.py:49`](https://github.com/NVIDIA/xr-ai/blob/68dcd6b7c4897822f360f56b83f81833eac8052b/agent-sdk/xr-ai-web-events/xr_ai_web_events/_server.py#L49) — Because loopback binding is the only protection for potentially sensitive transcripts, accepting every `Host` leaves `/api/events` exposed to DNS-rebinding pages; reject unrecognized hosts before serving any route and test the permitted loopback values. |
| 🚫 | [`agent-sdk/xr-ai-web-events/xr_ai_web_events/_models.py:13`](https://github.com/NVIDIA/xr-ai/blob/68dcd6b7c4897822f360f56b83f81833eac8052b/agent-sdk/xr-ai-web-events/xr_ai_web_events/_models.py#L13) — `WebEvent` accepts `NaN` and infinity inside `JsonValue`, but `_json()` uses `allow_nan=False`, so one valid published measurement makes `/api/events` terminate every response until eviction; reject non-finite values during model validation and cover them in tests. |
| 💡 | [`agent-sdk/xr-ai-web-events/xr_ai_web_events/static/app.js:46`](https://github.com/NVIDIA/xr-ai/blob/68dcd6b7c4897822f360f56b83f81833eac8052b/agent-sdk/xr-ai-web-events/xr_ai_web_events/static/app.js#L46) — Every incoming event replaces all topic panes, so disabling Follow resets each feed to the top instead of preserving the user’s scroll position; retain per-pane positions or append incrementally when Follow is off. |
| 🔍 | none |

- `agent-sdk/xr-ai-web-events/xr_ai_web_events/_server.py:49` — Validate the request `Host` against explicitly permitted listener names and addresses before serving health, API, or static routes, with loopback and rejection tests.
- `agent-sdk/xr-ai-web-events/xr_ai_web_events/_models.py:13` — Configure `WebEvent` to reject nested non-finite floats and add regression tests proving `NaN` and infinity cannot reach JSON serialization.
- `agent-sdk/xr-ai-web-events/xr_ai_web_events/static/app.js:46` — Preserve each feed’s scroll position across renders when Follow is disabled, or update the existing DOM incrementally.

yanziz-nvidia (approved):

wenxind-nvidia (commented):
Re-reviewed current head `0a97dc6`. The new finite-number validation resolves the JSON serialization failure. The following findings remain.

## Blockers

1. `agent-sdk/xr-ai-web-events/xr_ai_web_events/_agent.py:99`: `stop()` clears `_server` and `_thread` before its first cancellable await. If cancellation occurs during `server.shutdown()`, `server_close()` and `thread.join()` are skipped, while a later `stop()` returns immediately because the references are already gone. I reproduced the listener remaining bound after this sequence. Keep the references until cleanup completes and ensure close/join run even when shutdown is cancelled; please add a cancellation regression test.


## Suggestions

1. `docs/source/getting_started/networking.md:81` tells users to change `web_events_host` and `web_events_port` in sample worker configuration, but no sample configuration parser recognizes either key. Document the actual constructor wiring, or add the corresponding real sample configuration.

2. `agent-sdk/xr-ai-web-events/xr_ai_web_events/static/app.js:18`: `participantName()` maps an absent participant to the literal string `Global`. A real participant whose ID is `Global` is therefore merged with global events and cannot be filtered separately. Keep the internal participant/grouping key distinct from the display label.

nvddr:
Disposition for the latest Wenxin review, addressed in b0eb933:

- Cancellation-safe shutdown: WebEventsAgent now retains its listener references until cleanup completes, shields shutdown from caller cancellation, and guarantees server close plus thread join. Added a regression test that cancels stop during shutdown and verifies the port can be rebound.
- Concrete consumers: integrated the viewer into simple-vlm-example and xr-render-demo. Each application explicitly selects participant-scoped query and response events for publication.
- Real configuration: added web_events_host and web_events_port to both worker parsers and shipped YAML files, and updated the related sample and networking documentation.
- Participant identity: unscoped events now use a distinct internal key and display as Global (unscoped), so a real participant named Global remains independently grouped and filterable.

Validation: 141 relevant tests passed; Ruff, JavaScript syntax validation, and git diff checks passed.

wenxind-nvidia (approved):

PR #388, opened 2026-08-18
Fix render worker startup and align dependency rule

blongs-nv:
The render worker crashed at startup with `AttributeError: 'WorkerConfig' object has no attribute 'models_yaml'`: `__main__.py` still read the `models_yaml` field that the deployment-profile conversion (#329) removed from `WorkerConfig`, replacing it with the `load_models(config_path)` helper.

- The worker now builds its model clients through `load_models(config_path)`, extracted into `_make_model_services()` so the exact wiring `main()` uses is testable.
- Adds a worker-startup regression test that constructs every model client from the shipped worker YAML; it fails with the `AttributeError` on unfixed main.
- Aligns the curated dependency rule in `DEPENDENCIES.md` with the approved `xr-ai-models[riva]` exception already recorded in `AGENTS.md` (the curated rule predated that approval and still stated the unamended contract).

Verified: full non-GPU suite passes; the render demo stack was started end-to-end on hardware with the shipped default profile — worker reaches ready and serves.

### DIFF
--- a/DEPENDENCIES.md
+++ b/DEPENDENCIES.md
@@ -600,7 +600,9 @@ Keep non-obvious fan-out in the same change:
   server-side packages.
 - `agent-sdk/xr-ai-models/` depends only on `xr-ai-logging`, `httpx`, and
   `pyyaml`. In-tree backends use typed OpenAI-compatible HTTP rather than vendor
-  SDKs.
+  SDKs, with one scoped exception: the optional `riva` extra adds
+  `nvidia-riva-client` for gRPC-only Riva speech NIMs (see AGENTS.md); the
+  base install is unchanged.
 - `agent-sdk/xr-ai-tools/` keeps capability-specific dependencies optional;
   spatial math remains CPU-only.
 - Agent workers use public SDK packages and task-specific libraries. They never
--- a/agent-samples/xr-render-demo/worker/xr_render_demo_worker/__main__.py
+++ b/agent-samples/xr-render-demo/worker/xr_render_demo_worker/__main__.py
@@ -12,7 +12,7 @@ from pathlib import Path
 
 from loguru import logger
 from xr_ai_logging import setup_logging
-from xr_ai_models import ChatMessage, load_models_config, make_llm, make_stt, make_tts, make_vlm
+from xr_ai_models import ChatMessage, make_llm, make_stt, make_tts, make_vlm
 from xr_ai_runtime import AgentRuntime
 from xr_ai_tools.tool_calling import tool_definitions
 from xr_ai_voice import (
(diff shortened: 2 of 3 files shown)


wenxind-nvidia (commented):
Re-reviewed current head `55fe542`. The render-worker startup fix is correct: `main()` now resolves the selected profile through `load_models(config_path)`, so it no longer accesses the removed `WorkerConfig.models_yaml` field. The following issues remain.

## Blocker

1. `DEPENDENCIES.md:601` resolves the documentation contradiction by permitting `nvidia-riva-client`, but the current repository contract says `xr-ai-models` may depend only on `xr-ai-logging`, `httpx`, and `pyyaml`, with no vendor SDKs. Relaxing the dependency map does not resolve that architectural violation. Keep the dependency rule intact and isolate Riva behind an external/OpenAI-compatible service boundary instead of placing the vendor client in `xr-ai-models`.

## Suggestion

- `tests/test_xr_render_demo_wire.py:1221`: `test_worker_builds_model_services_from_shipped_config` tests only the extracted helper. It would still pass if `main()` stopped using that helper and reintroduced the original stale-field crash. Add a worker-entry test that verifies `main()` passes `config_path` through `_make_model_services`.

## Nit

- `tests/test_xr_render_demo_wire.py:1224`: the regression comment records the historical failure. Per the repository comment convention, reduce it to the current contract: the worker must construct every model client from the profile selected by its worker YAML.

Verification: 42 render-wire tests and 37 profile/launcher tests passed; Ruff and the generated dependency-map check passed.

blongs-nv:
Updated for @wenxind-nvidia comment's suggestion and nit, not the blocker as explained below.

The speech NIMs (parakeet-tdt, magpie-tts) are Riva servers: they only speak Riva's gRPC protocol, and their HTTP port serves health checks only. There is no OpenAI-compatible surface to use httpx against, so supporting them requires a Riva gRPC client.

The shipped shape keeps the footprint minimal: nvidia-riva-client is an optional [riva] extra, imported only inside make_stt/make_tts when a profile selects kind: riva_grpc. The base install is unchanged. This PR's DEPENDENCIES.md edit just brings the curated rule in line with AGENTS.md and the pyproject as merged in #329 — before it, the two contract documents contradicted each other.

wenxind-nvidia (approved):

nvddr (commented):
The startup fix and dependency-document reconciliation look correct; the earlier test-wiring and comment feedback are addressed. One remaining test-hygiene issue: `test_worker_main_resolves_models_from_config_path` aborts only after `main()` calls `logger.add(...)`, leaving a process-global Loguru sink registered and touching `/tmp/xr-agent-trace.log` after the test. Please stub `worker_main.logger.add` in this test, or remove the returned sink in a `finally`, so the test isolates only the `config_path` wiring.

nvddr (approved):

PR #389, opened 2026-08-18
feat(runtime): add live web events viewer

nvddr:
## Summary

- restore the final reviewed live web-events viewer change from PR #386 on the intended `main` base
- add the typed `WebEvent`/`WebEventsAgent` SDK package and bounded browser UI
- keep sample integrations out of this SDK-focused change
- retain loopback defaults, SSH-tunnel guidance, source-restricted trusted-LAN configuration, and explicit IPv4 listener validation
- bound each serialized event payload to 16 KiB in addition to the count-bounded history

## Why

PR #386 was accidentally based on and merged into PR #385 (`agent/voice-output-aggregation`) instead of `main`. This replacement reconstructs the reusable viewer SDK, browser assets, tests, and documentation directly on current `main`. It intentionally contains neither voice-aggregation code nor changes to existing samples.

PR #385 has already been restored to its voice-only scope. After this PR merges, the samples that use the viewer can be restacked without changing unrelated existing samples here.

## Validation

- web-events focused suite — 16 passed
- web-events plus unchanged Simple VLM/XR Render suites — 87 passed
- public API documentation contract — 13 passed
- strict Sphinx documentation build — passed
- source distribution and wheel build — passed, including static assets
- generated dependency map — current
- Ruff 0.15.16 check and formatting for the web-events package and tests — clean
- `git diff --check` — clean

## History

The original PR #386 will remain marked as historically merged into PR #385. This replacement corrects the actual branch topology and final code placement.

### DIFF
--- a/DEPENDENCIES.md
+++ b/DEPENDENCIES.md
@@ -456,7 +456,6 @@ change, even when an individual dependency publishes newer Python wheels.
   - `xr-ai-models[riva]` → [`xr-ai-models`](agent-sdk/xr-ai-models/) (local, editable)
   - `xr-ai-tools[frames,vision]` → [`xr-ai-tools`](agent-sdk/xr-ai-tools/) (local, editable)
   - `xr-ai-voice` → [`xr-ai-voice`](agent-sdk/xr-ai-voice/) (local, editable)
-  - `xr-ai-web-events` → [`xr-ai-web-events`](agent-sdk/xr-ai-web-events/) (local, editable)
   - `xr-ai-voicegate` → [`xr-ai-voicegate`](utils/xr-ai-voicegate/) (local, editable)
   - `loguru>=0.7`
   - `pyyaml>=6.0`
@@ -504,7 +503,6 @@ change, even when an individual dependency publishes newer Python wheels.
   - `xr-ai-models` → [`xr-ai-models`](agent-sdk/xr-ai-models/) (local, editable)
   - `xr-ai-tools[frames,services,vision]` → [`xr-ai-tools`](agent-sdk/xr-ai-tools/) (local, editable)
   - `xr-ai-voice` → [`xr-ai-voice`](agent-sdk/xr-ai-voice/) (local, editable)
-  - `xr-ai-web-events` → [`xr-ai-web-events`](agent-sdk/xr-ai-web-events/) (local, editable)
   - `xr-ai-voicegate` → [`xr-ai-voicegate`](utils/xr-ai-voicegate/) (local, editable)
   - `xr-ai-logging` → [`xr-ai-logging`](utils/xr-ai-logging/) (local, editable)
   - `xr-render-scene` → [`xr-render-scene`](agent-samples/xr-render-demo/scene/) (local, editable)
--- a/DEPENDENCIES.md
+++ b/DEPENDENCIES.md
@@ -126,6 +126,17 @@ change, even when an individual dependency publishes newer Python wheels.
 - Optional dependency groups: none
 - Commands: none
 
+#### `xr-ai-web-events` — [`agent-sdk/xr-ai-web-events/`](agent-sdk/xr-ai-web-events/)
+
+- Python: `>=3.11,<3.13`
+- Build dependencies:
+  - `hatchling`
+- Runtime dependencies:
+  - `pydantic>=2.10`
+  - `xr-ai-agent-runtime` → [`xr-ai-agent-runtime`](agent-sdk/xr-ai-runtime/) (local, editable)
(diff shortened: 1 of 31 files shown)


wenxind-nvidia:
## Review

Reviewed at `6734422`.

### Blockers

None.

### Suggestions

1. **Handle or explicitly reject IPv6 listener addresses.**

   [`WebEventsAgent.start()`](https://github.com/NVIDIA/xr-ai/blob/6734422bb70a728a6add0088951ec32aa9c653a7/agent-sdk/xr-ai-web-events/xr_ai_web_events/_agent.py#L55-L76) accepts `::1`, and host validation recognizes IPv6, but `ThreadingHTTPServer` remains IPv4-only. I reproduced `WebEventsAgent(host="::1", port=0).start()` failing with `gaierror`. The generated URL would also be invalid because IPv6 literals require brackets.

   Either select `AF_INET6` and format URLs as `http://[::1]:port`, or reject IPv6 clearly and remove the misleading allowances. Please add a real IPv6 bind test rather than only sending an IPv6 `Host` header to an IPv4 listener.

### Nits

- The PR description says Ruff formatting is clean, but `ruff format --check` still reports several modified config/test files. Formatting is not currently enforced by CI and much of the affected formatting predates this PR, so clarifying the scope of that validation would be enough.

### Validation

- The branch is correctly based directly on current `main` and contains no voice-aggregation implementation.
- The implementation matches the final PR 386 viewer code, aside from formatting corrections in `test_web_events.py`.
- `test_web_events.py`: 14 passed on Python 3.12.
- Both new sample event-publication tests passed.
- Ruff check passed.
- GitHub's Python 3.11/3.12 tests, build, dependency-map, lock, SPDX, DCO, Ruff, and CodeQL checks are passing.

Overall: approve with the IPv6 issue treated as a non-blocking follow-up.

wenxind-nvidia (approved):

blongs-nv:
Reviewed at `6734422bb70a728a6add0088951ec32aa9c653a7` by a bot. The `xr-ai-web-events` package, static UI, tests, and networking docs are byte-identical to the final reviewed content of PR #386 (differing only in test-file line wrapping), so the security and protocol verification from that review carries over: loopback bind in code, `Host` allowlist, non-finite payload rejection, `textContent`-only rendering under CSP, read-only endpoints, correct rollover cursors, and cancellation-safe `stop()`. The PR contains no voice-aggregation code and imports nothing from the unmerged aggregation branch. The sample integrations are new surface and are wired correctly: the viewer wraps the runtime so it outlives it, and forwarded events keep participant scope through `ctx.publish`.

## Suggestions

1. The event store is bounded by count, not by size: `WebEvent.payload` accepts arbitrarily large values while `topic` and `title` are capped ([`_models.py:19-20`](https://github.com/NVIDIA/xr-ai/blob/6734422bb70a728a6add0088951ec32aa9c653a7/agent-sdk/xr-ai-web-events/xr_ai_web_events/_models.py#L19-L20)), so `max_events` bounds nothing in bytes and a new client's `after=0` snapshot scales with payload size. Cap the serialized payload at validation, mirroring the existing field caps.

2. The SDK UI hardcodes the sample-specific `meter_reading` key in its headline heuristics ([`app.js:34`](https://github.com/NVIDIA/xr-ai/blob/6734422bb70a728a6add0088951ec32aa9c653a7/agent-sdk/xr-ai-web-events/xr_ai_web_events/static/app.js#L34)) — neither integrated sample uses it (both publish `text`). An optional `display` field on `WebEvent` would let producers choose the headline instead of the viewer guessing.

3. The two forwarding agents re-publish from inside voice-path subscriptions with no exception containment ([`app.py:44-69`](https://github.com/NVIDIA/xr-ai/blob/6734422bb70a728a6add0088951ec32aa9c653a7/agent-samples/simple-vlm-example/worker/simple_vlm_example_worker/app.py#L44-L69), [`__main__.py:50-81`](https://github.com/NVIDIA/xr-ai/blob/6734422bb70a728a6add0088951ec32aa9c653a7/agent-samples/xr-render-demo/worker/xr_render_demo_worker/__main__.py#L50-L81)): `runtime.publish` re-raises subscriber exceptions to the publisher, so a `RuntimeClosedError` from a forward during shutdown propagates into the query dispatch or the sample's mid-turn voice-output publish, which does not suppress it. Wrap the forwards in `suppress(RuntimeClosedError)` (or log-and-continue) so viewer plumbing can never surface into the voice path — these agents are the pattern other samples will copy.

No blockers. CI at this head was entirely queued at review time; conclusions here rest on the code and the carried-over verification, not on check results.

blongs-nv (approved):

PR #390, opened 2026-08-19
fix(vision): harden stream cleanup and warm VLM

wenxind-nvidia:
## Summary

- treat Relay image-sanitizer cleanup as complete only when its owning scope has already been removed
- warm the active simple-vlm model profile with a representative 1280x720 JPEG streaming request before readiness
- log image-stream request start and first-token latency for VLM-server correlation
- cover partial-stream teardown, unrelated cleanup errors, timing, and warmup request shape

## Validation

- `uv run pytest -q test_image_tools.py test_simple_vlm_example_worker.py test_config_documentation.py` (47 passed)
- full CPU suite (1,071 passed; 10 GPU tests and 2 local ignored-directory layout checks deselected)
- Ruff 0.15.16, SPDX header check, and `git diff --check`

### DIFF
--- a/agent-samples/simple-vlm-example/README.md
+++ b/agent-samples/simple-vlm-example/README.md
@@ -60,7 +60,9 @@ The worker and orchestrator consume the deployment profile selected by
 - `models.omni.json` reuses the Nemotron-Omni VLM service on port 8108.
 
 The same profile owns model behavior, endpoints, credentials, readiness, and
-launcher process ownership.
+launcher process ownership. Before announcing readiness, the worker completes a
+small streaming request with a 1280x720 JPEG so the first user query does not
+pay the multimodal initialization cost.
 
 ## Relay visibility
 
--- a/agent-samples/simple-vlm-example/worker/simple_vlm_example_worker/app.py
+++ b/agent-samples/simple-vlm-example/worker/simple_vlm_example_worker/app.py
@@ -5,6 +5,8 @@
 
 from __future__ import annotations
 
+import io
+import time
 from collections.abc import AsyncIterator
 from contextlib import asynccontextmanager
 from pathlib import Path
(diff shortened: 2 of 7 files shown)


blongs-nv:
Reviewed at `789284445c8fb5481f1a569763c3e97a3e11fb63` by a bot. The sanitizer-cleanup fix is correct and verified: with a parent scope popped while a stream is still open, closing the stream previously raised out of the deregistration and surfaced as a stream failure, and now completes cleanly (the new teardown test fails with the fix reverted, so it pins the behavior). The warmup is well placed: readiness probes are polled until they first succeed and then discarded, so the warmup runs to success exactly once, failures log and return False for the normal 5-second retry, and the request shape (one 1280x720 JPEG, `max_tokens=4`, 120 s bound) is within every shipped profile's image limits, including Omni's one-image default.

## Suggestions

1. The teardown tolerance matches the relay error by exact string ([`_vision.py:38-42`](https://github.com/NVIDIA/xr-ai/blob/789284445c8fb5481f1a569763c3e97a3e11fb63/agent-sdk/xr-ai-tools/xr_ai_tools/_vision.py#L38-L42)): `str(error) != f"not found: scope {handle.uuid} not found"`. If nemo_relay ever rewords that message, the suppression silently stops working and partial-stream teardown reverts to logging spurious "Image VLM stream failed" errors. Match on a stable prefix (`"not found: scope "`) or, if nemo_relay exposes one, a typed error, and note the coupling where the string is defined.

2. `_warm_vlm`'s failure branch is untested: the `except` path that logs and returns False ([`app.py:78-80`](https://github.com/NVIDIA/xr-ai/blob/789284445c8fb5481f1a569763c3e97a3e11fb63/agent-samples/simple-vlm-example/worker/simple_vlm_example_worker/app.py#L78-L80)) is what keeps a failed warmup from wedging readiness, but the worker test only exercises the success path. Add a case where `stream_images` raises and assert the probe returns False and is retried.

The first-token latency logging is proportionate (two INFO lines per image query, no event-loop cost), docs and quickstart describe the new readiness behavior, and the request-shape test pins the warmup contract. No blockers.

blongs-nv (approved):

nvddr (approved):
Reviewed current head 7ea6825. The follow-up addresses both non-blocking suggestions: missing-owner detection now matches the owning scope through a documented Relay error prefix while preserving unrelated failures, with coverage for both cases; the VLM warmup failure path now verifies that readiness retries and succeeds on the next attempt. The original cleanup, warmup, and latency changes have no blocking findings, and all current checks pass.

Approved as reviewed. GitHub currently reports a merge conflict with main, so the branch still needs to be rebased and revalidated before merge.

PR #393, opened 2026-08-19
Lift Android build's transitive Netty to 4.2.17

blongs-nv:
AGP's UTP test tooling pulls Netty 4.1.93.Final (via `grpc-netty`) onto the Android sample's build classpath; that line has known CVEs. Upgrading doesn't help: even AGP 9.3.0's UTP still resolves Netty 4.1.x.

Adds a resolution rule that lifts any `io.netty` 4.x below 4.2.17.Final up to it. It's a floor, not a pin: newer versions pass through, and non-4.x artifacts (`netty-tcnative`) are untouched. Build-time only; nothing Netty ships in the APK.

Verified: `buildEnvironment` and the UTP configurations resolve all Netty at 4.2.17.Final; a 4.2.16 request is lifted, 4.2.17+ and tcnative are left alone.

### DIFF
--- a/.github/workflows/android-build.yml
+++ b/.github/workflows/android-build.yml
@@ -46,3 +46,6 @@ jobs:
 
       - name: Build debug APK
         run: ./gradlew assembleDebug
+
+      - name: Verify Netty pin
+        run: ./gradlew verifyNettyPin
--- a/client-samples/android/README.md
+++ b/client-samples/android/README.md
@@ -138,4 +138,4 @@ The app requests the following Android permissions:
 
 **Build-time only:** `build.gradle.kts` lifts transitive `io.netty` (pulled in by AGP's
 UTP test tooling) past known CVEs in its 4.1 line; the version lives in the `netty` catalog
-entry. Re-check after an AGP bump: `gradle buildEnvironment | grep netty`.
+entry. Re-check after an AGP bump: `./gradlew verifyNettyPin` (also run by CI).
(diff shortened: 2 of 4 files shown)


yanziz-nvidia:
Should we bump to newer AGP instead?

blongs-nv:
> Should we bump to newer AGP instead?

It wouldn't fix this because UTP would still use an old Netty version.

yanziz-nvidia:
Reviewed by yanziz-review-bot

**Summary**  
This adds a Netty security-version floor across the Android build’s configurations and documents its maintenance path. All current CI checks pass, including `assembleDebug`; there are no merge blockers, but CI does not enforce the security invariant and the documented verification command bypasses the wrapper.

| | Finding |
|---|---|
| 🚫 | none |
| 💡 | [`client-samples/android/build.gradle.kts:26`](https://github.com/NVIDIA/xr-ai/blob/bdd917ca9cce480f11a002ae0055cd9cad8bfc04/client-samples/android/build.gradle.kts#L26) — `assembleDebug` does not assert the resolved Netty versions, so this security mitigation can silently stop applying after Gradle or AGP changes; add a verification task covering the buildscript and UTP configurations and invoke it from Android CI. |
| 🔍 | [`client-samples/android/README.md:141`](https://github.com/NVIDIA/xr-ai/blob/bdd917ca9cce480f11a002ae0055cd9cad8bfc04/client-samples/android/README.md#L141) — The maintenance command invokes a system `gradle`, so it may fail or inspect a different Gradle version; use the checked-in `./gradlew` wrapper. |

- `client-samples/android/build.gradle.kts:26` — Add a verification task that fails when resolved buildscript or UTP `io.netty` 4.x dependencies are below `nettyPin`, and run it in Android CI.
- `client-samples/android/README.md:141` — Replace `gradle buildEnvironment` with `./gradlew buildEnvironment`.

yanziz-nvidia (approved):

nvddr (approved):

PR #396, opened 2026-08-19
fix(launcher): harden GPU profile detection and diagnostics

yanziz-nvidia:
## Summary

Make GPU-profile selection fail safely and make model startup failures actionable.

- inventory each visible GPU independently, including compute capability and nullable total/free/used GPU memory
- remove the unsafe `dual_48G_ada` fallback when GPU inspection fails or the topology is unsupported
- select bundled profiles only when the complete topology is known to match, including rejecting undersized Blackwell GPUs
- preserve GB10/B10 Spark detection when `nvidia-smi` reports `[N/A]` for unified-memory fields
- add `model_servers --gpu-profile NAME` for explicitly reviewed, complete custom YAML profiles while keeping automatic detection strict
- add best-effort active-process context to unsupported-topology errors without making process inspection a startup dependency
- classify common vLLM startup failures, including insufficient GPU memory, while preserving original vLLM and NIM container logs
- pass the original vLLM arguments into diagnostics so configured `gpu_memory_utilization` appears in the failure report
- report GPU-detection failures as clean `model_servers` CLI errors without Python tracebacks
- keep the inventory records and query implementation private until a production caller requires a public contract

## Scope

This PR retains the existing named model-server profiles as a compatibility bridge. Measured per-service GPU-memory requirements, derived utilization, and capacity preflight are intentionally separate follow-up changes.

## Testing

- focused launcher, model-server, vLLM/NIM diagnostic/lifecycle, service-layout, and CLI-documentation suite: `156 passed`
- Ruff passes for every changed Python file
- regression coverage includes a 32 GiB RTX 5090, GB10 with unavailable memory telemetry, empty memory fields, best-effort process inspection, custom-profile validation, vLLM-only diagnostic arguments, and traceback-free CLI errors

Full non-GPU collection is currently blocked in this checkout by missing optional `stt_server` and `openxr_service` packages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,7 +189,10 @@ uv sync
 uv run model_servers
 ```
 
-GPU profiles are auto-detected (`dual_48G_ada` / `spark` / `96G_blackwell`).
+Known GPU profiles are auto-detected (`dual_48G_ada` / `spark` /
+`96G_blackwell`). Unsupported or ambiguous topologies stop with the complete
+inventory instead of assuming a profile; use `--gpu-profile NAME` only after
+copying and reviewing a custom YAML profile for that hardware.
 On first run the stack downloads tens of GB from Hugging Face and can take
 tens of minutes. On subsequent runs the containers restart in under a minute.
 
--- a/agent-samples/model-servers/main.py
+++ b/agent-samples/model-servers/main.py
@@ -45,6 +45,7 @@ import argparse
 from pathlib import Path
 
 from xr_ai_launcher import (
+    GPUInventoryError,
     Process,
     detect_gpu_config,
     load_deployment_profile,
@@ -55,6 +56,23 @@ from xr_ai_logging import setup_logging
 from xr_ai_vllm import stop_persistent_servers
 
 _BASE = Path(__file__).resolve().parent
+
+
+def _gpu_profile_names() -> tuple[str, ...]:
+    root = _BASE / "yaml"
+    if not root.is_dir():
+        return ()
+    return tuple(path.name for path in sorted(root.iterdir()) if path.is_dir())
+
+
+def _gpu_profile_name(value: str) -> str:
+    names = _gpu_profile_names()
+    if value not in names:
+        available = ", ".join(names) or "none"
+        raise argparse.ArgumentTypeError(
+            f"unknown GPU profile {value!r}; available profiles: {available}"
+        )
+    return value
 
 # service → (project, command, config basename, port). Order is launch
 # order: NIM containers precede local servers (speech NIMs allocate fixed
@@ -103,24 +121,33 @@ def _service_config(gpu_dir: str, config_base: str, profile_key: str) -> str:
     return f"{gpu_dir}/{config_base}.yaml"
 
 
-def _build_processes(selection: str) -> tuple[list[Process], tuple[str, ...]]:
+def _build_processes(
+    selection: str, gpu_profile: str | None = None,
+) -> tuple[list[Process], tuple[str, ...]]:
     profile_path = _profile_path(selection)
     deployment = load_deployment_profile(profile_path)
     unknown = deployment.services.keys() - _MODEL_SERVICES.keys()
     if unknown:
         raise ValueError(f"model profile declares unknown services: {sorted(unknown)}")
 
-    gpu_dir = f"yaml/{detect_gpu_config()}"
+    profile_name = gpu_profile or detect_gpu_config()
+    gpu_dir = f"yaml/{profile_name}"
     profile_key = profile_path.stem.removeprefix("models.")
-    processes = [
-        Process(
+    processes = []
+    for service, (project, command, config_base, port) in _MODEL_SERVICES.items():
+        if deployment.launch_mode(service) != "own":
+            continue
+        config = _service_config(gpu_dir, config_base, profile_key)
+        if not (_BASE / config).is_file():
+            raise ValueError(
+                f"GPU profile {profile_name!r} is incomplete: missing {config} "
+                f"for service {service!r}"
+            )
+        processes.append(Process(
             service, project, command,
-            config=_service_config(gpu_dir, config_base, profile_key),
+            config=config,
             launch_mode="persist", port=port,
-        )
-        for service, (project, command, config_base, port) in _MODEL_SERVICES.items()
-        if deployment.launch_mode(service) == "own"
-    ]
+        ))
     return processes, deployment.required_credentials
 
 
@@ -171,13 +198,26 @@ def run() -> None:
     p.add_argument("--allow-anonymous", action="store_true",
                    help="Start without HF_TOKEN (unauthenticated downloads "
                         "of the multi-GB checkpoints may stall indefinitely).")
+    p.add_argument(
+        "--gpu-profile", metavar="NAME", type=_gpu_profile_name,
+        help="Use a named YAML GPU profile instead of automatic detection. "
+             "Intended for explicitly reviewed custom hardware profiles.",
+    )
     ns, _ = p.parse_known_args()
 
     if ns.stop:
         _stop_models()
         return
 
-    processes, credentials = _build_processes(ns.models)
+    try:
+        processes, credentials = _build_processes(ns.models, ns.gpu_profile)
+    except GPUInventoryError as exc:
+        p.error(
+            f"{exc}\nUse --gpu-profile NAME to select an explicitly reviewed "
+            "custom YAML profile."
+        )
+    except ValueError as exc:
+        p.error(str(exc))
 
     # A missing HF_TOKEN silently stalls the multi-GB first-run download; see
     # docs/source/getting_started/credentials.md.
--- a/docs/source/getting_started/quickstart.md
+++ b/docs/source/getting_started/quickstart.md
@@ -46,9 +46,13 @@ uv sync
 uv run model_servers
 ```
 
-GPU profiles are auto-detected (`dual_48G_ada`, `spark`, `96G_blackwell`). These
-are presets for common configurations; to run on a different GPU, refer to
+GPU profiles are auto-detected (`dual_48G_ada`, `spark`, `96G_blackwell`). The
+profiles are presets for common configurations; to run on a different GPU, refer to
 {doc}`Running on other GPUs </getting_started/requirements>`.
+Detection inventories every visible device. If `nvidia-smi` fails, returns
+malformed data, or reports a topology without an existing model-server
+configuration, startup stops with the detected per-GPU capacity instead of
+assuming a fallback.
 On first run each model downloads from HuggingFace (tens of GB; can take
 tens of minutes). On subsequent runs the containers restart in under a minute.
 
--- a/docs/source/getting_started/requirements.md
+++ b/docs/source/getting_started/requirements.md
@@ -130,6 +130,18 @@ and adjust those knobs to your hardware:
 3. On lower-VRAM GPUs, run fewer models concurrently, or lower `max_model_len` on
    the LLM and VLM servers to reduce the KV-cache footprint.
 
+Then select the reviewed profile explicitly:
+
+```bash
+uv run --project agent-samples/model-servers model_servers \
+  --gpu-profile <profile-directory-name>
+```
+
+Automatic detection intentionally accepts only bundled profiles whose hardware
+requirements are known to match. `--gpu-profile` bypasses that selection for a
+profile you have explicitly copied and tuned; it does not validate that the model
+servers fit the selected devices.
+
 ```{note}
 The model weights are independent of the GPU. Any NVIDIA GPU with enough VRAM for
 the models you load will run the stack; the profiles only encode where each server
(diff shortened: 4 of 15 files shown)


nvddr (changes requested):
Thanks for tightening GPU detection and preserving the original container logs. I found three blocking correctness issues plus two API/inventory issues that should be addressed before merge.

1. **P1 — Small Blackwell GPUs are assigned the 96 GB profile.** `detect_gpu_config()` maps every single SM 10+ GPU below 120 GiB to `96G_blackwell`. I verified that a 32 GiB RTX 5090 selects that profile, recreating the unsafe selection this PR intends to remove. Require a suitable minimum/range (and, if necessary, a supported device identity) before selecting it.

2. **P1 — Unsupported/custom GPU profiles can no longer be launched.** The hard failure is appropriate for automatic detection, but `model-servers` always calls `detect_gpu_config()` and offers no GPU-profile override. That makes the documented workflow of copying and tuning a profile for other hardware impossible. Please add an explicit `--gpu-profile` or equivalent that bypasses auto-detection while keeping automatic selection strict.

3. **P1 — Configured utilization is never included in production diagnostics.** `run_container()` passes the full `docker run` argv, where `--gpu-memory-utilization 0.78` is embedded inside the final `bash -c` string rather than appearing as standalone list elements. The new parser therefore never finds it. Pass the original vLLM argv or explicit diagnostic context and add a test using `build_run_argv()` output.

4. **P2 — Failure to inspect processes is reported as “no processes.”** If the compute-process query fails, every `GPUDevice` receives `processes=()`, which is indistinguishable from a successful empty result. Either fail the promised exact inventory or represent process availability separately.

5. **P2 — The change unnecessarily broadens the public launcher API.** No production caller uses the new dataclasses or `query_gpu_inventory`; only the pre-existing `detect_gpu_config` path is consumed. Please keep these implementation details private unless this public API was explicitly requested and has a concrete external contract.

Validation: the 13 focused tests passed; the broader author-listed suite reached 110 passing tests with one unrelated failure caused by the conflicted primary checkout leaking through its editable environment. Ruff passed.

blongs-nv:
Reviewed at `abf544a53dcac0bf7b3c6a6b14f41dc5522c0280` by a bot. I verified all five findings from the review above as fixed at this head (the 32 GiB RTX 5090 now correctly raises, the diagnostics receive the original vLLM argv with tests, the override flag works, process-inventory failures fail loudly, and the inventory API is private). One item remains, plus two suggestions.

## Blockers

1. **A GPU reporting `[N/A]` for any memory field aborts the entire inventory, which can make the `spark` config unreachable on the hardware it exists for.**

   [`utils/xr-ai-launcher/xr_ai_launcher/_gpu.py:126-137`](https://github.com/NVIDIA/xr-ai/blob/abf544a53dcac0bf7b3c6a6b14f41dc5522c0280/utils/xr-ai-launcher/xr_ai_launcher/_gpu.py#L126-L137)
   With `nounits`, nvidia-smi prints `[N/A]` for unsupported fields, and `_parse_mib` raises on it, so the inventory fails with "unparseable nvidia-smi GPU row" — reproduced at this head with a `NVIDIA GB10` row carrying `[N/A]` memory fields. The pre-PR implementation and its removed test (`test_blackwell_no_mem_data_returns_spark`, input `N/A`) treated exactly this shape as a real Spark/unified-memory report and still selected `spark` by name. Tolerate `[N/A]` per field (store `None`) and fail only when a load-bearing value is actually needed, so the name-based `spark` match stays reachable.

## Suggestions

1. `GPUInventoryError` escapes `model_servers` as a raw Python traceback: `_build_processes` calls `detect_gpu_config()` ([`agent-samples/model-servers/main.py:127`](https://github.com/NVIDIA/xr-ai/blob/abf544a53dcac0bf7b3c6a6b14f41dc5522c0280/agent-samples/model-servers/main.py#L127)) with no handler in `run()`. The message text is actionable; catch it at the entry point and exit with it cleanly.

2. Nothing pins the `[N/A]` handling either way; whichever behavior blocker 1 resolves to, add a test with an `[N/A]` memory row so the strict parser's contract is explicit.

The blocker should be resolved before merge.

yanziz-nvidia:
Addressed all five review findings:

- Rejects undersized Blackwell GPUs; a 32 GiB RTX 5090 no longer selects 96G_blackwell.
- Adds model_servers --gpu-profile NAME for explicitly reviewed/custom YAML profiles.
- Passes the original vLLM arguments into failure diagnostics, preserving configured utilization.
- Treats compute-process inspection failures as inventory errors.
- Keeps inventory dataclasses and query functions private in PR1.

yanziz-nvidia:
> Reviewed at `abf544a53dcac0bf7b3c6a6b14f41dc5522c0280` by a bot. I verified all five findings from the review above as fixed at this head (the 32 GiB RTX 5090 now correctly raises, the diagnostics receive the original vLLM argv with tests, the override flag works, process-inventory failures fail loudly, and the inventory API is private). One item remains, plus two suggestions.
> 
> ## Blockers
> 1. **A GPU reporting `[N/A]` for any memory field aborts the entire inventory, which can make the `spark` config unreachable on the hardware it exists for.**
>    [`utils/xr-ai-launcher/xr_ai_launcher/_gpu.py:126-137`](https://github.com/NVIDIA/xr-ai/blob/abf544a53dcac0bf7b3c6a6b14f41dc5522c0280/utils/xr-ai-launcher/xr_ai_launcher/_gpu.py#L126-L137)
>    With `nounits`, nvidia-smi prints `[N/A]` for unsupported fields, and `_parse_mib` raises on it, so the inventory fails with "unparseable nvidia-smi GPU row" — reproduced at this head with a `NVIDIA GB10` row carrying `[N/A]` memory fields. The pre-PR implementation and its removed test (`test_blackwell_no_mem_data_returns_spark`, input `N/A`) treated exactly this shape as a real Spark/unified-memory report and still selected `spark` by name. Tolerate `[N/A]` per field (store `None`) and fail only when a load-bearing value is actually needed, so the name-based `spark` match stays reachable.
> 
> ## Suggestions
> 1. `GPUInventoryError` escapes `model_servers` as a raw Python traceback: `_build_processes` calls `detect_gpu_config()` ([`agent-samples/model-servers/main.py:127`](https://github.com/NVIDIA/xr-ai/blob/abf544a53dcac0bf7b3c6a6b14f41dc5522c0280/agent-samples/model-servers/main.py#L127)) with no handler in `run()`. The message text is actionable; catch it at the entry point and exit with it cleanly.
> 2. Nothing pins the `[N/A]` handling either way; whichever behavior blocker 1 resolves to, add a test with an `[N/A]` memory row so the strict parser's contract is explicit.
> 
> The blocker should be resolved before merge.

Addressed this as well:

- [N/A] memory fields are stored as None.
- GB10/B10 still selects spark by identity.
- Capacity-based profiles still require numeric total memory.
- Unsupported hardware with missing capacity gets a clear matching error.
- model_servers reports inventory failures as clean CLI errors without tracebacks.
- Added regression tests for both behaviors.

nvddr (approved):

blongs-nv:
Reviewed at `f2f78f776579eba4128730e5bee63cd1a618b063` by a bot. The blocker and both suggestions from my previous review are resolved at this head: `[N/A]` memory fields parse to `None`, a GB10 with no memory data selects `spark` again, and `model_servers` reports `GPUInventoryError` as a clean CLI error, each with a regression test. Two new items surfaced in the amended code, plus suggestions.

## Blockers

1. **The compute-process inventory is a fatal dependency whose data no production code reads.**

   [`utils/xr-ai-launcher/xr_ai_launcher/_gpu.py:123`](https://github.com/NVIDIA/xr-ai/blob/f2f78f776579eba4128730e5bee63cd1a618b063/utils/xr-ai-launcher/xr_ai_launcher/_gpu.py#L123)
   `_query_gpu_inventory` aborts on any `_query_compute_processes` failure (non-zero exit, unparseable row, non-integer pid), yet `_GPUDevice.processes` is read only by tests; neither `detect_gpu_config` nor its topology message uses it. A host where `--query-compute-apps` is restricted (vGPU, some container setups) now fails startup over data nothing consumes, against the repo rule of not adding an abstraction until two concrete use cases need it. Either surface the process list in the "no existing configuration matches" message, where it would actually help, or drop the query (or degrade its failure to a warning).

2. **An empty memory field in a GPU row escapes as a raw `IndexError` traceback.**

   [`utils/xr-ai-launcher/xr_ai_launcher/_gpu.py:49-52`](https://github.com/NVIDIA/xr-ai/blob/f2f78f776579eba4128730e5bee63cd1a618b063/utils/xr-ai-launcher/xr_ai_launcher/_gpu.py#L49-L52)
   `_parse_mib` does `value.strip().split()[0]`, which raises `IndexError` on an empty field and bypasses the `except ValueError` guard at [line 141](https://github.com/NVIDIA/xr-ai/blob/f2f78f776579eba4128730e5bee63cd1a618b063/utils/xr-ai-launcher/xr_ai_launcher/_gpu.py#L141). I reproduced with the row `0, GPU-0, 00000000:01:00.0, NVIDIA L40S, 8.9, , , `: every other malformed row raises `GPUInventoryError` and reaches `p.error`, while this one prints a traceback. Catch `(ValueError, IndexError)` or guard the empty case.

## Suggestions

1. NIM containers now get vLLM-worded diagnoses computed from the wrong argv: [`_nim.py:161`](https://github.com/NVIDIA/xr-ai/blob/f2f78f776579eba4128730e5bee63cd1a618b063/utils/xr-ai-vllm/xr_ai_vllm/_nim.py#L161) calls `run_container` without `diagnostic_argv`, so the [`diagnostic_argv or argv` fallback](https://github.com/NVIDIA/xr-ai/blob/f2f78f776579eba4128730e5bee63cd1a618b063/utils/xr-ai-vllm/xr_ai_vllm/_docker.py#L817-L818) scans a docker argv that never contains `--gpu-memory-utilization`, and a NIM OOM is told to "Reduce the configured model/context/concurrency" for knobs it does not have. Pass NIM-appropriate context or gate the classifier to the vLLM caller.

2. The terminal error at [`_gpu.py:192-196`](https://github.com/NVIDIA/xr-ai/blob/f2f78f776579eba4128730e5bee63cd1a618b063/utils/xr-ai-launcher/xr_ai_launcher/_gpu.py#L192-L196) never mentions `--gpu-profile`, and the topologies most likely to hit it are supersets that launched before this PR (a 4x L40S box previously selected `dual_48G_ada`, which only uses GPUs 0 and 1). Append the escape hatch at the [`p.error` site](https://github.com/NVIDIA/xr-ai/blob/f2f78f776579eba4128730e5bee63cd1a618b063/agent-samples/model-servers/main.py#L200-L203), or match on the devices a config actually consumes.

3. [`_parse_mib`](https://github.com/NVIDIA/xr-ai/blob/f2f78f776579eba4128730e5bee63cd1a618b063/utils/xr-ai-launcher/xr_ai_launcher/_gpu.py#L50) tolerates only `N/A`/`[N/A]`, but nvidia-smi also emits `Not Supported`/`[Not Supported]` on restricted or virtualized devices, and such a row aborts the inventory as "unparseable". Treat the `Not Supported` sentinels the same way.

4. `--gpu-profile` validates the directory name but never the directory's contents, so a half-populated custom profile launches processes and dies later on a missing config path; verify the resolved `gpu_dir`'s YAMLs exist in [`_build_processes`](https://github.com/NVIDIA/xr-ai/blob/f2f78f776579eba4128730e5bee63cd1a618b063/agent-samples/model-servers/main.py#L128). Also compute [`_GPU_PROFILE_NAMES`](https://github.com/NVIDIA/xr-ai/blob/f2f78f776579eba4128730e5bee63cd1a618b063/agent-samples/model-servers/main.py#L59-L61) lazily so a missing `yaml/` directory breaks the flag, not `import`, `--help`, and `--stop`.

5. Four branches could be deleted with the suite still green: the spark-by-capacity clause ([`_gpu.py:160-163`](https://github.com/NVIDIA/xr-ai/blob/f2f78f776579eba4128730e5bee63cd1a618b063/utils/xr-ai-launcher/xr_ai_launcher/_gpu.py#L160-L163), both spark tests use GB10/B10 names), the third diagnostics class ([`_diagnostics.py:56`](https://github.com/NVIDIA/xr-ai/blob/f2f78f776579eba4128730e5bee63cd1a618b063/utils/xr-ai-vllm/xr_ai_vllm/_diagnostics.py#L56)), the `diagnostic_argv or argv` fallback (no test runs the real classifier against real `build_run_argv` output), and invalid `--gpu-profile` rejection ([`main.py:64`](https://github.com/NVIDIA/xr-ai/blob/f2f78f776579eba4128730e5bee63cd1a618b063/agent-samples/model-servers/main.py#L64)). Add a case for each.

6. [`README.md:192`](https://github.com/NVIDIA/xr-ai/blob/f2f78f776579eba4128730e5bee63cd1a618b063/README.md#L192) still describes auto-detection as always succeeding while the quickstart was updated, and the [`requirements.md` snippet](https://github.com/NVIDIA/xr-ai/blob/f2f78f776579eba4128730e5bee63cd1a618b063/docs/source/getting_started/requirements.md#L134-L137) `uv run model_servers --gpu-profile ...` lacks `--project`/`cd`, so it fails from the repo root.

The strict inventory, the clean CLI failure path, and the diagnostic argv wiring all hold up well at this head. The two blockers above should be resolved before merge.

blongs-nv (approved):

PR #397, opened 2026-08-19
refactor(model-servers): read ports from YAML

yanziz-nvidia:
## Summary

- remove duplicated service-port values from the `model_servers` Python catalog
- read `port` or `http_port` from the exact resolved YAML passed to each launched service
- retain that resolved value as `Process.port` metadata for selected-port protection and cleanup
- discover current checked-in cleanup targets from the model-server and base-service YAML files
- fail clearly when a selected config is missing or does not declare a valid HTTP port

## Why

Before this change, a service could bind to the port in its selected YAML while `model_servers` used an independent Python constant for cleanup and selected-port protection. A profile-specific YAML edit could therefore make lifecycle operations inspect or stop the wrong port. Deriving the metadata from the configuration actually passed to the service removes that drift path.

The parser remains local to `model_servers`; this PR does not broaden the public `xr_ai_launcher` API.

## Scope

This PR does not make YAML a runtime ownership database. Cleanup still discovers the current checkout's known configurations, as it did conceptually with the previous current-checkout Python catalog. Durable discovery of persisted services across configuration or checkout changes requires runtime ownership records for pip/direct processes and label/name enumeration for containers; that is a separate lifecycle change.

GPU-memory planning, request-driven service resolution, configuration overlays, and generated local endpoints are also outside this focused change.

## Testing

- `uv tool run ruff==0.15.16 check agent-samples/model-servers/main.py tests/test_model_servers.py`
- combined affected suite: `221 passed`

Rebased onto current `main`. Focused commits: `ec49f40c`, `4e665361`, `0859663a`.

### DIFF
--- a/agent-samples/model-servers/main.py
+++ b/agent-samples/model-servers/main.py
@@ -49,7 +49,7 @@ from xr_ai_launcher import (
     Process,
     detect_gpu_config,
     load_deployment_profile,
-    read_service_port,
+    read_config_scalar,
     require_credentials,
     run_stack,
 )
@@ -112,6 +112,20 @@ def _profile_path(selection: str) -> Path:
     return _BASE / "yaml" / f"models.{selection}.json"
 
 
+def _read_service_port(path: Path) -> int | None:
+    """Read and validate the HTTP port used by one model-server config."""
+    raw = read_config_scalar(path, "port") or read_config_scalar(path, "http_port")
+    if not raw:
+        return None
+    try:
+        port = int(raw)
+    except ValueError as exc:
+        raise ValueError(f"{path}: port must be an integer, got {raw!r}") from exc
+    if not 1 <= port <= 65535:
+        raise ValueError(f"{path}: port must be between 1 and 65535, got {port}")
+    return port
+
+
 def _build_processes(
     selection: str, gpu_profile: str | None = None,
 ) -> tuple[list[Process], tuple[str, ...]]:
@@ -134,7 +148,7 @@ def _build_processes(
                 f"GPU profile {profile_name!r} is incomplete: missing {config} "
                 f"for service {service!r}"
             )
-        port = read_service_port(config)
+        port = _read_service_port(config)
         if port is None:
             raise ValueError(f"{config}: service config must declare port or http_port")
         processes.append(Process(
@@ -151,7 +165,7 @@ def _known_service_ports() -> list[tuple[str, int]]:
         configs = list((_BASE / "yaml").glob(f"*/{config_base}*.yaml"))
         configs.append((_BASE / project / f"{config_base}.yaml").resolve())
         for config in configs:
-            if config.is_file() and (port := read_service_port(config)) is not None:
+            if config.is_file() and (port := _read_service_port(config)) is not None:
                 targets.add((service, port))
     return sorted(targets)
 
--- a/agent-samples/model-servers/main.py
+++ b/agent-samples/model-servers/main.py
@@ -49,6 +49,7 @@ from xr_ai_launcher import (
     Process,
     detect_gpu_config,
     load_deployment_profile,
+    read_service_port,
     require_credentials,
     run_stack,
 )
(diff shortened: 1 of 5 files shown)


wenxind-nvidia:
**Blocker**

`agent-samples/model-servers/main.py:148`: Cleanup discovers ports only from the current YAML files, but model servers persist across repository updates. If a release changes a port across all YAML variants, `model_servers --stop` loses the old port. Docker services are eventually replaced by name on the next launch, but the detached STT server has no equivalent discovery and remains loaded while another STT instance starts on the new port. Cleanup needs ownership-based discovery independent of current configuration, such as Docker label enumeration plus persisted PID/port state for local services.

blongs-nv:
Reviewed at `f293e515719b104e223ebb773637b9625ced1d0e` by a bot.

## Suggestions

1. The two new failure paths in `_build_processes` have no tests: a selected service whose GPU config file is missing ([`main.py:134-138`](https://github.com/NVIDIA/xr-ai/blob/f293e515719b104e223ebb773637b9625ced1d0e/agent-samples/model-servers/main.py#L134-L138)) and a config that declares no `port`/`http_port` ([`main.py:140-141`](https://github.com/NVIDIA/xr-ai/blob/f293e515719b104e223ebb773637b9625ced1d0e/agent-samples/model-servers/main.py#L140-L141)). A tmp-profile test for each would pin the fail-clearly messages this PR adds.

Reading each port from the exact YAML the process launches with removes a real drift risk (the old constants could disagree with a selected per-profile variant), and the validation and test updates are clean. With the cleanup-discovery blocker already on the thread resolved, ready to merge.

blongs-nv (approved):

nvddr (changes requested):
What is the end goal design the PR description is referring to?

yanziz-nvidia:
> What is the end goal design the PR description is referring to?

discussed offline. We would like a better design about how ports are defined and used. But let's reduce the confusion layer with this refactoring PR.

nvddr (commented):
Reviewed at `53b1d1c`. Text-only review comment; I am not changing the existing review/approval state.

The direction is good: reading `Process.port` from the exact YAML selected for launch removes a real source of drift, and the port is still present on every persistent `Process`, so the existing selected-port protection remains intact.

I checked the other service-owning samples. There is no new immediate collision in the checked-in configurations:

- `simple-vlm-example` is the only overlapping sample that directly launches the shared VLM/STT services; its standalone configs still use VLM `8100` and STT `8103`, and on supported GPUs it already selects the same model-server VLM YAML.
- `lab-instrument-monitoring` and the local `xr-render-demo` profiles reuse the model-server VLM/STT/Omni services rather than launching competing instances.
- The sample-owned TTS service uses `8105` and is outside the model-server cleanup catalog.

However, I think Wenxin's earlier cleanup concern remains valid. `_known_service_ports()` discovers cleanup targets only from the **current checkout's** model-server YAML and base service YAML (`main.py:147-156`). Persisted services deliberately outlive both the launcher and repository updates, and `simple-vlm-example` can launch from a sample-local fallback YAML that is not scanned. If all checked-in variants move a service from one port to another, or that fallback config diverges, the old pip/direct service is no longer discoverable by `--stop` or `_stop_unselected_services()`. It can remain resident on the GPU while the new instance starts on the new port. Docker is eventually recoverable by stable container name, but pip vLLM and the direct STT process do not have equivalent name-based recovery.

YAML should own the **desired launch port**; it should not be the ownership database for **what is currently running**. Cleanup should discover actual owned runtime instances: enumerate the Docker ownership labels/stable names, and persist or otherwise enumerate service/port/PID identity for pip/direct processes. Then filter those discovered instances against the selected process ports. A cross-sample regression covering the simple-VLM fallback plus a simulated old-port instance would pin this invariant. The existing global behavior where `model_servers --stop` can stop a currently active sample's shared model service is not introduced here, but it should remain explicit because these are host-global singleton resources.

One smaller API issue: `read_service_port` is exported from `xr_ai_launcher.__all__`, but this PR has only one concrete consumer. It does not need to broaden the public launcher API; keep it private/local (using the existing scalar reader) until a second consumer establishes the shared contract.

Validation: `tests/test_launcher_config.py` and `tests/test_model_servers.py` pass locally (`91 passed`), and the PR checks are green.

nvddr (changes requested):
Please address comments about public api surface expansion

yanziz-nvidia:
removed read_service_port from the public xr_ai_launcher API; the parser is now local to model_servers.

The lifecycle concern should be out of this PR. We focus on avoiding drift between the exact YAML launched and the Process.port metadata used for selected-port protection and cleanup.

nvddr (approved):

nvddr (changes requested):
Fix DCO

nvddr (approved):

PR #398, opened 2026-08-19
Fix voice text delivery during playback pacing

nvddr:
## Summary

- finalize raw streamed voice text as soon as its final contribution arrives
- finalize finite and LLM-rewritten speech as soon as the text is available
- keep the playback-duration reservation only in the aggregation scheduler, so later routine speech still waits and coalesces
- document that client text delivery is independent from open-loop audio pacing

## Root cause

`VoiceAggregationAgent` kept downstream `voice.output` responses open for its estimated playback duration. `VoiceAgent` emits the configured `agent.response` data message only when that downstream response closes, so an overestimated playback window delayed the Agent panel even after TTS audio had finished.

## Impact

The Agent panel now receives completed raw or rewritten text immediately. Audio overlap prevention, urgent interruption, participant isolation, and routine-output coalescing continue to use the existing playback reservation.

## Validation

- `36 passed`: `tests/test_voice_aggregation.py`
- `3 passed`: focused StreamingTTS data-echo and trailing-text tests
- Ruff 0.15.16 check and format check
- strict Sphinx build and generated public API reference check
- `git diff --check`

### DIFF
--- a/agent-sdk/xr-ai-voice/README.md
+++ b/agent-sdk/xr-ai-voice/README.md
@@ -105,7 +105,10 @@ next batch instead of becoming separate queued utterances. This is open-loop
 pacing rather than a client playback acknowledgement. An estimate that is too
 short moves queuing downstream into TTS; one that is too long creates silence
 between utterances, but neither delays the completed-response data echo. Tune
-all three playback settings for the selected TTS voice.
+all three playback settings for the selected TTS voice. The echo represents the
+complete intended utterance, not a playback acknowledgement: a later urgent
+barge-in can interrupt its audio after the full text has already reached the
+client.
 
 Rewrites have an aggregator-enforced `rewrite_timeout_s` deadline, which is also
 passed to the model service. If a rewrite fails or times out, the original text
--- a/agent-sdk/xr-ai-voice/README.md
+++ b/agent-sdk/xr-ai-voice/README.md
@@ -95,14 +95,17 @@ rather than subscribing to one hard-coded topic name.
 
 Aggregation is participant-scoped. A lone finite contribution passes through
 without an LLM call after the short coalescing window. Two or more finite
-contributions in one batch are rewritten into a single concise utterance. The
-aggregator keeps each output open for an estimated spoken duration, based on
-`speech_rate_wpm` and clamped between `minimum_playback_s` and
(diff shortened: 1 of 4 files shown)


wenxind-nvidia:
Reviewed current head `0e3caa8`.

## Blockers

None.

## Suggestion

- `agent-sdk/xr-ai-voice/xr_ai_voice/_aggregation.py:512`: remove the optional `final` override from `_publish_stream`. Every call now uses finality from the contribution, and the stream-final call passes `contribution.output.final` explicitly. Keeping one source of finality would make this path simpler and prevent accidental divergence.

The implementation correctly finalizes completed text before the playback reservation while retaining participant-scoped scheduler pacing, coalescing, and urgent interruption behavior. The documentation matches the implementation.

Verification: 39 focused aggregation and StreamingTTS tests passed. A broader run passed 149 tests; one hub integration test could not collect under `--noconftest` because its fixtures live in the root conftest. Ruff, formatting, and `git diff --check` passed.

wenxind-nvidia (approved):

yanziz-nvidia (approved):

blongs-nv:
Reviewed at `ab934fd465d4d79c38ed8165a5eadd7623fefe04` by a bot.

Verified the behaviors that matter at this head: a completed batch finalizes downstream immediately while updates arriving during the reservation still coalesce into one follow-up; an urgent contribution still publishes at the moment of arrival with `interrupt=True`; a stream's last chunk closes its response in place, with the explicit terminator retained on the idle-timeout and discarded-stream paths that still need it. The runtime's `voice.output` consumer handles the one-shot `final=True` publications, and the `ab934fd` follow-up commit is behavior-neutral. Aggregation suite passes.

## Suggestion

1. Barge-in now creates a divergence the docs don't mention: the superseded response's full text was already echoed as completed, but its audio is cut mid-playback, so the transcript can show sentences the user never heard. The README's data-echo sentence should say so, so client authors don't treat echoed text as "was spoken". ([`README.md:95-108`](https://github.com/NVIDIA/xr-ai/blob/ab934fd465d4d79c38ed8165a5eadd7623fefe04/agent-sdk/xr-ai-voice/README.md#L95-L108))

No blockers; ready to merge. CI was still queued at review time and should be green before merge.

blongs-nv (approved):

PR #400, opened 2026-08-19
Add cross-platform client network telemetry

nvddr:
## Summary

- add a transport-neutral NetworkMetrics snapshot to StreamKit on web, Android, iOS/visionOS, and native C++
- report LiveKit connection quality, selected ICE candidate RTT, and inbound RTP jitter about once per second
- display the same Network panel in both web clients, Android, and iOS/visionOS; print the snapshot in the native sample
- use only existing LiveKit participant and RTCStats APIs; no hub messages, server changes, or telemetry service

## Scope

This intentionally starts with three high-value metrics. Frame delay, stutter detection, received FPS, resolution, and historical aggregation remain out of scope until there is evidence that LiveKit-native values are insufficient.

RTT and jitter remain unavailable until LiveKit has a matching RTCStats record.

## Validation

- node syntax checks and NetworkMetrics smoke assertion
- native CMake build and all 6 CTest tests
- docs public API check
- git diff check

Android compilation could not run because this host has no Android SDK. Swift tests could not run because this host has no Swift toolchain.


wenxind-nvidia:
## Blockers

1. **Native metric callbacks can terminate the process when they disconnect.** The telemetry `jthread` invokes `on_network_metrics` directly, while `Disconnect()` calls `StopNetworkMetricsReporting()` and joins that same thread. A normal callback such as `if (quality == lost) session.Disconnect()` therefore self-joins, throws `std::system_error`, and escapes the thread entry point into `std::terminate`. See [`StopNetworkMetricsReporting()`](https://github.com/NVIDIA/xr-ai/blob/5aa1744e3238c39b28d6942d979c31d0a355386a/client-samples/native/StreamKit/src/Backends/LiveKit/LiveKitBackend.cpp#L568-L572) and the [callback invocation](https://github.com/NVIDIA/xr-ai/blob/5aa1744e3238c39b28d6942d979c31d0a355386a/client-samples/native/StreamKit/src/Backends/LiveKit/LiveKitBackend.cpp#L611-L614). The reporting lifecycle must support callback-initiated disconnect, with a regression test.

2. **Native telemetry does not target a released C++ SDK compatible with the rest of this backend.** [`Room::getStats()`](https://github.com/NVIDIA/xr-ai/blob/5aa1744e3238c39b28d6942d979c31d0a355386a/client-samples/native/StreamKit/src/Backends/LiveKit/LiveKitBackend.cpp#L582) is absent from [LiveKit C++ v0.4.1](https://github.com/livekit/client-sdk-cpp/blob/v0.4.1/include/livekit/room.h), whose legacy `Connect()` and raw `localParticipant()` APIs match this code. It is present in v1.0+, but those releases removed the legacy APIs still used elsewhere in this file. The only upstream revision that compiled was the untagged intermediate commit [069b78d](https://github.com/livekit/client-sdk-cpp/commit/069b78d). Pin and document that revision, or migrate the complete backend to a released SDK and compile the real `STREAMKIT_HAVE_LIVEKIT=1` path in CI. Stub-mode tests do not validate this code.

3. **An unexpected web disconnect permanently leaks the polling interval and backend.** The terminal [`RoomEvent.ConnectionStateChanged("disconnected")` handler](https://github.com/NVIDIA/xr-ai/blob/5aa1744e3238c39b28d6942d979c31d0a355386a/client-samples/web/StreamKit/Backends/LiveKit/LiveKitBackend.js#L212-L214) only forwards state. The interval is cleared only by explicit `#tearDown()`. After a network/server disconnect, the app replaces the session on the next connect, but the old interval continues every second and retains the old backend, room, and track references. Stop reporting when the room reaches terminal `disconnected`.

4. **Web and Android can report RTT from an inactive ICE pair.** Both accept any candidate pair whose state is `succeeded`: [web](https://github.com/NVIDIA/xr-ai/blob/5aa1744e3238c39b28d6942d979c31d0a355386a/client-samples/web/StreamKit/Backends/LiveKit/LiveKitBackend.js#L530-L533), [Android](https://github.com/NVIDIA/xr-ai/blob/5aa1744e3238c39b28d6942d979c31d0a355386a/client-samples/android/app/src/main/java/com/nvidia/xrai/streamkitsample/streamkit/backends/livekit/LiveKitBackend.kt#L354-L359). Multiple pairs may succeed; only the nominated or transport-selected pair carries traffic. Resolve `transport.selectedCandidatePairId`, with `nominated`/`selected` as compatibility fallbacks, rather than using `state == "succeeded"`.

## Suggestions

1. Compare native quality events against the room's actual local participant rather than [`SessionConfig.identity`](https://github.com/NVIDIA/xr-ai/blob/5aa1744e3238c39b28d6942d979c31d0a355386a/client-samples/native/StreamKit/src/Backends/LiveKit/LiveKitBackend.cpp#L152-L157). With an inline JWT, the requested identity is not used during connection and may differ from the token identity, leaving quality permanently `unknown`.

2. Add focused tests for active-pair selection, terminal-disconnect timer cleanup, and callback-initiated native disconnect. Current additions mostly test DTO storage and callback forwarding.

## Verification

Reviewed head `5aa1744`. The iOS StreamKit and complete sample app build successfully against LiveKit Swift 2.13.0. Web camera tests pass, the changed native session test compiles and passes in stub mode, and `git diff --check` passes. Android could not be built because this machine has no usable JDK/Android SDK; its API usage was checked against LiveKit Android 2.7.0 source.

blongs-nv:
Reviewed at `5aa1744e3238c39b28d6942d979c31d0a355386a` by a bot.

## Blockers

1. **The web metrics interval is never stopped on a LiveKit-initiated disconnect and can be double-started.**

   [`client-samples/web/StreamKit/Backends/LiveKit/LiveKitBackend.js:502-505`](https://github.com/NVIDIA/xr-ai/blob/5aa1744e3238c39b28d6942d979c31d0a355386a/client-samples/web/StreamKit/Backends/LiveKit/LiveKitBackend.js#L502-L505) The timer is only cleared in `#tearDown()`, but a server kick or network drop only fires [`ConnectionStateChanged`](https://github.com/NVIDIA/xr-ai/blob/5aa1744e3238c39b28d6942d979c31d0a355386a/client-samples/web/StreamKit/Backends/LiveKit/LiveKitBackend.js#L212-L214), and the next Connect builds a fresh session, orphaning the old backend with its 1 Hz interval running for the page's lifetime. `#startNetworkMetricsReporting()` also never clears an existing timer, and a second Connect click during the token fetch creates a second backend with the first left running. The `t.after(() => backend.disconnect())` added in [`tests/javascript/web_camera_capture.test.mjs:199`](https://github.com/NVIDIA/xr-ai/blob/5aa1744e3238c39b28d6942d979c31d0a355386a/tests/javascript/web_camera_capture.test.mjs#L199) is this leak keeping Node's event loop alive.

2. **The Android metrics loop blocks the main thread twice per second.**

   [`.../streamkit/backends/livekit/LiveKitBackend.kt:171-176`](https://github.com/NVIDIA/xr-ai/blob/5aa1744e3238c39b28d6942d979c31d0a355386a/client-samples/android/app/src/main/java/com/nvidia/xrai/streamkitsample/streamkit/backends/livekit/LiveKitBackend.kt#L171-L176) The loop runs on `Dispatchers.Main` and `Room.getPublisherRTCStats`/`getSubscriberRTCStats` in livekit-android 2.7.0 are `runBlocking` bridges onto the single RTC thread, so the UI thread parks until the RTC executor drains its queue. During renegotiation, ICE restart, or teardown that is easily hundreds of milliseconds, which is jank at best and an ANR at worst; collect on a background dispatcher and hop to Main only for the callback.

3. **The native client's quality is pinned to `unknown` unless the JWT identity happens to equal `SessionConfig::identity`.**

   [`client-samples/native/StreamKit/src/Backends/LiveKit/LiveKitBackend.cpp:152-158`](https://github.com/NVIDIA/xr-ai/blob/5aa1744e3238c39b28d6942d979c31d0a355386a/client-samples/native/StreamKit/src/Backends/LiveKit/LiveKitBackend.cpp#L152-L158) The filter compares the event participant against `session_config_.identity`, but that string is never sent to the server: `Connect()` passes only the URL and an inline token, so the server-side identity comes from the JWT while the sample defaults to [`--identity native-client`](https://github.com/NVIDIA/xr-ai/blob/5aa1744e3238c39b28d6942d979c31d0a355386a/client-samples/native/App/main.cpp#L64). Read the room's local participant quality directly, as the other three platforms do.

4. **`Disconnect()` called from inside `on_network_metrics` deadlocks the native backend.**

   [`client-samples/native/StreamKit/src/Backends/LiveKit/LiveKitBackend.cpp:568-572`](https://github.com/NVIDIA/xr-ai/blob/5aa1744e3238c39b28d6942d979c31d0a355386a/client-samples/native/StreamKit/src/Backends/LiveKit/LiveKitBackend.cpp#L568-L572) The callback is invoked synchronously on the metrics jthread ([`:614`](https://github.com/NVIDIA/xr-ai/blob/5aa1744e3238c39b28d6942d979c31d0a355386a/client-samples/native/StreamKit/src/Backends/LiveKit/LiveKitBackend.cpp#L614)), and `Disconnect()` → `TearDown()` → `StopNetworkMetricsReporting()` joins that same thread, so a handler that disconnects on `quality == kLost` throws `std::system_error` (`resource_deadlock_would_occur`). This is also the only platform that delivers the callback off the main thread, and neither `StreamingBackend.h` nor `StreamSession.h` states a threading contract.

5. **`roundTripTimeMs` is selected by a different candidate-pair rule on each platform.**

   Web and Android accept any pair that is nominated or `succeeded` ([`LiveKitBackend.js:529-534`](https://github.com/NVIDIA/xr-ai/blob/5aa1744e3238c39b28d6942d979c31d0a355386a/client-samples/web/StreamKit/Backends/LiveKit/LiveKitBackend.js#L529-L534), [`LiveKitBackend.kt:353-359`](https://github.com/NVIDIA/xr-ai/blob/5aa1744e3238c39b28d6942d979c31d0a355386a/client-samples/android/app/src/main/java/com/nvidia/xrai/streamkitsample/streamkit/backends/livekit/LiveKitBackend.kt#L353-L359)), while iOS and C++ require `nominated` only ([`LiveKitBackend.swift:415-419`](https://github.com/NVIDIA/xr-ai/blob/5aa1744e3238c39b28d6942d979c31d0a355386a/client-samples/ios-visionos/StreamKit/Sources/StreamKit/Backends/LiveKit/LiveKitBackend.swift#L415-L419), [`LiveKitBackend.cpp:594-598`](https://github.com/NVIDIA/xr-ai/blob/5aa1744e3238c39b28d6942d979c31d0a355386a/client-samples/native/StreamKit/src/Backends/LiveKit/LiveKitBackend.cpp#L594-L598)), and none follows the transport's selected pair that [`docs/source/getting_started/clients.md:38`](https://github.com/NVIDIA/xr-ai/blob/5aa1744e3238c39b28d6942d979c31d0a355386a/docs/source/getting_started/clients.md#L38) promises. After an ICE restart, several succeeded-but-unselected pairs typically exist, so web/Android can report a stale pair's RTT while iOS/C++ report a different value or none, contradicting the parity claim in [`STREAMKIT_PARITY.md:20`](https://github.com/NVIDIA/xr-ai/blob/5aa1744e3238c39b28d6942d979c31d0a355386a/client-samples/STREAMKIT_PARITY.md#L20). Write one selection rule and implement it identically on all four.

6. **Web computes metrics from a different stats source than the other platforms.**

   [`client-samples/web/StreamKit/Backends/LiveKit/LiveKitBackend.js:515-520`](https://github.com/NVIDIA/xr-ai/blob/5aa1744e3238c39b28d6942d979c31d0a355386a/client-samples/web/StreamKit/Backends/LiveKit/LiveKitBackend.js#L515-L520) Stats come only from the local mic/camera tracks and remote audio tracks, so remote video jitter is invisible on web while Android/iOS/C++ include it, and in the post-connect idle state (mic off, camera off, no remote audio) the track list is empty and RTT/jitter stay "—" on web while the other platforms report values. Use connection-level stats (or include remote video tracks) so the three fields have one definition.

7. **The 1 Hz metrics callback rebuilds the received-messages list every second.**

   [`client-samples/web/App/core.js:410-413`](https://github.com/NVIDIA/xr-ai/blob/5aa1744e3238c39b28d6942d979c31d0a355386a/client-samples/web/App/core.js#L410-L413) `onNetworkMetrics` calls `render()`, and `renderBase()` reassigns `list.innerHTML` ([`:316`](https://github.com/NVIDIA/xr-ai/blob/5aa1744e3238c39b28d6942d979c31d0a355386a/client-samples/web/App/core.js#L316)), so every message node is destroyed and recreated once per second for the whole session, wiping text selection and scroll position in the messages panel. Update the three metric spans directly, or skip the list rebuild when `receivedMessages` is unchanged.

8. **A disconnected Android client keeps displaying stale metrics.**

   [`.../streamkit/backends/livekit/LiveKitBackend.kt:171-176`](https://github.com/NVIDIA/xr-ai/blob/5aa1744e3238c39b28d6942d979c31d0a355386a/client-samples/android/app/src/main/java/com/nvidia/xrai/streamkitsample/streamkit/backends/livekit/LiveKitBackend.kt#L171-L176) On a server-side drop, `RoomEvent.Disconnected` ([`:305-308`](https://github.com/NVIDIA/xr-ai/blob/5aa1744e3238c39b28d6942d979c31d0a355386a/client-samples/android/app/src/main/java/com/nvidia/xrai/streamkitsample/streamkit/backends/livekit/LiveKitBackend.kt#L305-L308)) doesn't cancel the scope, so a poll suspended inside `collectNetworkMetrics` resumes and fires one final callback after `AppViewModel` cleared `networkMetrics` ([`AppViewModel.kt:193-195`](https://github.com/NVIDIA/xr-ai/blob/5aa1744e3238c39b28d6942d979c31d0a355386a/client-samples/android/app/src/main/java/com/nvidia/xrai/streamkitsample/AppViewModel.kt#L193-L195)), and `NetworkSection` ([`MainActivity.kt:178`](https://github.com/NVIDIA/xr-ai/blob/5aa1744e3238c39b28d6942d979c31d0a355386a/client-samples/android/app/src/main/java/com/nvidia/xrai/streamkitsample/MainActivity.kt#L178)) has no connection gate, so the dead UI shows the stale values indefinitely. Re-check liveness after the stats await (iOS has the same post-await gap at [`LiveKitBackend.swift:393-401`](https://github.com/NVIDIA/xr-ai/blob/5aa1744e3238c39b28d6942d979c31d0a355386a/client-samples/ios-visionos/StreamKit/Sources/StreamKit/Backends/LiveKit/LiveKitBackend.swift#L393-L401)) and gate the panel on connection state like the web client does.

## Suggestions

1. The new row in [`STREAMKIT_PARITY.md:20`](https://github.com/NVIDIA/xr-ai/blob/5aa1744e3238c39b28d6942d979c31d0a355386a/client-samples/STREAMKIT_PARITY.md#L20) invents a **parity** tag the file doesn't define (rows document divergences tagged idiom/to-fix) while itself recording an asymmetry: only iOS has a stored `@Published networkMetrics` snapshot ([`StreamSession.swift:52`](https://github.com/NVIDIA/xr-ai/blob/5aa1744e3238c39b28d6942d979c31d0a355386a/client-samples/ios-visionos/StreamKit/Sources/StreamKit/StreamSession.swift#L52)). Either tag it **idiom** with the `@Published agentStatus` rationale or add a last-value property on the other platforms.

2. Docs drift: [`clients.md:34`](https://github.com/NVIDIA/xr-ai/blob/5aa1744e3238c39b28d6942d979c31d0a355386a/docs/source/getting_started/clients.md#L34) says "Every client displays the same three network values" but the native sample prints to stdout; the iOS README source tree ([`README.md:18-19`](https://github.com/NVIDIA/xr-ai/blob/5aa1744e3238c39b28d6942d979c31d0a355386a/client-samples/ios-visionos/README.md#L18-L19)) omits `NetworkMetrics.swift` while the Android README was updated; and the `@example` import block in [`index.js:11-20`](https://github.com/NVIDIA/xr-ai/blob/5aa1744e3238c39b28d6942d979c31d0a355386a/client-samples/web/StreamKit/index.js#L11-L20) wasn't extended with the new exports.

3. The Android loop exits permanently once `isConnected` goes false and `RoomEvent.Reconnected` does not restart it ([`LiveKitBackend.kt:301-304`](https://github.com/NVIDIA/xr-ai/blob/5aa1744e3238c39b28d6942d979c31d0a355386a/client-samples/android/app/src/main/java/com/nvidia/xrai/streamkitsample/streamkit/backends/livekit/LiveKitBackend.kt#L301-L304)), so telemetry stays dead after any disconnect/recovery cycle; C++ is the only platform whose loop resumes.

4. iOS re-issues `await track.set(reportStatistics: true)` for every track on every tick and never disables it ([`LiveKitBackend.swift:410`](https://github.com/NVIDIA/xr-ai/blob/5aa1744e3238c39b28d6942d979c31d0a355386a/client-samples/ios-visionos/StreamKit/Sources/StreamKit/Backends/LiveKit/LiveKitBackend.swift#L410)); enable it once per track at publish/subscribe and treat the loop as a read, since the SDK refreshes `statistics` on its own cadence anyway.

5. Only web clamps values ([`NetworkMetrics.js:31`](https://github.com/NVIDIA/xr-ai/blob/5aa1744e3238c39b28d6942d979c31d0a355386a/client-samples/web/StreamKit/NetworkMetrics.js#L31)): C++ applies no presence guard, so an unsampled pair reports `0 ms` where the others report null/nil, and its quality switch ([`LiveKitBackend.cpp:533-547`](https://github.com/NVIDIA/xr-ai/blob/5aa1744e3238c39b28d6942d979c31d0a355386a/client-samples/native/StreamKit/src/Backends/LiveKit/LiveKitBackend.cpp#L533-L547)) has no default/Unknown arm, latching the previous value. Mirror the clamp on Kotlin/Swift/C++ and add the default arm.

6. Register the cleanup in the shared helper `connectedBackend` ([`web_camera_capture.test.mjs:60`](https://github.com/NVIDIA/xr-ai/blob/5aa1744e3238c39b28d6942d979c31d0a355386a/tests/javascript/web_camera_capture.test.mjs#L60)) instead of the one test, and add a `timeout=` to the `subprocess.run` in [`test_web_camera_capture.py:13`](https://github.com/NVIDIA/xr-ai/blob/5aa1744e3238c39b28d6942d979c31d0a355386a/tests/test_web_camera_capture.py#L13): any assertion failure before the in-test disconnect leaves the leaked interval pinning Node, and the CI job hangs instead of failing.

7. The new metric logic is untested where it is testable: `finiteOrNull`, `mapQuality`, and the candidate-pair/jitter extraction can all be covered in the existing harness, and today the mock tracks lack `getRTCStatsReport`, so every tick throws into the swallowed catch at [`LiveKitBackend.js:503`](https://github.com/NVIDIA/xr-ai/blob/5aa1744e3238c39b28d6942d979c31d0a355386a/client-samples/web/StreamKit/Backends/LiveKit/LiveKitBackend.js#L503) and a broken extraction path would pass silently. The Swift test ([`StreamKitTests.swift:17`](https://github.com/NVIDIA/xr-ai/blob/5aa1744e3238c39b28d6942d979c31d0a355386a/client-samples/ios-visionos/StreamKit/Tests/StreamKitTests/StreamKitTests.swift#L17)) only asserts memberwise struct assignment; testing the quality mapping would carry real weight.

The transport-neutral `NetworkMetrics` type and the backend → session → app wiring follow the existing `onAgentStatus` pattern cleanly, and keeping telemetry out of the hub protocol is the right call. The blockers above should be resolved before merge.

nvddr:
Addressed the review feedback in `ee9fb1b`.

## Disposition

- **Native callback-initiated disconnect:** replaced the self-joining `jthread` shutdown path with a stop flag and self-thread detach path. Added a regression test whose metrics callback calls `Disconnect()`, and documented that native telemetry callbacks run on the worker thread.
- **Released LiveKit C++ compatibility:** removed the unavailable `Room::getStats()` call. Native telemetry now gathers `Track::getStats()` from local and remote tracks, uses `Room::connect()`, and documents LiveKit C++ v0.4.1 as the tested release. The real `STREAMKIT_HAVE_LIVEKIT=1` source path was compiled against the v0.4.1 headers locally.
- **Web lifecycle leaks / duplicate starts:** terminal LiveKit disconnects now stop the interval, starting reporting first clears any existing interval, and the app guards concurrent Connect attempts. The shared test helper always disconnects its backend and the Python wrapper has a subprocess timeout.
- **Android UI-thread blocking:** RTC stats collection now runs on `Dispatchers.Default`; only the callback is dispatched to Main.
- **Local quality identity:** native quality events are matched against `room.localParticipant()` rather than `SessionConfig.identity`.
- **Active ICE-pair selection:** all four clients now prefer `transport.selectedCandidatePairId`, with nominated/selected compatibility fallbacks where the SDK exposes them. `state == "succeeded"` is no longer accepted as an active-pair signal.
- **Consistent stats sources:** web now samples every local and remote track publication, including remote video. Android also leaves RTT/jitter unavailable when no media track exists, matching the other clients.
- **Focused web rendering:** the 1 Hz callback updates only the three network fields and no longer rebuilds the received-message list.
- **Stale reconnect/disconnect metrics:** Android and iOS re-check liveness after asynchronous collection, clear metrics while reconnecting, and gate their network panels on the connected state. Android's loop remains alive across reconnects.
- **iOS statistics lifecycle:** reporting is enabled once per track and disabled during teardown rather than being re-enabled every tick.
- **Value handling:** Kotlin, Swift, and C++ now reject non-finite/negative measurements; native treats an unsampled zero RTT as unavailable and resets unknown quality through a default switch arm.
- **Parity and docs:** changed the telemetry parity row to the documented `idiom` tag with the Swift observable-state rationale; corrected native stdout wording, added the omitted iOS source-tree entry and web exports example, and documented the native SDK version.
- **Tests:** added focused web coverage for transport-selected pair choice, inbound video jitter, quality mapping, invalid values, and terminal timer cleanup; added Swift quality-mapping coverage and the native callback-disconnect regression.

## Validation

- Web syntax checks and `node --test tests/javascript/web_camera_capture.test.mjs`
- Native CMake build and all 6 CTest tests
- Real native source compilation with `STREAMKIT_HAVE_LIVEKIT=1` against LiveKit C++ v0.4.1 headers
- `make -C docs api-check`
- `git diff --check`

Android compilation could not run locally because this host has no Android SDK. Swift compilation could not run locally because the host has no Swift toolchain.

blongs-nv:
Reviewed at `ee9fb1ba262a922cb7c616d778a44849328c94d7` by a bot.

All eight blockers and all seven suggestions from my previous review are resolved. One new item in the reworked native teardown:

## Suggestions

1. `StopNetworkMetricsReporting()` is unsynchronised, so two overlapping `Disconnect()` calls race on `network_metrics_thread_` ([`LiveKitBackend.cpp:588-600`](https://github.com/NVIDIA/xr-ai/blob/ee9fb1ba262a922cb7c616d778a44849328c94d7/client-samples/native/StreamKit/src/Backends/LiveKit/LiveKitBackend.cpp#L588-L600)). Since [`StreamingBackend.h:53-54`](https://github.com/NVIDIA/xr-ai/blob/ee9fb1ba262a922cb7c616d778a44849328c94d7/client-samples/native/StreamKit/include/streamkit/Backends/StreamingBackend.h#L53-L54) now invites callback-initiated disconnects, a metrics-callback `Disconnect()` overlapping an app-thread `Disconnect()` (or backend destruction) has the worker `detach()` the thread object while the other thread `join()`s it, which is undefined behavior; a small mutex around teardown closes it.

The recheck holds up well across all four platforms: the transport-selected-pair rule, the connection-state gating, and the new JS/Swift/C++ tests each address the underlying issue rather than the symptom. This is ready to merge; the suggestion above is not blocking.

blongs-nv (approved):

wenxind-nvidia:
Re-reviewed at `30a2f6763ad31edafcf6634fd526954d62d1c26f`. The new teardown commit serializes ownership changes to the telemetry stop flag and `std::thread`, but the following issues remain.

## Blockers

1. **Native can publish telemetry after entering reconnecting or disconnected.** [`PublishNetworkMetrics()`](https://github.com/NVIDIA/xr-ai/blob/30a2f6763ad31edafcf6634fd526954d62d1c26f/client-samples/native/StreamKit/src/Backends/LiveKit/LiveKitBackend.cpp#L621-L709) can spend up to 750 ms awaiting stats. During that wait, [`HandleConnectionStateChange()`](https://github.com/NVIDIA/xr-ai/blob/30a2f6763ad31edafcf6634fd526954d62d1c26f/client-samples/native/StreamKit/src/Backends/LiveKit/LiveKitBackend.cpp#L503-L507) may set `is_connected_` false and notify the application, but line 709 invokes the metrics callback without rechecking connection state or room generation. A late callback can therefore restore stale metrics after the disconnect/reconnecting handler cleared them. Web, Android, and Swift now perform this post-await liveness check; native should do the same, with a delayed-stats regression test.

2. **An exception from the native metrics callback terminates the process.** The telemetry worker has no exception boundary, and [`callback(metrics)`](https://github.com/NVIDIA/xr-ai/blob/30a2f6763ad31edafcf6634fd526954d62d1c26f/client-samples/native/StreamKit/src/Backends/LiveKit/LiveKitBackend.cpp#L705-L709) is outside the RTCStats `try` block. `on_network_metrics` is not documented as no-throw. I reproduced this against the current head in stub mode: a callback throwing `std::runtime_error` aborts with exit code 134. Contain callback exceptions at the worker/callback boundary and define the failure behavior.

3. **Overlapping `Disconnect()` calls still race through the rest of `TearDown()`.** The new mutex protects only `network_metrics_stop_` and `network_metrics_thread_` in [`StopNetworkMetricsReporting()`](https://github.com/NVIDIA/xr-ai/blob/30a2f6763ad31edafcf6634fd526954d62d1c26f/client-samples/native/StreamKit/src/Backends/LiveKit/LiveKitBackend.cpp#L600-L618). Once that returns, both the callback thread and application thread continue through [`TearDown()`](https://github.com/NVIDIA/xr-ai/blob/30a2f6763ad31edafcf6634fd526954d62d1c26f/client-samples/native/StreamKit/src/Backends/LiveKit/LiveKitBackend.cpp#L716-L735), concurrently resetting `room_`, `delegate_`, and the track/source `shared_ptr`s. Concurrent writes to the same `shared_ptr` objects are a data race. The new test runs in stub mode, where those LiveKit members are compiled out, and its 10 ms sleep does not deterministically establish which disconnect owns the worker. Make teardown single-owner/idempotent across the entire operation while avoiding the app-thread-joins-worker / worker-enters-teardown deadlock.

4. **The public Swift `StreamSession.networkMetrics` snapshot remains stale after an unexpected disconnect.** It is cleared only by an explicit [`disconnect()`](https://github.com/NVIDIA/xr-ai/blob/30a2f6763ad31edafcf6634fd526954d62d1c26f/client-samples/ios-visionos/StreamKit/Sources/StreamKit/StreamSession.swift#L98-L103). The backend state callback only updates `connectionState`, while the metrics callback assigns unconditionally ([callback wiring](https://github.com/NVIDIA/xr-ai/blob/30a2f6763ad31edafcf6634fd526954d62d1c26f/client-samples/ios-visionos/StreamKit/Sources/StreamKit/StreamSession.swift#L182-L208)). `AppModel` masks this with a second, gated snapshot, but SDK consumers observing the published session property retain old metrics through reconnecting or terminal disconnection. Clear it for every non-connected state and ignore late metrics unless the session is connected.

## Suggestions

1. Define `roundTripTimeMs` when both publisher and subscriber transports have selected pairs. The implementations collect both selected-pair IDs but return the first candidate-pair match, so the value can switch between uplink and downlink as track/report ordering changes. Select a defined direction, aggregate them, or expose both, then test the two-transport case.

2. Add native regressions for a delayed stats collection followed by disconnection and for a throwing callback. Extend the overlapping-disconnect test so it verifies the real-backend member teardown rather than only thread ownership in stub mode.

## Nits

1. Android truncates milliseconds with [`toInt()`](https://github.com/NVIDIA/xr-ai/blob/30a2f6763ad31edafcf6634fd526954d62d1c26f/client-samples/android/app/src/main/java/com/nvidia/xrai/streamkitsample/MainActivity.kt#L178-L180), while web and Swift round them. Use the same presentation rule for UI parity.

## Verification

- All five web tests pass.
- Native stub regression passes under ASan/UBSan and TSan.
- The real native source syntax-checks against the official LiveKit C++ v0.4.1 headers.
- The iOS sample builds against LiveKit Swift 2.13.0.
- All completed GitHub checks are green; the docs deployment and two pytest jobs were still running at review time.

nvddr:
Addressed Wenxin's re-review in `8419e22`.

## Disposition

- **Late native telemetry after reconnect/disconnect:** added a connection epoch and serialized delivery boundary. A sample collected for an earlier epoch is dropped after disconnect or reconnect, and state transitions cannot publish a callback after notifying the application. Added a stale-epoch regression.
- **Native callback exceptions:** `on_network_metrics` exceptions are contained at the delivery boundary. The callback contract now documents this behavior, and the regression verifies that reporting reaches a second callback after the first throws.
- **Overlapping native teardown:** teardown ownership now covers the complete operation, including room, delegate, tracks, and sources. Metrics-callback teardown is idempotent and does not block behind an application teardown that is joining the worker. The overlap test now uses an explicit teardown-state condition rather than a timing sleep.
- **Swift stale snapshot:** `StreamSession.networkMetrics` is cleared for every non-connected state, and late samples are ignored unless the session is connected. Added lifecycle coverage using a mock backend.
- **Publisher/subscriber RTT:** all clients now define RTT as the highest valid RTT across the active selected publisher and subscriber transport pairs, with the existing nominated/selected compatibility fallback. Web coverage includes two selected transports.
- **Android presentation parity:** millisecond values now round rather than truncate.
- **Real native path:** the changed source compiles with `STREAMKIT_HAVE_LIVEKIT=1` against the official LiveKit C++ v0.4.1 headers. I did not add a real-SDK runtime teardown harness because this environment has no linked LiveKit runtime and adding that infrastructure would substantially expand this review fix.

## Validation

- Native CMake build and all 6 CTest tests
- Native lifecycle regression repeated 10 times
- Native lifecycle regression under ASan/UBSan (leak detection disabled because the environment runs under ptrace)
- Real native source syntax-check against LiveKit C++ v0.4.1 headers
- Web syntax check and `node --test tests/javascript/web_camera_capture.test.mjs`
- `make -C docs api-check`
- `git diff --check`

Android compilation could not run locally because this host has no Android SDK. Swift compilation could not run locally because this host has no Swift toolchain.

wenxind-nvidia:
## Blockers

1. **The new Swift test target does not compile.** [`StreamKitTests.swift:10`](https://github.com/NVIDIA/xr-ai/blob/7041ff495bf6861c8201338d9012342f1c401e38/client-samples/ios-visionos/StreamKit/Tests/StreamKitTests/StreamKitTests.swift#L10) imports both `LiveKit` and `StreamKit`, so `ConnectionState` is ambiguous. `xcodebuild test` fails at this line. Qualify the StreamKit occurrences so the added tests actually build and run.

2. **Native `Connect()` continues after a re-entrant disconnect from a connection callback.** The new threading contract explicitly permits callbacks to call `Disconnect()`, but [`Connect()`](https://github.com/NVIDIA/xr-ai/blob/7041ff495bf6861c8201338d9012342f1c401e38/client-samples/native/StreamKit/src/Backends/LiveKit/LiveKitBackend.cpp#L220-L243) sets `teardown_complete_` false, fires `kConnecting`, and then unconditionally creates and connects the room. If that callback disconnects, teardown marks the backend complete; the outer call still connects a real LiveKit room, `ApplyConnectionState()` suppresses the connected transition, and later teardown skips cleanup. Treat a callback-initiated disconnect as cancellation/generation invalidation and add a regression.

3. **A terminal Android disconnect leaks the connection scope and telemetry coroutine.** [`RoomEvent.Disconnected`](https://github.com/NVIDIA/xr-ai/blob/7041ff495bf6861c8201338d9012342f1c401e38/client-samples/android/app/src/main/java/com/nvidia/xrai/streamkitsample/streamkit/backends/livekit/LiveKitBackend.kt#L316-L319) only clears `isConnected` and forwards the state. The telemetry loop at [lines 171-185](https://github.com/NVIDIA/xr-ai/blob/7041ff495bf6861c8201338d9012342f1c401e38/client-samples/android/app/src/main/java/com/nvidia/xrai/streamkitsample/streamkit/backends/livekit/LiveKitBackend.kt#L171-L185) remains active forever because the scope and room identity remain unchanged. The job captures the backend and room, so replacing the session after each terminal disconnect leaks another scope and polling coroutine. Cancel/clear the terminal connection resources while preserving the loop across `Reconnecting`/`Reconnected`.

4. **Swift retains every track ever enabled for statistics until the whole session disconnects.** [`collectNetworkMetrics()`](https://github.com/NVIDIA/xr-ai/blob/7041ff495bf6861c8201338d9012342f1c401e38/client-samples/ios-visionos/StreamKit/Sources/StreamKit/Backends/LiveKit/LiveKitBackend.swift#L418-L449) only adds tracks to `statisticsTracks`; they are removed and statistics reporting is disabled only during full teardown. Repeated camera toggles or remote-track churn therefore retain every old `Track` and leave statistics enabled. Reconcile the stored set with current publications and disable/remove departed tracks.

## Suggestions

1. Add the standalone Swift package tests to CI. The sample application build succeeds without compiling the test target, which is why blocker 1 is not caught by the current checks.

The previously reported native stale-delivery, callback-exception, overlapping-teardown, Swift stale-snapshot, selected-pair, and web lifecycle issues are resolved.

## Verification

Reviewed head `7041ff4`. Web tests pass, native stub lifecycle tests pass, the iOS sample builds against LiveKit Swift 2.13.0, strict docs and `git diff --check` pass, and the PR merges cleanly with current `main`. The standalone Swift test build fails as described above. Android could not be compiled locally because this environment lacks Java and the Android SDK.

nvddr:
Addressed Wenxin's latest review in `dee363f6`.

## Disposition

- **Swift test ambiguity:** Qualified the test backend and assertions with `StreamKit.ConnectionState`, avoiding the conflicting LiveKit type.
- **Reentrant native `Connect()` cancellation:** Added a private connect-generation guard so `Disconnect()` from the `kConnecting` callback invalidates the in-flight attempt. The teardown lock is released around the blocking SDK connect call, and a regression test verifies that the attempt never proceeds to `kConnected`.
- **Android terminal disconnect cleanup:** Scoped room events to the room that emitted them. A terminal `Disconnected` event now clears the active room and cancels the connection/telemetry scope, while `Reconnecting` and `Reconnected` preserve that scope across transient recovery.
- **Swift departed-track cleanup:** Each statistics pass now reconciles the currently published tracks, disables statistics on departed tracks, and removes their retained entries.
- **Standalone Swift tests in CI:** Deferred. This repository currently has no macOS/Apple CI workflow, so adding one would introduce a new hosted-runner dependency and broader CI infrastructure. I kept this review fix scoped; we can add that workflow separately with explicit approval.

## Validation

- Native CTest: 6/6 passed, including 10 repeated lifecycle runs.
- Native lifecycle test passed under ASan/UBSan (`detect_leaks=0` because leak detection is incompatible with this ptrace environment).
- Native source passed syntax checking against the real LiveKit C++ v0.4.1 headers.
- Web syntax and camera-capture tests passed.
- Documentation API check passed.
- 48 targeted layout, configuration, API, launcher, and LiveKit tests passed.
- `git diff --check` passed.

Android and Swift compilation were not available locally because this environment has neither an Android SDK nor a Swift toolchain; their CI checks will provide platform validation.

wenxind-nvidia (approved):

PR #402, opened 2026-08-20
Fix simple VLM reuse of running model servers

nvddr:
## What changed

- make `simple-vlm-example` select the same hardware-specific STT and VLM configs as `model_servers`
- remove the duplicate sample-local STT and VLM configs
- add regression coverage for Ada, Spark, and Blackwell profiles

## Why

The vLLM persistence safety check only reuses a healthy container when its complete launch fingerprint matches. `simple-vlm-example` used a generic VLM config while `model_servers` used hardware-specific configs, so launching the sample recreated an already-running Cosmos container instead of reusing it.

Sharing the canonical configs preserves the safe fingerprint validation and restores the intended launch-if-missing/reuse-if-compatible behavior.

## Validation

- `pytest`: reuse-contract regression plus affected config-documentation coverage (7 passed)
- Ruff 0.15.16 on all changed Python files
- `git diff --check`


wenxind-nvidia:
## Review

### Blockers

1. [`README.md:306`](https://github.com/NVIDIA/xr-ai/blob/6dab6bd07f8b82be890341ec4861b2dbdfd0a255/README.md#L306) and [`docs/source/getting_started/quickstart.md:155`](https://github.com/NVIDIA/xr-ai/blob/6dab6bd07f8b82be890341ec4861b2dbdfd0a255/docs/source/getting_started/quickstart.md#L155) still instruct users to edit `vlm_server.yaml` in the sample directory. This PR deletes that file, so the documented local-model customization workflow now points to a nonexistent path. Update both sections to reference the appropriate canonical hardware-profile YAML under `agent-samples/model-servers/yaml/<profile>/`, including the Cosmos-Reason1 instructions.

### Suggestions

None.

### Nits

None.

The implementation otherwise looks correct and focused. Verification: 56 affected tests passed, Ruff passed, dependency-map validation passed, and all CI checks are green.

wenxind-nvidia (approved):

wenxind-nvidia (approved):

blongs-nv:
Reviewed at `48cb9ad1dc3b8bb5b3088ac1a94816a4c85feb01` by a bot.

## Blockers

1. **On single-GPU non-Blackwell hosts the sample now pins STT to GPU 1, which doesn't exist, so Parakeet silently loads on CPU.**

   [`agent-samples/simple-vlm-example/main.py:89-98`](https://github.com/NVIDIA/xr-ai/blob/48cb9ad1dc3b8bb5b3088ac1a94816a4c85feb01/agent-samples/simple-vlm-example/main.py#L89-L98) `detect_gpu_config()` returns `dual_48G_ada` for every non-Blackwell host, including single-GPU ones, and on any nvidia-smi failure ([`_gpu.py:73-76`](https://github.com/NVIDIA/xr-ai/blob/48cb9ad1dc3b8bb5b3088ac1a94816a4c85feb01/utils/xr-ai-launcher/xr_ai_launcher/_gpu.py#L73-L76)), and that profile sets `cuda_visible_devices: "1"` ([`stt_server.yaml:17`](https://github.com/NVIDIA/xr-ai/blob/48cb9ad1dc3b8bb5b3088ac1a94816a4c85feb01/agent-samples/model-servers/yaml/dual_48G_ada/stt_server.yaml#L17)), so `device: auto` resolves to CPU with only an info log ([`stt_server/__main__.py:104-106`](https://github.com/NVIDIA/xr-ai/blob/48cb9ad1dc3b8bb5b3088ac1a94816a4c85feb01/services/stt-server/stt_server/__main__.py#L104-L106)). STT reuse is a bare health probe with no fingerprint ([`stt_server/__main__.py:241`](https://github.com/NVIDIA/xr-ai/blob/48cb9ad1dc3b8bb5b3088ac1a94816a4c85feb01/services/stt-server/stt_server/__main__.py#L241)), so rewriting the STT config gains no reuse while importing this pin; the deleted sample config had none.

2. **The standalone sample inherits `gpu_memory_utilization: 0.47`, which cannot load the 16.8 GiB Cosmos3 weights on the ~24 GB cards the docs advertise for it.**

   [`dual_48G_ada/vlm_server.yaml:31`](https://github.com/NVIDIA/xr-ai/blob/48cb9ad1dc3b8bb5b3088ac1a94816a4c85feb01/agent-samples/model-servers/yaml/dual_48G_ada/vlm_server.yaml#L31) The deleted sample config used the 0.85 default ([`vlm_server/__main__.py:60`](https://github.com/NVIDIA/xr-ai/blob/48cb9ad1dc3b8bb5b3088ac1a94816a4c85feb01/services/vlm-server/vlm_server/__main__.py#L60)), so a card matching the documented "simple-vlm-example (standalone) | ~23 GB" ([`README.md:65`](https://github.com/NVIDIA/xr-ai/blob/48cb9ad1dc3b8bb5b3088ac1a94816a4c85feb01/README.md#L65)) now gets a ~11 GiB budget and vLLM aborts at engine init. The 0.47 split exists to co-host omni and embedding, which the standalone sample never starts.

## Suggestions

1. The config filename rule duplicates model-servers' [`_service_config`](https://github.com/NVIDIA/xr-ai/blob/48cb9ad1dc3b8bb5b3088ac1a94816a4c85feb01/agent-samples/model-servers/main.py#L98-L104) minus its `<config>_<profile>.yaml` variant fallback, so the first `vlm_server_<profile>.yaml` variant silently diverges the fingerprints again. A shared resolver would keep the two sites from drifting; the new test only guards the `default` profile.

2. The composed path at [`main.py:95-97`](https://github.com/NVIDIA/xr-ai/blob/48cb9ad1dc3b8bb5b3088ac1a94816a4c85feb01/agent-samples/simple-vlm-example/main.py#L95-L97) is never checked to exist, and [`_config.py:51`](https://github.com/NVIDIA/xr-ai/blob/48cb9ad1dc3b8bb5b3088ac1a94816a4c85feb01/utils/xr-ai-vllm/xr_ai_vllm/_config.py#L51) silently ignores a missing `--config`, so a renamed profile directory surfaces as `'model' is required in config` inside the service. A one-line existence check at build time gives a clear error.

3. When the sample adopts a matching container that is still loading ([`_docker.py:755-761`](https://github.com/NVIDIA/xr-ai/blob/48cb9ad1dc3b8bb5b3088ac1a94816a4c85feb01/utils/xr-ai-vllm/xr_ai_vllm/_docker.py#L755-L761)), the abort handler that stops and removes the container by name ([`_docker.py:676-677`](https://github.com/NVIDIA/xr-ai/blob/48cb9ad1dc3b8bb5b3088ac1a94816a4c85feb01/utils/xr-ai-vllm/xr_ai_vllm/_docker.py#L676-L677)) stays installed until readiness, so a Ctrl-C of the sample removes the persistent container model-servers is still bringing up. Consider not removing a container this wrapper did not create.

4. The new test at [`tests/test_model_servers.py:60-88`](https://github.com/NVIDIA/xr-ai/blob/48cb9ad1dc3b8bb5b3088ac1a94816a4c85feb01/tests/test_model_servers.py#L60-L88) covers only `models.local.json`, while `models.hosted.json` and `models.omni.json` also mark `stt` managed and hit the rewrite, and path equality does not pin the launch fingerprint (an `HF_TOKEN` present in one launcher's environment but not the other's still diverges the digest and recreates the container). Parametrize over the sample's profiles and assert the `launch_fingerprint` payload matches.

5. The sample-authoring walkthrough still teaches copying reference YAML into the sample's own `yaml/` directory ([`ai-services.md:124`](https://github.com/NVIDIA/xr-ai/blob/48cb9ad1dc3b8bb5b3088ac1a94816a4c85feb01/docs/source/components/ai-services.md#L124)), and the sample's module docstring ([`main.py:14-21`](https://github.com/NVIDIA/xr-ai/blob/48cb9ad1dc3b8bb5b3088ac1a94816a4c85feb01/agent-samples/simple-vlm-example/main.py#L14-L21)) does not mention the shared hardware-profile configs. Both should describe the new layout.

Sharing the canonical hardware-profile configs is the right fix for the Cosmos container recreation, and the parametrized regression test genuinely fails without the change. The blockers above should be resolved before merge.

blongs-nv:
Reviewed at `06426848fc0cab539230fd2cbb5cef37806048e0` by a bot.

## Suggestions

1. [`main.py:103-104`](https://github.com/NVIDIA/xr-ai/blob/06426848fc0cab539230fd2cbb5cef37806048e0/agent-samples/simple-vlm-example/main.py#L103-L104) sums memory across GPUs for the single-GPU profiles, so a host with two 48 GiB Blackwell cards passes the 80 GiB gate and gets the 96G profile's [`gpu_memory_utilization: 0.23`](https://github.com/NVIDIA/xr-ai/blob/06426848fc0cab539230fd2cbb5cef37806048e0/agent-samples/model-servers/yaml/96G_blackwell/vlm_server.yaml#L21) on one 48 GiB GPU, an ~11 GiB budget below the 16.8 GiB weights. Gating these profiles on the largest single GPU matches the budget the config actually assumes.

2. A non-numeric `memory.total` (the unified-memory case [`detect_gpu_config` tolerates](https://github.com/NVIDIA/xr-ai/blob/06426848fc0cab539230fd2cbb5cef37806048e0/utils/xr-ai-launcher/xr_ai_launcher/_gpu.py#L67) when classifying Spark) raises `ValueError` at [`main.py:94-96`](https://github.com/NVIDIA/xr-ai/blob/06426848fc0cab539230fd2cbb5cef37806048e0/agent-samples/simple-vlm-example/main.py#L94-L96) and routes such a Spark host to the standalone config, losing reuse and applying the 24 GiB-sized 0.85 budget to unified memory. Mirroring that tolerant parsing keeps the two detectors agreeing.

3. Suggestion 3 from my previous review is unchanged: the signal handler ([`_docker.py:660-677`](https://github.com/NVIDIA/xr-ai/blob/06426848fc0cab539230fd2cbb5cef37806048e0/utils/xr-ai-vllm/xr_ai_vllm/_docker.py#L660-L677)) still stops and removes an adopted still-loading container ([`_docker.py:752-760`](https://github.com/NVIDIA/xr-ai/blob/06426848fc0cab539230fd2cbb5cef37806048e0/utils/xr-ai-vllm/xr_ai_vllm/_docker.py#L752-L760)), so a Ctrl-C of the sample removes the persistent container model-servers is bringing up.

4. The rewritten tests still pin config-path and command equality only; the fingerprint also digests `HF_TOKEN` resolved from the launcher's environment ([`_config.py:86`](https://github.com/NVIDIA/xr-ai/blob/06426848fc0cab539230fd2cbb5cef37806048e0/utils/xr-ai-vllm/xr_ai_vllm/_config.py#L86), [`_docker.py:116-118`](https://github.com/NVIDIA/xr-ai/blob/06426848fc0cab539230fd2cbb5cef37806048e0/utils/xr-ai-vllm/xr_ai_vllm/_docker.py#L116-L118)), so a token present in one launcher's env but not the other's still recreates the container. Asserting the `launch_fingerprint` payload matches would pin the full contract.

Both blockers from my previous review are resolved: STT keeps the sample-local config with no GPU pin, and hosts below a profile's budget fall back to the restored standalone VLM config with the 0.85 budget, with regression tests covering both. Ready to merge; the suggestions above are non-blocking.

blongs-nv (approved):

yanziz-nvidia (approved):

PR #403, opened 2026-08-20
Support extra SANs in self-signed cert for NAT and proxy hosts

blongs-nv:
Behind NAT, clients dial a public IP that is on no local interface, so the auto-generated cert's SAN never covers it and iOS/Android reject the connection even with the profile installed.

- Add `web_server_extra_sans` to `xr_media_hub.yaml` for addresses clients dial that the hub can't detect (public IP, forwarding proxy, DNS name)
- Regenerate the cached cert when a configured entry is missing from its SAN; regeneration is additive so entries are never dropped
- Validate config input: scalar/empty/invalid yaml values and unencodable entries are warned about and skipped instead of crashing startup
- Run cert generation in a worker thread
- Documentation and tests

Tested on LAN with iOS end to end; NAT validation pending.

### DIFF
--- a/client-samples/ios-visionos/README.md
+++ b/client-samples/ios-visionos/README.md
@@ -238,18 +238,19 @@ is:
 
 If you installed the profile and toggled Full Trust on but the wss
 handshake still errors out with NSURLErrorDomain `-1202` and a message
-like *"pretending to be 10.29.90.196"*, the cert's SubjectAlternativeName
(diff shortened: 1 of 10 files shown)


yanziz-nvidia (approved):

nvddr (approved):
Reviewed the SAN normalization and regeneration path, preservation of cached SANs, config coercion, off-event-loop certificate generation, documentation, and regression coverage. The 14 focused TLS/config tests pass locally and CI is green. No blockers.

PR #404, opened 2026-08-20
Apple client keyboard fix

blongs-nv:
Apple clients used numpad keyboard for IP address fields, which prevents users from entering an IP address on devices with a regional format that doesn't use a period as a decimal point (e.g. France).


yanziz-nvidia (approved):

nvddr (approved):
Reviewed the iOS/visionOS keyboard changes for host/IP, port, and token URL fields. The platform guards and keyboard choices address the locale-specific decimal separator issue without changing connection behavior. CI is green. No blockers.

PR #276, opened 2026-06-30
Fix strict YAML boolean parsing

fallintoplace:
## Summary

Replace truthiness casts for YAML-backed boolean config values with strict parsers. Quoted values like `"false"`, `"no"`, and `"0"` now parse as false instead of becoming true because they are non-empty strings.

This covers the vLLM service wrappers, Piper `use_cuda`, VLM MCP `enable_thinking`, model `health_check`, and voicegate `listening_chime`. Nemotron Omni's `use_bf16` path is included too because it had the same string truthiness issue.

Invalid boolean strings now raise `ValueError` with the config key name instead of silently picking a truthy value.

## Validation

- `PYTHONPATH=server-runtime:utils/xr-ai-vllm:agent-sdk/xr-ai-models:agent-mcp-servers/vlm-mcp:utils/xr-ai-voicegate:ai-services/tts/piper:utils/xr-ai-logging:tests uv run --python 3.12 --project agent-sdk --with pytest --with pytest-asyncio --with numpy --with aiohttp --with fastmcp --with pillow --with httpx --with uvicorn --with pyyaml --with loguru pytest tests/test_vllm_lifecycle.py tests/test_models_config.py tests/test_vlm_mcp.py tests/test_xr_ai_voicegate.py tests/test_piper_tts.py::test_piper_config_parses_quoted_use_cuda_false tests/test_piper_tts.py::test_piper_config_rejects_unknown_use_cuda_string`
- `uvx ruff check utils/xr-ai-vllm/xr_ai_vllm/_config.py utils/xr-ai-vllm/xr_ai_vllm/__init__.py ai-services/vlm-server/vlm_server/__main__.py ai-services/llm/llama_nemotron/llama_nemotron_llm_server/__main__.py ai-services/llm/nemotron3_nano/nemotron3_nano_llm_server/__main__.py ai-services/llm/nemotron_omni/nemotron_omni_llm_server/__main__.py ai-services/tts/piper/piper_tts_server/__main__.py agent-sdk/xr-ai-models/xr_ai_models/config.py agent-mcp-servers/vlm-mcp/vlm_mcp_server/__main__.py utils/xr-ai-voicegate/xr_ai_voicegate/config.py tests/test_vllm_lifecycle.py tests/test_models_config.py tests/test_vlm_mcp.py tests/test_xr_ai_voicegate.py tests/test_piper_tts.py`
- `git diff --check`

PR #263, opened 2026-06-23
fix(cloudxr): start runtime with spawn context

WilliamK112:
## Summary

- start the native CloudXR runtime process from an explicit `spawn` multiprocessing context instead of the platform-default fork context
- preserve the existing `multiprocessing.Process` handle contract used by readiness polling and `terminate_or_kill_runtime`
- add a focused unit test that stubs native CloudXR imports and verifies the spawn context
- document the process-model decision in the architecture changelog

## Rationale

Issue #40 calls out that the current `multiprocessing.Process(target=runtime_run)` path can fork from inside an already-running asyncio event loop. Using `spawn` starts the runtime in a fresh interpreter, avoiding inherited asyncio lock and signal-handler state while keeping the current readiness and cleanup APIs intact.

Fixes #40

## Validation

Refreshed onto current `NVIDIA/xr-ai@6de6ca73cec931c5833fe7c418fb30e1df5b4172`; GitHub reports `behind_by: 0`, `ahead_by: 1`.

- `pytest -q tests/test_cloudxr_runtime_process.py tests/test_cloudxr_service_layout.py` — 4 passed
- Ruff 0.15.16 on both changed Python files — passed
- repository SPDX checker on source, test, and changelog — passed
- signed-off commit / DCO — present
- `git diff --check upstream/main...HEAD` — passed

The unit test imports the runtime from its current `services/cloudxr-runtime` location and replaces the platform-specific Isaac Teleop modules with fakes, so this process-launch contract is verified without requiring CloudXR hardware.

PR #408, opened 2026-08-20
fix(runtime): consolidate pending reliability fixes

nvddr:
## Summary

Consolidates and rebases the still-needed fixes from #261, #262, #263, #275, #276, and #277 onto the current repository structure:

- remove a stale shared-memory ring when connector re-registration cannot open its replacement
- require explicit LiveKit credentials, with environment injection support
- start the native CloudXR runtime with the `spawn` multiprocessing context
- reject invalid shared-memory frame sizes before claiming or reading a slot, including non-byte memoryviews
- parse surviving YAML boolean fields strictly instead of relying on string truthiness
- bound each participant's return-audio backlog by audio duration and drop the oldest queued frames under overload

The old VLM MCP portions of #276 are no longer applicable because that server was retired. The current model configuration already validates `health_check`, so this applies strict parsing only to the remaining affected service fields. The shared parser stays internal rather than expanding the public API.

Original work is attributed in the commit to @WilliamK112 and @fallintoplace.

## Validation

- Ruff 0.15.16 on every changed Python file
- `git diff --check`
- Python 3.11: 1,213 passed, 10 GPU tests deselected
- Python 3.12: 1,213 passed, 10 GPU tests deselected


wenxind-nvidia:
## Review — PR #408 (`0a9cfa8`)

**Verdict: Request changes.** The rebase onto the DeviceIOHub rename is clean and the current CI matrix is green. One shared-memory integrity issue remains blocking.

### Blockers

1. **`read_slot()` still trusts the ZMQ signal's size and metadata instead of the slot header.**

   `ShmRingBuffer.read_slot()` checks `signal.data_sz` only against slot capacity, then slices using that value. It does not validate `signal.slot` before calculating the offset or compare the signal with the header's actual `data_sz`, sequence, dimensions, or format.

   This still accepts an invalid in-capacity signal: after writing `b"ABCD"` to an 8-byte slot, a signal declaring `data_sz=8` returns `b"ABCD\x00\x00\x00\x00"` with the signal's forged metadata. Depending on slot reuse, those extra bytes can be stale frame data. That contradicts this PR's stated guarantee to reject invalid frame sizes before reading a slot.

   Please validate `0 <= signal.slot < self._num_slots`, the slot magic/state, the header's own size against capacity, and the signal/header agreement before returning a view (or construct the returned metadata from the canonical header). Add regressions for an in-capacity size mismatch and negative/out-of-range slot indices.

### Suggestions

1. **Restore #277's configuration hook for the return-audio limit.** The consolidation keeps a fixed three-second `_RETURN_AUDIO_MAX_BUFFER_S` and creates `_ReturnAudioPipe` without reading a `LiveKitConnectorConfig` field. Reintroducing `return_audio_max_buffer_s` would let latency-sensitive and long-form deployments choose an appropriate bound. An integration-level test through `RoomClient` would also catch wiring omissions while verifying that overflow remains participant-local.

2. **Finish the strict-YAML sweep.** `agent-samples/lab-instrument-monitoring/worker/lab_instrument_monitoring_worker/config.py` still has `capture_marker_scans=bool(data.get("capture_marker_scans", False))`, so quoted YAML `"false"` evaluates to `True`.

3. **Make the credential failure actionable.** `LiveKitConnectorConfig.__post_init__()` currently says only to set `api_key` and `api_secret`; include `LIVEKIT_API_KEY`, `LIVEKIT_API_SECRET`, and the DeviceIOHub YAML alternative so startup failures point directly to every supported configuration path.

4. Add a changelog entry for the explicit-credential migration and the return-audio overload/drop policy, since both change operator-visible behavior.

### Nits

- `_ReturnAudioPipe.close()` says a "bounded queue could drop the sentinel," but the underlying `asyncio.Queue` is unbounded; queued audio duration is bounded manually. Please rephrase this to describe the actual cancellation invariant.
- `_validate_return_audio_max_buffer_s(True)` accepts `True` as `1.0`. Explicitly reject booleans if the limit becomes YAML-configurable.

### Validation

- Current GitHub checks pass, including Python 3.11/3.12, Ruff, SPDX, dependency-map, lockfile, build, and CodeQL checks.
- Local targeted tests: 112 configuration/audio/service tests passed; 12 shared-memory/IPC tests passed.

blongs-nv:
Reviewed at `0a9cfa8d5077a08d4f304da724e2c464fe44b7f0` by a bot.

## Blockers

1. **The 3-second drop-oldest bound in `_ReturnAudioPipe.push` truncates ordinary TTS replies, not just pathological backlog.**

   [`services/device-io-hub/device_io_hub/transport/livekit/_room_client.py:100-123`](https://github.com/NVIDIA/xr-ai/blob/0a9cfa8d5077a08d4f304da724e2c464fe44b7f0/services/device-io-hub/device_io_hub/transport/livekit/_room_client.py#L100-L123) Nothing between TTS and this pipe paces audio: pipecat's without-mixer output loop calls `write_audio_frame` as fast as frames arrive and the XR transport forwards them non-blocking over ZMQ, so a whole utterance lands in the queue in milliseconds while `_drain` plays it at 1×. Any reply longer than ~3 s of audio loses everything but a fragment of the start and the final 3 s. Barge-in staleness is already handled by `flush()` + `clear_queue()`; if a bound is still wanted it needs to be utterance-aware or based on wall-clock staleness, not drop-oldest.

2. **A FRAME_SIGNAL rejected by the new `read_slot` validation permanently leaks its ring slot on the hub side.**

   [`services/device-io-hub/device_io_hub/ipc/_hub.py:309-316`](https://github.com/NVIDIA/xr-ai/blob/0a9cfa8d5077a08d4f304da724e2c464fe44b7f0/services/device-io-hub/device_io_hub/ipc/_hub.py#L309-L316) The previously held slot is popped and released before `ring.read_slot(msg)`, and the `ValueError` this PR adds escapes to `run()`'s generic handler with no `release_slot`, leaving the signalled slot READY forever. `_claim_slot` skips non-FREE slots, so a handful of bad signals exhausts the ring (default 10 slots) and every later frame from that connector fails with "all slots occupied" — a permanent video outage instead of a dropped frame.

3. **The three sample `device_io_hub.yaml` copies were not updated alongside the credential change.**

   [`DEPENDENCIES.md:630`](https://github.com/NVIDIA/xr-ai/blob/0a9cfa8d5077a08d4f304da724e2c464fe44b7f0/DEPENDENCIES.md#L630) requires hub configuration changes to fan out to "Its reference YAML and every sample `device_io_hub.yaml`" in the same change. [`agent-samples/xr-render-demo/yaml/device_io_hub.yaml:10-11`](https://github.com/NVIDIA/xr-ai/blob/0a9cfa8d5077a08d4f304da724e2c464fe44b7f0/agent-samples/xr-render-demo/yaml/device_io_hub.yaml#L10-L11) and the simple-vlm-example and lab-instrument-monitoring copies still ship bare `devkey`/`devsecret…` with no mention of `LIVEKIT_API_KEY`/`LIVEKIT_API_SECRET`, and all three are auto-published into the generated config reference alongside the updated guidance.

## Suggestions

1. Re-export `parse_config_bool` from the `xr_ai_vllm` package root and import it from there in the five service `__main__` files ([`services/embedding-server/embedding_server/__main__.py:42`](https://github.com/NVIDIA/xr-ai/blob/0a9cfa8d5077a08d4f304da724e2c464fe44b7f0/services/embedding-server/embedding_server/__main__.py#L42)): `load_config`, `resolve_model_cache`, and `setup_hf_env` live in the same `_config.py` and reach these files through the package `__all__`, which is how the import block directly above consumes them.

2. A non-contiguous memoryview makes `data.cast("B")` raise `TypeError` ([`agent-sdk/xr-ai-hub/xr_ai_hub/_shm.py:169`](https://github.com/NVIDIA/xr-ai/blob/0a9cfa8d5077a08d4f304da724e2c464fe44b7f0/agent-sdk/xr-ai-hub/xr_ai_hub/_shm.py#L169)), outside the `ValueError` contract that frame producers rely on (e.g. [`_room_client.py:482`](https://github.com/NVIDIA/xr-ai/blob/0a9cfa8d5077a08d4f304da724e2c464fe44b7f0/services/device-io-hub/device_io_hub/transport/livekit/_room_client.py#L482)), so one strided view escalates a droppable frame into a stream-ending error. Raise the same `ValueError` for non-contiguous input.

3. `_frame_duration_s` raises out of `push()` for a chunk with non-positive `samples`/`sample_rate` ([`services/device-io-hub/device_io_hub/transport/livekit/_room_client.py:151-159`](https://github.com/NVIDIA/xr-ai/blob/0a9cfa8d5077a08d4f304da724e2c464fe44b7f0/services/device-io-hub/device_io_hub/transport/livekit/_room_client.py#L151-L159)), aborting dispatch of that return-audio message; record it as a drop instead, matching the oversized-frame policy.

4. Store stripped credential values in [`services/device-io-hub/device_io_hub/_config_loader.py:43-49`](https://github.com/NVIDIA/xr-ai/blob/0a9cfa8d5077a08d4f304da724e2c464fe44b7f0/services/device-io-hub/device_io_hub/_config_loader.py#L43-L49): a trailing newline from `LIVEKIT_API_KEY=$(cat file)` passes the `.strip()` validation but is signed into JWTs verbatim, failing auth with no startup clue.

5. Two test gaps: [`tests/test_livekit_config_loader.py`](https://github.com/NVIDIA/xr-ai/blob/0a9cfa8d5077a08d4f304da724e2c464fe44b7f0/tests/test_livekit_config_loader.py) never covers a YAML that exists but has no credential keys and no env vars (the likeliest real misconfiguration; every failure case goes through the no-file branch), and [`tests/test_participant_events.py:180-191`](https://github.com/NVIDIA/xr-ai/blob/0a9cfa8d5077a08d4f304da724e2c464fe44b7f0/tests/test_participant_events.py#L180-L191) seeds `_latest_slots = {}` so the failed-re-registration path never exercises the held-slot purge it sits next to.

6. Strict parsing now rejects values that previously worked, e.g. `enforce_eager: 1`/`0` and a null `listening_chime:` become startup-fatal; the new [`docs/source/reference/migrations.md`](https://github.com/NVIDIA/xr-ai/blob/0a9cfa8d5077a08d4f304da724e2c464fe44b7f0/docs/source/reference/migrations.md) row covers only the credential change, so add a line for this since existing configs can hit it.

7. The spawned runtime child no longer inherits `setup_logging("cloudxr")` ([`services/cloudxr-runtime/cloudxr_runtime/__main__.py:65-70`](https://github.com/NVIDIA/xr-ai/blob/0a9cfa8d5077a08d4f304da724e2c464fe44b7f0/services/cloudxr-runtime/cloudxr_runtime/__main__.py#L65-L70)), so Python-level errors from `runtime.run` land on a default stderr sink instead of the cloudxr log; wrap the target in a small function that calls `setup_logging` first.

This is a careful consolidation: the registry pop, the spawn switch, the write-side frame validation, and the strict bool parsing all land correctly on the current code, with genuine regression tests. The blockers above should be resolved before merge.

nvddr:
Addressed in `5478f42`:

- `read_slot()` now validates slot bounds, header magic/state, the header-owned size, and signal/header agreement for sequence, timestamp, dimensions, format, and size; the returned view is sliced using the canonical header size. Added regressions for in-capacity size forgery, every stored metadata field, negative/out-of-range slots, corrupt magic/size, and non-ready state.
- Restored `return_audio_max_buffer_s` (default 3 seconds) in `LiveKitConnectorConfig` and the reference YAML, wired it through `RoomClient` to every participant pipe, rejected booleans/non-finite values, and added a RoomClient-level participant-isolation/overflow test.
- Replaced the remaining `capture_marker_scans` string-truthiness conversion with strict parsing and tests.
- Expanded the missing-credential error to name `device_io_hub.yaml`, `LIVEKIT_API_KEY`, and `LIVEKIT_API_SECRET`.
- Documented the credential migration and return-audio drop policy in the release migration and server-runtime docs.
- Corrected the queue cancellation comment to describe the duration-bounded backlog over an unbounded `asyncio.Queue`.

Validation: Python 3.11 and 3.12 each pass 1,231 non-GPU tests locally; all GitHub checks are green.

blongs-nv:
Reviewed at `7e5737d98f285fb9691b55c4f95b507c5ab8088c` by a bot.

Blockers 2 and 3 and suggestions 2 through 7 from my previous review are resolved, with genuine regression tests. Blocker 1's fix is close but leaves a residual case, and the stricter `release_slot` introduces one new issue.

## Blockers

1. **Measuring return-audio staleness from enqueue time still drops mid-reply audio whenever TTS delivery leads playback by more than the threshold.**

   [`services/device-io-hub/device_io_hub/transport/livekit/_room_client.py:107`](https://github.com/NVIDIA/xr-ai/blob/7e5737d98f285fb9691b55c4f95b507c5ab8088c/services/device-io-hub/device_io_hub/transport/livekit/_room_client.py#L107) A frame's age clock starts at push while the queue drains at 1×, so a reply whose synthesis runs more than `max_buffer_s` ahead of playback ages out queued frames as later frames of the same utterance keep arriving, and the drops interleave with keeps (stutter, not a clean skip). I reproduced with the real pipe at `max_buffer_s=0.05`: one 60-frame utterance delivered at 5× real time lost 24 mid-reply frames in an every-4th-frame-kept pattern; at the 3 s default, a reply delivered at 2× real time starts losing audio once it exceeds ~12 s. Measure staleness from when a frame reaches the queue head (a healthy drain never parks a frame there), or drop only while `capture_frame` is stalled.

## Suggestions

1. `release_slot` now raises on non-READY or bad-magic headers ([`agent-sdk/xr-ai-hub/xr_ai_hub/_shm.py:302`](https://github.com/NVIDIA/xr-ai/blob/7e5737d98f285fb9691b55c4f95b507c5ab8088c/agent-sdk/xr-ai-hub/xr_ai_hub/_shm.py#L302)), and the four hub sites releasing hub-held slots call it unguarded: a raise at [`services/device-io-hub/device_io_hub/ipc/_hub.py:336`](https://github.com/NVIDIA/xr-ai/blob/7e5737d98f285fb9691b55c4f95b507c5ab8088c/services/device-io-hub/device_io_hub/ipc/_hub.py#L336) escapes `_dispatch` after `prev` was popped and before the new view is stored, leaking both slots, and a raise in the participant-leave purge skips the leave publication. These trip only on the same misbehaving-producer shared memory `read_slot` now defends against, so tolerate the failure at held-slot release sites the way the rejected-signal path does.

2. Suggestion 1 from my previous review is not yet addressed: the five service `__main__` files still import `parse_config_bool` from the private `xr_ai_vllm._config` module (e.g. [`services/embedding-server/embedding_server/__main__.py:42`](https://github.com/NVIDIA/xr-ai/blob/7e5737d98f285fb9691b55c4f95b507c5ab8088c/services/embedding-server/embedding_server/__main__.py#L42)) instead of a package-root re-export.

The `read_slot` validation now treats the slot header as canonical with thorough regressions, and the FRAME_SIGNAL error path correctly frees rejected slots without touching held views. The blocker above should be resolved before merge.

wenxind-nvidia:
## Review — PR #408 (`7e5737d`)

**Verdict: Request changes.** The previous header-validation and slot-leak findings are fixed, but two blockers remain.

### Blockers

1. **A duplicate valid `FRAME_SIGNAL` leaves the hub exposing a freed shared-memory slot.**

   The hub reads the new view before releasing the previous one, but does not handle both views referring to the same ring and slot. A replayed signal therefore succeeds, the previous entry releases that same slot, and the new view is stored after its slot becomes reusable. A subsequent write mutates the “latest” frame underneath its old metadata.

   My one-slot reproduction produced:

   ```text
   reused_slot=0, held_signal_seq=1, held_bytes_after_reuse=b'WXYZ'
   ```

   Treat a `(same ring, same slot)` signal as a duplicate without releasing the slot, and add a regression covering replay followed by slot reuse.

2. **`return_audio_max_buffer_s` no longer bounds the backlog—or even its maximum age.**

   Eviction happens only during `push()` and compares enqueue wall-clock age. A fast producer can therefore enqueue an arbitrary amount before time advances; if no later frame arrives, nothing subsequently checks the age.

   With `max_buffer_s=0.03`, a zero-time burst accepted 10,000 frames representing 100 seconds of audio with zero drops. This contradicts the PR’s stated goal of bounding participant backlog and leaves memory and playback latency unbounded.

   Please retain a real hard bound—by bytes, frames, or audio duration—while handling normal long replies through pacing or utterance-aware segmentation. Add a test asserting a burst much larger than the configured bound cannot grow the queue indefinitely.

### Suggestions

1. Finish the strict-boolean sweep for `complete_on_skip`. It still uses `bool(raw.get(...))`, so quoted YAML `"false"` becomes `True` in `agent-samples/tea-making-sample/worker/tea_making_worker/spec.py`.

2. Re-export `parse_config_bool` from `xr_ai_vllm`, as the original #276 did. The service entry points currently import the private `xr_ai_vllm._config` module while neighboring helpers use the public package surface.

### Nits

- The ignored failure around `release_slot()` has an empty `except`. Add a concise explanation or debug log because failure means the rejected slot may remain occupied.

### Validation

- Current GitHub checks are green.
- `git diff --check` passes.
- 51 focused shared-memory, IPC, audio, credential, and CloudXR tests passed locally on Python 3.12.
- Both blocker reproductions were run against the current head.

nvddr:
Disposition for the reviews against `7e5737d`, addressed in rebased head `23b6046`:

- Return audio now has both protections: the built-in voice transport paces each participant before IPC so ordinary long replies do not flood the hub, while DeviceIOHub enforces a hard per-participant queued-audio duration bound for custom or faulty producers. Regressions cover a paced 60-frame reply and a 10,000-frame burst that cannot grow beyond the configured limit.
- Duplicate valid frame signals for the same ring and slot are ignored without releasing the held slot. The regression replays a signal, fills every other slot, and verifies that the held frame cannot be reused or mutated.
- Held-slot cleanup is centralized and tolerant of corrupt headers during frame replacement, participant departure, connector re-registration, and shutdown. Rejected-slot release failures now emit diagnostic context instead of using an empty exception handler.
- The tea workflow now strictly parses `complete_on_skip`; quoted `"false"` no longer becomes true, and invalid values fail with a targeted error.
- The suggested `parse_config_bool` package-root re-export is intentionally not adopted. It is a YAML implementation helper rather than an `xr_ai_vllm` capability, and keeping it private avoids broadening the supported public API.

Validation: the complete non-GPU suite passed on Python 3.11 and 3.12 (`1,281 passed`, `10 deselected` on each), repository-wide Ruff and whitespace checks passed, and the patch-identical rebase onto current `main` passed the focused suite plus the new base test on both versions (`151 passed` each).

blongs-nv:
Reviewed at `23b6046886efb40eecb6c5d255a936c9ec043f77` by a bot.

The return-audio blocker from my previous review is resolved by the pacing-plus-hard-bound redesign, and `release_slot` failures are now tolerated at every held-slot release site. One gap remains in the new duplicate-signal guard.

## Blockers

1. **The duplicate `FRAME_SIGNAL` guard checks only the same `(participant, track)` key, so a replayed signal carrying a different `track_id` still leads to the held slot being freed and reused.**

   [`services/device-io-hub/device_io_hub/ipc/_hub.py:361-364`](https://github.com/NVIDIA/xr-ai/blob/23b6046886efb40eecb6c5d255a936c9ec043f77/services/device-io-hub/device_io_hub/ipc/_hub.py#L361-L364) The slot header carries no track identity, so a replayed signal with an altered `track_id` passes `read_slot`, misses the same-key `prev` check, and stores a second view of the already-held slot; the next genuine frame for the original track then releases the slot out from under the other view. I reproduced it: after replaying slot 0's signal as track `camB` and advancing `camA` by one frame, the producer reused slot 0 and `camB`'s held frame mutated from `b'AAAA'` to `b'WXYZ'`. Any accepted signal aliasing any held slot is necessarily a duplicate (a held slot is never FREE for a new frame), so the any-entry scan the rejection path already uses at [`_hub.py:339-342`](https://github.com/NVIDIA/xr-ai/blob/23b6046886efb40eecb6c5d255a936c9ec043f77/services/device-io-hub/device_io_hub/ipc/_hub.py#L339-L342) is the correct guard here too.

## Suggestions

1. Suggestion 2 from my previous review remains open: the five service `__main__` files still import `parse_config_bool` from the private `xr_ai_vllm._config` module (e.g. [`services/embedding-server/embedding_server/__main__.py:42`](https://github.com/NVIDIA/xr-ai/blob/23b6046886efb40eecb6c5d255a936c9ec043f77/services/embedding-server/embedding_server/__main__.py#L42)) instead of a package-root re-export.

The pacing redesign is the right shape: the built-in voice transport now paces per participant before IPC so long replies play in full, while the per-participant duration bound caps faulty-producer bursts, both with solid regression tests. The blocker above should be resolved before merge.

nvddr:
Disposition for the additional review:

- Fixed the remaining blocker in `6474e8c`: duplicate `FRAME_SIGNAL` handling now checks the incoming ring/slot against every held view, so replaying a held slot under another `track_id` cannot create a stale shared-memory alias.
- Added a regression covering the reported `camA`/`camB` replay, advancing `camA`, and reusing the original slot. It verifies no `camB` view is retained and the live `camA` bytes remain stable.
- Kept `parse_config_bool` private. Re-exporting a YAML-only helper would unnecessarily broaden the public API.

Validation: `tests/test_participant_events.py` passes on Python 3.11 and 3.12 (10 tests each); Ruff passes for both changed files.

blongs-nv (approved):

wenxind-nvidia:
## Suggestions

1. **Route participant-scoped interruptions to the paced participant sender.**

   [`DeviceIOHubOutputTransport.write_audio_frame()`](https://github.com/NVIDIA/xr-ai/blob/6474e8c9c94a1ec74e6ae95153c2ef1e9cb2e428/agent-sdk/xr-ai-voice/xr_ai_voice/_transport.py#L337-L383) now paces audio before IPC, so the remainder of a reply stays queued in that participant\x27s Pipecat `MediaSender`. `InterruptionFrame` identifies the participant through `transport_source`, but `_handle_frame()` routes by `transport_destination`, which is `None`; the interruption therefore reaches only the default sender. I reproduced this with Alice and Bob senders: `default=1, alice=0, bob=0`. The hub queue is flushed, but Alice\x27s sender then resumes sending the old reply.

   Route a scoped interruption to the sender named by `transport_source`, route a participant-less interruption to every sender, and clear the corresponding pacing deadline. Please add a regression showing Alice stops while Bob continues.

2. **Normalize a non-integer `FrameSignal.data_sz` to the existing rejection path.**

   [`ShmRingBuffer.read_slot()`](https://github.com/NVIDIA/xr-ai/blob/6474e8c9c94a1ec74e6ae95153c2ef1e9cb2e428/agent-sdk/xr-ai-hub/xr_ai_hub/_shm.py#L244-L253) compares `signal.data_sz` numerically without first validating its type. A msgpack signal with `data_sz="4"` raises `TypeError`, which is not caught by the hub\x27s `(RuntimeError, ValueError)` handler, so the READY slot is never released. In a two-slot reproduction, the malformed signal raised `TypeError`, and after one more frame the next write failed with `ShmRingBuffer: all slots occupied`.

   Validate that `data_sz` is an integer excluding booleans and raise `ValueError`, then add a dispatch-level regression confirming that the malformed signal is dropped and its slot becomes reusable.

wenxind-nvidia (approved):

nvddr:
Disposition for the latest suggestions, addressed in `05e7458`:

- Participant-scoped `InterruptionFrame`s now route by `transport_source` to the matching participant `MediaSender` and clear only that participant pacing deadline. Participant-less interruptions fan out to every sender and clear all pacing deadlines. Regressions verify Alice stops while Bob remains untouched, plus the global fallback.
- `ShmRingBuffer.read_slot()` now rejects non-integer and boolean `data_sz` values with `ValueError`, keeping malformed signals in the hub rejection-and-release path. Unit coverage includes string, float, and boolean sizes; a dispatch-level regression proves a rejected READY slot is reusable.
- No rebase was needed: current `main` (`782f67f`) is already an ancestor of the PR head.

Validation: the full non-GPU suite passes on Python 3.11 and 3.12 (`1,288 passed`, `10 deselected` each), and Ruff/whitespace checks pass.

PR #410, opened 2026-08-21
fix(native): address Sonar telemetry findings

yanziz-nvidia:
## Summary

- fix the actionable Sonar findings in native network telemetry: explicit optional checks and lambda captures, const-correct parameters, CTAD/scoped locks, and a dedicated test exception
- group the telemetry synchronization and lifecycle fields into one cohesive state object
- hide the mutable thread-local callback marker behind an accessor
- retain std::thread/self-detach and catch-all callback boundaries where they are required; annotate those exact lines because std::jthread cannot safely self-join and user callbacks can throw non-standard exceptions

## Sonar audit

This addresses the 36 C++ findings from the August 20 scan. The separate Python S2083 report in lab-instrument-monitoring is not a vulnerability: Sonar followed YAML contents into Path.write_text() as file content, while the destination is a constant filename in an internally created TemporaryDirectory. I documented that flow and resolved the issue as FALSE_POSITIVE in Sonar.

## Validation

- compiled the native sample and all six StreamKit tests directly with Apple clang 21 in C++20 stub mode
- ran all six StreamKit test binaries successfully
- git diff --check

CMake was not installed in the local environment; the direct compiler commands covered the same stub-mode sources and warning flags.

### DIFF
--- a/client-samples/native/App/main.cpp
+++ b/client-samples/native/App/main.cpp
@@ -126,21 +126,26 @@ int main(int argc, char** argv) {
     };
 
     session.on_network_metrics = [](const streamkit::NetworkMetrics& metrics) {
-        const auto quality = [&]() -> std::string_view {
+        const auto quality = [&metrics]() -> std::string_view {
+            using enum streamkit::NetworkQuality;
             switch (metrics.quality) {
-                case streamkit::NetworkQuality::kExcellent: return "excellent";
-                case streamkit::NetworkQuality::kGood: return "good";
-                case streamkit::NetworkQuality::kPoor: return "poor";
-                case streamkit::NetworkQuality::kLost: return "lost";
-                case streamkit::NetworkQuality::kUnknown: return "unknown";
+                case kExcellent: return "excellent";
+                case kGood: return "good";
+                case kPoor: return "poor";
+                case kLost: return "lost";
+                case kUnknown: return "unknown";
             }
             return "unknown";
         }();
         std::cout << "[network] quality=" << quality << " rtt=";
-        if (metrics.round_trip_time_ms) std::cout << *metrics.round_trip_time_ms << "ms";
+        if (metrics.round_trip_time_ms.has_value()) {
+            std::cout << *metrics.round_trip_time_ms << "ms";
+        }
         else std::cout << "—";
         std::cout << " jitter=";
-        if (metrics.receive_jitter_ms) std::cout << *metrics.receive_jitter_ms << "ms";
+        if (metrics.receive_jitter_ms.has_value()) {
+            std::cout << *metrics.receive_jitter_ms << "ms";
+        }
         else std::cout << "—";
         std::cout << "\n";
     };
--- a/client-samples/native/App/main.cpp
+++ b/client-samples/native/App/main.cpp
@@ -140,13 +140,15 @@ int main(int argc, char** argv) {
         std::cout << "[network] quality=" << quality << " rtt=";
         if (metrics.round_trip_time_ms.has_value()) {
             std::cout << *metrics.round_trip_time_ms << "ms";
+        } else {
+            std::cout << "—";
         }
-        else std::cout << "—";
         std::cout << " jitter=";
         if (metrics.receive_jitter_ms.has_value()) {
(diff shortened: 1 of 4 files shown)


blongs-nv:
Reviewed at `590f828ee2cca55de18d38f8533cf2a883a160c7` by a bot.

## Suggestions

1. [`LiveKitBackend.h:222`](https://github.com/NVIDIA/xr-ai/blob/590f828ee2cca55de18d38f8533cf2a883a160c7/client-samples/native/StreamKit/include/streamkit/Backends/LiveKit/LiveKitBackend.h#L222) `NetworkMetricsState` puts two near-identically named locks side by side: `network_metrics_.mutex` (plain) and `network_metrics_.delivery->mutex` (recursive), where the old globally distinct names made a mix-up a compile error. Renaming the struct field (e.g. `thread_mutex`) and adding a one-line lock-order note would keep a future edit that grabs the wrong sibling from compiling into a silent self-deadlock on the re-entrant callback → `Disconnect()` path.

2. [`LiveKitBackendTests.cpp:233`](https://github.com/NVIDIA/xr-ai/blob/590f828ee2cca55de18d38f8533cf2a883a160c7/client-samples/native/StreamKit/Tests/StreamKitTests/LiveKitBackendTests.cpp#L233) The new `CallbackFailure` still derives from `std::runtime_error`, so the "user callbacks may throw any exception type" contract at [`LiveKitBackend.cpp:797`](https://github.com/NVIDIA/xr-ai/blob/590f828ee2cca55de18d38f8533cf2a883a160c7/client-samples/native/StreamKit/src/Backends/LiveKit/LiveKitBackend.cpp#L797) stays unpinned: a regression to `catch (const std::exception&)` would still pass this test. Throwing a non-`std::exception` (e.g. `throw 42;`) on the first delivery and `CallbackFailure` on the second would pin the full `catch (...)` contract for one extra line.

3. [`LiveKitBackendTests.cpp:179`](https://github.com/NVIDIA/xr-ai/blob/590f828ee2cca55de18d38f8533cf2a883a160c7/client-samples/native/StreamKit/Tests/StreamKitTests/LiveKitBackendTests.cpp#L179) The `std::jthread` swap is inert: the lambda takes no `stop_token`, the explicit `join()` at line 191 remains, and `Expect` failures call `std::exit` without unwinding, so the auto-join never fires. It is also the repo's only `jthread`, which libc++ ships non-experimentally only from LLVM 18 while nothing else here needs more than LLVM 17, so keeping `std::thread` avoids raising the Apple toolchain floor for nothing.

4. [`LiveKitBackendTests.cpp:32`](https://github.com/NVIDIA/xr-ai/blob/590f828ee2cca55de18d38f8533cf2a883a160c7/client-samples/native/StreamKit/Tests/StreamKitTests/LiveKitBackendTests.cpp#L32) `CallbackFailure` is a test-only type declared in the library's `namespace streamkit`, where the other test files keep helpers in anonymous namespaces and `streamkit` exceptions are `*Error` types deriving from `StreamError`. Move it to an anonymous namespace.

5. [`LiveKitBackend.cpp:698`](https://github.com/NVIDIA/xr-ai/blob/590f828ee2cca55de18d38f8533cf2a883a160c7/client-samples/native/StreamKit/src/Backends/LiveKit/LiveKitBackend.cpp#L698) `tracks_mutex_` is now taken with `std::scoped_lock` here but with `std::lock_guard<std::mutex>` at its six other sites in this file, so the same mutex uses two idioms. Convert the rest or leave this one as `lock_guard`.

6. [`main.cpp:141-149`](https://github.com/NVIDIA/xr-ai/blob/590f828ee2cca55de18d38f8533cf2a883a160c7/client-samples/native/App/main.cpp#L141-L149) Both optional prints now pair a braced `if` with an unbraced `else`, the exact construct Sonar's brace rule targets. Brace both arms.

The refactor is faithful: the single delivery critical section, the teardown try_lock/thread-local escape, the stop-flag/detach-or-join path, the three-stage Connect re-checks, the shared-ownership delivery state, and member destruction order all survive the field grouping intact. No blockers; this is ready to merge.

blongs-nv (approved):

wenxind-nvidia (approved):

nvddr (approved):

PR #411, opened 2026-08-21
fix(models): let vLLM select NVFP4 MoE backend

nvddr:
## Summary

- remove the Blackwell Omni profile's unsupported `moe_backend: triton` override
- let vLLM automatically select a backend compatible with the NVFP4 checkpoint
- prevent all bundled Omni profiles from reintroducing a forced MoE backend
- document the automatic backend-selection behavior

## Root cause

The Blackwell profile selected the NVFP4 Nemotron Omni checkpoint while also
passing `--moe-backend triton`. vLLM v0.20.0 rejects Triton for NVFP4 MoE before
model startup. With no override, vLLM's default `auto` path probes its supported
NVFP4 backends and selects one compatible with the deployment.

This restores the same narrow configuration fix that existed in the earlier
tea-making branch but was not present in current `main`.

Upstream selection logic:
https://github.com/vllm-project/vllm/blob/v0.20.0/vllm/model_executor/layers/fused_moe/oracle/nvfp4.py

## Validation

- `tests/test_model_servers.py`: 53 passed
- configuration, API documentation, and service-layout tests: 27 passed
- `make -C docs api-check`
- `git diff --check`

GPU model startup was not run locally because this environment does not have
the target Blackwell deployment and checkpoint loaded.

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,8 +211,6 @@ avoiding GPU overcommit.
 The default profile starts Nemotron-3 Nano Omni (8108, serving both the
 reactive and agent LLM roles), Cosmos3 Nano Reasoner (8100), STT (8103),
 and embeddings (8109).
-On Blackwell, the Omni profile leaves the MoE kernel backend unset so vLLM can
-select an implementation compatible with the model's NVFP4 quantization.
 
 ```bash
 uv run model_servers --models vlm_llm_nim
--- a/README.md
+++ b/README.md
@@ -211,6 +211,8 @@ avoiding GPU overcommit.
 The default profile starts Nemotron-3 Nano Omni (8108, serving both the
 reactive and agent LLM roles), Cosmos3 Nano Reasoner (8100), STT (8103),
 and embeddings (8109).
+On Blackwell, the Omni profile leaves the MoE kernel backend unset so vLLM can
+select an implementation compatible with the model's NVFP4 quantization.
 
 ```bash
 uv run model_servers --models vlm_llm_nim
--- a/agent-samples/model-servers/yaml/96G_blackwell/nemotron_omni_llm_server.yaml
+++ b/agent-samples/model-servers/yaml/96G_blackwell/nemotron_omni_llm_server.yaml
@@ -2,8 +2,9 @@
 # SPDX-License-Identifier: Apache-2.0
 
 # Nemotron-3-Nano-Omni — single Blackwell 96 GB.
-# NVFP4 is selected automatically. This standalone Omni stack leaves the
-# remainder of the GPU for STT, rendering, and runtime overhead.
+# NVFP4 is selected automatically. Leave moe_backend unset so vLLM can select
+# a compatible kernel for the checkpoint and hardware. This standalone Omni
+# stack leaves the remainder of the GPU for STT, rendering, and runtime overhead.
 
 host: "0.0.0.0"
 port: 8108
--- a/agent-samples/model-servers/yaml/96G_blackwell/nemotron_omni_llm_server.yaml
+++ b/agent-samples/model-servers/yaml/96G_blackwell/nemotron_omni_llm_server.yaml
@@ -19,8 +19,6 @@ enforce_eager: false
 video_pruning_rate: 0.5
 video_fps: 2
 video_num_frames: 64
-# RTX Pro Blackwell requires Triton for this MoE architecture.
-moe_backend: triton
 
 vllm_backend: docker
 vllm_image: vllm/vllm-openai:v0.20.0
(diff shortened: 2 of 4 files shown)


blongs-nv:
Reviewed at `dd7f3442d0a8d7931be9e11dac50229aa66d1c2c` by a bot.

## Blockers

1. **The rationale for leaving `moe_backend` unset moved into narrative docs, where the repo's change contract says it belongs as a YAML comment.**

   [`docs/source/components/ai-services.md:34`](https://github.com/NVIDIA/xr-ai/blob/dd7f3442d0a8d7931be9e11dac50229aa66d1c2c/docs/source/components/ai-services.md#L34) [`AGENTS.md:106`](https://github.com/NVIDIA/xr-ai/blob/dd7f3442d0a8d7931be9e11dac50229aa66d1c2c/AGENTS.md#L106-L108) keeps sample-config field guidance in YAML comments (generated verbatim into the config reference), and [`AGENTS.md:101`](https://github.com/NVIDIA/xr-ai/blob/dd7f3442d0a8d7931be9e11dac50229aa66d1c2c/AGENTS.md#L100-L103) reserves README and `docs/source/` edits for concept, workflow, or architecture changes, which dropping one tuning key is not. The new sentence also narrates the removed override ("instead of receiving a backend override that may be valid only for other quantization formats"), which [`AGENTS.md:127`](https://github.com/NVIDIA/xr-ai/blob/dd7f3442d0a8d7931be9e11dac50229aa66d1c2c/AGENTS.md#L127-L128) forbids, and asserts a quantization-validity cause the change does not establish. Record the rationale beside where the key was (e.g. extend the "NVFP4 is selected automatically" header comment in [`nemotron_omni_llm_server.yaml:5`](https://github.com/NVIDIA/xr-ai/blob/dd7f3442d0a8d7931be9e11dac50229aa66d1c2c/agent-samples/model-servers/yaml/96G_blackwell/nemotron_omni_llm_server.yaml#L5)) and drop the two doc paragraphs.

## Suggestions

1. The deleted comment claimed "RTX Pro Blackwell requires Triton for this MoE architecture" and nothing in the tree or the PR body rebuts it, so the next person debugging Blackwell MoE failures has no record that the override was tried and rejected. Include the vLLM v0.20.0 rejection error in the PR description and confirm the Blackwell profile reaches healthy without the override.

2. The guard in [`tests/test_model_servers.py:388`](https://github.com/NVIDIA/xr-ai/blob/dd7f3442d0a8d7931be9e11dac50229aa66d1c2c/tests/test_model_servers.py#L388-L402) hardcodes the three profile directories and reads only the base YAML, while the launcher discovers profile dirs dynamically ([`main.py:63`](https://github.com/NVIDIA/xr-ai/blob/dd7f3442d0a8d7931be9e11dac50229aa66d1c2c/agent-samples/model-servers/main.py#L63)) and resolves `<config>_<profile>.yaml` variants ([`main.py:145`](https://github.com/NVIDIA/xr-ai/blob/dd7f3442d0a8d7931be9e11dac50229aa66d1c2c/agent-samples/model-servers/main.py#L145), an idiom already live in `dual_48G_ada/nim_vlm_server_vlm_speech_nim.yaml`), so a new profile dir or an Omni variant file silently escapes the guard. Globbing `yaml/*/nemotron_omni_llm_server*.yaml` makes it self-maintaining.

3. [`test_omni_forwards_configured_moe_backend`](https://github.com/NVIDIA/xr-ai/blob/dd7f3442d0a8d7931be9e11dac50229aa66d1c2c/tests/test_model_servers.py#L640) only covers the key-present case; add the absent case asserting `--moe-backend` is not in `extra_serve_args`, since that omitted-flag launch is the behavior this fix ships and is currently proven only by static YAML inspection.

4. [`README.md:214`](https://github.com/NVIDIA/xr-ai/blob/dd7f3442d0a8d7931be9e11dac50229aa66d1c2c/README.md#L214) "On Blackwell, the Omni profile leaves the MoE kernel backend unset" implies the other profiles set one, but after this change none does. If any prose survives blocker 1, the per-server **nemotron-omni-llm** bullet at [`ai-services.md:500`](https://github.com/NVIDIA/xr-ai/blob/dd7f3442d0a8d7931be9e11dac50229aa66d1c2c/docs/source/components/ai-services.md#L500) is its home, and it should note that `moe_backend` remains a supported YAML knob for user profiles.

Removing an override the pinned image rejects and extending the profile guard to all three Omni profiles is the right call, and the launch path degrades cleanly when the key is absent. The blocker above should be resolved before merge.

wenxind-nvidia (approved):

blongs-nv (approved):

PR #412, opened 2026-08-21
fix(native): make network quality update const

nvddr:
## Summary

- make the private network-quality callback logically const
- mark only the atomic telemetry cache as mutable
- resolve the remaining unsuppressed `cpp:S5817` finding from the exported Sonar report

The other 35 exported findings are already addressed on current `main` by #410, including the deliberate `NOSONAR` annotations for the self-detaching telemetry thread and catch-all callback boundary.

## Validation

- `git diff --check`
- `cmake -S client-samples/native -B /tmp/xr-ai-sonar-native-build -DSTREAMKIT_BUILD_TESTS=ON`
- `cmake --build /tmp/xr-ai-sonar-native-build --parallel`
- `ctest --test-dir /tmp/xr-ai-sonar-native-build --output-on-failure` (6/6 passed)

### DIFF
--- a/client-samples/native/StreamKit/include/streamkit/Backends/LiveKit/LiveKitBackend.h
+++ b/client-samples/native/StreamKit/include/streamkit/Backends/LiveKit/LiveKitBackend.h
@@ -169,7 +169,7 @@ private:
 
     void ApplyConnectionState(ConnectionState state);
     void BlockNetworkMetricsDelivery();
-    void HandleNetworkQualityChange(int lk_quality) const;
+    void HandleNetworkQualityChange(int lk_quality);
     void StartNetworkMetricsReporting();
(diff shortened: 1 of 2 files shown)


blongs-nv (approved):

wenxind-nvidia:
**Suggestion**

`client-samples/native/StreamKit/include/streamkit/Backends/LiveKit/LiveKitBackend.h:214`: `quality` is externally observable state, published by `PublishNetworkMetrics()`. Marking it `mutable` solely to make the state-changing handler `const` weakens const correctness without helping any call site; the delegate already owns a non-const backend pointer. Keep both the handler and atomic non-const.

wenxind-nvidia (approved):

PR #413, opened 2026-08-21
Adopt technical writing standards

yanziz-nvidia:
## Summary
- establish XR AI technical-writing standards
- keep the authoritative style in a dedicated documentation guide
- link to the guide from AGENTS.md and contributor documentation
- revise customer-facing Sphinx pages, the root README, and included SDK READMEs
- preserve literal UI text, paths, endpoints, logs, and established technical terms

## Testing
- strict Sphinx build with warnings treated as errors
- generated public API reference check
- git diff --check

PR #416, opened 2026-08-21
XR render demo perception improvements

blongs-nv:
Perception fixes to the xr-render-demo sample (#345 follow-up). Color requests
that depend on the real world now observe it or fail honestly, with the
guarantees in structure rather than prompt heuristics (per review feedback):

- **Typed color source.** The mutating tools take `color_kind`
  (`literal` | `scene_object` | `physical`) plus `color_value`; the model
  declares the source through the schema and code dispatches on it with no
  phrase reinterpretation. Literal words are repaired ("teel"), scene copies
  resolve against the scene only (failures surface as the ask-back, never a
  camera query), and physical sources ("my shirt", "the wall") run one camera
  observation.
- **Closed observation grammar.** `resolve_physical_color`'s VLM must answer
  `VISIBLE r g b` (each 0..1) or `UNKNOWN`, full-match; prose, hedges, and
  out-of-range replies fail closed. One observation per phrase per turn.
- **Evidence-backed success.** A per-turn record counts scene writes actually
  applied (and recolors that found the color already in place). Without
  evidence, a completion claim is replaced with a fixed honest no-change
  sentence and any other non-question reply gets the no-change fact appended;
  clarifying questions pass through. Evidence is per turn, not per delegated
  task (documented).
- **Transport outages classified by exception type**, including Relay-wrapped
  `FrameUnavailable`/`RPCError` where the cause chain is erased; ordinary
  English ("disabled") and local bugs no longer match.
- **Truncation detection works on unpunctuated input**: leading wh-questions
  and progressive constructions ("the wall I'm staring at") read as complete;
  punctuation, embedded wh-words, and bare -ing nouns ("the ceiling in") do
  not exempt a dangling preposition.
- **Eval/test hardening**: prompt vocabulary disjoint from all eval text
  (audit also scans the model-visible tool text in code), grammar-faithful
  physical-color fake that validates the requested source, and unit matrices
  for source dispatch, the grammar, the evidence gate, and truncation.

## Test plan

Unit 85/85, ruff and prompt audit clean. Offline: subagents 48/49, routing
21/23, utterances 32/34 (remainders are documented pre-existing failures or
accepted probabilistic routing misses). Live: manipulation 13/13, garble
10/10; wall-recolor probe with no camera 8+ consecutive honest no-change
replies (previously false success); with an injected camera frame, recolors
land with the color measured from the frame.

### DIFF
--- a/agent-samples/xr-render-demo/README.md
+++ b/agent-samples/xr-render-demo/README.md
@@ -54,9 +54,9 @@ voice query
        └─ SceneSupervisor.handle()           one turn, per-participant lock
             └─ run_tool_loop (LLM + subagent tools)
                  ├─ make_placement_agent()   SceneTools + TrackingTools
-                 ├─ make_appearance_agent()  SceneTools + physical-color camera tool
+                 ├─ make_appearance_agent()  SceneTools + camera-backed color resolver
                  ├─ make_object_agent()      SceneTools + TrackingTools
-                 │                           + physical-color camera tool
+                 │                           + camera-backed color resolver
                  ├─ make_vision_agent()      CurrentFrameTool + ImageQueryTool
                  │                           + VideoMemoryTools (optional)
                  └─ make_memory_agent()      TextMemoryTools
--- a/agent-samples/xr-render-demo/README.md
+++ b/agent-samples/xr-render-demo/README.md
@@ -54,8 +54,9 @@ voice query
        └─ SceneSupervisor.handle()           one turn, per-participant lock
             └─ run_tool_loop (LLM + subagent tools)
                  ├─ make_placement_agent()   SceneTools + TrackingTools
-                 ├─ make_appearance_agent()  SceneTools
+                 ├─ make_appearance_agent()  SceneTools + physical-color camera tool
                  ├─ make_object_agent()      SceneTools + TrackingTools
+                 │                           + physical-color camera tool
                  ├─ make_vision_agent()      CurrentFrameTool + ImageQueryTool
                  │                           + VideoMemoryTools (optional)
                  └─ make_memory_agent()      TextMemoryTools
--- a/agent-samples/xr-render-demo/eval/README.md
+++ b/agent-samples/xr-render-demo/eval/README.md
@@ -91,7 +91,7 @@ fast gate manually after each edit:
 uv run xr_render_demo_eval utterances
 ```
 
-25 cases, ~3 minutes. The `scenarios` and `precision` tiers catch
+34 cases, a few minutes. The `scenarios` and `precision` tiers catch
 regressions but take longer; run them before calling a tuning round done.
 
 ## Writing a case
@@ -105,8 +105,9 @@ exemplified). Precision and utterances cases are `Case` dataclasses in
 ## Don't train on the test set
 
 Prompt worked examples and case fixtures share the same model, so the
-harness audits every worker prompt at startup (all three offline tiers run
-it) and warns on:
+harness audits every worker prompt plus the model-visible tool text in
+`spatial_ops.py`, `_physical_color.py`, and the agent modules at startup
+(all three offline tiers run it) and warns on:
 
 1. A case utterance from any tier appearing verbatim in a prompt.
 2. A case fixture id appearing in a prompt.
@@ -124,4 +125,5 @@ while the prompt contains its vocabulary is scoring recall, not skill.
 - The live worker pipeline (VAD, STT, TTS, history bookkeeping); the live
   tier covers it.
 - Real scene-service / LOVR effects (fixture-succeeded).
-- Real visual queries (`look_at_current_frame`, `look_at_past_frame`): stubbed.
+- Real visual queries (`look_at_current_frame`, `look_at_past_frame`,
+  `resolve_physical_color`): stubbed.
--- a/agent-samples/xr-render-demo/eval/xr_render_demo_eval/cases.py
+++ b/agent-samples/xr-render-demo/eval/xr_render_demo_eval/cases.py
@@ -1265,8 +1265,8 @@ CASES = [
     },
 
     # ── perception gating: real-world colour must come from the camera ───────
-    # The colour word is never in the utterance; the model must call
-    # look_at_current_frame FIRST and read the colour out of the answer.
+    # The colour word is never in the utterance; a perception tool (either of
+    # PERCEPTION_TOOL) must observe the camera before any mutation.
     # `vlm_answer` is what the mocked camera sees; `must_call_first` fails
     # the case if the model mutates before (or without) looking.
     {
--- a/agent-samples/xr-render-demo/eval/xr_render_demo_eval/cases.py
+++ b/agent-samples/xr-render-demo/eval/xr_render_demo_eval/cases.py
@@ -78,7 +78,7 @@ def _stacked_vertically(mutations: list[tuple[str, dict]]) -> tuple[bool, str]:
     return True, f"stacked at y={ys}"
 
 
-PERCEPTION_TOOL = "look_at_current_frame"
+PERCEPTION_TOOL = ("look_at_current_frame", "resolve_physical_color")
 
 CASES = [
     # ── direct render ops ─────────────────────────────────────────────────────
--- a/agent-samples/xr-render-demo/eval/xr_render_demo_eval/cases.py
+++ b/agent-samples/xr-render-demo/eval/xr_render_demo_eval/cases.py
@@ -78,7 +78,7 @@ def _stacked_vertically(mutations: list[tuple[str, dict]]) -> tuple[bool, str]:
     return True, f"stacked at y={ys}"
 
 
-PERCEPTION_TOOL = ("look_at_current_frame", "resolve_physical_color")
+PERCEPTION_TOOLS = ("look_at_current_frame", "resolve_physical_color")
 
 CASES = [
     # ── direct render ops ─────────────────────────────────────────────────────
@@ -1266,7 +1266,7 @@ CASES = [
 
     # ── perception gating: real-world colour must come from the camera ───────
     # The colour word is never in the utterance; a perception tool (either of
-    # PERCEPTION_TOOL) must observe the camera before any mutation.
+    # PERCEPTION_TOOLS) must observe the camera before any mutation.
     # `vlm_answer` is what the mocked camera sees; `must_call_first` fails
     # the case if the model mutates before (or without) looking.
     {
@@ -1274,7 +1274,7 @@ CASES = [
         "scene": [],
         "user":  "Make a sphere the same color as the thing I'm holding.",
         "vlm_answer": "The user is holding a bright red apple.",
-        "must_call_first": PERCEPTION_TOOL,
+        "must_call_first": PERCEPTION_TOOLS,
         "result": [
             {"tool": "add_primitive",
              "args": {"prim_type": "sphere",
@@ -1287,7 +1287,7 @@ CASES = [
                    "pos": [0.0, 1.6, -1.5], "color": [1, 1, 1], "size": 0.1}],
         "user":  "Make the sphere the same color as my shirt.",
         "vlm_answer": "The user's shirt is blue.",
-        "must_call_first": PERCEPTION_TOOL,
+        "must_call_first": PERCEPTION_TOOLS,
         "result": [
             {"tool": "update_primitive",
              "args": {"obj_id": "sphere-0",
@@ -1299,7 +1299,7 @@ CASES = [
         "scene": [],
         "user":  "Add a cube that matches the color of the wall I'm looking at.",
         "vlm_answer": "The wall is green.",
-        "must_call_first": PERCEPTION_TOOL,
+        "must_call_first": PERCEPTION_TOOLS,
         "result": [
             {"tool": "add_primitive",
              "args": {"prim_type": "box",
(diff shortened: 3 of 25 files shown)


wenxind-nvidia:
I reviewed the cumulative PR at `520e90c` against base `2bf0786`. The current CI suite is green, but I recommend **request changes**.

## Blockers

1. **The physical-color VLM is forced to invent a color when the referenced object is not visible.**

   `_physical_color.py` requires exactly three numbers and provides no `UNKNOWN`/`NOT_VISIBLE` response. `ImageQueryResult.available` only indicates that inference succeeded; it does not mean the requested object was found. An absent or occluded shirt can therefore produce a plausible RGB triple and mutate the scene.

   The named-color fallback compounds this: a refusal such as “I cannot determine whether the wall is white” resolves to white.

   Add an explicit not-visible/ambiguous sentinel, instruct the VLM never to infer typical colors, parse the response strictly, and add tests proving no mutation occurs when the source cannot be grounded.

2. **Physical objects named with renderer shapes can silently resolve to XR objects.**

   `_Leaves.color()` tries `find(color_words)` before the camera fallback. For example, with one XR `cone-0`, “make the ring match the cone I’m holding” copies `cone-0`’s color and never observes the held cone.

   The documentation says the model “marks the source as physical,” but the schema contains only an untyped phrase—there is no such marker. Introduce an explicit source kind or separate scene-color and physical-color paths. Add coverage with both a virtual cone and a differently colored physical cone.

3. **Complete punctuation-free physical-color commands are rejected as truncated.**

   `_is_truncated()` classifies any non-interrogative transcript ending in a preposition as incomplete unless it has terminal punctuation. Consequently, `Make a sphere match what I'm looking at` is rejected before perception runs. Typed input and ASR output cannot be assumed to include punctuation. The test only covers the same construction with a trailing period. Add punctuation-free imperative cases and avoid using punctuation as the deciding signal for subordinate clauses such as “what I’m looking at.”

## Suggestions

- `as_unavailable()` should classify exported exception types such as `xr_ai_hub.FrameUnavailable`, rather than relying on case-sensitive message fragments. The standard “No camera frame available” message is not itself matched unless it retains a timeout cause.

- Strengthen the physical-color eval fake to validate the requested source and support an explicit “not visible” result. It currently returns the configured color for every query, so an incorrect or truncated `source_words` value still passes.

- Remove or rewrite the “Live flail” and “proven live” comments in the eval harness. They record debugging history rather than the invariant being tested, contrary to the repository’s comment conventions.

- Add a PR description and test plan. This is a 656-line behavioral change spanning routing, vision, tool semantics, and evaluation, but the PR body is empty.

## Nits

- Export a public color palette/helper instead of importing private `_COLOR_WORDS` across modules.
- Constrain `ResolvedColor.r/g/b` with `ge=0, le=1`, and give `image_query` an explicit protocol/type.
- Update the stale corpus comment that says `look_at_current_frame` must run first; the accepted gate now includes `resolve_physical_color`.

wenxind-nvidia:
Verdict: **request changes**. Reviewed PR head `ca20fcc` against current `main` (`5d321c8`). The latest commits address several earlier concerns, but four correctness gaps remain.

## Blockers

1. **Verification can still return a false success.**

   After the verification pass, `result2.content` replaces the response without checking whether the second pass called a mutating agent or changed the scene ([`supervisor.py` lines 319–325](https://github.com/NVIDIA/xr-ai/blob/ca20fcc0befbfea59f341190f6a96b38fbf8640b/agent-samples/xr-render-demo/worker/xr_render_demo_worker/supervisor.py#L319-L325)).

   The new test actually returns `"Recolored it to match the wall."` from both passes with no tool calls, but only checks the wording of the nudge—not the returned reply ([test lines 143–163](https://github.com/NVIDIA/xr-ai/blob/ca20fcc0befbfea59f341190f6a96b38fbf8640b/tests/test_xr_render_demo_supervisor.py#L143-L163)). The handler therefore still returns the unsupported success.

   Additionally, `_wants_mutation()` misses verbs such as `match`, `copy`, `duplicate`, `color`, and `stack`, allowing some failed mutation requests to skip verification entirely ([`supervisor.py` lines 66–76](https://github.com/NVIDIA/xr-ai/blob/ca20fcc0befbfea59f341190f6a96b38fbf8640b/agent-samples/xr-render-demo/worker/xr_render_demo_worker/supervisor.py#L66-L76)).

   Please recheck the scene after the second loop and deterministically return a no-change response when the requested mutation remains unverified. The test should assert `reply.response` and final scene state.

2. **Physical-versus-scene color selection is still inferred from an incomplete word heuristic.**

   `_PHYSICAL_CUE_WORDS` treats any first-person pronoun as physical, while omitting explicit terms such as `physical` and `real`; `_physical_clause()` only recovers held/worn/carried shapes ([`spatial_ops.py` lines 54–89](https://github.com/NVIDIA/xr-ai/blob/ca20fcc0befbfea59f341190f6a96b38fbf8640b/agent-samples/xr-render-demo/worker/xr_render_demo_worker/spatial_ops.py#L54-L89)). Consequently:

   - `"the cube I created"` is an XR reference but routes to the camera because it contains `I`.
   - `"the real cone"` routes to the XR cone because neither `real` nor `physical` is recognized.

   With same-shaped physical and XR objects, this can silently apply the wrong color ([`spatial_ops.py` lines 283–301](https://github.com/NVIDIA/xr-ai/blob/ca20fcc0befbfea59f341190f6a96b38fbf8640b/agent-samples/xr-render-demo/worker/xr_render_demo_worker/spatial_ops.py#L283-L301)). Please pass an explicit source kind or expose separate scene-copy and physical-observation operations instead of rediscovering source semantics from the phrase.

3. **The physical-color parser remains fail-open.**

   The VLM contract requires exactly an RGB triple or `UNKNOWN`, but the parser scans arbitrary prose for any triple and then falls back to the last named color ([prompt contract](https://github.com/NVIDIA/xr-ai/blob/ca20fcc0befbfea59f341190f6a96b38fbf8640b/agent-samples/xr-render-demo/worker/xr_render_demo_worker/_physical_color.py#L26-L30), [parser](https://github.com/NVIDIA/xr-ai/blob/ca20fcc0befbfea59f341190f6a96b38fbf8640b/agent-samples/xr-render-demo/worker/xr_render_demo_worker/_physical_color.py#L59-L75)). I confirmed these results:

   - `"The object may be occluded; likely RGB 1 0 0"` → red
   - `"The requested item is outside the frame; the couch is blue"` → blue

   Neither describes a reliably observed requested object. Please accept only an anchored, explicitly allowed RGB response or exact `UNKNOWN`; remove the prose color-word fallback. Add occluded, outside-frame, ambiguous, and unrelated-background-color regressions.

4. **The truncation exemptions classify incomplete commands as complete.**

   Terminal punctuation and the presence of any wh-word currently override a dangling preposition ([`supervisor.py` lines 86–100](https://github.com/NVIDIA/xr-ai/blob/ca20fcc0befbfea59f341190f6a96b38fbf8640b/agent-samples/xr-render-demo/worker/xr_render_demo_worker/supervisor.py#L86-L100)). I confirmed `_is_truncated()` returns `False` for:

   - `"Can you put the sphere on."`
   - `"Can you tell me what color to"`
   - `"Move what I selected to"`

   These are incomplete despite punctuation or an embedded `what`. Please narrow the exemptions to recognized complete tail constructions rather than treating punctuation or any wh-word as sufficient, and add these cases to the wire tests.

## Suggestions

- Rebase or merge current `main` before final validation. The branch is behind PRs #413 and #417. A synthetic merge showed no textual conflicts, but CI has not exercised the combined result.
- Add end-to-end assertions for the claimed guarantees: no scene mutation from an unobserved color and no successful mutation response without a verified scene change.

## Nits

- Replace historical comments such as “defeated earlier versions” and “the live false-success bug” with the invariant or failure mode, per the repository comment conventions.
- `PERCEPTION_TOOL` now contains multiple tool names; rename it to `PERCEPTION_TOOLS`.

All 15 GitHub checks are green, and `git diff --check` is clean. The pure helper probes reproduced blockers 3 and 4; the full focused suite could not be resolved locally on macOS because the tests project includes the Linux-only `pynvvideocodec` package.

nvddr:
I took another pass at this from the abstraction boundary rather than from the individual failing phrases. I would **not** block this PR on making every English construction route perfectly; occasional model mistakes are expected, and a structured prompt plus representative evals is sufficient there. I would block only where the code claims a hard guarantee but currently derives it from free text, lexical cue lists, or an unverified model reply.

| Priority | Area | Why this matters | Durable direction / exit criteria |
|---|---|---|---|
| **Blocker** | **Make color source and observation result typed end to end** | The current path starts with one untyped `color_words` string, then reconstructs meaning through `_PHYSICAL_CUE_WORDS`, `_physical_clause()`, `current_instruction`, fuzzy scene lookup, refusal words, and a prose color fallback ([routing](https://github.com/NVIDIA/xr-ai/blob/ca20fcc0befbfea59f341190f6a96b38fbf8640b/agent-samples/xr-render-demo/worker/xr_render_demo_worker/spatial_ops.py#L54-L89), [dispatch](https://github.com/NVIDIA/xr-ai/blob/ca20fcc0befbfea59f341190f6a96b38fbf8640b/agent-samples/xr-render-demo/worker/xr_render_demo_worker/spatial_ops.py#L271-L315), [VLM parser](https://github.com/NVIDIA/xr-ai/blob/ca20fcc0befbfea59f341190f6a96b38fbf8640b/agent-samples/xr-render-demo/worker/xr_render_demo_worker/_physical_color.py#L26-L75)). Adding more cue or refusal words will keep moving the gaps around. | Give the mutating tool a discriminated internal color reference, for example `LiteralColor`, `SceneObjectColor`, or `PhysicalColor`. Let the subagent choose that variant through the tool schema and a structured prompt; dispatch in code without reinterpreting the phrase. For physical observation, use a closed response grammar such as `VISIBLE <r> <g> <b>` or `UNKNOWN` and require a full match. Any malformed response fails closed. Tests should prove that each variant calls only its intended dependency, a failed scene lookup never falls through to the camera, and `UNKNOWN`/malformed output cannot mutate the scene. The model may occasionally choose the wrong variant; that is an eval-quality issue, not something to repair with another lexical layer. |
| **Blocker** | **Make mutation success an evidence-backed outcome** | The verification pass currently replaces the final answer with `result2.content` without establishing that a mutating tool succeeded or that the requested state was reached ([supervisor](https://github.com/NVIDIA/xr-ai/blob/ca20fcc0befbfea59f341190f6a96b38fbf8640b/agent-samples/xr-render-demo/worker/xr_render_demo_worker/supervisor.py#L302-L325)). The action-verb set is also being used as a proxy for intent. A better nudge cannot make the final success claim trustworthy. | Have mutation subagents return a structured outcome derived from actual tool results, e.g. `changed`, `already_satisfied`, or `rejected`, alongside the user-facing detail. A final response may claim success only from `changed`/`already_satisfied`; no successful outcome means an honest no-change response. The scene diff can corroborate `changed`, but should not be the sole signal because an idempotent request may already be satisfied. Cover four end-to-end cases: no delegation, rejected tool call, already-satisfied state, and an actual mutation. The verb list can remain a prompt/routing hint, but not a correctness boundary. |
| **Lower priority** | **Stop growing the English truncation parser** | `_is_truncated()` is accumulating punctuation rules, wh-word exemptions, and noun exceptions ([code](https://github.com/NVIDIA/xr-ai/blob/ca20fcc0befbfea59f341190f6a96b38fbf8640b/agent-samples/xr-render-demo/worker/xr_render_demo_worker/supervisor.py#L78-L103)). There will always be counterexamples in both directions. These are occasional interaction failures, not data-corruption risks. | Keep deterministic handling only for very high-confidence fragments (for example, a final determiner), and let the supervisor make the semantic ask-back decision under a structured prompt. Alternatively return a typed `needs_clarification` decision from the routing step. Do not add another phrase or exception for each observed miss. |
| **Lower priority / follow-up** | **Preserve typed availability across Relay** | `as_unavailable()` still has to recognize exception class names embedded in text because the Relay boundary erases the original type. The current narrow matching is reasonable locally, but expanding the token list would become the same pattern again. | Move toward a typed unavailable/error result or stable error code at the shared tool/Relay boundary. Until that exists, keep the local matcher deliberately small and treat unmatched failures as real errors. This need not expand the scope of this PR. |
| **Lower priority** | **Test contracts, not an ever-growing phrase catalog** | The new tests are useful, but phrase-by-phrase regressions can lock in the heuristics rather than the intended behavior. | Center unit coverage on the typed source/result and mutation-outcome matrices above. Keep a small, disjoint language eval set for representative ambiguous phrases and accept that it will be probabilistic. |

The earlier punctuation-free command, typed exception, source-validating fake, bounded RGB model, and PR-description concerns are otherwise addressed. The branch is mergeable and all checks on `ca20fcc` are green, but it is behind current `main`; rerun CI after updating it.

nvddr:
Re-reviewed current head `8c0b23d`. The remaining mutation-evidence edge cases are pre-existing, model-dependent failures rather than regressions, so I would not block this PR on them. One concrete blocker remains:

| Priority | Finding | Exit criteria |
|---|---|---|
| **Blocker** | **Undo the unintended public API expansion.** `spatial_ops.__all__` now exports `COLOR_WORDS` and `ColorKind`, but the repository contract says not to broaden public API without an explicit request. The only external consumer added here is the eval harness, which does not justify making these worker implementation details supported API. | Keep the palette and color-kind alias private, or move shared color definitions into a private module used by both the worker and eval harness. Remove both new names from `spatial_ops.__all__`; retain the typed internal dispatch and keep tests/evals green. |

Non-blocking suggestion: as part of the same mechanical cleanup, annotate `resolve_color()` with the private color-kind alias instead of `str`, so static checking preserves the closed set internally without exposing it publicly.

The typed source dispatch, closed VLM response grammar, and associated coverage otherwise address the earlier physical-color concerns. All current checks are green.

wenxind-nvidia (approved):

nvddr (approved):

PR #418, opened 2026-08-21
fix(tea): reject unrelated active-guide queries

nvddr:
## Summary
- keep active tea guidance state unchanged for unrelated requests
- require explicit whole-request intent before calling workflow controls
- add live-model routing coverage for active steps, completed timer state, control-word collisions, and valid controls

## Design note
A prompt-only mode that answered unrelated questions was evaluated but not included. It still selected `workflow__reset` or `workflow__advance` in two of four adversarial cases, so it did not satisfy the state-preservation contract. A reliable answer mode would need a separate intent or control guard and should be designed independently.

## Testing
- `PYTHONPATH=services/device-io-hub .../pytest tests/test_tea_making_sample.py -q` — 37 passed
- `PYTHONPATH=agent-samples/tea-making-sample/worker .../python agent-samples/tea-making-sample/eval/eval.py` — 16 passed against the local Omni endpoint
- Ruff and `git diff --check`

### DIFF
--- a/agent-samples/tea-making-sample/README.md
+++ b/agent-samples/tea-making-sample/README.md
@@ -63,6 +63,29 @@ finalized when its rewrite completes; playback pacing does not delay the Agent
 panel. An urgent interruption may stop audio after the complete intended text
 has already appeared.
 
+## Foreground behavior
+
+While tea guidance is active, foreground requests remain limited to the current
+step, tea-guide status and controls, and the independent background
+applications. An unrelated request receives exactly “I can only help with the
+active tea guide right now.” without calling a tool or changing guide state.
+When guidance is idle, ordinary questions are not subject to that refusal.
+
+## Foreground routing eval
+
+The live-model eval checks idle and active first actions, exact unrelated-query
+refusal, valid guide controls, current-step questions, visual routing, and
+background-application controls. Start `model-servers` so the configured Omni
+LLM endpoint on port 8108 is healthy, then run from the repository root:
+
+```bash
+uv run --project agent-samples/tea-making-sample/worker \
+  python agent-samples/tea-making-sample/eval/eval.py
+```
+
+The command prints one `PASS` or `FAIL` line per case and exits nonzero when a
+tool choice, argument schema, or response contract does not match.
+
 ## File outputs
 
 Background task output is retained under the sample's `artifacts/` directory:
--- a/agent-samples/tea-making-sample/eval/cases.yaml
+++ b/agent-samples/tea-making-sample/eval/cases.yaml
@@ -4,6 +4,10 @@
 - name: idle-conversation
   query: Hello there.
   expected_tool: null
+- name: idle-answers-unrelated-general-knowledge
+  query: What is the capital of France?
+  expected_tool: null
+  forbidden_response: I can only help with the active tea guide right now.
 - name: start-tea-guide
   query: Help me make a cup of tea.
   expected_tool: workflow__start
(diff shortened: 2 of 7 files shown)


wenxind-nvidia:
## Suggestions

1. **Make the rejection eval enforce the promised exact response.**

   The prompt requires the model to reply exactly with the refusal sentence ([prompt](https://github.com/NVIDIA/xr-ai/blob/e8cf09a47cf37c2f9715020d87f6adcc76dba816/agent-samples/tea-making-sample/worker/tea_making_worker/prompts/foreground_prompt.txt#L5-L8)), but every case uses an unanchored substring pattern ([cases](https://github.com/NVIDIA/xr-ai/blob/e8cf09a47cf37c2f9715020d87f6adcc76dba816/agent-samples/tea-making-sample/eval/cases.yaml#L29-L57)) evaluated with `re.search()` ([eval](https://github.com/NVIDIA/xr-ai/blob/e8cf09a47cf37c2f9715020d87f6adcc76dba816/agent-samples/tea-making-sample/eval/eval.py#L167-L173)).

   For example, a response that translates “good morning” and then appends the refusal passes. Only the France case has any forbidden-answer check, and even that only rejects “Paris.” Consider comparing the normalized response with the exact required sentence or using a fully anchored expression for every rejection case. Otherwise, the reported 16/16 result does not prove “decline without answering.”

2. **Add positive active-guide coverage.**

   The new instruction could over-reject legitimate active interactions. Please consider adding cases for:

   - asking what to do at the current step;
   - requesting active workflow status;
   - a step-specific visual question;
   - starting, stopping, or querying a background application while guidance is active.

   The last category is important because production deliberately merges background tools into the active tool set, while the new “unrelated requests” wording could conflict with the following background-control instruction.

3. **Exercise the production foreground path.**

   The eval reconstructs active prompt/tool assembly through private fields and calls `llm.chat()` once with `max_tokens=256`; production uses `_answer()` with `run_tool_loop` and `max_tokens=512`. Prefer exercising the production path, or centralize preparation of the system prompt and tool set so the eval cannot drift.

4. **Document the new live-model eval.**

   `eval/eval.py` adds a new manually operated workflow, but the exact command, model-server prerequisite, and expected endpoint are only in the PR description. Please add a short section to the tea sample documentation.

5. **Refresh CI against current main.**

   The checked merge ref uses main at `9b6bda24`, while current main is `895fefeb`. The newer main also changed `tests/test_tea_making_sample.py`. The merge is conflict-free locally, but checks should be refreshed against the actual merge target.

## Nits

- `_active_route()` can return `tuple[str, ToolSet]` instead of `tuple[str, Any]`.

The case organization, active-step setup, argument-schema validation, and prompt/test pairing otherwise look sound.

wenxind-nvidia (approved):

blongs-nv:
Reviewed at `e8cf09a47cf37c2f9715020d87f6adcc76dba816` by a bot.

## Blockers

1. **The new refusal behavior and the new live-model eval are undocumented.**

   [`agent-samples/tea-making-sample/README.md:82`](https://github.com/NVIDIA/xr-ai/blob/e8cf09a47cf37c2f9715020d87f6adcc76dba816/agent-samples/tea-making-sample/README.md#L82) AGENTS.md requires narrative README/docs updates when behavior or operations change: the Behavior section says nothing about the new mid-guide refusal contract, and nothing in the sample documents how to run `eval/eval.py` (the lab-instrument-monitoring sample documents its equivalent eval in both its README and `eval/README.md`). [`docs/source/reference/tea-making-sample.md:381`](https://github.com/NVIDIA/xr-ai/blob/e8cf09a47cf37c2f9715020d87f6adcc76dba816/docs/source/reference/tea-making-sample.md#L381) also still lists routing evals as a future adaptation step.

## Suggestions

1. The decline rule is a free-standing sentence in a prompt that also serves the idle root route, so an idle model can plausibly answer an unrelated question with "I can only help with the active tea guide right now" when no guide is active, and no root case would catch it because root cases assert no response patterns. Scope it under the "While the tea guide is active" clause ([`foreground_prompt.txt:5-8`](https://github.com/NVIDIA/xr-ai/blob/e8cf09a47cf37c2f9715020d87f6adcc76dba816/agent-samples/tea-making-sample/worker/tea_making_worker/prompts/foreground_prompt.txt#L5-L8), e.g. by moving the new lines directly after the line-3 active-route sentence) and add an idle-route unrelated-question case.

2. Nothing ties the prompt text to the case regexes: rewording the rejection sentence in `foreground_prompt.txt` passes every unit test and only breaks the manually run live eval. Extend [`tests/test_tea_making_sample.py:1316-1339`](https://github.com/NVIDIA/xr-ai/blob/e8cf09a47cf37c2f9715020d87f6adcc76dba816/tests/test_tea_making_sample.py#L1316-L1339) to assert the prompt contains the exact sentence and that each unrelated case's `expected_response_pattern` matches it.

The prompt rule lands with genuinely discriminating eval coverage (the old prompt produces no tool call but can never emit the sentinel sentence, so all four rejection cases fail without the fix), and the eval's workflow-store driving reaches each target step correctly, including the completed-timer state. The blocker above should be resolved before merge.

nvddr:
Disposition for review feedback (511d30fe):

- Made the active-guide refusal an exact response contract after whitespace normalization; all four unrelated active cases now require the full sentence.
- Added positive active-route coverage for current-step answers, workflow status, current-view routing, and starting a background watch, plus an idle general-knowledge case that forbids the active-guide refusal.
- Centralized prompt/tool/route preparation in the private ForegroundAgent._prepare_route helper so the live-model eval exercises the production routing path; aligned its token limit with production.
- Scoped the prompt explicitly to unrelated requests while a tea guide is active, including background applications as valid scope.
- Added README and reference documentation for the behavior, prerequisites, command, and expected eval result.
- Tied prompt wording and eval-case coverage together in the deterministic tests, and resolved the _active_route typing nit by making it state setup only.
- Rebased the branch onto current main before pushing.

Validation: 37 focused tests passed; Ruff passed; strict Sphinx docs build passed. The live-model eval was not run locally because the Omni endpoint on port 8108 was unavailable; the documented command and import environment were validated.

Final CI disposition: all PR checks passed, including DCO, Ruff, SPDX, lock validation, strict docs, Python 3.11 and 3.12 pytest, CodeQL, dependency-map generation, and Pages publishing.

blongs-nv (approved):

PR #419, opened 2026-08-22
docs: make published docs canonical

wenxind-nvidia:
## Summary

- make docs/source the canonical home for user-facing guidance and codify README ownership in AGENTS.md
- give every sample README a useful introduction, lifecycle explanation, prerequisites, and expected outputs
- standardize sample README commands on the sample directory, including starting ../model-servers and then launching the sample from the same terminal
- move detailed package, client, service, sample, evaluation, and testing guidance into published documentation
- align the rewritten guidance with the shared Piper model stack and the repository documentation style
- correct stale commands, configuration guidance, prerequisites, and the xr-render-demo Web-XR vendor path

## Testing

- sample README convention and generated CLI tests (15 passed)
- final-base targeted documentation, layout, and tea tests (79 passed)
- non-GPU suite on the immediately preceding main (1,281 passed)
- strict Sphinx build with warnings treated as errors
- Ruff on changed Python files


nvddr:
## Review disposition: request changes

I audited all 66 changed files at head `28b7299c` against base `cbb6342f`, including command execution context, local links and anchors, configuration behavior, and a deletion ledger for guidance moved out of READMEs. CI is green, but there are multiple semantic and operational failures that the current documentation tests do not exercise.

### Introduced blockers

1. **P1 — `docs/source/reference/agent-sdk-voice.md:43-52`: false lossless transcript guarantee.** The guide says every final transcript is published, but the 32-entry queue drops the oldest transcript on overflow and discards pending transcripts during shutdown. Restore the lossy-delivery and shutdown semantics and qualify “Every.”

2. **P1 — `agent-samples/xr-render-demo/main.py:205-208`: incomplete Web-XR vendor readiness check.** The corrected directory is right, but readiness checks only `cloudxr-sdk.esm.mjs`. `build.sh` creates CloudXR before copying LiveKit, so an interrupted build can leave CloudXR present and `livekit-client.esm.mjs` absent. Every later launch then skips repair and the client receives a permanent 404. Check both outputs before returning.

3. **P2 — six SDK package READMEs link to a nonexistent generated-API source path.** `agent-sdk/xr-ai-{hub,models,runtime,tools,voice,web-events}/README.md` all link to `../../docs/source/reference/python/index.md`. AutoAPI creates `reference/python/` only in Sphinx build output; GitHub and package readers get 404s.

4. **P2 — `docs/source/reference/agent-sdk-tools.md:67-103`: missing required extras.** The canonical example imports `CurrentFrameTool` and `ImageQueryTool`, but the rewrite removed the only instructions to install `xr-ai-tools[frames]` and `[vision]`. A base install lacks NumPy, Pillow, `xr-ai-hub-client`, and `xr-ai-models`, so the example raises `ModuleNotFoundError`.

5. **P2 — `docs/source/reference/agent-sdk-models.md:62-72`: advertised video configuration is unusable as shipped.** The table advertises video for the Cosmos presets without preserving the requirement that `vlm-server` set `max_videos_per_prompt >= 1`; every shipped profile sets it to zero. The rewrite also drops Nemotron Omni’s thinking-disabled-unless-requested default.

6. **P2 — `docs/source/reference/agent-sdk-web-events.md:18-35`: incoherent example.** Removing the enclosing callback left `await ctx.publish(...)` with both `ctx` and `reading` undefined. Restore a function/callback wrapper so the public example is runnable.

7. **P2 — `docs/source/guides/testing.md:16-33`: broken GPU test command.** After `cd tests`, `bash tests/run_local_gpu_tests.sh` resolves to `tests/tests/run_local_gpu_tests.sh`. Use `bash run_local_gpu_tests.sh` or explicitly return to the repository root.

8. **P2 — `docs/source/guides/customizing-model-servers.md:225-240`: broken validate-then-switch workflow.** The first block now runs `cd tests`; the following model-server commands remain repository-root-relative and resolve under `tests/`. Keep validation root-relative or return to root before switching profiles.

9. **P2 — `docs/source/getting_started/quickstart.md:351-356`: broken Hub-only working directory.** The page promises every procedure begins with a `cd` from the repository root, but Hub-only has no reset after the XR procedure and uses `--project services/device-io-hub`. Sequential use resolves inside `agent-samples/xr-render-demo/`.

10. **P2 — `docs/source/reference/xr-render-demo.md:341-370` and `agent-samples/xr-render-demo/eval/README.md:8-19`: live eval tier is no longer runnable from the docs.** Five valid commands were replaced by the literal non-command `xr_render_demo_live_*`: `xr_render_demo_live_smoke`, `xr_render_demo_live_pose_matrix`, `xr_render_demo_live_manip`, `xr_render_demo_live_garble`, and `xr_render_demo_live_explore`. They survive only in `eval/pyproject.toml` and are absent from generated CLI docs.

11. **P2 — `README.md:20-29`: version-routing contradiction.** The README says documentation defaults to the latest release, but all four primary links hard-code `/main/`. Tagged-release readers are sent to potentially incompatible development documentation; use the site root or `/latest/`.

12. **P2 — `client-samples/web-xr-build/README.md:8`: false reproducibility claim.** It calls both bundles pinned, but `package.json` declares `livekit-client: ^2.0.0` and no lockfile is tracked. Exact-pin LiveKit or remove “pinned.”

13. **P2 — `docs/source/getting_started/clients.md:132-143`: missing Apple build prerequisite.** `Package.swift` requires Swift tools 6.2 and visionOS 26, which require Xcode 26, yet the canonical platform-requirements section does not specify that minimum.

14. **P2 — `docs/source/components/ai-services.md:514-517`: incomplete RAG operational migration.** The shortened RAG README points here for startup order and caching, but the destination no longer explains that embedding must start before RAG and the worker, or why `cache_key` must change when a remote endpoint changes its backing model without changing the profile.

### Retained blockers in rewritten canonical content

These existed before PR 419, but are reasserted in changed canonical pages while the PR claims to correct stale commands and prerequisites:

- **P1 — native C++ cannot use the documented connection path** (`docs/source/getting_started/clients.md:201-206`, `client-samples/native/README.md:11-15`). The command defaults to insecure port 7880, `LiveKitConfig.secure` remains false, and the CLI has no `--secure`. There is no CLI invocation for the guide’s required `wss://host:8080` path. Add secure/8080 support or document the direct-7880 exception and firewall/security implications.
- **P1/P2 — Android audio processing modes are documented but not implemented** (`clients.md:83-85,117-119`). The backend distinguishes only `DISABLED`; Voice, Software, and Raw all call `setMicrophoneEnabled(true)` with identical options.
- **P2 — Android Studio minimum is wrong** (`clients.md:89-94`). The guide says Hedgehog, while AGP 8.5.2 requires Koala 2024.1.1 or newer.
- **P2 — DGX Spark profile path is nonexistent** (`docs/source/getting_started/requirements.md:63`). The actual path is `agent-samples/model-servers/yaml/spark/`, not `xr-render-demo/yaml/spark/`.
- **P2 — container registry claim is false** (`requirements.md:37`). It says all checked-in model services use `nvcr.io/nvidia/vllm`, while Nemotron Omni uses `vllm/vllm-openai:v0.20.0`.

### Material information lost during consolidation

The following still-valid guidance was deleted without a canonical replacement, contrary to the new preservation rule in `AGENTS.md`:

- XR Render eval costs, case-name filtering, garble/explore semantics, repeat-three-times and full-log rules, fast-gate size/runtime, case schema, the `anchor_words` code-first tuning precedent, and concrete overlap-audit examples.
- SDK tools’ stronger privacy contract: only controlled Relay headers are forwarded and every image location is redacted from VLM events while provider input is preserved.
- Voice lifecycle contracts: join publication after the gate greeting, duplicate-roster suppression until leave/reconnect, and transcript shutdown cancellation.
- iOS guidance for removing the low-latency entitlement and using the fallback when a team cannot provision it, plus App Store/TestFlight provisioning requirements for main-camera access.
- Simple VLM’s developer source map, the lab/tea explicit native-runtime boundary, native C++’s move-overload performance warning, two useful iOS media diagnostics, and SDK-model testing guidance.

### Verification

- All 16 current CI checks pass.
- `git diff --check` passes.
- The relative-link scan found exactly the six broken generated-API links above; all other changed local paths and anchors in scope resolve.
- Service YAML/runtime changes, token-flow corrections, certificate guidance, sample configuration mappings, and the Web-XR output-directory correction otherwise matched the implementation.

Given the number of newly introduced unusable commands, false behavioral guarantees, dead entry-point links, and material guidance losses, my disposition is **request changes** before merging.

yanziz-nvidia:
Reviewed by yanziz-review-bot

**Summary**  
At the current head, the documentation consolidation still contains three implementation mismatches. Targeted documentation/layout tests pass (23 tests), and 15 CI checks pass with GitHub Pages pending; the advertised coding-agent setup route should be fixed before merge.

| | Finding |
|---|---|
| 🚫 | [`skills/getting-started/SKILL.md:30`](https://github.com/NVIDIA/xr-ai/blob/775e5eff4b666482058b09050153f39cba530f31/skills/getting-started/SKILL.md#L30) — The setup skill sends local users to the root README’s now-deleted Requirements and VRAM section and later says `simple-vlm-example` runs directly, although the new quickstart requires `model-servers` first; update both routes to the versioned requirements and quickstart pages. |
| 💡 | [`docs/source/reference/tea-making-sample.md:227`](https://github.com/NVIDIA/xr-ai/blob/775e5eff4b666482058b09050153f39cba530f31/docs/source/reference/tea-making-sample.md#L227) — The multi-image adaptation guidance tells users to set `--limit-mm-per-prompt`, but the Omni wrapper exposes neither that flag nor a corresponding YAML field and repository policy prohibits ad hoc `vllm serve`; add a wrapper-owned setting before documenting this migration path. |
| 💡 | [`docs/source/reference/simple-vlm-example.md:97`](https://github.com/NVIDIA/xr-ai/blob/775e5eff4b666482058b09050153f39cba530f31/docs/source/reference/simple-vlm-example.md#L97) — The guide says TTS emits one scope for the completed operation, but `_dispatch_sentence()` creates a separate `voice.tts` scope for every synthesized sentence, so document the per-sentence granularity to keep trace interpretation accurate. |
| 🔍 | none |

- `skills/getting-started/SKILL.md:30` — Replace the deleted root-README requirements route with the versioned requirements page and correct the quickstart order so `model-servers` starts before `simple-vlm-example`.
- `docs/source/reference/tea-making-sample.md:227` — Add and document an Omni-wrapper configuration field for the per-prompt image limit, or remove the unsupported adaptation instruction.
- `docs/source/reference/simple-vlm-example.md:97` — State that TTS emits one `voice.tts` scope per synthesized sentence rather than one scope per complete response.

blongs-nv:
Reviewed at `a5dd46628d3c0a36560d79dc52c4ff9429ce8c5c` by a bot.

## Blockers

1. **The canonical install command for the tools extras cannot resolve.**

   [`docs/source/reference/agent-sdk-tools.md:73`](https://github.com/NVIDIA/xr-ai/blob/a5dd46628d3c0a36560d79dc52c4ff9429ce8c5c/docs/source/reference/agent-sdk-tools.md#L73)
   None of `xr-ai-tools`, `xr-ai-models`, or `xr-ai-hub-client` is published to PyPI, and the `[tool.uv.sources]` entries in the package's own `pyproject.toml` are development metadata a consuming project does not resolve. I ran `uv add "xr-ai-tools[frames,vision]"` in a fresh `uv init` project and it failed with `Because there are no versions of xr-ai-tools[frames] ... requirements are unsatisfiable`. Every real consumer declares the extra alongside its own editable path source, as in [`agent-samples/simple-vlm-example/worker/pyproject.toml:18,30`](https://github.com/NVIDIA/xr-ai/blob/a5dd46628d3c0a36560d79dc52c4ff9429ce8c5c/agent-samples/simple-vlm-example/worker/pyproject.toml#L18-L30).

2. **The root README's primary entry points now point at unrendered MyST source.**

   [`README.md:20-27`](https://github.com/NVIDIA/xr-ai/blob/a5dd46628d3c0a36560d79dc52c4ff9429ce8c5c/README.md#L20-L27)
   "Set up with a coding agent" resolves to [`docs/source/getting_started/skills.md:11-15`](https://github.com/NVIDIA/xr-ai/blob/a5dd46628d3c0a36560d79dc52c4ff9429ce8c5c/docs/source/getting_started/skills.md#L11-L15), whose entire payload is a ` ```{literalinclude} /_snippets/agent-setup-prompt.txt ` block, so a reader on GitHub sees the directive instead of the prompt; `quickstart.md` and `architecture.md` similarly carry 13 and 10 inert `{doc}` roles. The links resolve as file paths, so a link checker passes. `skills/getting-started/SKILL.md:25-28` already uses the published `https://nvidia.github.io/xr-ai/<ref>/...` form; use that here.

3. **Every sample README sends readers to a build-time-only stub for "every field and default".**

   [`agent-samples/simple-vlm-example/README.md:41`](https://github.com/NVIDIA/xr-ai/blob/a5dd46628d3c0a36560d79dc52c4ff9429ce8c5c/agent-samples/simple-vlm-example/README.md#L40-L42)
   The whole payload of [`docs/source/reference/configuration.rst`](https://github.com/NVIDIA/xr-ai/blob/a5dd46628d3c0a36560d79dc52c4ff9429ce8c5c/docs/source/reference/configuration.rst#L32) is the `.. xr-ai-config-reference::` directive on line 32, registered only during a Sphinx build, so the 27 generated config listings are absent for a README reader. This is the surviving member of the class you closed by removing the six `reference/python/index.md` links; the other four sites are `agent-samples/{model-servers:44,xr-render-demo:53,tea-making-sample:42,lab-instrument-monitoring:42}/README.md`. [`tests/test_docs_versions.py:131`](https://github.com/NVIDIA/xr-ai/blob/a5dd46628d3c0a36560d79dc52c4ff9429ce8c5c/tests/test_docs_versions.py#L131) entrenches the link without checking the destination renders.

4. **Four service READMEs give repository-root-relative commands without stating the working directory.**

   [`services/embedding-server/README.md:11-14`](https://github.com/NVIDIA/xr-ai/blob/a5dd46628d3c0a36560d79dc52c4ff9429ce8c5c/services/embedding-server/README.md#L11-L14)
   `uv run --project services/embedding-server ...` resolves only from the repository root, but the README lives in `services/embedding-server/` and no prose anchors the directory, so running the block from the README's own location fails. The same shape is in [`services/llama-nemotron-llm/README.md:12-13`](https://github.com/NVIDIA/xr-ai/blob/a5dd46628d3c0a36560d79dc52c4ff9429ce8c5c/services/llama-nemotron-llm/README.md#L12-L13), [`services/nemotron3-nano-llm/README.md:13-14`](https://github.com/NVIDIA/xr-ai/blob/a5dd46628d3c0a36560d79dc52c4ff9429ce8c5c/services/nemotron3-nano-llm/README.md#L13-L14), and [`services/rag-service/README.md:13`](https://github.com/NVIDIA/xr-ai/blob/a5dd46628d3c0a36560d79dc52c4ff9429ce8c5c/services/rag-service/README.md#L13).

5. **The live-eval setup step names a path that does not exist from the working directory the page just pinned.**

   [`docs/source/reference/xr-render-demo.md:372-373`](https://github.com/NVIDIA/xr-ai/blob/a5dd46628d3c0a36560d79dc52c4ff9429ce8c5c/docs/source/reference/xr-render-demo.md#L372-L373)
   Line 347 states "Run all commands from `agent-samples/xr-render-demo/eval/`", but `openxr_service.yaml` lives one level up at `agent-samples/xr-render-demo/yaml/`, so `yaml/openxr_service.yaml` misses. A reader who follows it never sets `allow_sim_pose`, and all five live drivers then run against a stack that rejects simulated pose.

6. **The native client's TLS trust instruction has no mechanism behind it.**

   [`client-samples/native/README.md:19-23`](https://github.com/NVIDIA/xr-ai/blob/a5dd46628d3c0a36560d79dc52c4ff9429ce8c5c/client-samples/native/README.md#L19-L23)
   `LiveKitConfig` exposes no CA path, verification toggle, or trust callback ([`BackendConfiguration.h:57-61`](https://github.com/NVIDIA/xr-ai/blob/a5dd46628d3c0a36560d79dc52c4ff9429ce8c5c/client-samples/native/StreamKit/include/streamkit/Config/BackendConfiguration.h#L57-L61)), and [`docs/source/getting_started/networking.md:148-165`](https://github.com/NVIDIA/xr-ai/blob/a5dd46628d3c0a36560d79dc52c4ff9429ce8c5c/docs/source/getting_started/networking.md#L148-L165) documents trusting the hub certificate only for Chrome/Edge, Firefox, Android, and iOS. A reader following the now-primary `--port 8080 --secure` command has no documented way to complete the trust step on a Linux host.

7. **The web-events example still publishes nothing.**

   [`docs/source/reference/agent-sdk-web-events.md:19-35`](https://github.com/NVIDIA/xr-ai/blob/a5dd46628d3c0a36560d79dc52c4ff9429ce8c5c/docs/source/reference/agent-sdk-web-events.md#L19-L35)
   `report_reading` is defined and never called, and `ctx` cannot be obtained where the example places it: a `RuntimeContext` is supplied by the runtime to `@subscribe`-decorated agent methods, as in [`web_events.py:38-41`](https://github.com/NVIDIA/xr-ai/blob/a5dd46628d3c0a36560d79dc52c4ff9429ce8c5c/agent-samples/lab-instrument-monitoring/worker/lab_instrument_monitoring_worker/web_events.py#L38-L41). Pasted as written the block runs and emits no web event.

8. **The new sample-README guard cannot fail for the regression it is named after.**

   [`tests/test_docs_versions.py:113-135`](https://github.com/NVIDIA/xr-ai/blob/a5dd46628d3c0a36560d79dc52c4ff9429ce8c5c/tests/test_docs_versions.py#L113-L135)
   All ten checks are unanchored substring tests over the whole file, tied to no fenced block, heading, or each other. Running a byte-identical replica of the assertion body, a README whose only occurrences of the required strings sit inside an HTML comment marked "do not follow", and whose real instructions are `cd ../..` followed by `uv run --directory agent-samples/xr-render-demo xr_render_demo`, passes every assertion. Parse the fenced `bash` blocks and track the declared working directory instead.

9. **Twelve published heading fragments were reworded away without the compatibility anchors the style guide requires.**

   [`docs/source/getting_started/clients.md:13-248`](https://github.com/NVIDIA/xr-ai/blob/a5dd46628d3c0a36560d79dc52c4ff9429ce8c5c/docs/source/getting_started/clients.md#L13-L248)
   `#which-clients-exist`, `#the-connect-flow`, `#self-signed-certificate-trust`, `#build-and-run`, `#connect`, `#android-xr` and five more are gone, as is `#worker-configuration` in [`docs/source/reference/xr-render-demo.md:102`](https://github.com/NVIDIA/xr-ai/blob/a5dd46628d3c0a36560d79dc52c4ff9429ce8c5c/docs/source/reference/xr-render-demo.md#L102). [`docs/source/guides/documentation-style.md:19-21`](https://github.com/NVIDIA/xr-ai/blob/a5dd46628d3c0a36560d79dc52c4ff9429ce8c5c/docs/source/guides/documentation-style.md#L19-L21) requires preserving a reworded heading's previous fragment with an explicit compatibility anchor, which this PR does correctly at [`tea-making-sample.md:447`](https://github.com/NVIDIA/xr-ai/blob/a5dd46628d3c0a36560d79dc52c4ff9429ce8c5c/docs/source/reference/tea-making-sample.md#L447). `clients.md` is the page every rewritten client README now points into, and `conf.py` sets no linkcheck, so the strict build cannot catch it.

## Suggestions

1. Add the mechanical guards the hand-fixed defects fall into: a relative-link resolver, a `uv run <name>` cross-check against the union of all `[project.scripts]`, and a working-directory tracer over fenced `bash` blocks. I ran all three over the changed files at this SHA and each reports zero hits, so they land green and lock the fixes in.

2. [`tests/test_docs_versions.py:114-120`](https://github.com/NVIDIA/xr-ai/blob/a5dd46628d3c0a36560d79dc52c4ff9429ce8c5c/tests/test_docs_versions.py#L114-L120) hardcodes five sample directories, so a sixth sample is silently exempt from the contract this PR codifies. Derive the list the way [`tests/test_config_documentation.py:18-19`](https://github.com/NVIDIA/xr-ai/blob/a5dd46628d3c0a36560d79dc52c4ff9429ce8c5c/tests/test_config_documentation.py#L18-L19) does.

3. [`agent-samples/xr-render-demo/main.py:226-232`](https://github.com/NVIDIA/xr-ai/blob/a5dd46628d3c0a36560d79dc52c4ff9429ce8c5c/agent-samples/xr-render-demo/main.py#L226-L232) logs "Web vendor bundle ready" without re-checking the two outputs after `build.sh` exits 0, and the new test's `fake_run` always creates `livekit-client.esm.mjs`, so that state is unreachable. Verify both files after the build and add a third phase to [`test_xr_render_repairs_an_incomplete_web_xr_vendor_bundle`](https://github.com/NVIDIA/xr-ai/blob/a5dd46628d3c0a36560d79dc52c4ff9429ce8c5c/tests/test_service_layout.py#L204-L227).

4. [`AGENTS.md:104-105`](https://github.com/NVIDIA/xr-ai/blob/a5dd46628d3c0a36560d79dc52c4ff9429ce8c5c/AGENTS.md#L104-L105) keeps the generated CLI reference as the owner of sample invocations, and [`docs/source/_cli_reference.py:100-103`](https://github.com/NVIDIA/xr-ai/blob/a5dd46628d3c0a36560d79dc52c4ff9429ce8c5c/docs/source/_cli_reference.py#L100-L103) emits them as `uv run --project agent-samples/<dir> <program>` — the exact string [`tests/test_docs_versions.py:126`](https://github.com/NVIDIA/xr-ai/blob/a5dd46628d3c0a36560d79dc52c4ff9429ce8c5c/tests/test_docs_versions.py#L126) now forbids in READMEs. Both commands work, but two canonical surfaces teach opposite shapes; scope the new rule to READMEs explicitly or teach the generator the sample-directory form.

5. The README-ownership rule is now stated in four places and has already drifted: [`docs/source/guides/contributing-and-conventions.md:34-42`](https://github.com/NVIDIA/xr-ai/blob/a5dd46628d3c0a36560d79dc52c4ff9429ce8c5c/docs/source/guides/contributing-and-conventions.md#L34-L42) says configuration guidance belongs in the sample's reference guide where [`AGENTS.md:145-148`](https://github.com/NVIDIA/xr-ai/blob/a5dd46628d3c0a36560d79dc52c4ff9429ce8c5c/AGENTS.md#L145-L148) requires it in the README, and `CONTRIBUTING.md:29-33` keeps a third copy of the test commands with no link to the new canonical testing page. Have the others link the contract file rather than paraphrase it.

6. [`client-samples/STREAMKIT_PARITY.md:17`](https://github.com/NVIDIA/xr-ai/blob/a5dd46628d3c0a36560d79dc52c4ff9429ce8c5c/client-samples/STREAMKIT_PARITY.md#L17) still says Android "relies on the `MicrophoneMode` preset" for DSP and tags the divergence **idiom**, which the audio-mode collapse in this PR makes false. Update the row or retag it **to-fix**.

7. [`docs/source/getting_started/networking.md:16`](https://github.com/NVIDIA/xr-ai/blob/a5dd46628d3c0a36560d79dc52c4ff9429ce8c5c/docs/source/getting_started/networking.md#L16) calls 8080 the single entry point "for browser, Android, iOS, and visionOS clients", but the native C++ sample is now documented on 8080 too. Add it to that list.

8. [`AudioConfig.SOFTWARE`](https://github.com/NVIDIA/xr-ai/blob/a5dd46628d3c0a36560d79dc52c4ff9429ce8c5c/client-samples/android/app/src/main/java/com/nvidia/xrai/streamkitsample/streamkit/config/AudioConfig.kt#L63-L66) and `AudioConfig.RAW` now have zero callers anywhere in the repository, and [`MainActivity.kt:624-628`](https://github.com/NVIDIA/xr-ai/blob/a5dd46628d3c0a36560d79dc52c4ff9429ce8c5c/client-samples/android/app/src/main/java/com/nvidia/xrai/streamkitsample/MainActivity.kt#L624-L628) leaves a dropdown with a single option. Either collapse the enum or drop the picker.

9. [`docs/source/reference/agent-sdk-voice.md:55-57`](https://github.com/NVIDIA/xr-ai/blob/a5dd46628d3c0a36560d79dc52c4ff9429ce8c5c/docs/source/reference/agent-sdk-voice.md#L55-L57) credits DeviceIOHub with suppressing duplicate roster joins, but the hub forwards every participant event and deliberately replays a join per connected participant on each roster request ([`ipc/_hub.py:516-533`](https://github.com/NVIDIA/xr-ai/blob/a5dd46628d3c0a36560d79dc52c4ff9429ce8c5c/services/device-io-hub/device_io_hub/ipc/_hub.py#L516-L533)); the dedup lives in `ProcessorEndpoint` ([`_processor.py:708-711`](https://github.com/NVIDIA/xr-ai/blob/a5dd46628d3c0a36560d79dc52c4ff9429ce8c5c/agent-sdk/xr-ai-hub/xr_ai_hub/_processor.py#L708-L711)), as that hub docstring itself says. Attribute it to the endpoint.

10. [`services/rag-service/README.md:14`](https://github.com/NVIDIA/xr-ai/blob/a5dd46628d3c0a36560d79dc52c4ff9429ce8c5c/services/rag-service/README.md#L14) points `--config` at `agent-samples/my-agent/yaml/rag_service.yaml`, which does not exist, after the shrink deleted the prose that explained it was a placeholder. Use a real path or restore the explanation.

Making `docs/source/` canonical is the right call, the restructure is coherent across all 29 READMEs, the strict Sphinx build is clean with zero warnings, and the vendor-bundle regression test is genuinely good work. The blockers above should be resolved before merge.

nvddr:
Second-pass disposition on `a5dd46628d3c0a36560d79dc52c4ff9429ce8c5c`: **request changes**.

I re-audited the full base-to-head documentation diff and deletion ledger, split across samples/getting-started, SDK/services, and clients/tooling, then reconciled the results against the implementation. All current CI checks are green, but the following user-facing blockers remain:

1. `docs/source/reference/agent-sdk-tools.md:69-74` gives `uv add "xr-ai-tools[frames,vision]"`, but `xr-ai-tools` is an internal local package and a fresh consumer cannot resolve it. I reproduced the exact command in an empty `uv` project; resolution fails. The example needs an explicit local source/path and cwd.
2. `docs/source/getting_started/clients.md:208-223` now correctly routes native C++ through `--secure --port 8080`, but neither it nor the networking guide gives native/Linux users an actionable procedure for installing the default self-signed hub certificate. The native config has no CA-path or trust callback, so the documented primary flow stops at “trust the certificate.”
3. `README.md:20-27` links the primary coding-agent setup to raw MyST source. `docs/source/getting_started/skills.md` supplies the prompt only through `literalinclude`, so GitHub displays the directive rather than the prompt users are told to paste. Other raw source targets also contain inert `{doc}` roles.
4. The Configure sections in all five top-level sample READMEs link raw `docs/source/reference/configuration.rst` while promising a generated field/default catalog. On GitHub the target contains only the unexpanded `xr-ai-config-reference` directive. Four READMEs additionally call hardware-specific example values “defaults,” contrary to the reference itself.
5. `docs/source/reference/xr-render-demo.md:347,370-373` establishes `agent-samples/xr-render-demo/eval/` as the cwd, then tells users to edit `yaml/openxr_service.yaml`. The valid path from that cwd is `../yaml/openxr_service.yaml`; otherwise `allow_sim_pose` remains disabled and live evaluations fail.
6. The embedding, Llama, Nano, and RAG service READMEs replaced their self-contained `cd` quickstarts with repo-root-relative `--project services/...` commands but never establish the repository root as cwd. Running them from their service directories resolves nonexistent nested paths.
7. `services/rag-service/README.md:12-15` presents nonexistent `agent-samples/my-agent/yaml/rag_service.yaml` as its only command. The deleted base text explained that users must copy the reference YAML and update `documents_dir` and `models_config`; those prerequisites need to be restored.
8. The embedding, Llama, Nano, and video-memory READMEs claim their exact standalone YAML fields are generated. `docs/source/_config_reference.py` enrolls only sample-owned configurations under `agent-samples`; standalone Llama and Nano YAMLs are absent, while the embedding/video entries are different sample-owned copies.
9. Adding generic NIM to the service inventories makes two blanket claims false: `services/README.md:29-31` says model servers expose OpenAI-compatible HTTP even though speech NIM inference is gRPC, and `docs/source/components/ai-services.md:35-38` says all weights use `model_cache` although NIM uses `nim_cache`.
10. `services/llama-nemotron-llm/README.md:16-18` promises model-swap guidance at the destination, but the destination has no swap procedure. The consolidation also drops the compatibility/tuning changes needed for a swap and the operationally important default that no-toggle behavior reasons by default with an unstripped preamble.
11. The rewrite removes eleven established fragments from `clients.html` and `worker-configuration` from the XR reference without compatibility anchors. This conflicts with the new `documentation-style.md` rule that published fragments are stable.
12. The new ownership contracts disagree: `AGENTS.md:136-149` requires a compact configuration entry point in every top-level sample README, while `contributing-and-conventions.md:34-42` says sample configuration guidance belongs in the reference guide without preserving the compact-versus-exhaustive distinction.

Lower-severity information gaps: the `nemotron3_nano` thinking-disabled preset default is omitted; video-memory disabled-mode behavior was lost; the web-events 16 KiB statement applies to `payload`, not the full serialized event; and the new documentation tests use unanchored substring checks that do not validate cwd placement or rendered directive content.

I also found current defects that predate this PR and am not attributing them to this change: voice roster deduplication is assigned to DeviceIOHub instead of `ProcessorEndpoint`; the web-events example never calls/registers `report_reading`; troubleshooting requires host NVCC and names nonexistent `xr-render-demo/yaml/96G_blackwell/`; the XR reference still describes an `os.execvp` vLLM shim rather than persistent Docker supervision; arbitrary `--models` paths are described as repository-relative although they resolve from the shell cwd; and the Android parity table still claims `MicrophoneMode` controls DSP.

The previous findings around client commands, Apple/Android prerequisites, vendor-bundle repair, voice queue-loss semantics, Cosmos video configuration, live eval command restoration, iOS fallbacks, model-server validation cwd, and RAG runtime behavior are fixed in this head.

Validation: exact base/head deletion audit, relative-link scan, heading-fragment audit, `git diff --check`, 24 targeted documentation/service-layout tests, implementation checks, and a final refresh confirming all 16 GitHub checks are green.

yanziz-nvidia:
Reviewed by yanziz-review-bot

**Summary**  
Other than existing comments that are not addressed yet, this pass found one new publishing blocker: the rendered links target pages absent from the release used for `/latest/`. All 16 CI checks and 28 targeted local tests pass; the new Linux certificate procedure also has an incomplete regeneration sequence.

| | Finding |
|---|---|
| 🚫 | [`.github/workflows/docs.yaml:111`](https://github.com/NVIDIA/xr-ai/blob/666394e51e3220c814f34563e6af4e3cb650db2c/.github/workflows/docs.yaml#L111) — Non-PR publishing copies the existing release tag (`v0.1.0-beta`) into `/latest/`, but the READMEs now link pages introduced after that tag, so representative setup, quickstart, sample, and configuration URLs currently return 404 and remain broken after a main-branch publish; ensure the selected source contains every linked page or target a published ref that does. |
| 💡 | [`docs/source/getting_started/networking.md:189`](https://github.com/NVIDIA/xr-ai/blob/666394e51e3220c814f34563e6af4e3cb650db2c/docs/source/getting_started/networking.md#L189) — The procedure installs the certificate before instructing users to add a SAN, but SAN regeneration creates a new certificate and key, leaving the native host trusting the obsolete certificate; configure SANs before copying the certificate or explicitly require copying and installing it again after regeneration. |
| 🔍 | none |

- `.github/workflows/docs.yaml:111` — Make `/latest/` contain every page targeted by the new README links and add a build-time check that those files exist in the selected source.
- `docs/source/getting_started/networking.md:189` — Move SAN configuration before certificate copying or tell users to copy and reinstall the newly generated certificate after restarting the hub.

nvddr:
Third-pass disposition on `d29cda99616543638f3aa943425cb4fde69956da` (base `986ccaebb97083385e820e8f3a57819c0a85b2ea`): **request changes**.

I re-audited all 81 changed files and deletions, split independently across samples/getting-started, SDK/services, and clients/tooling/publishing, then reconciled the results against the implementation and rendered preview. The latest commits fix almost every second-pass blocker, but the following issues remain.

## Remaining blockers

1. **Published fragment compatibility is still broadly broken.** `docs/source/guides/documentation-style.md:19-21` requires published fragments to remain stable.

   - The old Agent SDK reference pages included package README headings. The rewrites remove 18 generated fragments and two explicit aliases without compatibility anchors. Confirmed examples include `agent-sdk-hub.html#frames`, `agent-sdk-models.html#profile-contract`, `agent-sdk-tools.html#native-tools-and-model-tool-calls`, and `agent-sdk-voice.html#multiple-voice-producers`: they exist on the published pre-PR `main` build and are absent from the PR preview. The full losses are hub `subscription-and-roster-contract`, `frames`, `return-path`, `readiness`; models `contract`, `quickstart`, `profile-contract`, `deployment-profiles`, `protocols`, `remote-and-hosted-nim-endpoints`, `riva-grpc-speech-nim-stttts`, plus aliases `remote--hosted-nim-endpoints` and `remote-hosted-nim-endpoints`; tools `native-tools-and-model-tool-calls`, `typed-capability-services`, `image-selection-and-vlm-query-tools`, `magenta-polygon-image-editing`; and voice `usage`, `multiple-voice-producers`, `voice-tuning-and-data-echo`.
   - A mechanical base/head audit also finds approximately 157 removed GitHub README heading fragments without aliases. This includes 47 in root/sample/skill READMEs and 62 in client READMEs. Representative losses include `README.md#quickstart`, `tea-making-sample/README.md#configuration`, `client-samples/native/README.md#running-the-tests`, and `client-samples/web-xr-build/README.md#bumping-the-cloudxr-sdk-version`. There are no surviving internal references, but public bookmarks and external links break. The new fragment regression test covers only selected physical Sphinx headings, so it misses both classes.

2. **The native networking guide contradicts the shipped executable.** `docs/source/getting_started/networking.md:16,124-126` calls port 8080 the single/only path native C++ uses and says clients never reach 7880 directly. `client-samples/native/App/main.cpp:42-71` still defaults to insecure port 7880, and `client-samples/native/README.md:24-26` explicitly documents that direct debugging path. Qualify the networking text as the supported external path or change the executable default.

3. **The XR evaluation reference is inaccurate.** `docs/source/reference/xr-render-demo.md:445` says the utterance gate has 25 cases, while the current `UTTERANCES` tuple contains 35. The same page says unknown filter names fail, but the harness rejects them only when every supplied name is unknown; a valid name plus a typo silently skips the typo and can create false confidence in a filtered run.

4. **The generic NIM inventory conflicts with the shared YAML instruction.** `docs/source/components/ai-services.md:546-549` says every service has a sample YAML in its service directory and instructs users to copy it and edit `model_cache`. `services/nim-server/` has no YAML; its configurations live under `agent-samples/model-servers/yaml/<gpu-profile>/` and use `nim_cache`. The page's own NIM section correctly describes that different layout.

5. **The safe tool-loop recovery contract was lost.** The old tools guide documented that `ToolLoopError.messages` is a valid resumable transcript, the rejected turn remains separately available as `rejected_response`, malformed/duplicate IDs are rejected before batch execution, and unary side-effect tools produce model-visible `null`. `docs/source/reference/agent-sdk-tools.md:48-58` now says only that typed errors carry a partial audit. AutoAPI does not restore these dynamically assigned recovery fields, so callers no longer have the published contract needed to recover safely.

## Remaining preservation gaps

These remain contrary to the consolidation's no-information-loss contract:

- `agent-sdk-models.md:55` no longer enumerates accepted compatibility formats such as direct role mappings, legacy `health_check`, or `kind: preset:<name>`, although `adding-a-sample.md:58-60` promises those formats there.
- The exact searchable iOS error `io.livekit.swift-sdk Code=101 "Timed out"` after approximately five seconds and the expected microphone-activation sound after XR exit remain absent; only the general mechanism is preserved.
- Tea's eval output contract—one `PASS`/`FAIL` line per case and nonzero exit on mismatch—was deleted.
- The native C++ test coverage table and targeted/rerun CTest commands were not moved.
- The iOS bundle-ID/UserDefaults reset behavior and Web-XR's implemented IsaacTeleop tarball reuse path remain undocumented.

## Lower-severity gaps

- `.github/scripts/check_latest_docs_links.py` scans only `README.md` files, and the docs workflow path filters omit `CONTRIBUTING.md`, even though `CONTRIBUTING.md:29` now owns a `/latest/` documentation link.
- `docs/source/_static/root-redirect.html` still calls `/latest/` a “release” even when the new fallback intentionally serves `main` or PR `current`.

## Confirmed fixed

The new `/latest/` fallback is sound for the current repository. `v0.1.0-beta` lacks required current targets, so it is rejected and the complete `main/current` build is copied. The Linux SAN regeneration/install order is also fixed.

The revised tools TOML is runnable: I reproduced it exactly and `uv sync` successfully resolved the transitive local `xr-ai-logging` source. I also rejected a suspected `connect-1`/`build-and-run-1` fragment regression after checking the actual published pre-PR HTML; those duplicate headings used generic `id1`/`id2`, not the alleged suffixed fragments.

The earlier raw-MyST links, build-only configuration links, service cwd problems, RAG placeholder, NIM protocol/cache claims, Llama swap guidance, web-events example, voice attribution, native trust procedure, XR live YAML path, contributor ownership conflict, and initial Sphinx anchor set are fixed in this head.

## Validation

- Exact base/head per-line and deletion audit across all 81 changed files
- Current rendered preview and published pre-PR fragment checks
- Release-tree and `/latest/` fallback checks
- `git diff --check`
- 29 targeted documentation/layout tests locally, plus the broader TLS/model suites in the independent audits
- Exact tools-consumer `uv sync` reproduction
- Final refresh: all 16 GitHub checks green

Despite green CI, the remaining compatibility breaks, documentation/implementation contradictions, and material recovery/operational guidance losses should be resolved before merge.

nvddr:
Fourth-pass disposition on `db92246cd5e1870ee5c154dc81618a51028a20ae` (base `986ccaebb97083385e820e8f3a57819c0a85b2ea`): **request changes**.

I re-audited the full documentation consolidation after the latest fixes, including every changed/deleted documentation hunk, the implementation behind operational claims, exact GitHub-rendered legacy fragments, the published preview, and focused tests. Most third-pass findings are resolved, but four blockers remain.

## Remaining blockers

### 1. Port 7880 is security-relevantly misdocumented

[`networking.md`](https://github.com/NVIDIA/xr-ai/blob/db92246cd5e1870ee5c154dc81618a51028a20ae/docs/source/getting_started/networking.md#L13) says LiveKit is bound to `127.0.0.1`, and line 30 says it stays there. The implementation does not enforce that:

- [`_LIVEKIT_YAML`](https://github.com/NVIDIA/xr-ai/blob/db92246cd5e1870ee5c154dc81618a51028a20ae/services/device-io-hub/device_io_hub/transport/livekit/_docker.py#L42) does not set `bind_addresses`.
- The container runs with [`--network host`](https://github.com/NVIDIA/xr-ai/blob/db92246cd5e1870ee5c154dc81618a51028a20ae/services/device-io-hub/device_io_hub/transport/livekit/_docker.py#L109).
- LiveKit v1.13.5 maps missing bind addresses to an empty host and listens on all interfaces ([upstream source](https://github.com/livekit/livekit/blob/v1.13.5/pkg/service/server.go#L525-L533)).
- The repository's connector contract explicitly supports LAN clients at `ws://<host>:7880`.

The changed text also permits trusted-LAN native debugging, which is impossible if the port really stays on loopback. [`clients.md`](https://github.com/NVIDIA/xr-ai/blob/db92246cd5e1870ee5c154dc81618a51028a20ae/docs/source/getting_started/clients.md#L38) blanket-prohibits direct 7880 connections while its native section permits them.

This can cause operators to under-firewall a plaintext signaling endpoint. Either bind LiveKit explicitly to loopback and remove remote 7880 instructions, or document its actual reachability and provide source-restricted firewall guidance.

### 2. Fifteen legacy GitHub fragments remain broken

The new compatibility aliases and their test expectations normalize punctuation to one hyphen, but GitHub's actual renderer preserves repeated separators. I compared the exact base and head rendered HTML; these published fragments are still missing:

- Root README: `simple-vlm-example-vision-qa-over-voice--text`, `step-1--start-the-server`, `step-2--connect-a-client`, `lab-instrument-monitoring-marker-associated-readings--foreground-voice`, `step-1--start-model-servers-once`, `step-2--start-the-demo`
- iOS README: `6-visionos-passthrough-camera--device-only`, `connection-fails-with-errsslbadcert---1202-after-the-cert-is-trusted`, `mic--camera-go-dead-while-the-ui-still-says-on`, `cert--trust-notes`
- Native README: `streamkit-for-native-c--livekit-backed-client`
- Web-XR build README: `web-xr-build--web-vendor-bundles`
- XR-render README: `add-a-scene-tool--function-group`
- Llama README: `reasoning-toggle--per-turn-via-system-prompt`
- Tests README: `gpu--docker--nvenc-tests`

The new [`test_consolidated_readme_headings_keep_github_compatibility_anchors`](https://github.com/NVIDIA/xr-ai/blob/db92246cd5e1870ee5c154dc81618a51028a20ae/tests/test_docs_versions.py#L266) passes because its hard-coded expected IDs contain the same normalized mistakes; it does not establish equality with the old rendered IDs. Replace those aliases and expectations with the exact published fragments above.

### 3. The restored tool-loop section introduces a nonexistent Relay behavior

[`agent-sdk-tools.md`](https://github.com/NVIDIA/xr-ai/blob/db92246cd5e1870ee5c154dc81618a51028a20ae/docs/source/reference/agent-sdk-tools.md#L62) claims that “NeMo Relay AutoAPI exception serialization” does not reconstruct dynamically assigned fields. NeMo Relay 0.7.3 has no AutoAPI or exception-serialization surface. Sphinx `autoapi.extension` is this repository's API-documentation generator, while [`ToolLoopError`](https://github.com/NVIDIA/xr-ai/blob/db92246cd5e1870ee5c154dc81618a51028a20ae/agent-sdk/xr-ai-tools/xr_ai_tools/tool_calling.py#L63) is a local `RuntimeError` subclass.

The preceding recovery guidance is valid. Delete the Relay sentence, or accurately describe the Sphinx documentation limitation if that was intended.

### 4. Service consolidation still deletes current, unique guidance

Neither `llama-nemotron-llm` nor `nemotron3-nano-llm` was removed. Both retain launchers, YAML, package entry points, model presets, GPU integration tests, service inventory entries, and canonical documentation. Nemotron 3 is superseded as the default by Omni but remains a supported standalone/custom-profile service; Llama Nemotron remains an optional standalone service.

The consolidation deletes the only repository copies of:

- Llama Nemotron's current model-card sampling guidance (`temperature=0.6`, `top_p=0.95` for reasoning-on; greedy reasoning-off), governing-license references, commercial-use statement, and authoritative model-card link.
- Nemotron 3 Nano's current sampling guidance (`1.0/1.0` for reasoning, `0.6/0.95` for tool calling), governing-license reference, commercial-use statement, and authoritative model-card link.

The current NVIDIA cards still contain that information: [Llama-3.1-Nemotron-Nano-8B-v1](https://huggingface.co/nvidia/Llama-3.1-Nemotron-Nano-8B-v1) and [NVIDIA-Nemotron-3-Nano-30B-A3B-NVFP4](https://huggingface.co/nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-NVFP4). Current [`ai-services.md`](https://github.com/NVIDIA/xr-ai/blob/db92246cd5e1870ee5c154dc81618a51028a20ae/docs/source/components/ai-services.md#L475) preserves neither set.

Restore the authoritative links, current license references, and current sampling recommendations. The stale Nemotron BF16-on-Hopper/Ampere advice should remain deleted; current code selects FP8 there.

## Additional preservation notes

- The deleted Apple guide stated that native `auto-native` CloudXR does not use port 48322, requires no additional certificate installation, and uses CloudXR transport encryption. The consolidated guide only says “native transport,” while networking lists 48322 without this qualification.
- The direct native test-binary command was deleted. CTest filtering and rerun-failure commands are now restored, so this is a minor convenience loss.
- Workflow comments/notices still say “current README” although latest-version selection now also covers `CONTRIBUTING.md`; behavior is correct.

## Resolved since the third pass

- XR utterance count is now 35 and mixed valid/unknown filters fail.
- Exact iOS timeout and post-XR microphone-sound diagnostics are preserved.
- Tea eval PASS/FAIL and nonzero-exit behavior is preserved.
- NIM YAML location/cache guidance, model compatibility formats, and the substantive tool-loop recovery contract are restored.
- Native test coverage, iOS bundle-ID behavior, and IsaacTeleop CloudXR cache reuse are restored.
- `/latest/` selection, `CONTRIBUTING.md` coverage, root-redirect wording, Linux native certificate ordering, and Sphinx compatibility aliases are fixed.
- The prior `xr-ai-logging` dependency allegation remains retracted; exact `uv sync` succeeds.

## Verification

- All 16 GitHub checks at this head pass, including strict docs, Python 3.11/3.12, DCO, lint, lock, SPDX, CodeQL, Android, and Pages publication.
- Strict Sphinx build and rendered `/latest/` link checking pass.
- Focused suites passed across three independently audited scopes (49, 62, and 117 tests; overlapping, not additive).
- `git diff --check` passes; changed Markdown/RST local-link validation found no missing targets.
- The review worktree remained clean; no source changes were made.

nvddr:
Disposition: **approve — no blocking findings** at `49ff082db1c66b72798c4755bc515d8963dae6af`.

I re-audited the follow-up changes against the prior review. The four outstanding blockers are resolved:

- The port `7880` documentation now accurately describes the host-networked LiveKit listener, warns against public exposure, provides source-restricted firewall guidance, and consistently scopes direct `7880` access to the native trusted-network debugging exception.
- All 15 exact legacy GitHub README fragment IDs are now preserved, including punctuation-generated double/triple-hyphen forms. I compared GitHub-rendered IDs across every README covered by the compatibility map and found no remaining missing legacy fragments.
- The unsupported NeMo Relay AutoAPI exception-serialization statement was removed while retaining the valid `ToolLoopError` recovery guidance.
- The active Llama Nemotron and Nemotron3 service documentation again preserves the authoritative model-card links, current sampling recommendations, license references, and commercial-use status. Stale hardware guidance was not restored.

The additional preservation issues from the prior pass are also addressed: native CloudXR port/certificate/encryption behavior is explicit, the direct native test binary is documented again, and the docs workflow wording correctly covers repository entry points beyond README files.

Validation performed on this head:

- `git diff --check 986ccaebb97083385e820e8f3a57819c0a85b2ea..49ff082db1c66b72798c4755bc515d8963dae6af`
- `62 passed` across docs-version, service-layout, LiveKit TLS SAN, and XR-render wire tests
- Exact GitHub-rendered legacy-fragment comparison across the 20 compatibility-map READMEs: no missing fragments
- All 16 GitHub checks are green, including strict docs, Python 3.11/3.12, DCO, lint, locks, SPDX, CodeQL, Android, and Pages

One non-blocking, pre-existing edge remains: with `web_server_tls=False`, `/token` advertises a direct `ws://host:7880` URL to custom consumers that choose to honor that field. The shipped web client does not use that URL and the documented sample flow remains on the port-8080 proxy, so this does not block this documentation consolidation.

nvddr (approved):

blongs-nv:
Reviewed at `49ff082db1c66b72798c4755bc515d8963dae6af` by a bot.

## Blockers

1. **The riva anchor on the docs page uses the GitHub slug, not the published one, and the test locks the mistake in.** [docs/source/reference/agent-sdk-models.md#L127](https://github.com/NVIDIA/xr-ai/blob/49ff082db1c66b72798c4755bc515d8963dae6af/docs/source/reference/agent-sdk-models.md#L127) has `(riva-grpc-speech-nim-stttts)=`, but the published fragment is `riva-grpc-speech-nim-stt-tts` (live on the pre-PR build and reproduced with the repo's pinned myst-parser). The same slug is correct in the package README, where GitHub's slugger drops the parentheses and slash without substitution, so only the Sphinx label and the expectations at [tests/test_docs_versions.py#L300](https://github.com/NVIDIA/xr-ai/blob/49ff082db1c66b72798c4755bc515d8963dae6af/tests/test_docs_versions.py#L300) and [#L428](https://github.com/NVIDIA/xr-ai/blob/49ff082db1c66b72798c4755bc515d8963dae6af/tests/test_docs_versions.py#L428) need the multi-hyphen form (or an added alias), and the test currently keeps CI green while the published fragment stays broken.

## Suggestions

1. [docs/source/_cli_reference.py#L103](https://github.com/NVIDIA/xr-ai/blob/49ff082db1c66b72798c4755bc515d8963dae6af/docs/source/_cli_reference.py#L103) emits every sample command as `uv run --project agent-samples/<dir> <program>`, the exact shape the sample-README guard forbids, and the rendered page never states these commands run from the repository root. Either emit the README-approved form or state the repo-root working directory on the page.

2. The compatibility anchors `(requirements)=`, `(build-and-run)=`, and `(connect)=` in [docs/source/getting_started/clients.md#L95-L114](https://github.com/NVIDIA/xr-ai/blob/49ff082db1c66b72798c4755bc515d8963dae6af/docs/source/getting_started/clients.md#L95-L114) are bare global Sphinx labels, and this PR adds more generic ones (`contract`, `quickstart`, `usage`, `frames`, `readiness`, `protocols`), so a future `{ref}` from another page could resolve into an unintended page. No duplicates exist today, but page-scoped names (e.g. `clients-requirements`) would remove the hazard.

3. The link checker's regex ([.github/scripts/check_latest_docs_links.py#L16-L19](https://github.com/NVIDIA/xr-ai/blob/49ff082db1c66b72798c4755bc515d8963dae6af/.github/scripts/check_latest_docs_links.py#L16-L19)) matches only inline `[text](url)` markdown, so reference-style links, bare URLs, and HTML anchors are invisible to it.

4. The port table now calls 8080 "the normal external entry point for every client" ([docs/source/getting_started/networking.md#L16](https://github.com/NVIDIA/xr-ai/blob/49ff082db1c66b72798c4755bc515d8963dae6af/docs/source/getting_started/networking.md#L16)) while two rows down 48322 remains a client-facing connection for `auto-webrtc` clients. Scoping the 48322 row was the bigger fix; "every client" on the 8080 row is the last residue of the overreach.

## Closing

The fourth push resolves everything else and introduces no new problems. The remaining gate is the riva Sphinx label and its two test expectations, a one-line rename each.

yanziz-nvidia (approved):
LGTM!

blongs-nv (approved):

PR #422, opened 2026-08-25
fix(voice): eliminate return audio underruns

wenxind-nvidia:
## Summary

- Preserve the one-time 320 ms return-track pre-roll, but send its eight 40 ms silence chunks at audio cadence so a supported 120 ms hub queue retains all of them.
- Run pre-roll as a participant-scoped output task: join, greeting synthesis, and application work proceed concurrently, while real speech remains ordered behind the silence.
- Maintain a bounded 120 ms downstream speech reserve and catch up after event-loop or IPC stalls instead of preserving them as audible gaps.
- Use one participant-scoped text-response boundary for greetings, stop acknowledgements, and assistant responses, ordered behind synthesis so every final partial Pipecat chunk is padded and flushed independently.
- Require `pipecat-ai>=1.6`, serialize participant lifecycle work, clear pacing state safely across interruptions, and prevent wake chimes from mid-response pad-flushing.
- Document the pre-roll and compatible hub-buffer behavior, with regression coverage for bounded hub queues, failure, concurrency, leave/shutdown teardown, incremental output, pacing stalls, and utterance boundaries.

## Root cause

Return audio originally had no jitter reserve, so ordinary scheduling and IPC delays reached the hub as late chunks and caused playout underruns. The return track was also created lazily from the first real speech, allowing setup to consume the beginning of a response, and partial aggregation chunks were not explicitly flushed after every source of synthesized text.

The first pre-roll implementation burst all 320 ms through IPC and modeled every chunk as retained. A 120 ms hub queue could drop 200 ms while the sender still delayed speech against a 320 ms deadline, creating an empty-track gap. Pacing the complete 320 ms pre-roll at 40 ms cadence keeps the hub queue and local deadline aligned without delaying text or greeting synthesis.

The fix lives in the shared xr-ai-voice and hub return-audio path, so voice-enabled agent samples receive the same behavior without sample-specific changes.

## Validation

- From `tests/`: `uv run pytest test_voice_pipeline.py -q`: 104 passed
- From `tests/`: `uv run pytest test_return_audio_pacing.py -q`: 17 passed
- From `tests/`: `uv run pytest -q -m "not gpu"`: 1,297 passed, 10 deselected
- Real 120 ms `_ReturnAudioPipe`: all eight pre-roll chunks retained, zero drops
- At the declared floor: Pipecat 1.6.0 boundary/chime cases: 5 passed
- Documentation API check, configuration-documentation tests, dependency map, and Python compilation: passed
- Manual simple-vlm-example session: response audio was materially smoother, including response starts

### DIFF
--- a/DEPENDENCIES.md
+++ b/DEPENDENCIES.md
@@ -119,7 +119,7 @@ change, even when an individual dependency publishes newer Python wheels.
   - `xr-ai-models` → [`xr-ai-models`](agent-sdk/xr-ai-models/) (local, editable)
   - `xr-ai-vad` → [`xr-ai-vad`](utils/xr-ai-vad/) (local, editable)
   - `xr-ai-voicegate` → [`xr-ai-voicegate`](utils/xr-ai-voicegate/) (local, editable)
-  - `pipecat-ai>=1.3`
+  - `pipecat-ai>=1.6`
   - `nltk!=3.10.1`
   - `numpy>=1.24`
   - `scipy>=1.11`
--- a/agent-sdk/xr-ai-voice/pyproject.toml
+++ b/agent-sdk/xr-ai-voice/pyproject.toml
@@ -18,7 +18,7 @@ dependencies = [
     "xr-ai-models",
     "xr-ai-vad",
     "xr-ai-voicegate",
-    "pipecat-ai>=1.3",
+    "pipecat-ai>=1.6",
     # NLTK 3.10.1 rejects its dependencies from in-project virtual environments.
     "nltk!=3.10.1",
     "numpy>=1.24",
--- a/agent-sdk/xr-ai-voice/xr_ai_voice/_frames.py
+++ b/agent-sdk/xr-ai-voice/xr_ai_voice/_frames.py
@@ -6,8 +6,8 @@
 Everything pipecat already ships (``InputAudioRawFrame``,
 ``OutputAudioRawFrame``, ``TranscriptionFrame``, ``UserStartedSpeakingFrame``,
 ``UserStoppedSpeakingFrame``, ``InterruptionFrame``, ``TextFrame``) is reused
-directly — only the participant lifecycle and the voicegate-emitted query
-frame live here.
+directly — only participant lifecycle, voice-gate queries, and synthesized-text
+response boundaries live here.
 """
 from __future__ import annotations
 
@@ -55,7 +55,20 @@ class GatedQueryFrame(DataFrame):
 
 
 @dataclass
-class AssistantResponseEndFrame(DataFrame):
+class TextResponseEndFrame(DataFrame):
+    """One participant-scoped sequence of synthesized text has ended.
+
+    Every producer that emits ``TextFrame`` objects for one utterance follows
+    them with this frame. ``StreamingTtsProcessor`` uses it to flush trailing
+    text and place the audio boundary behind all synthesis already queued for
+    the participant.
+    """
+
+    pid: str
+
+
+@dataclass
+class AssistantResponseEndFrame(TextResponseEndFrame):
     """A single assistant turn finished emitting ``TextFrame``s.
 
     Emitted by the private assistant processor when a query completes
@@ -70,6 +83,5 @@ class AssistantResponseEndFrame(DataFrame):
     was interrupted out of saying.
     """
 
-    pid: str
     text: str
     pts_us: int
--- a/agent-sdk/xr-ai-voice/xr_ai_voice/_pipeline.py
+++ b/agent-sdk/xr-ai-voice/xr_ai_voice/_pipeline.py
@@ -70,6 +70,9 @@ def _build_voice_pipeline(
         transport  = transport,
         text_topic = text_topic,
     )
+    voice_gate_proc.set_tts_response_active_probe(
+        streaming_tts.has_active_response,
+    )
     vad_stt = VadSttProcessor(
         stt=stt,
         vad_cfg=vad_cfg,
--- a/agent-sdk/xr-ai-voice/xr_ai_voice/_processors/streaming_tts.py
+++ b/agent-sdk/xr-ai-voice/xr_ai_voice/_processors/streaming_tts.py
@@ -77,6 +77,7 @@ class _TtsPidState:
     """Per-participant streaming state: pending text + its ordered sender."""
 
     __slots__ = (
+        "active_responses",
         "pending",
         "response_sentences",
         "sender_task",
@@ -85,6 +86,7 @@ class _TtsPidState:
     )
 
     def __init__(self) -> None:
+        self.active_responses: int = 0
         self.pending: str = ""
         self.response_sentences: int = 0
         self.sender_task: asyncio.Task | None = None
@@ -186,6 +188,11 @@ class StreamingTtsProcessor(FrameProcessor):
             )
         return st.sender_queue
 
+    def has_active_response(self, pid: str) -> bool:
+        """Return whether ordered response audio is still open for ``pid``."""
+        st = self._by_pid.get(pid)
+        return bool(st is not None and st.active_responses)
+
     async def _handle_text(self, frame: TextFrame) -> None:
         if not frame.text:
             return
@@ -266,6 +273,8 @@ class StreamingTtsProcessor(FrameProcessor):
         logger.info("tts sentence dispatch pid={!r} len={}", pid, len(sentence))
         queue = self._ensure_sender(pid)
         st = self._state(pid)
+        if not st.response_sentences:
+            st.active_responses += 1
         st.response_sentences += 1
         st.synth_seq += 1
         task  = asyncio.create_task(
@@ -296,6 +305,9 @@ class StreamingTtsProcessor(FrameProcessor):
                     stopped = TTSStoppedFrame()
                     stopped.transport_destination = item.pid
                     await self.push_frame(stopped)
+                    st = self._by_pid.get(item.pid)
+                    if st is not None and st.active_responses:
+                        st.active_responses -= 1
                     continue
                 task, pid = item
                 try:
--- a/agent-sdk/xr-ai-voice/xr_ai_voice/_processors/streaming_tts.py
+++ b/agent-sdk/xr-ai-voice/xr_ai_voice/_processors/streaming_tts.py
@@ -28,6 +28,7 @@ from __future__ import annotations
 
 import asyncio
 import re
+from dataclasses import dataclass
 from typing import TYPE_CHECKING
 
 import nemo_relay
@@ -38,6 +39,7 @@ from pipecat.frames.frames import (
     Frame,
     InterruptionFrame,
     TextFrame,
+    TTSStoppedFrame,
 )
 from pipecat.processors.frame_processor import FrameDirection, FrameProcessor
 
@@ -60,13 +62,27 @@ _SENTENCE_END = re.compile(r"(?<=[.!?])\s+")
 _TRAILING_SENTENCE_END = re.compile(r"""[.!?]["')\]]*\s*$""")
 
 
+@dataclass(frozen=True, slots=True)
+class _TtsResponseBoundary:
+    """Ordered marker placed behind every sentence in one assistant response."""
+
+    pid: str
+
+
 class _TtsPidState:
     """Per-participant streaming state: pending text + its ordered sender."""
 
-    __slots__ = ("pending", "sender_task", "sender_queue", "synth_seq")
+    __slots__ = (
+        "pending",
+        "response_sentences",
+        "sender_task",
+        "sender_queue",
+        "synth_seq",
+    )
 
     def __init__(self) -> None:
         self.pending: str = ""
+        self.response_sentences: int = 0
         self.sender_task: asyncio.Task | None = None
         self.sender_queue: asyncio.Queue | None = None
         self.synth_seq: int = 0
@@ -187,6 +203,15 @@ class StreamingTtsProcessor(FrameProcessor):
             st.pending = ""
             await self._dispatch_sentence(sentence, pid=frame.pid)
 
+        # The output transport aggregates small frames into 40 ms chunks. Put
+        # the boundary on the same FIFO as synthesis so it arrives only after
+        # every WAV in this response; Pipecat then flushes and silence-pads the
+        # final partial chunk instead of carrying it into the next response.
+        if st is not None and st.response_sentences:
+            queue = self._ensure_sender(frame.pid)
+            await queue.put(_TtsResponseBoundary(pid=frame.pid))
+            st.response_sentences = 0
+
         if not self._text_topic or self._transport is None:
             return
         if not frame.text:
@@ -233,6 +258,7 @@ class StreamingTtsProcessor(FrameProcessor):
         logger.info("tts sentence dispatch pid={!r} len={}", pid, len(sentence))
         queue = self._ensure_sender(pid)
         st = self._state(pid)
+        st.response_sentences += 1
         st.synth_seq += 1
         task  = asyncio.create_task(
             self._synthesize(sentence, pid=pid),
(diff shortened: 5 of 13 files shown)


blongs-nv:
Reviewed at `1cacf7fae37c924d924fc9603af6e5fc57d3af60` by a bot.

## Blockers

1. **The declared `pipecat-ai>=1.3` floor predates the `TTSStoppedFrame` flush behavior the whole fix depends on.**

   [`agent-sdk/xr-ai-voice/pyproject.toml:21`](https://github.com/NVIDIA/xr-ai/blob/1cacf7fae37c924d924fc9603af6e5fc57d3af60/agent-sdk/xr-ai-voice/pyproject.toml#L21)
   `MediaSender.handle_tts_stopped` and `_enqueue_flushed_audio_buffer` first appear in pipecat 1.6.0; in 1.3.0 through 1.5.0 the frame routes to `handle_sync_frame` and `_bot_stopped_speaking` discards the partial buffer, so any environment resolving inside the declared range truncates up to one 40 ms chunk off the tail of every greeting, stop-ack, and response. `tests/uv.lock` resolves 1.7.0, so CI never sees the floor. Raise it to `pipecat-ai>=1.6`.

2. **An interruption landing during a paced send resurrects the full reserve estimate for a hub queue that was just flushed, so the next response plays with no real reserve.**

   [`agent-sdk/xr-ai-voice/xr_ai_voice/_transport.py:370`](https://github.com/NVIDIA/xr-ai/blob/1cacf7fae37c924d924fc9603af6e5fc57d3af60/agent-sdk/xr-ai-voice/xr_ai_voice/_transport.py#L370), [`agent-sdk/xr-ai-voice/xr_ai_voice/_transport.py:421-455`](https://github.com/NVIDIA/xr-ai/blob/1cacf7fae37c924d924fc9603af6e5fc57d3af60/agent-sdk/xr-ai-voice/xr_ai_voice/_transport.py#L421-L455)
   Once the reserve is full, `write_audio_frame` holds the per-pid lock and sleeps in `_sleep_s` for roughly a full chunk per chunk, so a barge-in `InterruptionFrame` almost always arrives mid-sleep; its handler pops `_return_audio_deadline_s` without the lock while `_flush_hub` empties the hub queue, and the sleeping writer then wakes and writes back `max(buffered_until_s, sent_s) + duration_s`, reinstating a ~120 ms estimate against an empty queue. The estimate never resyncs downward during continuous playback, so the entire post-interrupt response is paced as if the reserve existed and ordinary jitter underruns exactly as before this fix. Before this change the write-back was `_monotonic_s() + duration_s`, indistinguishable from a fresh start, so the unlocked pop was harmless; clearing the deadline under the per-pid lock, or re-reading it after the sleep, closes the window.

## Suggestions

1. The 320 ms pre-roll is sent by the input transport ([`_transport.py:196-230`](https://github.com/NVIDIA/xr-ai/blob/1cacf7fae37c924d924fc9603af6e5fc57d3af60/agent-sdk/xr-ai-voice/xr_ai_voice/_transport.py#L196-L230)), bypassing the output transport's per-pid lock and `_return_audio_deadline_s` accounting, so the pacer's reserve estimate is blind to that real downstream backlog at greeting time. Priming from the output transport's `ParticipantJoinedFrame` handling keeps the return path single-writer and lets the deadline map see the silence.

2. The join callback now awaits eight IPC sends before pushing `ParticipantJoinedFrame` ([`_transport.py:196`](https://github.com/NVIDIA/xr-ai/blob/1cacf7fae37c924d924fc9603af6e5fc57d3af60/agent-sdk/xr-ai-voice/xr_ai_voice/_transport.py#L196)) while the endpoint dispatches join and leave callbacks as detached tasks ([`_processor.py:727`](https://github.com/NVIDIA/xr-ai/blob/1cacf7fae37c924d924fc9603af6e5fc57d3af60/agent-sdk/xr-ai-hub/xr_ai_hub/_processor.py#L727)), so a quick join-then-leave can deliver `ParticipantLeftFrame` first: `_release_destination` no-ops, and the late join frame then creates a `MediaSender` plus deadline and lock entries for a departed participant that are retained until shutdown. Pushing the join frame before the pre-roll (or running the pre-roll concurrently) restores the previously negligible window.

3. `play_wav` now follows every chime with a `TTSStoppedFrame` ([`voice_gate.py:130-133`](https://github.com/NVIDIA/xr-ai/blob/1cacf7fae37c924d924fc9603af6e5fc57d3af60/agent-sdk/xr-ai-voice/xr_ai_voice/_processors/voice_gate.py#L130-L133)), including the early wake-ack chime, which `handle_partial_transcript` can fire while assistant TTS for the same pid is still streaming; that mid-utterance stop marker pad-flushes a partial response chunk, inserting up to 40 ms of silence mid-word. Consider suppressing the marker when a response is in flight for that pid.

4. No test reaches teardown with a `_TtsResponseBoundary` parked in the sender queue ([`streaming_tts.py:349`](https://github.com/NVIDIA/xr-ai/blob/1cacf7fae37c924d924fc9603af6e5fc57d3af60/agent-sdk/xr-ai-voice/xr_ai_voice/_processors/streaming_tts.py#L349)); if that `isinstance` skip regressed, `synth_task, _ = item` would raise on every interrupt. A slow-TTS test that sends text plus a response end and then interrupts the same pid would cover it.

5. The new README paragraph names LiveKit ([`README.md:153-156`](https://github.com/NVIDIA/xr-ai/blob/1cacf7fae37c924d924fc9603af6e5fc57d3af60/agent-sdk/xr-ai-voice/README.md#L153-L156)) and this file is included verbatim in the published SDK docs, while AGENTS.md keeps LiveKit internal to the hub. Reword in hub terms, e.g. "so the hub can publish and subscribe to the return track before the greeting or first response."

The bounded-reserve pacing math is sound (capped at 120 ms, stalls drain the estimate so playback catches up) and the participant-scoped `TextResponseEndFrame` to `TTSStoppedFrame` boundary is a clean contract adopted consistently by all three text producers; once the pipecat floor is raised and the interrupt-versus-writer race is closed, this is good to merge.

nvddr (approved):

yanziz-nvidia:
Reviewed by yanziz-review-bot

**Summary**  
The current head passes all 101 voice-pipeline tests locally, but the pre-roll accounting assumes the hub retains every silence chunk. A supported smaller hub buffer can drop part of that burst, causing the sender to delay speech based on nonexistent queued audio and reintroduce an audible gap.

| | Finding |
|---|---|
| 🚫 | [`prime_return_audio()` unconditionally accounts for all 320 ms](https://github.com/NVIDIA/xr-ai/blob/fd73ceb9171f7019049e3622b79c084c9d206e34/agent-sdk/xr-ai-voice/xr_ai_voice/_transport.py#L290-L305), but the hub’s configurable [`return_audio_max_buffer_s`](https://github.com/NVIDIA/xr-ai/blob/fd73ceb9171f7019049e3622b79c084c9d206e34/services/device-io-hub/device_io_hub/transport/livekit/config.py#L105-L109) accepts any positive duration and drops oldest frames beyond that bound. With a 120 ms bound, processing the eight ready IPC chunks as one burst retains three and drops five, while `_return_audio_deadline_s` still advances by 320 ms. The first 40 ms speech chunk then waits about 240 ms despite at most 120 ms of pre-roll remaining, creating the underrun this PR is intended to eliminate. Pace the silent chunks through a buffer-aware path, or enforce/document a compatible minimum and cover the smaller-buffer case. |
| 💡 | none |
| 🔍 | [`the hub can publish and subscribe`](https://github.com/NVIDIA/xr-ai/blob/fd73ceb9171f7019049e3622b79c084c9d206e34/agent-sdk/xr-ai-voice/README.md#L153-L155) assigns both sides of the operation to the hub, but the hub publishes the return track and the participant subscribes. “The first audible output therefore follows that one-time pre-roll” also overstates the behavior when no immediate greeting is emitted and the pre-roll drains long before the first response. Describe it as triggering publication and giving the participant time to subscribe before an immediate greeting or response. |

- Make pre-roll pacing/accounting correct when `return_audio_max_buffer_s` is below 320 ms. Prefer pacing the silence rather than bursting eight chunks; otherwise enforce a compatible hub minimum. Add a regression test demonstrating that retained audio and `_return_audio_deadline_s` cannot diverge.
- Rewrite the README paragraph so the hub publishes the return track and the participant subscribes, without claiming that a later first response necessarily follows the join-time pre-roll.

nvddr:
Re-reviewed current head `fd73ceb` after reading the other reviews, with particular attention to keeping the fix below the application/aggregation boundary.

## Blocker

| Priority | Finding | Exit criteria |
|---|---|---|
| **Blocker** | **Pre-roll accounting can diverge from the supported hub buffer.** [`prime_return_audio()`](https://github.com/NVIDIA/xr-ai/blob/fd73ceb9171f7019049e3622b79c084c9d206e34/agent-sdk/xr-ai-voice/xr_ai_voice/_transport.py#L277-L305) advances the local deadline by all 320 ms sent, while the hub accepts any positive [`return_audio_max_buffer_s`](https://github.com/NVIDIA/xr-ai/blob/fd73ceb9171f7019049e3622b79c084c9d206e34/services/device-io-hub/device_io_hub/transport/livekit/config.py#L105-L130) and drops oldest queued frames above that bound. With a 120 ms bound, the hub can retain only three of the eight ready 40 ms silence chunks while the sender still models 320 ms. The first speech frame then waits about 240 ms, after the real pre-roll has already drained, leaving an avoidable empty-track gap and defeating the continuity guarantee for a documented supported configuration. | Keep the correction low-level: pace pre-roll through a path whose deadline follows elapsed/retained audio, or enforce and document a compatible hub minimum. Add the smaller-buffer regression so retained audio and the local deadline cannot diverge. |

## Aggregation and scope

`VoiceAggregationAgent`, `VoiceOutput`, and their scheduling semantics are unchanged. Finite and incremental aggregated responses still close through the existing `AssistantResponseEndFrame`; this PR consumes that internal marker to flush the final partial Pipecat audio chunk. The 120 ms reserve sends initial chunks immediately, so it neither double-paces aggregation nor adds recurring response-onset latency.

An output published immediately from a participant-joined handler can start the aggregation agents open-loop playback estimate before the one-time pre-roll drains. That may move up to 320 ms of queueing downstream at join, but participant-scoped ordering and the new boundaries prevent chopping or cross-response merging. I consider that bounded behavior non-blocking and consistent with the documented open-loop estimator.

The changes above transport are justified by the tail-clipping mode: raw audio cannot reveal where an utterance ends. Greetings, stop acknowledgements, and assistant/aggregated responses therefore need a private end marker, and the asynchronous chime path needs the narrow activity check to avoid pad-flushing an active response mid-word. These remain private pipeline mechanics; the public aggregation API is not broadened.

Non-blocking suggestion: add one integration test carrying a final incremental `VoiceAggregationAgent` output through `VoiceAgent`/streaming TTS to verify exactly one independently flushed final chunk. Existing unit coverage is strong, but no new test crosses that complete boundary.

Validation: all current CI checks are green; locally, 101 voice-pipeline tests and 52 aggregation/hub-pacing tests passed.

nvddr:
Re-reviewed current head `4272ab8`. Good to go from my review. The pre-roll is now paced through the low-level return-audio path, stays within the documented hub buffer, and keeps real speech ordered behind the primer without adding recurring response latency. The previous blocker is resolved, and I found no remaining blockers.

nvddr (approved):

blongs-nv:
Reviewed at `4272ab8bf50f9cfd2311f178b29b94f06b76263f` by a bot.

## Blockers

1. **Cancelled pre-roll kills the sender's audio task through `asyncio.shield`.** [`_wait_return_audio_preroll`](https://github.com/NVIDIA/xr-ai/blob/4272ab8bf50f9cfd2311f178b29b94f06b76263f/agent-sdk/xr-ai-voice/xr_ai_voice/_transport.py#L334-L339) awaits the pre-roll task through a bare shield, so when [`_release_destination`](https://github.com/NVIDIA/xr-ai/blob/4272ab8bf50f9cfd2311f178b29b94f06b76263f/agent-sdk/xr-ai-voice/xr_ai_voice/_transport.py#L437) or [`stop()`/`cancel()`](https://github.com/NVIDIA/xr-ai/blob/4272ab8bf50f9cfd2311f178b29b94f06b76263f/agent-sdk/xr-ai-voice/xr_ai_voice/_transport.py#L454-L463) cancels that task, `CancelledError` re-raises into every `write_audio_frame` waiting on it, and pipecat's `MediaSender._audio_task_handler` guards `write_audio_frame` with only `except Exception`, so the sender's audio task dies with nothing logged. The departing participant's own sender masks this via the immediate `sender.cancel()`, but for the shared default (`destination=None`) sender that carries fallback-routed frames the death is permanent: every later unaddressed audio frame queues into a dead sender until process restart, with the window being the roughly 280 ms pre-roll after every join plus any pre-roll active at shutdown. Reproduced with a faithful emulation of pipecat's guard: after `_release_destination("alice")` during an active pre-roll the writer task ends cancelled and the `except Exception` handler never fires, while [the new test](https://github.com/NVIDIA/xr-ai/blob/4272ab8bf50f9cfd2311f178b29b94f06b76263f/tests/test_voice_pipeline.py#L3792-L3830) asserts the `CancelledError` escape as intended behavior but only against a bare task, never through pipecat's handler. Suggested shape: catch `CancelledError` in `_wait_return_audio_preroll`, re-raise only when the pre-roll task was not itself cancelled, and have `write_audio_frame` return `False` for the pre-roll-cancelled case.

## Suggestions

1. [`_run_return_audio_preroll`](https://github.com/NVIDIA/xr-ai/blob/4272ab8bf50f9cfd2311f178b29b94f06b76263f/agent-sdk/xr-ai-voice/xr_ai_voice/_transport.py#L289-L316) holds the per-pid lock for the full roughly 280 ms of paced sends, and the [`InterruptionFrame` path](https://github.com/NVIDIA/xr-ai/blob/4272ab8bf50f9cfd2311f178b29b94f06b76263f/agent-sdk/xr-ai-voice/xr_ai_voice/_transport.py#L486-L501) acquires that lock (the global path takes every pid's lock in turn), so an interruption overlapping any participant's join pre-roll stalls output-transport frame handling by up to roughly 280 ms where the old burst held it for microseconds.
2. The `return_audio_max_buffer_s` minimum of 0.12 stated in the [README](https://github.com/NVIDIA/xr-ai/blob/4272ab8bf50f9cfd2311f178b29b94f06b76263f/agent-sdk/xr-ai-voice/README.md#L161), server-runtime.md, and migrations.md is not enforced by [the hub validator](https://github.com/NVIDIA/xr-ai/blob/4272ab8bf50f9cfd2311f178b29b94f06b76263f/services/device-io-hub/device_io_hub/transport/livekit/config.py#L14-L29), which still accepts any finite value above zero. Below 0.04 every 40 ms chunk individually exceeds the bound and the hub drops 100% of both pre-roll and speech, so built-in voice output goes completely silent with only rate-limited warnings; a validation floor or a loud startup warning would catch it.
3. In [`_release_destination`](https://github.com/NVIDIA/xr-ai/blob/4272ab8bf50f9cfd2311f178b29b94f06b76263f/agent-sdk/xr-ai-voice/xr_ai_voice/_transport.py#L435-L450) the per-pid lock is popped only after release, so a fallback writer already queued on that lock proceeds after the pop, sends audio for the departed participant, and re-inserts a deadline entry. The consequence is bounded to roughly 120 ms of phantom reserve on a quick rejoin, and the path is untested.
4. [`stop()` and `cancel()`](https://github.com/NVIDIA/xr-ai/blob/4272ab8bf50f9cfd2311f178b29b94f06b76263f/agent-sdk/xr-ai-voice/xr_ai_voice/_transport.py#L454-L463) clear the deadline and lock maps without holding the locks and before the senders drain, so an in-flight writer can repopulate both maps after the clear. Harmless at teardown, but it is the one remaining lock-free mutation of those maps.
5. In [`voice_gate.py`](https://github.com/NVIDIA/xr-ai/blob/4272ab8bf50f9cfd2311f178b29b94f06b76263f/agent-sdk/xr-ai-voice/xr_ai_voice/_processors/voice_gate.py#L139-L147) the TTS-active probe is read when the chime is pushed, while the chime's own raw frames are still traversing the pipeline, so a sub-chunk chime tail can sit unflushed until the next stop marker. That file is untouched by this push.

The pacing fix is correct, including the pre-roll accounting fix: reproduction confirms zero drops and accounting that matches retained audio at every bound at or above the documented 0.12 minimum, with the join path undelayed and audible greeting timing unchanged. Once the shielded cancellation can no longer kill a sender's audio task, this looks good to merge.

yanziz-nvidia (approved):

blongs-nv (approved):

PR #424, opened 2026-08-25
Fix idle tea assistant routing

nvddr:
## Summary

- split common, idle-only, and active-only foreground policies so the active-guide refusal is absent from idle model context
- make idle mode explicitly general-purpose while routing only visible current-scene questions to `current_view`
- limit active visual-tool use to the tea guide and its current step
- always append the packaged idle or active route policy after the configurable shared foreground prompt, without adding configuration fields or constructor parameters
- remove worked examples from all model-visible prompt fragments and audit the complete idle and active prompt compositions for eval overlap
- document `foreground_prompt` and `foreground_prompt_file` semantics in the checked-in worker YAML, sample README, and published reference
- add symmetric idle and active shirt-color regression cases plus real-constructor coverage for packaged and overridden prompts

## Validation

- `UV_PROJECT_ENVIRONMENT=../.venv uv run pytest test_tea_making_sample.py -q` — 39 passed
- live Omni foreground routing evaluation — all 23 cases passed
- `ruff 0.15.16 check` on changed Python files
- strict Sphinx documentation build with warnings treated as errors
- wheel build includes all three foreground prompt files
- SPDX header check on changed source files
- `git diff --check`

### DIFF
--- a/agent-samples/tea-making-sample/README.md
+++ b/agent-samples/tea-making-sample/README.md
@@ -118,8 +118,11 @@ development network or put it behind an authenticated TLS proxy.
   and workflow paths.
 - `yaml/workflow.yaml` defines the tea steps, typed state, evidence gates, and
   user-facing messages.
-- `worker/tea_making_worker/prompts/` is the default source for model prompts;
-  explicit inline YAML values override those files.
+- `worker/tea_making_worker/prompts/` is the default source for model prompts.
+  An inline `foreground_prompt` or `foreground_prompt_file` replaces the shared
+  foreground instructions, while the packaged idle or active route policy is
+  always appended. Other prompt and prompt-file settings replace their matching
+  packaged files.
 - `yaml/voice_gate.yaml` is the checked-in wake-word configuration. Set
   `voice_gate_yaml: voice_gate.always-on.yaml` in the worker YAML to dispatch
   every finalized utterance without a wake phrase.
--- a/agent-samples/tea-making-sample/eval/cases.yaml
+++ b/agent-samples/tea-making-sample/eval/cases.yaml
@@ -8,6 +8,10 @@
   query: What is the capital of France?
   expected_tool: null
   forbidden_response: I can only help with the active tea guide right now.
+- name: idle-routes-visible-shirt-color
+  query: What color is the shirt?
+  expected_tool: current_view
+  forbidden_response: I can only help with the active tea guide right now.
 - name: start-tea-guide
   query: Help me make a cup of tea.
   expected_tool: workflow__start
@@ -81,6 +85,13 @@
   route: active
   step: start_steeping
   query: Please translate good morning into Japanese.
+  expected_tool: null
+  expected_response: I can only help with the active tea guide right now.
+
+- name: active-rejects-unrelated-visual-question
+  route: active
+  step: fill_water
+  query: What color is the person's shirt?
   expected_tool: null
   expected_response: I can only help with the active tea guide right now.
 
--- a/agent-samples/tea-making-sample/eval/eval.py
+++ b/agent-samples/tea-making-sample/eval/eval.py
@@ -13,6 +13,7 @@ from typing import Any
 import yaml
 from tea_making_worker.background_context import BackgroundContextAgent
 from tea_making_worker.change_watch import ChangeWatchAgent
+from tea_making_worker.config import load_config
 from tea_making_worker.foreground import ForegroundAgent
 from tea_making_worker.spec import load_workflow
 from tea_making_worker.transcript import TranscriptAgent
@@ -23,10 +24,10 @@ from xr_ai_tools.image import ImageRegistry
 from xr_ai_tools.tool_calling import tool_definitions
 
 _SAMPLE = Path(__file__).resolve().parents[1]
-_PROMPTS = _SAMPLE / "worker" / "tea_making_worker" / "prompts"
 
 
 def _build_agents(llm: LLMService) -> tuple[ForegroundAgent, GuidanceAgent]:
+    config = load_config(_SAMPLE / "yaml" / "tea_making_worker.yaml")
     images = SimpleNamespace(
         images=ImageRegistry(),
         get_current_frame=SimpleNamespace(),
@@ -70,7 +71,7 @@ def _build_agents(llm: LLMService) -> tuple[ForegroundAgent, GuidanceAgent]:
         change_watch=change_watch,
         transcript=transcript,
         video_log=video_log,
-        prompt=(_PROMPTS / "foreground_prompt.txt").read_text(encoding="utf-8"),
+        prompt=config.foreground_prompt,
     )
     return foreground, guidance
 
--- a/agent-samples/tea-making-sample/worker/tea_making_worker/config.py
+++ b/agent-samples/tea-making-sample/worker/tea_making_worker/config.py
@@ -14,6 +14,10 @@ import yaml
 _PACKAGE = Path(__file__).resolve().parent
 
 
+class _PackagedPrompt(str):
+    """Mark prompt text loaded from a package default rather than an override."""
+
+
 @dataclass(frozen=True, slots=True)
 class WorkerConfig:
     models_config: Path
@@ -71,7 +75,8 @@ def _prompt(data: dict[str, Any], config_path: Path | None, name: str) -> str:
         if configured is not None
         else _PACKAGE / "prompts" / f"{name}.txt"
     )
-    return path.read_text(encoding="utf-8").strip()
+    text = path.read_text(encoding="utf-8").strip()
+    return _PackagedPrompt(text) if configured is None else text
 
 
 def _positive(data: dict[str, Any], name: str, default: float) -> float:
--- a/agent-samples/tea-making-sample/worker/tea_making_worker/config.py
+++ b/agent-samples/tea-making-sample/worker/tea_making_worker/config.py
@@ -14,10 +14,6 @@ import yaml
 _PACKAGE = Path(__file__).resolve().parent
 
 
-class _PackagedPrompt(str):
-    """Mark prompt text loaded from a package default rather than an override."""
-
-
 @dataclass(frozen=True, slots=True)
 class WorkerConfig:
     models_config: Path
@@ -75,8 +71,7 @@ def _prompt(data: dict[str, Any], config_path: Path | None, name: str) -> str:
         if configured is not None
         else _PACKAGE / "prompts" / f"{name}.txt"
     )
-    text = path.read_text(encoding="utf-8").strip()
-    return _PackagedPrompt(text) if configured is None else text
+    return path.read_text(encoding="utf-8").strip()
 
 
 def _positive(data: dict[str, Any], name: str, default: float) -> float:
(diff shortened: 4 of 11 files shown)


blongs-nv:
Reviewed at `3e1ce3aa0e961f4438c39cd269f400909d6a4bb8` by a bot.

## Blockers

1. **The sample's documented prompt-override contract is now false in both directions, and no doc was updated.**

   [`agent-samples/tea-making-sample/README.md:121`](https://github.com/NVIDIA/xr-ai/blob/3e1ce3aa0e961f4438c39cd269f400909d6a4bb8/agent-samples/tea-making-sample/README.md#L121)
   The README says `prompts/` is the default source and explicit inline YAML values override those files, but `foreground_idle.txt` and `foreground_active.txt` are read at module import ([`foreground.py:53-55`](https://github.com/NVIDIA/xr-ai/blob/3e1ce3aa0e961f4438c39cd269f400909d6a4bb8/agent-samples/tea-making-sample/worker/tea_making_worker/foreground.py#L53-L55)) with no config keys, so no YAML value can touch them, while an explicit `foreground_prompt`/`foreground_prompt_file` now silently discards the packaged idle/active route policy, including the active-mode refusal. Update the README and the [`prompts/` row and routing section of the docs page](https://github.com/NVIDIA/xr-ai/blob/3e1ce3aa0e961f4438c39cd269f400909d6a4bb8/docs/source/reference/tea-making-sample.md#L154) to state the new semantics, or wire the new prompts through `WorkerConfig` like every other prompt.

2. **The config-marker to init-flag wiring is untested, so an inverted or dropped flag assignment breaks the feature in production while the suite stays green.**

   [`tests/test_tea_making_sample.py:1429`](https://github.com/NVIDIA/xr-ai/blob/3e1ce3aa0e961f4438c39cd269f400909d6a4bb8/tests/test_tea_making_sample.py#L1429)
   Both new route tests build the agent with `object.__new__(ForegroundAgent)` and hand-set `_uses_default_route_policy`, bypassing the assignment at [`foreground.py:88`](https://github.com/NVIDIA/xr-ai/blob/3e1ce3aa0e961f4438c39cd269f400909d6a4bb8/agent-samples/tea-making-sample/worker/tea_making_worker/foreground.py#L88). Inverting that `isinstance` check makes default users lose the refusal policy and override users get fragments appended, yet all four new and changed tests still pass. Add one test that constructs `ForegroundAgent` through `__init__` with `load_config(...).foreground_prompt` (and once with an override string) and asserts on `_prepare_route` output.

## Suggestions

1. [`config.py:17`](https://github.com/NVIDIA/xr-ai/blob/3e1ce3aa0e961f4438c39cd269f400909d6a4bb8/agent-samples/tea-making-sample/worker/tea_making_worker/config.py#L17): `_PackagedPrompt(str)` smuggles the override bit through the type system: `WorkerConfig.foreground_prompt` is typed plain `str`, any string operation silently strips the marker, and correctness depends on the `isinstance` check at [`foreground.py:88`](https://github.com/NVIDIA/xr-ai/blob/3e1ce3aa0e961f4438c39cd269f400909d6a4bb8/agent-samples/tea-making-sample/worker/tea_making_worker/foreground.py#L88) landing before the `.strip()` on the next line. An explicit boolean field on `WorkerConfig`, or per-mode prompt fields like the lab-instrument-monitoring sample uses, carries the contract visibly and survives refactors.

2. [`foreground.py:290`](https://github.com/NVIDIA/xr-ai/blob/3e1ce3aa0e961f4438c39cd269f400909d6a4bb8/agent-samples/tea-making-sample/worker/tea_making_worker/foreground.py#L290): `getattr(self, "_uses_default_route_policy", False)` maps a missing attribute to override mode, silently dropping the active refusal in exactly the case that should fail loudly. It exists only to serve the `object.__new__` test style, so once the tests construct through `__init__`, a plain attribute read turns wiring bugs into loud `AttributeError`s.

3. [`foreground_prompt.txt:1`](https://github.com/NVIDIA/xr-ai/blob/3e1ce3aa0e961f4438c39cd269f400909d6a4bb8/agent-samples/tea-making-sample/worker/tea_making_worker/prompts/foreground_prompt.txt#L1): the shared fragment's `current_view` example ("the color of visible clothing") and `foreground_active.txt`'s "visual tools only for facts relevant to the active tea guide" pull opposite directions for exactly the active shirt-color case your new eval case pins. Moving the clothing/holding example into `foreground_idle.txt` keeps the shared section route-neutral.

4. [`foreground_prompt.txt:1`](https://github.com/NVIDIA/xr-ai/blob/3e1ce3aa0e961f4438c39cd269f400909d6a4bb8/agent-samples/tea-making-sample/worker/tea_making_worker/prompts/foreground_prompt.txt#L1) is the natural template for a `foreground_prompt_file` override, but it no longer contains the idle/active split or the refusal, so an operator who copies and tweaks it silently gets no refusal in active mode. A one-line note in the file that an override must supply its own complete route policy would prevent the trap.

5. [`config.py:78`](https://github.com/NVIDIA/xr-ai/blob/3e1ce3aa0e961f4438c39cd269f400909d6a4bb8/agent-samples/tea-making-sample/worker/tea_making_worker/config.py#L78): an empty override (`foreground_prompt: ""` or a whitespace-only file) yields an empty system prompt with no policy and no error. Rejecting empty prompt text in `_prompt` turns that misconfiguration into a startup failure.

The common/idle/active split is the right factoring, the override precedence is correct as written, and the new eval cases cover both route policies in both directions. Once the docs match the new override semantics and one test exercises the real constructor path, this is good to merge.

nvddr:
Thanks @blongs-nv. Addressed the review feedback in `868c8d19284e8a5bd5620f0f0762314c3ad813c4`.

### Blockers addressed

1. **Prompt override contract and documentation:** `foreground_prompt` and `foreground_prompt_file` now replace only the shared foreground instructions. The packaged idle or active route policy is always appended, so an override cannot silently remove the active refusal or workflow-control safeguards. The sample README, published tea reference, prompt source-map row, and shared prompt text now document that behavior.
2. **Production constructor coverage:** replaced the `object.__new__` route tests with tests that call `ForegroundAgent.__init__` using both `load_config(...).foreground_prompt` and an inline override, then verify idle and active `_prepare_route` output.

### Suggestions addressed

- Removed `_PackagedPrompt`, `_uses_default_route_policy`, the `isinstance` provenance check, and the missing-attribute `getattr` fallback. No public configuration field or constructor parameter was added.
- Moved the holding/clothing examples from the shared prompt into the idle-only prompt, eliminating the conflict with active visual restrictions.
- Made the shared prompt explicitly state that a packaged route policy follows it; the README and published docs explain the same override boundary.
- The empty-prompt validation concern predates this PR and applies to the shared private prompt loader. It is tracked separately in #425. With this change, an empty foreground override still receives the route-specific policy rather than producing an empty foreground system prompt.

### Validation

- tea regression suite: 39 passed
- live Omni routing evaluation: all 23 cases passed
- Ruff check passed
- strict Sphinx build passed with warnings treated as errors
- SPDX and `git diff --check` passed

wenxind-nvidia (commented):
## Review

### Suggestions

1. **Remove the worked-example overlap from the idle prompt.** Moving the examples out of the shared prompt fixed their conflict with the active policy, but the idle model still receives `foreground_idle.txt`, which explicitly teaches “what someone is holding” and “the color of visible clothing.” The applicable idle eval cases then ask “What am I holding right now?” and “What color is the shirt?” That makes those results partly a recall check. Keep the eval cases unchanged and express the prompt rule generically without those examples.

2. **Audit each route’s complete model-visible prompt for eval overlap.** Check the shared prompt plus the applicable idle or active policy, rather than only the shared fragment. That would catch an example being relocated between prompt files while remaining visible to the same eval case.

3. **Expose the documented override fields in the checked-in worker YAML.** Add commented `foreground_prompt` and `foreground_prompt_file` guidance to `yaml/tea_making_worker.yaml` so these supported fields and their shared-policy-only semantics also appear in the generated configuration reference.

### Nits

1. The first line of `foreground_prompt.txt` describes how prompt fragments are packaged rather than model behavior. The README and published reference now document that boundary, so this implementation detail does not need to consume model context.

Reviewed at `868c8d19284e8a5bd5620f0f0762314c3ad813c4`.

wenxind-nvidia (approved):

nvddr:
Thanks @wenxind-nvidia. Addressed the current-head suggestions in `02c9873eb8a66016c1b74826aaeec93b28bb5134`.

- Removed the holding/clothing worked examples from the idle prompt entirely; the existing eval queries remain unchanged.
- Updated the regression test to inspect each complete model-visible composition (`shared + idle` and `shared + active`) and reject overlap with the worked-example terms.
- Added commented `foreground_prompt` and `foreground_prompt_file` guidance, including their shared-policy-only semantics, to the checked-in worker YAML for the generated configuration reference.
- Removed the prompt-packaging implementation note from `foreground_prompt.txt`; that information remains in the README and published reference instead of consuming model context.

Validation: 39 tea tests passed, all 23 live Omni routing cases passed, Ruff passed, strict Sphinx passed with warnings treated as errors, and SPDX plus `git diff --check` passed.

blongs-nv (approved):

yanziz-nvidia:
Reviewed by yanziz-review-bot

**Summary**  
The route-policy split is structurally sound, and the focused suite passes all 39 tests. However, the shared RAG instruction conflicts with the active-route contract and available tools in most workflow steps, leaving a core routing case ambiguous and untested.

| | Finding |
|---|---|
| 🚫 | [`foreground_prompt.txt:3`](https://github.com/NVIDIA/xr-ai/blob/02c9873eb8a66016c1b74826aaeec93b28bb5134/agent-samples/tea-making-sample/worker/tea_making_worker/prompts/foreground_prompt.txt#L3) tells both routes to use `rag_lookup` for every tea or hot-water fact, but the active policy says to answer from the current step and sparse state, and most active steps do not expose `rag_lookup` (`fill_water`, `heat_water`, `start_steeping`, and `steep_timer`). For example, asking the existing white-tea-temperature question during `fill_water` gives the model an instruction to call a tool that is absent while the documented active contract limits it to the current step. The eval covers that query only on the idle route, so the conflicting active composition is not exercised. Move the broad RAG rule into the idle policy, rely on the `identify` step’s narrower retrieval instructions while active, and add an active non-current-step tea-fact case that verifies the intended response. |
| 💡 | none |
| 🔍 | none |

- Remove the unconditional `rag_lookup` instruction from the shared foreground prompt and place idle-only retrieval guidance in `foreground_idle.txt`.
- Keep active retrieval governed by the current step’s voice policy and exposed tool set.
- Add an active-route eval case for a tea fact unrelated to the current step—such as the existing white-tea-temperature query during `fill_water`—and assert the intended no-tool response.

nvddr:
Thanks @yanziz-nvidia. I checked this against base `986ccaebb97083385e820e8f3a57819c0a85b2ea`.

The RAG/tool mismatch is real, but it predates this PR: the base foreground prompt already combined “answer from the current step” with “use `rag_lookup` for tea or hot-water facts,” while active steps such as `fill_water` already omitted `rag_lookup`. PR #424 preserves that active composition; it does not introduce or worsen the mismatch.

Moving the broad RAG rule into the idle policy and adding the proposed active-route eval would be a reasonable adjacent cleanup, but it is not a regression from this change. I’m therefore not treating it as a blocker for #424; it should be handled as a self-contained follow-up issue/PR.

yanziz-nvidia (approved):
not newly introduced

PR #427, opened 2026-08-25
chore(xr-render): bump Isaac Teleop to 1.3.131

nvddr:
## Summary

- require Isaac Teleop 1.3.131 or newer in the XR Render Demo's CloudXR and OpenXR services
- align the WebXR CloudXR SDK and Apple CloudXR framework pins with CloudXR 6.2.0 shipped by that release
- version downloaded WebXR SDK tarballs and rebuild generated bundles when their recorded SDK version is stale
- reconcile the duplicate Swift package resolutions
- regenerate the repository dependency map

## Why

Isaac Teleop v1.3.131 is the latest published release and updates the native CloudXR runtime and supported clients to CloudXR 6.2.0. Keeping the Python services and both XR client paths on the same release family avoids stale service locks and client/runtime version skew.

## Impact

Fresh and existing service environments now resolve at least Isaac Teleop 1.3.131. The XR Render Demo's WebXR vendor build and iOS/visionOS sample resolve their matching CloudXR 6.2.0 clients. Existing WebXR installations automatically replace cached 6.1 artifacts instead of serving them with the upgraded runtime.

## Validation

- `uv lock --project services/cloudxr-runtime`
- `uv lock --project services/openxr-service`
- `uv run --project tests pytest tests/test_cloudxr_runtime_process.py tests/test_cloudxr_service_layout.py tests/test_launcher_cloudxr_env.py tests/test_openxr_service_config.py -q` (35 passed)
- imported all Isaac Teleop CloudXR APIs used by `cloudxr-runtime` from the resolved 1.3.131 environment
- imported `isaacteleop.deviceio` and `isaacteleop.oxr` from the resolved 1.3.131 environment
- built the WebXR vendor bundle from CloudXR SDK 6.2.0 with `client-samples/web-xr-build/build.sh`
- `uv run --project tests pytest tests/test_service_layout.py tests/test_cloudxr_runtime_process.py tests/test_cloudxr_service_layout.py tests/test_openxr_service_config.py -q` (29 passed)
- `uv run --project tests pytest tests/test_docs_versions.py -q` (16 passed)
- `uvx ruff check agent-samples/xr-render-demo/main.py tests/test_service_layout.py`
- `bash -n client-samples/web-xr-build/build.sh`

Xcode validation was not run on this Linux host; the Apple package resolution uses the official CloudXR Framework 6.2.0 tag and commit.


yanziz-nvidia (approved):

yanziz-nvidia:
change the title?

blongs-nv:
Reviewed at `d00675b77230124fc2dfb10d1570ba89b537422b` by a bot.

## Suggestions

1. The two [`Package.resolved`](https://github.com/NVIDIA/xr-ai/blob/d00675b77230124fc2dfb10d1570ba89b537422b/client-samples/ios-visionos/StreamKit/Package.resolved#L30-L40) copies disagree on swift-protobuf (1.36.1 against 1.37.0 in the [xcworkspace copy](https://github.com/NVIDIA/xr-ai/blob/d00675b77230124fc2dfb10d1570ba89b537422b/client-samples/ios-visionos/StreamKitSample.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved#L30-L40)), which predates this change and survives it because the cloudxr-framework pins were edited directly. A real Xcode resolve would reconcile both files, so the next one will produce unrelated churn on top of whatever it is doing.

No blockers. `isaacteleop` 1.3.131 is the first stable release shipping CloudXR 6.2.0, so the floor matches the 6.2.0 client pins exactly, and the revision in both `Package.resolved` copies is the actual 6.2.0 tag. Good to merge.

blongs-nv (approved):

wenxind-nvidia:
## Blocker

The WebXR CloudXR 6.2 upgrade does not invalidate previously generated 6.1 artifacts. After updating, `agent-samples/xr-render-demo/main.py::_ensure_web_vendor()` returns whenever both bundles exist, without checking `.sdk-version`. Even a manual rebuild reuses the unversioned `client-samples/web-xr-build/sdk.tgz`, regardless of which SDK it contains. Existing installations can therefore run the upgraded 6.2 runtime while continuing to serve the 6.1 WebXR client.

Please record the SDK version alongside the generated bundle and rebuild when it differs from `.sdk-version`. The tarball should also be versioned or automatically replaced on mismatch. Add a regression test covering an existing bundle produced with the previous version.

## Suggestions

None.

## Nits

None.

The dependency changes and Swift lockfiles are consistent. The visionOS build succeeded against CloudXRKit 6.2.0, `git diff --check` and SPDX validation passed, and all current GitHub checks are green.

PR #432, opened 2026-08-26
Gate xr render sample perception claims on turn observation evidence

blongs-nv:
Follow-up to #416, extending evidence-backed success from mutations to perception. Live, "What am I holding?" was answered "You are holding a camera." in 0.4 s with no vision call, while the user held a mouse.

- `TurnEvidence` counts camera observations by kind: a successful live `look_at_current_frame` or `resolve_physical_color` records `observed`; `look_at_past_frame` records `observed_recorded`. Only a live observation licenses a present-tense claim.
- A reply asserting the user's physical state (present-tense hold/carry/wear/grip, "in your hand") with no live observation this turn gets one nudged retry, then an honest fallback. When only a recorded frame was consulted, the nudge steers the answer into past tense about the recorded moment, and the fallback sentence says the recorded-only situation plainly. The check runs per clause over the agent's own reply and skips questions; a hedge exempts only its own clause.

Test plan: unit 1360/1360 (`-m "not gpu"`), ruff and prompt audit clean, system prompt unchanged (retry nudge text reworked); regression tests for clause-scoped hedges and recorded-frame turns; eval cases `basics_bare_holding_question` and `historical_vision`; live manip 13/13; offline tiers (subagents 48/49, delegation 20/23, utterances 33/36) verified unchanged from the pre-PR base on the same stack.

### DIFF
--- a/agent-samples/xr-render-demo/eval/xr_render_demo_eval/harness.py
+++ b/agent-samples/xr-render-demo/eval/xr_render_demo_eval/harness.py
@@ -303,8 +303,11 @@ CASES = (
     Case(
         name="historical_vision",
         request="What color was the object I held ten seconds ago?",
-        vision="The previously held object was purple.",
+        # Phrased to lure a present-tense reply; the gate must steer the
+        # historical answer into past tense, never swallow it.
+        vision="In the recorded frame, the user is holding a purple mug.",
         required_tools=frozenset({"look_at_past_frame"}),
+        reply_contains="purple",
         # The supervisor's own conversation recall is the single expected
         # call; the camera's past is video, never the transcript.
         expected_call_counts=(("recall_conversation", 1),),
--- a/agent-samples/xr-render-demo/eval/xr_render_demo_eval/harness.py
+++ b/agent-samples/xr-render-demo/eval/xr_render_demo_eval/harness.py
@@ -787,6 +787,16 @@ UTTERANCES = (
         forbidden_tools=_FORBID_MUTATIONS,
     ),
     Case(
+        name="basics_bare_holding_question",
+        # The perception gate's live regression: a bare hands question got
+        # answered from the model's prior instead of the camera.
+        request="What am I holding right now?",
+        vision="A hand holding a stapler.",
+        required_tools=frozenset({"look_at_current_frame"}),
+        forbidden_tools=_FORBID_MUTATIONS,
+        reply_contains="stapler",
+    ),
+    Case(
         name="basics_look_for_yourself",
         # Answering a clarifying question with "go observe" must re-enter
         # the pending request through the camera, never echo those words.
--- a/agent-samples/xr-render-demo/worker/xr_render_demo_worker/_physical_color.py
+++ b/agent-samples/xr-render-demo/worker/xr_render_demo_worker/_physical_color.py
@@ -21,7 +21,7 @@ from xr_ai_tools.current_frame import CurrentFrameRequest, CurrentFrameTool
 from xr_ai_tools.vision import ImageQueryRequest, ImageQueryResult
 
 from ._tolerant import reraise_unavailable
-from ._trace import current_participant_id, current_trace_id
+from ._trace import current_participant_id, current_trace_id, record_evidence
 
 IMAGE_QUERY_SYSTEM_PROMPT = (
     'Reply with exactly "VISIBLE r g b", where r, g, b are numbers between 0 '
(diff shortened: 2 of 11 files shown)


wenxind-nvidia:
## Suggestions

1. `agent-samples/xr-render-demo/worker/xr_render_demo_worker/supervisor.py:403`: Preserve historical observations instead of forcing a live-camera retry. `look_at_past_frame` intentionally records no live evidence, but `_claims_perception` classifies a common history-grounded answer such as "In the recorded frame, you are holding a purple mug." as a present-tense claim. The supervisor then queries the current frame and may replace a correct historical answer with the user's current state or the fixed can't-say response. Please track historical evidence separately or account for the request/claim's temporal context so recorded evidence can support recorded-frame answers without satisfying present-time claims. Add a regression case whose historical answer uses this common phrasing; the current fixture avoids the path by saying "previously held."

2. `agent-samples/xr-render-demo/worker/xr_render_demo_worker/supervisor.py:141`: Scope hedges to the claim they qualify instead of exempting the entire sentence. `_PERCEPTION_HEDGES.search(sentence)` makes `_claims_perception("I can't see clearly, but you are holding a camera.")` return false, so with zero observations this reply bypasses both the retry and the fixed fallback and ships the unsupported physical claim. Please distinguish inability constructions such as "I can't tell what you're holding" from a separate positive clause, and add a mixed-clause regression test.

yanziz-nvidia:
Reviewed by yanziz-review-bot

**Summary**  
This head adds turn-scoped camera evidence plus retry/fallback enforcement and resolves the earlier recorded-frame and cross-clause-hedge cases. Common claim forms can still bypass classification, and unrelated color observations can license claims they do not support. The focused suite passes (77 tests); 13 of 15 GitHub checks pass, with both pytest jobs failing in an untouched voice-transport test reproducible locally.

| | Finding |
|---|---|
| 🚫 | [`agent-samples/xr-render-demo/worker/xr_render_demo_worker/supervisor.py:154`](https://github.com/NVIDIA/xr-ai/blob/a18ba671bfc10554a23cbdb94f654aa7ddaafc25/agent-samples/xr-render-demo/worker/xr_render_demo_worker/supervisor.py#L154) — `_claims_perception` still returns false for positive replies such as “I see you holding a camera” and “You are holding a camera I cannot identify clearly,” allowing them to bypass retry and fallback; recognize first-person observation constructions and exempt only hedges that negate the physical claim · previously raised |
| 💡 | [`agent-samples/xr-render-demo/worker/xr_render_demo_worker/_physical_color.py:118`](https://github.com/NVIDIA/xr-ai/blob/a18ba671bfc10554a23cbdb94f654aa7ddaafc25/agent-samples/xr-render-demo/worker/xr_render_demo_worker/_physical_color.py#L118) — `resolve_physical_color` increments the generic live-observation counter despite returning only RGB, so an unrelated color lookup can license a fabricated holding, carrying, or wearing claim; separate color evidence from claim-grounding evidence or bind evidence to the observed question and result. |
| 🔍 | none |

- `agent-samples/xr-render-demo/worker/xr_render_demo_worker/supervisor.py:154` — Recognize first-person positive observation claims and restrict hedge exemptions to language that negates the physical claim itself.
- `agent-samples/xr-render-demo/worker/xr_render_demo_worker/_physical_color.py:118` — Prevent color-only observations from authorizing unrelated physical-state claims, either with distinct evidence types or claim-bound evidence.

nvddr (changes requested):
Reviewed current head `a18ba671`. The recorded-frame handling and compound hedge/assertion cases now address the earlier feedback, but two blockers remain:

1. `agent-samples/xr-render-demo/worker/xr_render_demo_worker/supervisor.py:154` — The gate still misses ordinary positive perception claims. On this head, `_claims_perception` returns false for “I see you holding a camera,” “Your hand is holding a camera,” “You hold a camera,” and “You are holding a camera I cannot identify clearly.” Those replies therefore bypass both the retry and fallback when no camera observation occurred. Please cover these common positive constructions and make hedge handling exempt only language that negates the physical claim itself, with regression cases.

2. `agent-samples/xr-render-demo/worker/xr_render_demo_worker/_physical_color.py:118` / `supervisor.py:416` — Evidence is still turn-wide rather than claim-bound. `resolve_physical_color` increments the same `observed` counter after answering only a color query, and any nonzero value disables the gate for every holding/carrying/wearing claim in the final reply. A color lookup can therefore license an unrelated fabricated physical-state assertion. Please separate color evidence from claim-grounding evidence or bind observation evidence to the question/result being asserted.

The focused predicate checks above reproduce on the current head.

PR #431, opened 2026-08-26
ci: support Python 3.13 and 3.14

nvddr:
## Summary

- expand all 37 Python projects from `>=3.11,<3.13` to `>=3.11,<3.15`
- add Python 3.13 and 3.14 to the pytest CI matrix
- declare `importlib-resources` directly for `xr-ai-vad`, fixing Silero model-resource loading on Python 3.13+
- update the canonical requirements, testing guidance, sample templates, and generated dependency inventory

## Validation

- Python 3.11: `1348 passed, 1 skipped, 10 deselected`
- Python 3.12: `1348 passed, 1 skipped, 10 deselected`
- Python 3.13: `1348 passed, 1 skipped, 10 deselected`
- Python 3.14: `1348 passed, 1 skipped, 10 deselected`
- `uv lock` passed for all 37 projects across the complete declared range
- generated dependency-map check passed
- strict Sphinx documentation build passed
- Ruff, SPDX, merge-marker, and `git diff --check` checks passed

GPU-, Docker-, and NVENC-dependent tests remain outside the standard CI matrix and were not run locally.

PR #434, opened 2026-08-27
build: freeze dependency qualification by upload date

nvddr:
## Summary

- Add a repository-wide uv cutoff for package artifacts uploaded after `2026-08-27T17:15:48Z`.
- Apply that policy explicitly to lock-check, CPU test, and nightly GPU workflows, and include it in their cache keys.
- Document the intentional dependency-refresh process while keeping every generated `uv.lock` gitignored.

## Why

PR #433 showed that a fresh CI resolution can fail without any repository change when Pipecat publishes a new release. Adding exact upper bounds to every direct dependency would miss transitive/build dependencies, duplicate policy across 37 standalone projects, and leak repository qualification state into published library metadata. uv's `exclude-newer` setting provides one reviewable qualification point for the whole package-index graph instead.

Nested standalone projects do not inherit a parent uv config automatically, so CI exports `UV_CONFIG_FILE` explicitly and the contributor guidance uses `--config-file uv.toml`.

## Validation

- `uv lock` under the shared config: 37/37 projects resolved, 0 failures
- `xr-ai-voice` resolved Pipecat 1.8.1, uploaded before the cutoff
- Generated dependency map is current
- Focused dependency-map tests: 11 passed
- All workflow YAML parsed successfully
- No `uv.lock` is tracked or included in this PR

### DIFF
--- a/.github/workflows/lock-check.yml
+++ b/.github/workflows/lock-check.yml
@@ -9,9 +9,11 @@ name: lock-check
 # so a single `uv lock` per project covers every supported interpreter
 # without needing a Python-version matrix here.
 #
-# This is the repo-wide check on Python-version + dependency consistency:
-# a project bumping `requires-python` (or pinning a dep that drops a
-# Python version) only passes if the constraints stay jointly satisfiable.
+# The root uv.toml limits package-index candidates to the last repository-wide
+# qualification point. This is the repo-wide check on Python-version +
+# dependency consistency: a project bumping `requires-python`, advancing that
+# cutoff, or pinning a dependency that drops a Python version only passes if
+# the constraints stay jointly satisfiable.
 
 on:
   push:
@@ -21,6 +23,9 @@ on:
 concurrency:
   group: lock-check-${{ github.ref }}
   cancel-in-progress: true
+
+env:
+  UV_CONFIG_FILE: ${{ github.workspace }}/uv.toml
 
 jobs:
   dependency-map:
@@ -53,9 +58,10 @@ jobs:
           enable-cache: true
           # Lockfiles are gitignored repo-wide, so the default `**/uv.lock`
           # cache-dependency-glob matches nothing in CI. Key the cache off
-          # every committed pyproject.toml instead.
+          # every committed pyproject.toml and the shared resolution policy.
           cache-dependency-glob: |
             **/pyproject.toml
+            uv.toml
 
       - name: Resolve every pyproject.toml
         run: |
--- a/.github/workflows/lock-check.yml
+++ b/.github/workflows/lock-check.yml
@@ -52,6 +52,18 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
+      - name: Validate dependency cutoff
+        run: |
+          cutoff=$(sed -nE 's/^exclude-newer = "([^"]+)"$/\1/p' uv.toml)
+          if [[ -z "$cutoff" ]] || ! cutoff_epoch=$(date -u -d "$cutoff" +%s); then
+            echo "::error file=uv.toml::exclude-newer must be a valid timestamp"
+            exit 1
+          fi
+          if (( cutoff_epoch > $(date -u +%s) )); then
+            echo "::error file=uv.toml::exclude-newer is in the future: $cutoff"
+            exit 1
+          fi
+
       - name: Install uv
         uses: astral-sh/setup-uv@caf0cab7a618c569241d31dcd442f54681755d39 # v3.2.4
         with:
--- a/.github/workflows/nightly-xr-ai-test.yml
+++ b/.github/workflows/nightly-xr-ai-test.yml
@@ -3,6 +3,9 @@
 
 name: nightly XR AI test
 
+env:
+  UV_CONFIG_FILE: ${{ github.workspace }}/uv.toml
+
 # Runs the `gpu`-marked pytest suite on a self-hosted GPU runner — these
 # tests need real GPU / Docker / NVENC hardware, so they're filtered out of
 # the default `tests` workflow (ubuntu-latest). Mirrors the local
@@ -91,8 +94,8 @@ jobs:
         with:
           enable-cache: true
           cache-dependency-glob: |
+            uv.toml
             tests/pyproject.toml
-            tests/uv.lock
             agent-sdk/xr-ai-hub/pyproject.toml
             services/device-io-hub/pyproject.toml
 
--- a/.github/workflows/nightly-xr-ai-test.yml
+++ b/.github/workflows/nightly-xr-ai-test.yml
@@ -25,6 +25,7 @@ on:
   pull_request:
     paths:
       - .github/workflows/nightly-xr-ai-test.yml
+      - uv.toml
 
 # Queue overlapping runs instead of cancelling; the previous nightly
 # may still be running when the next cron fires on a busy GPU host.
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -3,6 +3,9 @@
 
 name: tests
 
+env:
+  UV_CONFIG_FILE: ${{ github.workspace }}/uv.toml
+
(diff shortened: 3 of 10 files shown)


wenxind-nvidia (changes requested):
## Blockers

- Add `uv.toml` to the GPU workflow’s PR path filter. Otherwise, future cutoff changes can admit new GPU dependencies without pre-merge GPU validation.

- Ensure local test entry points load the root config. `tests/run_local_gpu_tests.sh` and `docs/source/guides/testing.md` currently invoke uv without `--config-file`, causing the qualification cutoff to be ignored.

## Suggestions

- Consider adding a small CI assertion that representative nested-project commands retain the configured `exclude-newer` value. This would prevent future workflow or documentation changes from silently bypassing the policy.

## Nits

- None.

yanziz-nvidia:
can we document process to update this date/time (not purely relying on agent to remember it) somewhere so we don't forget about this needs to be updated some day

yanziz-nvidia (approved):

nvddr:
Addressed the review feedback in `b4eceba4`:

- added `uv.toml` to the GPU workflow's PR path filter;
- made the local GPU script and both testing entry points load the root config explicitly;
- corrected the testing guide's repo-root GPU script path;
- documented the cutoff refresh procedure, including how to generate the UTC timestamp, resolve every project, inspect version changes, and run CPU/GPU validation.

I did not add a separate static assertion for uv's config semantics: cutoff changes now trigger the real GPU workflow, while lock-check already resolves every project with the same root config.

Validation: `git diff --check`, shell syntax, workflow YAML parse, and 27 focused dependency-policy/docs tests passed.

blongs-nv:
Reviewed at `b4eceba4d72033486d04be08c46e70b646eb9a99` by a bot.

## Suggestions

1. "Standalone nested projects do not inherit the root config implicitly" is not the rule uv follows, and the wrong reason undermines the right instruction. Testing uv 0.10.7: a nested project inherits an ancestor `uv.toml` by upward discovery, and inheritance stops only when the project's own `pyproject.toml` defines a `[tool.uv]` table, which here means `[tool.uv.sources]` in 31 of the 37 projects. A contributor who checks the stated rule against one of the six projects without that table ([`utils/xr-ai-launcher`](https://github.com/NVIDIA/xr-ai/blob/b4eceba4d72033486d04be08c46e70b646eb9a99/utils/xr-ai-launcher/pyproject.toml) among them) sees the cutoff apply anyway and may drop the flag everywhere. Naming `[tool.uv.sources]` as what blocks discovery keeps the instruction standing: [`AGENTS.md:118`](https://github.com/NVIDIA/xr-ai/blob/b4eceba4d72033486d04be08c46e70b646eb9a99/AGENTS.md#L118), [`DEPENDENCIES.md:47`](https://github.com/NVIDIA/xr-ai/blob/b4eceba4d72033486d04be08c46e70b646eb9a99/DEPENDENCIES.md#L47), [`contributing-and-conventions.md:60`](https://github.com/NVIDIA/xr-ai/blob/b4eceba4d72033486d04be08c46e70b646eb9a99/docs/source/guides/contributing-and-conventions.md#L60).

2. The CI-mirror block in [`testing.md:22-26`](https://github.com/NVIDIA/xr-ai/blob/b4eceba4d72033486d04be08c46e70b646eb9a99/docs/source/guides/testing.md#L22-L26) no longer matches what CI runs. [`tests.yml:72`](https://github.com/NVIDIA/xr-ai/blob/b4eceba4d72033486d04be08c46e70b646eb9a99/.github/workflows/tests.yml#L72) runs `uv run pytest ...` from `working-directory: tests` with `UV_PROJECT_ENVIRONMENT: ../.venv`, relying on the workflow-level `UV_CONFIG_FILE` rather than the flag; the documented form differs in all three respects, and the workflow comment above it explains that the venv location specifically is load-bearing for nltk. Before this change the block matched the workflow verbatim.

3. A future-dated cutoff is the one misconfiguration nothing catches. A malformed value fails every uv invocation and a far-past value fails resolution across all 37 projects, but a single wrong digit in the year at [`uv.toml:5`](https://github.com/NVIDIA/xr-ai/blob/b4eceba4d72033486d04be08c46e70b646eb9a99/uv.toml#L5) resolves cleanly and silently restores the exact behaviour this PR removes. Asserting the value parses and is not in the future, before the resolve loop at [`lock-check.yml:66`](https://github.com/NVIDIA/xr-ai/blob/b4eceba4d72033486d04be08c46e70b646eb9a99/.github/workflows/lock-check.yml#L66), closes it.

4. [`contributing-and-conventions.md:23`](https://github.com/NVIDIA/xr-ai/blob/b4eceba4d72033486d04be08c46e70b646eb9a99/docs/source/guides/contributing-and-conventions.md#L23) leaves `uv tool run ruff check` bare while the two commands above it carry `--config-file`, and `uv tool run` is the one form that never picks the config up from the working directory. CI pins ruff to 0.15.16 at [`lint.yml:43`](https://github.com/NVIDIA/xr-ai/blob/b4eceba4d72033486d04be08c46e70b646eb9a99/.github/workflows/lint.yml#L43), so `ruff@0.15.16` here would make local lint match CI exactly.

5. The refresh procedure is now well documented, but nothing can detect that it is due. There is no dependabot or renovate config, no scheduled refresh job, and no dependency vulnerability scan, and the nightly resolves under the same frozen timestamp, so it stays green by construction while post-cutoff security releases remain invisible. A cutoff-age check that warns and then fails past an agreed window is the piece that turns the documented procedure into something that actually fires.

6. [`DEPENDENCIES.md:42`](https://github.com/NVIDIA/xr-ai/blob/b4eceba4d72033486d04be08c46e70b646eb9a99/DEPENDENCIES.md#L42) allows advancing the timestamp "only in a dedicated dependency refresh," which #431 cannot satisfy if any dependency's 3.13 or 3.14 support postdates the cutoff. Naming interpreter qualification as a second legitimate trigger avoids a rule that forbids the work already in flight.

The mechanism holds up under testing: the timestamp parses, the cutoff binds `lock`, `sync`, `run`, and `tool run` including transitive and build dependencies, a missing config file fails loudly rather than skipping the policy, every uv step in all three workflows runs after checkout, and no project loses its `[tool.uv.sources]` under `UV_CONFIG_FILE`. Nothing here blocks merging.

wenxind-nvidia (commented):

## Suggestions

- `DEPENDENCIES.md:42-49` — **Make future refreshes produce an auditable upgrade diff.** `uv lock` uses existing lockfile versions as resolution preferences and does not upgrade unchanged packages unless `--upgrade` is supplied. Because every `uv.lock` is gitignored, a clean checkout has no old resolution to compare, while a working checkout can retain old versions after the cutoff advances. Consider documenting a deterministic sequence: resolve and snapshot every project under the old cutoff, advance `exclude-newer`, re-lock every project with `--upgrade`, inspect the before/after lock diffs, and then run the CPU and GPU suites.

Reviewed head: b4eceba4d72033486d04be08c46e70b646eb9a99

wenxind-nvidia (approved):

nvddr:
Applied the scoped follow-up in `dc109cba`:

- corrected the uv discovery explanation: upward discovery stops at a nearer `[tool.uv]` table, which most nested projects have via `[tool.uv.sources]`;
- clarified that the documented CPU command is the repository-root equivalent of CI, whose root-level environment placement is intentional for nltk;
- added a lock-check guard that rejects malformed or future `exclude-newer` timestamps before resolving projects;
- added `--upgrade` to the documented dependency-refresh command.

Deferred the remaining suggestions to avoid broadening this PR:

- local Ruff pinning is separate lint-tool/version maintenance;
- cutoff-age enforcement needs an agreed refresh cadence and security policy;
- interpreter qualification should not weaken the dedicated dependency-refresh boundary.

Validation: 27 focused tests passed; workflow YAML, diff checks, current/future timestamp branches, and a network-backed `uv lock --upgrade --dry-run` all passed.

PR #435, opened 2026-08-27
Fix development TLS certificate interoperability

yanziz-nvidia:
## Summary

- replace the CA-as-server development certificate with a stable root CA and a signed `CA:FALSE` server leaf
- rotate only the leaf when SANs change, preserving the installed root across hostname and public-IP changes
- serve only the public root CA from `/cert`; neither private key can be returned
- detect and clearly log migration of the legacy single CA-as-leaf cache
- preserve externally supplied `cert_file`/`key_file` support; `/cert` is disabled when DeviceIOHub does not own the root
- document the updated trust and one-time migration flows for browsers, Android, Apple, and native clients

No certificate verification is bypassed or disabled.

## Validation

- `52 passed`: TLS unit tests, external-certificate compatibility, Docker config tests, docs-version tests, and API documentation tests
- Docker-backed LiveKit Python 1.1.8/Rust TLS regression connects through the generated HTTPS/WSS chain and verifies `/cert` returns only the root CA
- `openssl verify -CAfile root-ca.crt web-server.crt` succeeds
- `curl --cacert root-ca.crt https://localhost:PORT/token?...` returns a valid token response
- Ruff, SPDX, and `git diff --check` pass

The full locked test environment cannot be synchronized on macOS ARM because PyNvVideoCodec 2.2.0 has no macOS wheel; the focused CPU and Docker-backed regression suites above pass.

### DIFF
--- a/client-samples/android/app/src/main/java/com/nvidia/xrai/streamkitsample/CertInstall.kt
+++ b/client-samples/android/app/src/main/java/com/nvidia/xrai/streamkitsample/CertInstall.kt
@@ -16,7 +16,7 @@ import javax.net.ssl.SSLContext
 import javax.net.ssl.X509TrustManager
 
 /**
- * Fetches the hub's development root CA and returns a [KeyChain] install [Intent], or null on
+ * Fetches the hub's self-signed CA cert and returns a [KeyChain] install [Intent], or null on
  * failure. The fetch uses a single-connection trust-all SSLSocketFactory because the cert being
  * downloaded is the one that is not yet trusted.
  */
--- a/client-samples/android/app/src/main/java/com/nvidia/xrai/streamkitsample/CertInstall.kt
+++ b/client-samples/android/app/src/main/java/com/nvidia/xrai/streamkitsample/CertInstall.kt
@@ -16,7 +16,7 @@ import javax.net.ssl.SSLContext
 import javax.net.ssl.X509TrustManager
 
 /**
- * Fetches the hub's self-signed CA cert and returns a [KeyChain] install [Intent], or null on
+ * Fetches the hub's development root CA and returns a [KeyChain] install [Intent], or null on
  * failure. The fetch uses a single-connection trust-all SSLSocketFactory because the cert being
  * downloaded is the one that is not yet trusted.
  */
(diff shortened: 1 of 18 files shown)


wenxind-nvidia (commented):

## Suggestions

- `services/device-io-hub/device_io_hub/transport/livekit/_tls.py:255` — **Account for long-running certificate renewal.** The new leaf expires after 397 days, but renewal is checked only when `WebServer.start()` calls `ensure_development_certificates()`. A hub that remains running past `notAfter` continues serving the expired certificate; the base certificate lasted 3650 days. Consider reloading TLS before expiration or documenting the required restart cadence for generated development certificates.

- `services/device-io-hub/device_io_hub/transport/livekit/_tls.py:311-317` — **Tell operators when the trust anchor changes.** A missing, corrupt, mismatched, or near-expiry root causes automatic root replacement, after which every previously enrolled client rejects the new leaf. The warning currently says only that the root is being replaced; the reinstall guidance is emitted only during legacy leaf migration. Include the reinstall instruction whenever an existing root is replaced and document that recovery path.

Reviewed head: 169d4cdb00a357be66dafe09aedaff1dd1dfaa0a

wenxind-nvidia (approved):

nvddr:
Reviewed head `169d4cdb00a357be66dafe09aedaff1dd1dfaa0a`.

The root/leaf split, SAN-only leaf rotation, `/cert` exposure boundary, external-certificate behavior, and client trust paths look correct. I found no blockers.

| Priority | Area | Feedback |
|---|---|---|
| Lower | Long-running renewal | Renewal is checked only at web-server startup, so a hub that remains up beyond the 397-day leaf lifetime will continue serving an expired certificate. Consider documenting a restart cadence or adding a deliberate reload/renewal lifecycle. |
| Lower | Validity horizon | A newly issued 397-day leaf can outlive its issuer when the root has 31–396 days remaining. Either cap the leaf's `notAfter` to the root's lifetime or rotate the root before issuing such a leaf. |
| Lower | Root replacement recovery | Whenever an existing root is replaced because it is incomplete, invalid, mismatched, or near expiry, make the warning explicitly state that enrolled clients must reinstall the new root. |

Focused validation: all 18 TLS tests passed locally, as did `openssl verify -x509_strict -purpose sslserver` against the generated chain.

nvddr (approved):

yanziz-nvidia:
Addressed all three points in d7eff758 and 90892a28:

1. Generated certificates are checked daily, and renewed leaves are reloaded into the active TLS context without restarting the hub.
2. A leaf’s notAfter is now capped at the root CA’s notAfter.
3. Root replacement warnings explicitly tell enrolled clients to reinstall the new root-ca.crt, and the recovery path is documented.

Added regression coverage for near-expiry renewal, issuer validity, runtime TLS reload, and root replacement. The focused suite passes all 22 tests.

PR #438, opened 2026-08-28
Redact stale perception answers from inline context

blongs-nv:
Repeat "What am I holding?" asks were answered from the prior sighting in history instead of a fresh look. Supersedes the gate in #432.

Vision-turn replies show in inline history as a placeholder (tagged via a shadow transcript source recall never returns), so answering a perception question requires looking again. The store is cleared at worker startup so memory never outlives the run, and the VLM is told to say when the subject isn't visible instead of guessing.

Test plan: vision tier 10/10, offline tiers at baseline, new eval + restart-survival regressions, live headset testing across object changes, empty hands, and idle recycles.

### DIFF
--- a/agent-samples/xr-render-demo/eval/xr_render_demo_eval/harness.py
+++ b/agent-samples/xr-render-demo/eval/xr_render_demo_eval/harness.py
@@ -14,8 +14,11 @@ from xr_ai_tools import Tool
 from xr_ai_tools.text_memory import (
     AddTranscriptRequest,
     ConversationEntry,
+    QueryTranscriptsRequest,
+    QueryTranscriptsResult,
     RecallConversationRequest,
     RecallConversationResult,
+    TranscriptSegment,
 )
 from xr_ai_tools.types import SpatialFrame, Vector3
 from xr_render_demo_worker.config import load_config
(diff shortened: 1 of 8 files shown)


wenxind-nvidia:
## Blockers

1. **Startup recursively deletes an unrestricted configuration path.** `agent-samples/xr-render-demo/worker/xr_render_demo_worker/app.py:75` passes `text_memory_dir` directly to `shutil.rmtree()`. Because `config.py:41` accepts arbitrary absolute or relative paths, `text_memory_dir: .` deletes the YAML directory and `..` can delete the sample tree. It also removes unrelated files placed in that directory. Please use a per-run owned subdirectory or delete only recognized transcript artifacts after validating an ownership marker.

2. **Provenance failures fail open and restore the original stale-answer bug.** If the shadow-source query fails, `supervisor.py:255-266` sets the tag set to empty and injects every original vision answer. Persistence is also non-atomic: the normal agent reply can succeed before the shadow write fails at lines 288-314. For this invariant, failures should omit agent inline history or otherwise fail closed, and the answer plus provenance should be stored atomically.

## Suggestions

1. **Redaction does not prevent unsupported perception answers.** A first-turn “What am I holding?” can still return a physical claim without calling `vision_agent`: `needs_verification` remains mutation-only in `supervisor.py:388`. Redaction only affects later history and cannot provide the same guarantee as PR 432’s observation-evidence gate. Consider retaining an equivalent gate requiring a current observation before present-tense physical claims, with retry and an honest fallback.

2. `vision_reply = vision_ran and not mutation_ran` does not mean “pure perception.” A compound vision-plus-memory turn has no mutating agent, so its entire response, including stable recalled information, is redacted. Track all delegated agent kinds and tag only truly perception-only replies.

3. Add regressions for an unsupported first-turn physical claim, shadow-query failure, partial shadow persistence, compound vision-plus-memory turns, and dangerous `text_memory_dir` values.

4. Document the startup-clearing behavior beside `text_memory_dir` in the sample YAML, where configuration field behavior is defined.

GitHub’s Python 3.11/3.12 tests and other checks are green. Local Ruff and `git diff --check` pass; local pytest resolution is blocked on macOS ARM because `PyNvVideoCodec` has no compatible wheel.

yanziz-nvidia:
reviewed by yanziz-review-bot

**Summary**
This PR adds timestamp-based redaction of vision replies, startup transcript cleanup, and regression coverage. Other than the existing findings in previous comments, I found one new merge-blocking regression affecting historical-frame follow-ups. [GitHub checks](https://github.com/NVIDIA/xr-ai/pull/438/checks) are green; 49 relevant local tests passed, while a focused two-turn reproduction exposed the issue below.

| Severity | Finding |
|---|---|
| 🚫 BLOCKER | [`supervisor.py:451`](https://github.com/NVIDIA/xr-ai/blob/9fd54a9ec7523c94200cfba03ef3df9e2ebcbad5/agent-samples/xr-render-demo/worker/xr_render_demo_worker/supervisor.py#L451) - Every pure `vision_agent` reply is tagged for redaction, but that agent handles both live and immutable recorded frames. In a supported flow such as “What was I holding ten seconds ago?” followed by “Was it striped?”, the second turn receives only the placeholder, losing the historical answer and its original time anchor; retrying can inspect the wrong frame. Propagate whether the inner agent used `look_at_current_frame` or `look_at_past_frame`, and redact only live-frame answers. |
| 💡 SUGGESTION | none |
| 🧹 NIT | none |

- `agent-samples/xr-render-demo/worker/xr_render_demo_worker/supervisor.py:451` - Carry current-versus-historical observation provenance out of `vision_agent`, write the redaction tag only for live-frame answers, and add a two-turn recorded-frame follow-up regression proving the historical answer remains available inline.

PR #439, opened 2026-08-29
fix(services): migrate Hugging Face transfers to Xet

yanziz-nvidia:
## Summary

  - migrate Hugging Face downloads from the deprecated `HF_HUB_ENABLE_HF_TRANSFER` path to Xet high-performance mode across vLLM, STT, Magpie TTS, and Piper TTS
  - update the Hugging Face dependencies in eight service `pyproject.toml` files
  - forward effective Xet settings into vLLM containers and include them in the launch fingerprint
  - remove the unconditional `hf_transfer` installation from Docker startup
  - validate the in-container Xet stack and conditionally install a compatible `hf-xet` version when validation fails
  - honor `HF_HUB_DISABLE_XET` after `extra_env` overrides by skipping Xet validation and repair

  ## Why

  The existing Docker startup path unconditionally installs the deprecated `hf_transfer` helper from PyPI before starting vLLM. This makes startup depend on live
  package access and can fail in restricted or offline deployments.

  This change uses the supported Xet path instead. Docker mode validates the image’s Xet stack, conditionally repairs it when necessary, and fails clearly if the
  integration remains incompatible. Operators can set `HF_HUB_DISABLE_XET=1` to disable Xet and skip bootstrap validation and installation entirely.

  Because the launch contract changes, existing persistent vLLM containers are recreated once on their next launch.

  ## Testing

  - 95 focused vLLM Docker and lifecycle tests passed
  - 1,368 non-GPU full-suite tests passed
  - all three shipped vLLM image stacks were validated on dual Ada
  - all eight affected service lockfiles resolved successfully
  - dependency-map and whitespace checks passed

### DIFF
--- a/DEPENDENCIES.md
+++ b/DEPENDENCIES.md
@@ -238,7 +238,8 @@ change, even when an individual dependency publishes newer Python wheels.
 - Runtime dependencies:
   - `vllm>=0.14.0`
   - `pyyaml>=6.0`
-  - `hf-transfer>=0.1.4`
+  - `huggingface-hub>=0.32.0`
+  - `hf-xet>=1.1.2,<2.0.0`
   - `xr-ai-logging` → [`xr-ai-logging`](utils/xr-ai-logging/) (local, editable)
   - `xr-ai-vllm` → [`xr-ai-vllm`](utils/xr-ai-vllm/) (local, editable)
 - Optional dependency groups: none
@@ -253,7 +254,8 @@ change, even when an individual dependency publishes newer Python wheels.
 - Runtime dependencies:
   - `vllm>=0.12.0`
   - `pyyaml>=6.0`
-  - `hf-transfer>=0.1.4`
+  - `huggingface-hub>=0.32.0`
+  - `hf-xet>=1.1.2,<2.0.0`
   - `xr-ai-logging` → [`xr-ai-logging`](utils/xr-ai-logging/) (local, editable)
   - `xr-ai-vllm` → [`xr-ai-vllm`](utils/xr-ai-vllm/) (local, editable)
 - Optional dependency groups: none
@@ -272,7 +274,8 @@ change, even when an individual dependency publishes newer Python wheels.
   - `numpy>=1.24`
   - `fastapi>=0.111`
   - `uvicorn[standard]>=0.29`
-  - `hf-transfer>=0.1.4`
+  - `huggingface-hub>=0.32.0`
+  - `hf-xet>=1.1.2,<2.0.0`
   - `pyyaml>=6.0`
   - `xr-ai-logging` → [`xr-ai-logging`](utils/xr-ai-logging/) (local, editable)
 - Optional dependency groups: none
@@ -287,7 +290,8 @@ change, even when an individual dependency publishes newer Python wheels.
 - Runtime dependencies:
   - `vllm>=0.12.0`
   - `pyyaml>=6.0`
-  - `hf-transfer>=0.1.4`
+  - `huggingface-hub>=0.32.0`
+  - `hf-xet>=1.1.2,<2.0.0`
   - `xr-ai-logging` → [`xr-ai-logging`](utils/xr-ai-logging/) (local, editable)
   - `xr-ai-vllm` → [`xr-ai-vllm`](utils/xr-ai-vllm/) (local, editable)
 - Optional dependency groups: none
@@ -302,7 +306,8 @@ change, even when an individual dependency publishes newer Python wheels.
 - Runtime dependencies:
   - `vllm>=0.12.0`
   - `pyyaml>=6.0`
-  - `hf-transfer>=0.1.4`
+  - `huggingface-hub>=0.32.0`
+  - `hf-xet>=1.1.2,<2.0.0`
   - `xr-ai-logging` → [`xr-ai-logging`](utils/xr-ai-logging/) (local, editable)
   - `xr-ai-vllm` → [`xr-ai-vllm`](utils/xr-ai-vllm/) (local, editable)
 - Optional dependency groups: none
@@ -344,7 +349,8 @@ change, even when an individual dependency publishes newer Python wheels.
   - `hatchling`
 - Runtime dependencies:
   - `piper-tts>=1.4.0`
-  - `huggingface-hub>=0.22`
+  - `huggingface-hub>=0.32.0`
+  - `hf-xet>=1.1.2,<2.0.0`
   - `fastapi>=0.111`
   - `uvicorn[standard]>=0.29`
   - `pyyaml>=6.0`
@@ -379,6 +385,8 @@ change, even when an individual dependency publishes newer Python wheels.
   - `fastapi>=0.111`
   - `uvicorn[standard]>=0.29`
   - `python-multipart>=0.0.9`
+  - `huggingface-hub>=0.32.0`
+  - `hf-xet>=1.1.2,<2.0.0`
   - `pyyaml>=6.0`
   - `xr-ai-logging` → [`xr-ai-logging`](utils/xr-ai-logging/) (local, editable)
 - Optional dependency groups: none
@@ -409,7 +417,8 @@ change, even when an individual dependency publishes newer Python wheels.
 - Runtime dependencies:
   - `vllm>=0.23.0`
   - `pyyaml>=6.0`
-  - `hf-transfer>=0.1.4`
+  - `huggingface-hub>=0.32.0`
+  - `hf-xet>=1.1.2,<2.0.0`
   - `xr-ai-logging` → [`xr-ai-logging`](utils/xr-ai-logging/) (local, editable)
   - `xr-ai-vllm` → [`xr-ai-vllm`](utils/xr-ai-vllm/) (local, editable)
 - Optional dependency groups: none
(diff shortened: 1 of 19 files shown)


nvddr (commented):
Beta2 triage: I found no correctness blocker in the diff, and this looks like a good beta2 candidate: the behavioral change is compact, the edits across service manifests are mechanical, and all CI checks are green.

Please rebase/cherry-pick these two commits onto `release/beta-v0.2.0rc1` and make that the PR base so the change targets the release branch directly. Please do not only switch the base on the current head: it is based on `main` and contains main-only ancestry, including #434/#435, while the release line has its own divergent commits.

Please also add a short **Why** section to the PR description. The current summary says what is changing, but not the concrete failure/deprecation impact this prevents for beta2 users. It would also be useful to record that the checked-in vLLM images already provide a compatible `huggingface-hub`/`hf-xet`, since Docker startup no longer installs the transfer helper at runtime.

blongs-nv:
Reviewed at `6d05a9a85995d87ef28c5121dcfdd9039f6f3a91` by a bot. This applies equally to the #456 cherry-pick, whose diff is identical apart from hunk offsets.

## Blockers

1. **Docker mode loses its accelerator guarantee, and the failure is now silent.**

   [`_docker.py:164`](https://github.com/NVIDIA/xr-ai/blob/6d05a9a85995d87ef28c5121dcfdd9039f6f3a91/utils/xr-ai-vllm/xr_ai_vllm/_docker.py#L164) drops the in-container `pip install -q hf_transfer` whose own comment said the NGC image does not ship it, and installs nothing in its place; the new `hf-xet` and `huggingface-hub` pins are host-side `pyproject.toml` dependencies that govern pip mode only. Every shipped profile runs docker mode (`vllm_backend: docker` across all of `agent-samples/model-servers/yaml/{96G_blackwell,dual_48G_ada,spark}/`), over three images including `vllm/vllm-openai:v0.20.0`. The old flag hard-errored when its package was missing, which is why the install existed; an image without a Xet-capable `huggingface_hub` simply ignores [`HF_XET_HIGH_PERFORMANCE`](https://github.com/NVIDIA/xr-ai/blob/6d05a9a85995d87ef28c5121dcfdd9039f6f3a91/utils/xr-ai-vllm/xr_ai_vllm/_docker.py#L147) and falls back to plain HTTPS, so a no-op migration looks identical to a working one on 30-100 GB first-run downloads. The assumption that the pinned images already carry a Xet-capable stack is the load-bearing one here and is nowhere recorded; either evidence it per image, or restore a guarded in-container install or fail-fast import check so a missing stack is loud.

## Suggestions

1. The `huggingface-hub>=0.32.0` floor is below the release that introduced the setting this change depends on. I pulled the wheels: `HF_XET_HIGH_PERFORMANCE` is absent from `huggingface_hub/constants.py` in 0.32.0, 0.33.0, 0.34.0, and 0.35.0, and first appears in 1.0.0. `exclude-newer` currently resolves to 1.29.0 so nothing is broken today, but the declared constraint does not express what the code needs; `>=1.0.0` would.

2. The unconditional [`os.environ.pop`](https://github.com/NVIDIA/xr-ai/blob/6d05a9a85995d87ef28c5121dcfdd9039f6f3a91/utils/xr-ai-vllm/xr_ai_vllm/_config.py#L112) (also `stt_server/__main__.py:235` and `magpie_tts_server/__main__.py:197`) discards a deliberate operator opt-in with no message. `huggingface_hub` still honors `HF_HUB_ENABLE_HF_TRANSFER` when `hf_transfer` is installed, which is the usual workaround where the Xet CAS endpoints are unreachable.

3. Relatedly, `HF_XET_HIGH_PERFORMANCE=0` is not an off switch for Xet; `HF_HUB_DISABLE_XET` is a separate setting, and [`_docker.py:145`](https://github.com/NVIDIA/xr-ai/blob/6d05a9a85995d87ef28c5121dcfdd9039f6f3a91/utils/xr-ai-vllm/xr_ai_vllm/_docker.py#L145) forwards neither it nor the popped variable into the container. Docker mode therefore has no way to turn Xet off.

4. Adding `hf_xet_high_performance` to the launch payload at [`_docker.py:113`](https://github.com/NVIDIA/xr-ai/blob/6d05a9a85995d87ef28c5121dcfdd9039f6f3a91/utils/xr-ai-vllm/xr_ai_vllm/_docker.py#L113) changes every container fingerprint, so the first launch after this lands stops and recreates every persistent container, reloading models and rebuilding `extra_pip` wheels. That is correct here, since old containers have the `hf_transfer` install baked into their `bash -c`, but it is a one-time multi-minute restart per service worth stating in the description.

5. The `extra_env` branch at [`_docker.py:99`](https://github.com/NVIDIA/xr-ai/blob/6d05a9a85995d87ef28c5121dcfdd9039f6f3a91/utils/xr-ai-vllm/xr_ai_vllm/_docker.py#L99) is redundant: `env_vars.update(extra_env)` already gives `extra_env` precedence in the argv, and the payload already carries it into the fingerprint under its own key.

6. [`piper-tts/pyproject.toml:14`](https://github.com/NVIDIA/xr-ai/blob/6d05a9a85995d87ef28c5121dcfdd9039f6f3a91/services/piper-tts/pyproject.toml#L14) is the one remaining Hugging Face downloader left on `huggingface-hub>=0.22` with no `hf-xet`. Defensible for small ONNX voices, but it is the gap if the intent is Xet everywhere.

7. `_expected_fingerprint` in `tests/test_vllm_docker.py` has no callers and its payload no longer mirrors `build_run_argv`, so reviving it would assert a stale fingerprint. `TestSetupHfEnv` in `tests/test_vllm_lifecycle.py` also leaks `HF_HOME` and `TRANSFORMERS_CACHE` into the rest of the session, since `setup_hf_env` writes them but the test only guards the two variables it monkeypatched.

The pip-mode migration is clean and fixes a real latent bug along the way: `stt-server` was setting `HF_HUB_ENABLE_HF_TRANSFER=1` while never depending on `hf_transfer`, which hard-errors on any hub download. `DEPENDENCIES.md` also regenerates byte-exactly from the generator. The blocker above is about the deployment mode every sample actually uses; it should be resolved before merge.

yanziz-nvidia:
Thanks for the detailed review. All three shipped images already have working Xet stacks, so the reported deployment failure did not reproduce, but the dependency was implicit. This update adds guarded in-container validation/repair with fail-fast behavior, forwards operator controls, documents the one-time container recreation, and fixes the Piper and test gaps. I retained huggingface-hub>=0.32.0 because the flag is consumed by hf-xet/xet-core and is supported by hf-xet==1.1.2; requiring Hub 1.x would conflict with the 26.04 image. Validation on dualada passed all three image checks, 94 focused tests, and 1,368 non-GPU tests.

nvddr:
Please create a follow-up issue to reduce duplication of the environment flag.

yanziz-nvidia:
> Please create a follow-up issue to reduce duplication of the environment flag.

done, created https://github.com/NVIDIA/xr-ai/issues/457

nvddr:
🧹 **Nit — refresh the PR description to match the final revision.**

- The summary mentions vLLM, STT, and Magpie TTS, but this revision also migrates Piper TTS.
- “Without a startup install” is now slightly misleading: the unconditional install is gone, but Docker mode performs a conditional `hf-xet` repair install when validation fails.
- Eight service `pyproject.toml` files changed, rather than seven affected lockfiles.
- The Testing section still has the pre-update validation counts; please replace them with the latest focused/full-suite results.

No code blocker from my re-review.

nvddr (approved):

yanziz-nvidia:
Updated description.

wenxind-nvidia:
## Blockers

None.

## Suggestions

1. [piper_tts_server/__main__.py:330-331](https://github.com/NVIDIA/xr-ai/blob/3861206fec9b6e0fe9ff497326bb36b67e1cd15b/services/piper-tts/piper_tts_server/__main__.py#L330-L331) — Keep Piper’s Xet cache inside model_cache. Piper enables Xet but sets neither HF_HOME nor HF_XET_CACHE; hf-xet therefore stores its chunk cache under ~/.cache/huggingface/xet, outside the configured cache tree. This conflicts with the documented cache ownership at [ai-services.md:35-50](https://github.com/NVIDIA/xr-ai/blob/3861206fec9b6e0fe9ff497326bb36b67e1cd15b/docs/source/components/ai-services.md#L35-L50). Set HF_XET_CACHE to an appropriate model_cache subdirectory before importing Hub, and document that location. The fallback order is defined in [hf-xet v1.1.2](https://github.com/huggingface/xet-core/blob/v1.1.2/data/src/data_client.rs#L37-L47).

2. [ai-services.md:412-418](https://github.com/NVIDIA/xr-ai/blob/3861206fec9b6e0fe9ff497326bb36b67e1cd15b/docs/source/components/ai-services.md#L412-L418) — Clarify that using hf_transfer as a workaround requires both HF_HUB_DISABLE_XET=1 and HF_HUB_ENABLE_HF_TRANSFER=1, plus an installed hf_transfer package. For Xet-backed files, Hub chooses xet_get whenever Xet remains enabled; the legacy flag alone is not a fallback from an unreachable Xet CAS endpoint. See the [Hub 0.32 download selection](https://github.com/huggingface/huggingface_hub/blob/v0.32.0/src/huggingface_hub/file_download.py#L1708-L1725).

## Nits

1. [_docker.py:48-54](https://github.com/NVIDIA/xr-ai/blob/3861206fec9b6e0fe9ff497326bb36b67e1cd15b/utils/xr-ai-vllm/xr_ai_vllm/_docker.py#L48-L54) duplicates >=1.1.2,<2.0.0 in the installation requirement and validation script. Keep the version specification in one constant so a future bump cannot make installation and validation disagree.

Reviewed head: 3861206fec9b6e0fe9ff497326bb36b67e1cd15b

wenxind-nvidia (approved):

yanziz-nvidia:
addressed it 

Piper’s Xet cache now defaults to <model_cache>/piper/xet/ and is documented.
The hf_transfer workaround now specifies both required flags and package installation.
The hf-xet version range has one source constant.

PR #441, opened 2026-08-31
fix(tea): restore reliable visual monitoring controls

nvddr:
## Summary

- restore concise, structured change-watch, video-log, and transcript decisions; treat free-form model prose as a durable classifier failure
- keep raw background observations out of the event viewer while preserving actionable notifications
- route scoped commands such as "stop visual monitoring" to the tea agent instead of treating them as global speech interruptions
- preserve common conversational interruption phrases without allowing an ambiguous partial "stop" to swallow a longer scoped command
- add prompt, event projection, voice-gate, pipeline, and live routing regressions

## Root causes

The background classifiers could return prose without invoking their typed commit tools, which then failed typed validation. Their prompts now require exactly one typed commit; a remaining prose response follows the existing warning and durable error-record path rather than being silently skipped.

Separately, the voice gate classified `stop` followed by any two words as an immediate speech interruption, so "stop visual monitoring" produced the canned acknowledgement without calling `change_watch__stop`. The final-transcript matcher now accepts only known interruption forms, and the partial STT probe waits for final routing when it sees an ambiguous bare `stop`.

## Validation

- focused voice-gate, classifier, event-projection, durable-error, baseline, prompt, and routing tests — 100 passed
- `uv run --project worker python eval/eval.py` from `agent-samples/tea-making-sample` against the running model — all 24 routing cases passed, including the paraphrased `stop-change-watch` case
- live Omni background-classifier sampling — 8/8 change, no-change, irrelevant, viewpoint, and uncertain cases emitted exactly one schema-valid commit tool call with no assistant prose
- `git diff --check`


blongs-nv:
Reviewed at `2e7fe428ca34d9aa213a80f4bd2d280e92d381f5` by a bot.

## Blockers

1. **The narrowed `STOP_RE` stops matching the most common interruption phrases, and in the phrase-gated samples those utterances are dropped entirely.**

   [`_phrases.py:15`](https://github.com/NVIDIA/xr-ai/blob/2e7fe428ca34d9aa213a80f4bd2d280e92d381f5/utils/xr-ai-voicegate/xr_ai_voicegate/_phrases.py#L15) I ran both patterns over an interruption corpus: `stop it`, `stop that`, `wait, stop`, `no, stop`, `stop stop`, `stop stop stop`, `just stop`, and `stop already` all matched on main and none match here, because the new form admits only an enumerated filler set and the suffixes `please|now|talking|speaking`. The tea, lab-instrument, and simple-vlm samples all ship `magic_phrases: [agent, hey agent]`, so a mid-response "stop it" matches neither this pattern nor a magic phrase and falls through to the default drop at [`gate.py:286`](https://github.com/NVIDIA/xr-ai/blob/2e7fe428ca34d9aa213a80f4bd2d280e92d381f5/utils/xr-ai-voicegate/xr_ai_voicegate/gate.py#L286): the agent keeps talking and the utterance is discarded. Allowing repeated `stop` tokens and the pronoun objects `it`/`that` keeps the scoped-command exclusions intact.

2. **The same narrowing disables the early barge-in probe for every voice sample, not just this one.**

   [`vad_stt.py:345`](https://github.com/NVIDIA/xr-ai/blob/2e7fe428ca34d9aa213a80f4bd2d280e92d381f5/agent-sdk/xr-ai-voice/xr_ai_voice/_processors/vad_stt.py#L345) gates the partial-transcript probe that emits an early stop before the full utterance is transcribed. A user shouting "stop stop stop" no longer takes that low-latency path, so the interruption is delayed to the full-transcript path and then, per blocker 1, may not arrive at all. This consumer sits outside `xr-ai-voicegate` and outside this sample.

3. **A classifier that answers in prose now fails silently instead of leaving a record.**

   [`change_watch.py:379`](https://github.com/NVIDIA/xr-ai/blob/2e7fe428ca34d9aa213a80f4bd2d280e92d381f5/agent-samples/tea-making-sample/worker/tea_making_worker/change_watch.py#L379) and [`video_log.py:335`](https://github.com/NVIDIA/xr-ai/blob/2e7fe428ca34d9aa213a80f4bd2d280e92d381f5/agent-samples/tea-making-sample/worker/tea_making_worker/video_log.py#L335) log at debug, publish nothing, and leave the caption unappended. Before this change the prose reached `model_validate_json`, raised, and was caught into a durable error record. A model that persistently skips the commit tool, the exact failure this PR is fixing, now disables monitoring while `change_watch__status` still reports it running, with nothing in the artifacts and nothing in the log at default level.

4. **Broadening `_is_no_change` to a prefix match suppresses deltas that carry real information.**

   [`video_log.py:73`](https://github.com/NVIDIA/xr-ai/blob/2e7fe428ca34d9aa213a80f4bd2d280e92d381f5/agent-samples/tea-making-sample/worker/tea_making_worker/video_log.py#L73) went from exact equality to `startswith`, so "No meaningful visual change, but a person entered the room." is now classified as no-change and its fact never published. With the new viewer filtering, facts are the only route to the pane. The sibling classifier already solves this structurally: [`ChangeDecision`](https://github.com/NVIDIA/xr-ai/blob/2e7fe428ca34d9aa213a80f4bd2d280e92d381f5/agent-samples/tea-making-sample/worker/tea_making_worker/change_watch.py#L61) carries a typed `important: bool` and the code branches on the boolean, so `VideoDelta` can carry the same field and the sentinel can go.

## Suggestions

1. The prose-parse guard was applied to two of the three sites that share the pattern. [`transcript.py:359`](https://github.com/NVIDIA/xr-ai/blob/2e7fe428ca34d9aa213a80f4bd2d280e92d381f5/agent-samples/tea-making-sample/worker/tea_making_worker/transcript.py#L359) runs the identical `run_tool_loop` plus `model_validate_json(result.content)` with no `return_direct` check and no prompt rework, so the same prose response still produces an error record there.

2. [`tea-making-sample.md:313`](https://github.com/NVIDIA/xr-ai/blob/2e7fe428ca34d9aa213a80f4bd2d280e92d381f5/docs/source/reference/tea-making-sample.md#L313) now claims every-frame captions, no-change observations, unavailable frames, and errors "remain in the durable records and worker log", but the skip path this PR adds writes to neither at default log level. Either the skip should leave a record or the sentence should name it.

3. The new `stop-change-watch` eval case does not exercise the path that was broken: [`eval/cases.yaml:27`](https://github.com/NVIDIA/xr-ai/blob/2e7fe428ca34d9aa213a80f4bd2d280e92d381f5/agent-samples/tea-making-sample/eval/cases.yaml#L27) runs through `_prepare_route` and `llm.chat` with no voice gate involved, so it passes on main too. Its query "Stop visual monitoring." is also near-verbatim to the `change_watch__stop` description "Stop visual change monitoring." ([`change_watch.py:155`](https://github.com/NVIDIA/xr-ai/blob/2e7fe428ca34d9aa213a80f4bd2d280e92d381f5/agent-samples/tea-making-sample/worker/tea_making_worker/change_watch.py#L155)), where the sibling `start-change-watch` case deliberately paraphrases. A paraphrase would measure routing rather than lexical overlap.

4. [`video_log.py:308`](https://github.com/NVIDIA/xr-ai/blob/2e7fe428ca34d9aa213a80f4bd2d280e92d381f5/agent-samples/tea-making-sample/worker/tea_making_worker/video_log.py#L308) writes the first frame to the durable record with a synthesized `delta="No meaningful visual change."`, but a first frame is a baseline, not an absence of change. `change_watch` already models this with a [`"baseline"` record type](https://github.com/NVIDIA/xr-ai/blob/2e7fe428ca34d9aa213a80f4bd2d280e92d381f5/agent-samples/tea-making-sample/worker/tea_making_worker/change_watch.py#L341).

5. That same line is a third copy of the sentinel, alongside `_NO_CHANGE` and the sentence the prompt dictates; deriving it from `_NO_CHANGE` keeps the producer and the matcher from drifting apart.

6. Neither the first-frame path nor the change-watch lifecycle filter in `web_events.py` has a test: both can be deleted with the suite still green. The video-log twin of the filter is covered, so the pattern to copy is already there.

7. The new `_phrases.py` branches have no positive tests. Deleting the entire filler-prefix group leaves every voicegate test passing, and none of the phrase shapes the rewrite changed ("stop it", "stop stop stop", "please stop") is pinned in either direction.

8. [`web_events.py:69`](https://github.com/NVIDIA/xr-ai/blob/2e7fe428ca34d9aa213a80f4bd2d280e92d381f5/agent-samples/tea-making-sample/worker/tea_making_worker/web_events.py#L69) returns silently for any `application` outside the two mapped values, so a future `BackgroundFact` producer disappears from the viewer with no log line.

9. The prompt assertions at [`test_tea_making_sample.py:1478`](https://github.com/NVIDIA/xr-ai/blob/2e7fe428ca34d9aa213a80f4bd2d280e92d381f5/tests/test_tea_making_sample.py#L1478) pin exact prompt wording as a proxy for behavior; they break on innocent rewording and catch no regression.

Gating the classifiers on the typed commit tool is the right fix for the prose problem, and moving the voice gate to a whitelist is the right generalization: scoped commands reach the agent without leaking any sample's intent into shared code. The blockers above should be resolved before merge.

wenxind-nvidia:
## Blockers

1. **Always-on scoped stop commands can still be intercepted by the partial STT probe.**

   `agent-sdk/xr-ai-voice/xr_ai_voice/_processors/vad_stt.py:345` applies `STOP_RE` every 250 ms. While saying `stop visual monitoring`, an intermediate transcript of just `stop` triggers `_emit_early_stop`, suppressing the final utterance before VoiceGate sees the scoped command. This affects the supported always-on preset. Please use a separate partial-probe matcher that does not immediately accept ambiguous bare `stop`, and add a pipeline regression test with partial `stop` followed by final `stop visual monitoring`.

2. **Missing commit-tool calls now silently remove captions from the durable record.**

   `agent-samples/tea-making-sample/worker/tea_making_worker/change_watch.py:379` and `video_log.py:335` return after a debug message without publishing the caption or an error record. Previously, invalid model output reached the exception path and produced a durable error containing the caption. This contradicts the new documentation durable-record guarantee. Please publish a controlled `error` record containing the caption and add `_observe`-level coverage.

## Suggestions

1. `web_events.py:84` and `web_events.py:107` also filter `unavailable` and `error` records. A monitor can therefore remain visibly `started` while its camera or model is failing. Consider filtering routine observations while retaining concise operational failure states.

2. `_phrases.py:18` no longer recognizes common interrupt forms previously accepted, including `stop it`, `stop this`, and `I said stop`. Consider preserving intentional conversational forms while excluding application lifecycle commands.

3. `video_log.py:73` uses `startswith` for no-change detection. That can suppress a contradictory result such as `No meaningful visual change, but a person entered.` Prefer a structured boolean or an anchored set of accepted no-change responses.

## Verification

Reviewed at `2e7fe428ca34d9aa213a80f4bd2d280e92d381f5`. Ruff and `git diff --check` passed. Focused pytest resolution was blocked locally because `PyNvVideoCodec==2.2.2` has no macOS ARM wheel; GitHub checks are green.

nvddr:
@blongs-nv @wenxind-nvidia Thanks for the detailed review. Addressed at `a44474ae4c94e3d5ee8808e26085e78734233164`:

- preserved common interruption forms (`stop it/that/this`, repeated `stop`, `wait/no/just ... stop`, `I said stop`, and related polite forms) while keeping scoped application commands out of the global-stop path;
- added a distinct partial-probe rule so an ambiguous partial bare `stop` waits for the final transcript instead of suppressing a longer command such as `stop visual monitoring`;
- removed the new prose-skip guards. Missing commit-tool calls now remain real model failures and flow through the existing warning plus durable error-record paths with the caption attached;
- tightened the transcript summarizer to the same tool-only contract rather than adding inconsistent prose hiding there;
- replaced prefix-based no-change detection with a complete normalized allowlist and covered contradictory suffixes;
- records the first video frame as a baseline observation with no synthesized delta, without adding a new event type;
- paraphrased the foreground stop-watch eval and added focused coverage for the first-frame path, durable classifier failures, change-watch viewer filtering, interruption forms, and partial-stop preservation.

Kept the viewer's intentional concise presentation policy: routine observations, unavailable frames, and errors remain in durable records/logs rather than repeatedly filling the live panes. Also retained the closed background-fact presentation whitelist and the prompt contract assertions.

Validation: 100 focused tests passed, all 24 live model routing evals passed, and `git diff --check` passed. The PR is now rebased onto and targets `release/beta-v0.2.0rc1`.

blongs-nv:
Reviewed at `503ab1df3163a36567931b3d7e09250b7904eec0` by a bot.

All four blockers from my previous review are resolved. I re-ran the interruption corpus against the new `STOP_RE`: `stop it`, `stop that`, `stop this`, `wait, stop`, `no, stop`, `stop stop`, `stop stop stop`, `just stop`, `stop already`, and `I said stop` all match again, while every scoped command still routes to the agent. `_is_unambiguous_partial_stop` keeps a bare partial `stop` out of the early-stop path. Removing the prose-skip guards outright is a better answer than the warning I asked for, and it makes the durable-record sentence in the reference page true again. The no-change allowlist, the baseline first-frame record, and the paraphrased eval query all address what I raised.

## Suggestions

1. [`test_voice_pipeline.py:632`](https://github.com/NVIDIA/xr-ai/blob/503ab1df3163a36567931b3d7e09250b7904eec0/tests/test_voice_pipeline.py#L632) raises a bare `asyncio.sleep` from 0.05 to 1.0 to get `test_vad_stt_bare_partial_stop_preserves_scoped_final_command` green. A longer fixed sleep moves the flake rather than removing it, and it is the regression test for the partial-stop fix, so a loaded CI box can turn that guarantee back off. The file already waits on events for exactly this: `stt.first_started.wait()` at [line 747](https://github.com/NVIDIA/xr-ai/blob/503ab1df3163a36567931b3d7e09250b7904eec0/tests/test_voice_pipeline.py#L747) and `stt.probe_started.wait()` at [line 817](https://github.com/NVIDIA/xr-ai/blob/503ab1df3163a36567931b3d7e09250b7904eec0/tests/test_voice_pipeline.py#L817).

2. The filler set at [`_phrases.py:19`](https://github.com/NVIDIA/xr-ai/blob/503ab1df3163a36567931b3d7e09250b7904eec0/utils/xr-ai-voicegate/xr_ai_voicegate/_phrases.py#L19) is a closed enumeration, so lead-ins outside it still lose interrupt status relative to the previous pattern: `alright stop` used to match and no longer does, with `sorry`, `whoa`, and `hang on` in the same family. The same enumeration is duplicated at [`vad_stt.py:57`](https://github.com/NVIDIA/xr-ai/blob/503ab1df3163a36567931b3d7e09250b7904eec0/agent-sdk/xr-ai-voice/xr_ai_voice/_processors/vad_stt.py#L57), so each new phrasing needs adding twice.

Splitting the ambiguity rule out from `STOP_RE` rather than overloading one pattern is the right shape, and the parametrized interruption cases pin the forms that were previously unguarded. Nothing here blocks the merge.

blongs-nv (approved):

wenxind-nvidia (commented):
## Suggestions

1. `utils/xr-ai-voicegate/xr_ai_voicegate/_phrases.py:20` — The narrowed suffix list drops conversational interruptions accepted by the base branch, including `stop right now`, `stop doing that`, and `stop for now`. Wake-phrase configurations discard these utterances, leaving the agent speaking. Please add explicit legacy conversational forms and regression tests.

2. `agent-sdk/xr-ai-voice/xr_ai_voice/_processors/vad_stt.py:61` — `STOP_RE` recognizes `stop, stop`, but token counting sees `"stop,"` and only one exact `"stop"`, so the partial probe waits for final VAD instead of interrupting immediately. Please normalize punctuation on every token or count `stop` using word boundaries, with a punctuated repeated-stop test.

Reviewed head: 216dcf433ad2c1a7f665df577c0002675ada02f3

wenxind-nvidia (approved):

PR #454, opened 2026-08-31
fix(tea): ground idle visual questions

nvddr:
## Summary

- scope current-scene routing to idle turns without changing active-guide behavior
- require `current_view` before answering questions about the current scene, surroundings, or something currently visible
- align both private `current_view` tool descriptions with broad and specific scene questions
- distinguish current spatial context from observations recorded by background applications
- add routing evals for open scene, current visual-reference, and visible-background questions

## Motivation

In an idle-state run, the foreground model sometimes promised a visual check without calling `current_view`, and sometimes routed a question about the visible background to recorded application context. The foreground LLM receives the camera image only through `current_view`, so the idle prompt now states that current-scene questions are in scope while requiring the tool before answering.

The change does not claim general past-scene memory. `application_context__query` can retrieve only textual facts produced by active background applications.

## Validation

- `5 passed` — focused tea foreground tests
- `27 passed` — live-model routing eval against the local Omni endpoint
- Ruff passed
- `git diff --check` passed

### DIFF
--- a/agent-samples/tea-making-sample/eval/cases.yaml
+++ b/agent-samples/tea-making-sample/eval/cases.yaml
@@ -13,7 +13,7 @@
   expected_tool: current_view
   forbidden_response: I can only help with the active tea guide right now.
 - name: idle-routes-open-ended-scene-question
-  query: What do you see?
+  query: Describe my surroundings.
   expected_tool: current_view
 - name: idle-routes-visual-reference
   query: What is this?
--- a/agent-samples/tea-making-sample/eval/cases.yaml
+++ b/agent-samples/tea-making-sample/eval/cases.yaml
@@ -12,6 +12,15 @@
   query: What color is the shirt?
   expected_tool: current_view
   forbidden_response: I can only help with the active tea guide right now.
+- name: idle-routes-open-ended-scene-question
+  query: What do you see?
+  expected_tool: current_view
+- name: idle-routes-visual-reference
+  query: What is this?
+  expected_tool: current_view
+- name: idle-routes-visible-background
+  query: What is in the background?
+  expected_tool: current_view
 - name: start-tea-guide
   query: Help me make a cup of tea.
   expected_tool: workflow__start
--- a/agent-samples/tea-making-sample/worker/tea_making_worker/foreground.py
+++ b/agent-samples/tea-making-sample/worker/tea_making_worker/foreground.py
@@ -324,7 +324,7 @@ class ForegroundAgent(Agent):
 
         return Tool(
             "current_view",
-            "Inspect this participant's current camera frame for a specific visible fact.",
+            "Inspect this participant's current camera frame to answer a question about the current scene.",
             CurrentViewRequest,
             ImageQueryResult,
             inspect,
--- a/agent-samples/tea-making-sample/worker/tea_making_worker/prompts/foreground_idle.txt
+++ b/agent-samples/tea-making-sample/worker/tea_making_worker/prompts/foreground_idle.txt
@@ -1,3 +1,6 @@
 The tea guide is idle. Act as a general-purpose assistant and answer unrelated
-questions normally. Start tea guidance only when the user explicitly asks to
-make or start tea. Do not apply restrictions that belong to an active guide.
+questions normally. Questions about the current scene, surroundings, or a
+visual reference are in scope. You cannot see the scene without calling
+current_view, so call it before answering such a request. Start tea guidance
+only when the user explicitly asks to make or start tea. Do not apply
+restrictions that belong to an active guide.
--- a/agent-samples/tea-making-sample/worker/tea_making_worker/prompts/foreground_idle.txt
+++ b/agent-samples/tea-making-sample/worker/tea_making_worker/prompts/foreground_idle.txt
@@ -1,6 +1,6 @@
 The tea guide is idle. Act as a general-purpose assistant and answer unrelated
 questions normally. Questions about the current scene, surroundings, or a
-visual reference are in scope. You cannot see the scene without calling
-current_view, so call it before answering such a request. Start tea guidance
-only when the user explicitly asks to make or start tea. Do not apply
-restrictions that belong to an active guide.
+reference to something currently visible are in scope. You cannot see the scene
+without calling current_view, so call it before answering such a request. Start
+tea guidance only when the user explicitly asks to make or start tea. Do not
+apply restrictions that belong to an active guide.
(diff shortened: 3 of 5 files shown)


blongs-nv:
Reviewed at `2655289d0cad9774654bd4c8d54aedb97e97e0ea` by a bot.

## Blockers

1. **The new mandate sits in the shared route policy, so it now argues against the active route's decline rule and the eval case that pins it.**

   [`foreground.py:287`](https://github.com/NVIDIA/xr-ai/blob/2655289d0cad9774654bd4c8d54aedb97e97e0ea/agent-samples/tea-making-sample/worker/tea_making_worker/foreground.py#L287) prepends this prompt to both `_IDLE_PROMPT` and `_ACTIVE_PROMPT`, so "Any broad or specific request about the current scene or a visual reference must call current_view before any response" also governs the active route, where [`foreground_active.txt:3`](https://github.com/NVIDIA/xr-ai/blob/2655289d0cad9774654bd4c8d54aedb97e97e0ea/agent-samples/tea-making-sample/worker/tea_making_worker/prompts/foreground_active.txt#L3) says to use visual tools "only for facts relevant to the active tea guide" and to decline unrelated requests "without answering them, calling a tool". [`cases.yaml:100`](https://github.com/NVIDIA/xr-ai/blob/2655289d0cad9774654bd4c8d54aedb97e97e0ea/agent-samples/tea-making-sample/eval/cases.yaml#L100) requires exactly that decline for "What color is the person's shirt?", and `current_view` is in the active tool set at several steps ([`workflow.yaml:124`](https://github.com/NVIDIA/xr-ai/blob/2655289d0cad9774654bd4c8d54aedb97e97e0ea/agent-samples/tea-making-sample/yaml/workflow.yaml#L124)), so the model can obey either sentence and nothing states precedence. The failure being fixed is idle-route; scoping the rule there keeps it from fighting the active policy.

2. **"The current camera view is inherent conversation context" is not true of the context the model receives, and it points at the failure this PR exists to fix.**

   [`foreground.py:229`](https://github.com/NVIDIA/xr-ai/blob/2655289d0cad9774654bd4c8d54aedb97e97e0ea/agent-samples/tea-making-sample/worker/tea_making_worker/foreground.py#L229) sends the foreground model a system message and a user message, both text; the frame reaches only the VLM inside the `current_view` tool. Telling the model the view is already context invites the exact behavior the next sentence forbids, answering as though it can see without calling the tool. Stating it as scope rather than context ("scene questions are always in scope, but you cannot see the scene without calling current_view") carries the intent without the false premise.

## Suggestions

1. `current_view`'s own description still reads "Inspect this participant's current camera frame for a specific visible fact" ([`foreground.py:327`](https://github.com/NVIDIA/xr-ai/blob/2655289d0cad9774654bd4c8d54aedb97e97e0ea/agent-samples/tea-making-sample/worker/tea_making_worker/foreground.py#L327), [`workflow_tools.py:151`](https://github.com/NVIDIA/xr-ai/blob/2655289d0cad9774654bd4c8d54aedb97e97e0ea/agent-samples/tea-making-sample/worker/tea_making_worker/workflow_tools.py#L151)). The model sees the description and the prompt together, and the new rule now claims broad questions like "What do you see?" that the description scopes out.

2. The narrowed `application_context__query` rule collides with that tool's description the same way: the prompt says "only when explicitly asked about recorded or past background observations", while [`background_context.py:69`](https://github.com/NVIDIA/xr-ai/blob/2655289d0cad9774654bd4c8d54aedb97e97e0ea/agent-samples/tea-making-sample/worker/tea_making_worker/background_context.py#L69) says "Use only when those facts help answer the current request". Two different "only" scopes reach the model at once.

3. All three new cases are idle-route, so the widened rule's effect on the active route is unpinned. An active-route case asserting the decline still holds would cover the axis the change actually broadened.

4. Dropping "without calling current_view" from the general-knowledge sentence moves the over-call direction closer to the decision boundary, while the new absolute pushes the same way. `idle-answers-unrelated-general-knowledge` pins one phrasing of it; the three new cases likewise pin three phrasings rather than the rule, so a paraphrase such as "Describe my surroundings" is claimed by the rule but untested.

5. The fixed failure is still only detectable at eval time. [`ForegroundRecord`](https://github.com/NVIDIA/xr-ai/blob/2655289d0cad9774654bd4c8d54aedb97e97e0ea/agent-samples/tea-making-sample/worker/tea_making_worker/foreground.py#L174) already carries the query, the response, and the per-turn tool-call list, which is what a runtime guard or log-time detector would need to flag an answer that describes the scene on a turn with no tool call.

The eval additions are honest: generic phrasings with no eval-specific entities leaking into the prompt, and `idle-routes-visible-background` deliberately stresses the narrowed `application_context__query` rule rather than confirming it. The blockers above are about where the new rule lives and what it asserts, not about the routing intent. They should be resolved before merge.

nvddr:
@blongs-nv Thanks for the detailed review. Addressed at `9167fb0e264ac8e05943f6fa3cacc3e9df3b3dab`:

- moved the camera-routing mandate out of the shared prompt and into the idle-route prompt, so it no longer conflicts with the active guides unrelated-request policy;
- replaced the false claim that the camera image is already model context with the accurate contract: scene questions are in scope, but the model must call `current_view` to see the scene;
- aligned both private `current_view` tool descriptions with broad as well as specific scene questions;
- clarified `application_context__query` as recall of observations recorded by background applications, removing the spatial-background ambiguity without a worked example;
- changed the open-ended eval to the paraphrased `Describe my surroundings.` case. The existing active unrelated-visual eval continues to pin the opposite active-route behavior.

I did not add a runtime guard in this prompt-focused PR; that would introduce a separate enforcement policy beyond the reported routing regression.

Validation after rebasing onto `release/beta-v0.2.0rc1`: all 27 live-model routing cases passed, 5 focused foreground tests passed, Ruff passed, and `git diff --check` passed.

wenxind-nvidia (commented):
## Suggestions

1. `agent-samples/tea-making-sample/worker/tea_making_worker/prompts/foreground_idle.txt:2` — “A visual reference” is not explicitly limited to something visible now. A request such as “What was that object I showed you earlier?” can match both the `current_view` and recorded-observation rules. Please qualify the reference as current and add a natural past-scene eval expecting `application_context__query`; the existing case explicitly names “background watchers” and does not test this boundary.

## Nits

1. The PR description is stale after `9167fb0`: it still says the fix remains in the shared prompt and reports four focused tests. The implementation now scopes the rule to the idle prompt, updates two tool descriptions, and the author reports five focused tests.

Reviewed head: 9167fb0e264ac8e05943f6fa3cacc3e9df3b3dab

wenxind-nvidia (approved):

yanziz-nvidia (approved):

nvddr:
@wenxind-nvidia Addressed the supported part at `9da1c2b7`: the idle policy now applies only to a reference to something currently visible. I also updated the stale PR description.

I did not add a past-scene eval expecting `application_context__query`. The sample has no general past-frame memory: `current_view` reads only the latest frame, while application context contains only bounded textual facts emitted when `change_watch`, `video_log`, or `transcript` is active. A natural past-scene query therefore is not reliably answerable and that eval would assert a capability the sample does not provide.

Validation: all 27 live-model routing cases and 5 focused foreground tests pass; `git diff --check` passes.

PR #455, opened 2026-09-01
fix(voice): preserve final intent after stop probe

nvddr:
## Summary

- treat a stop-like partial STT result only as an early output interruption
- classify partial STOP on raw text and on the tail after a configured wake phrase, even when the listening chime is disabled
- keep bounded probing after wake acknowledgement so later partial STT can revise `hey agent` or `hey agent st` to `hey agent stop`
- keep chime playback presentation-only; STOP classification alone terminates probing and `VadSttProcessor` alone emits early interruption
- keep the completed transcript authoritative for global-stop versus scoped-command routing
- keep a closed global-STOP grammar while accepting punctuation in approved forms such as `I said, stop` and `can you, stop`
- document the exact partial/final routing boundary and cover it through the VAD-to-gate pipeline

## Intent boundary

Direct imperatives such as `stop`, `stop it`, `stop talking`, `be quiet`, and wake-prefixed equivalents such as `hey agent stop` interrupt active output early. Negations, questions, reported speech, unconfigured arbitrary prefixes, and scoped or multi-action commands such as `stop monitoring xyz` are not global stops.

A partial `hey agent stop` may interrupt playback immediately. If final STT revises it to `hey agent stop monitoring xyz`, the final transcript wins and dispatches `stop monitoring xyz` to the agent; the early interruption is not treated as committed intent.

## Why

Partial STT hypotheses are revisable. Treating a stop-like partial as the final command can discard the user's completed scoped request, while waiting for final STT makes genuine stop commands feel unresponsive. The split behavior preserves low-latency interruption without replacing final routing.

With wake phrases configured, the bounded probe can issue up to three partial STT requests and finalization issues the authoritative request. Chime configuration, initialization, and playback do not affect that control flow.

This follow-up is separate from merged #441: it corrects stop-probe routing and its gate-aware edge cases without redesigning the voice pipeline or expanding its public API.

## Validation

- `uv run --project tests pytest -q tests/test_voice_pipeline.py tests/test_xr_ai_voicegate.py` — 228 passed
- `uv run --project tests pytest -q tests/test_docs_versions.py tests/test_config_documentation.py tests/test_cli_documentation.py` — 26 passed
- `ruff 0.15.16 check .` — passed
- `git diff --check` — passed

### DIFF
--- a/agent-samples/simple-vlm-example/yaml/voice_gate.yaml
+++ b/agent-samples/simple-vlm-example/yaml/voice_gate.yaml
@@ -8,7 +8,9 @@
 # with one of these or contain one after '.', '?', or '!' followed by
 # whitespace or a closing quote. Matching is case-insensitive. Text through
 # the matched phrase is stripped before the agent sees the query; commas and
-# other mid-sentence mentions do not match.
+# other mid-sentence mentions do not match. The same stripped tail is used for
+# early global-STOP detection; the complete final transcript still determines
+# whether a stop-like opening is a global stop or a scoped agent command.
 magic_phrases:
   - "agent"
   - "hey agent"
--- a/agent-sdk/xr-ai-voice/xr_ai_voice/_pipeline.py
+++ b/agent-sdk/xr-ai-voice/xr_ai_voice/_pipeline.py
@@ -78,7 +78,7 @@ def _build_voice_pipeline(
         vad_cfg=vad_cfg,
         on_partial_transcript=(
             voice_gate_proc.handle_partial_transcript
-            if voice_gate_proc.early_wake_ack_enabled
+            if voice_gate_cfg.magic_phrases
             else None
         ),
         on_final_transcript=on_final_transcript,
--- a/agent-sdk/xr-ai-voice/xr_ai_voice/_processors/vad_stt.py
+++ b/agent-sdk/xr-ai-voice/xr_ai_voice/_processors/vad_stt.py
@@ -46,7 +46,7 @@ from xr_ai_voicegate._phrases import STOP_RE
 from .._frames import ParticipantLeftFrame
 
 
-# True means the gate handled the partial; False requests another bounded probe.
+# True means the gate matched STOP; False requests another bounded probe.
 PartialTranscriptHandler = Callable[[str, str], Awaitable[bool]]
 FinalTranscriptHandler = Callable[[str, str, int], Awaitable[None]]
 _MAX_PARTIAL_PROBES = 3
@@ -64,8 +64,9 @@ class VadConfig:
     ``stop_probe_after_s`` — cadence in seconds for up to three STT probes of
     the partial audio buffer. This gives STOP commands a fast interruption path
     without replacing final transcription, and lets a configured wake phrase
-    be acknowledged before the utterance ends. Set to ``0`` or negative to
-    disable probes.
+    be acknowledged before the utterance ends. A phrase-gated utterance can
+    make up to three partial STT requests plus the final request. Set to ``0``
+    or negative to disable probes.
     """
     silence_duration:   float = 0.8
     """Seconds of silence that finalize an utterance."""
@@ -339,7 +340,8 @@ class VadSttProcessor(FrameProcessor):
                         logger.exception("partial-transcript handler failed pid={!r}", pid)
                         return
                     if decision is True:
-                        logger.info("early partial transcript handled pid={!r}", pid)
+                        logger.info("gate-aware partial STOP matched pid={!r}", pid)
+                        await self._emit_early_interruption(pid, text)
                         return
             finally:
                 self._probe_inflight.discard(pid)
--- a/agent-sdk/xr-ai-voice/xr_ai_voice/_processors/vad_stt.py
+++ b/agent-sdk/xr-ai-voice/xr_ai_voice/_processors/vad_stt.py
@@ -53,15 +53,6 @@ _PARTIAL_PROBE_TAIL_S = 0.12
 _PARTIAL_PROBE_FINISH_GRACE_S = 0.15
 
 
-def _is_unambiguous_partial_stop(text: str) -> bool:
-    """Reserve an ambiguous bare STOP for final-transcript routing."""
-
-    if not STOP_RE.match(text):
-        return False
-    words = text.rstrip(" \t\r\n.!?").casefold().split()
-    return not (words[-1] == "stop" and words.count("stop") == 1)
-
-
 @dataclass(frozen=True)
 class VadConfig:
     """Tuning knobs for the Silero-VAD utterance detector.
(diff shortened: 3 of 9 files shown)


wenxind-nvidia:
## Blocker

1. [`vad_stt.py:321`](https://github.com/NVIDIA/xr-ai/blob/04a7a1c904dd68d6983a0ac07bf1a6d8501b3544/agent-sdk/xr-ai-voice/xr_ai_voice/_processors/vad_stt.py#L321-L341) — **Wake-prefixed STOP commands lose the early interruption path.** The new closed [`STOP_RE`](https://github.com/NVIDIA/xr-ai/blob/04a7a1c904dd68d6983a0ac07bf1a6d8501b3544/utils/xr-ai-voicegate/xr_ai_voicegate/_phrases.py#L15-L40) does not recognize `agent stop` or `hey agent stop`. Those are the configured forms in the phrase-gated samples. Instead, `handle_partial_transcript()` recognizes the wake phrase, plays the chime, returns `True`, and terminates further probes without emitting an `InterruptionFrame`. The active response therefore continues until VAD finalization and final STT, whereas the base pattern accepted the one- and two-word wake prefixes and interrupted immediately. Make partial-stop classification gate-aware—such as testing the magic-phrase-stripped tail—and add full-pipeline regressions for `agent stop`, `hey agent stop`, and a revised final command such as `hey agent stop monitoring xyz`.

## Suggestion

1. [`_phrases.py:21`](https://github.com/NVIDIA/xr-ai/blob/04a7a1c904dd68d6983a0ac07bf1a6d8501b3544/utils/xr-ai-voicegate/xr_ai_voicegate/_phrases.py#L21-L22) — Permit punctuation after the approved multiword prefixes. STT output such as `I said, stop` and `can you, stop` no longer matches, although the unpunctuated versions do and the base pattern accepted both. Using `[,\s]+` after `said` and `you` would preserve the closed grammar without restoring arbitrary filler words.

## Nit

1. [`test_voice_pipeline.py:973`](https://github.com/NVIDIA/xr-ai/blob/04a7a1c904dd68d6983a0ac07bf1a6d8501b3544/tests/test_voice_pipeline.py#L973-L978) and [`test_voice_pipeline.py:999`](https://github.com/NVIDIA/xr-ai/blob/04a7a1c904dd68d6983a0ac07bf1a6d8501b3544/tests/test_voice_pipeline.py#L999-L1000) still describe the removed suppression flag and say final transcription occurs only if suppression fails. Final transcription is now intentional and authoritative; update these comments accordingly.

Reviewed head: `04a7a1c904dd68d6983a0ac07bf1a6d8501b3544`

nvddr:
@wenxind-nvidia Addressed all three items on `3f4f0927`.

### Disposition

- **Blocker — addressed.** Partial STOP classification now checks both raw STT text and the tail after the configured wake phrase. The internal partial handler is wired whenever wake phrases exist, independently of the optional listening chime. A partial emits only the early `InterruptionFrame`; final STT remains authoritative for global-stop versus scoped-command routing.
- **Blocker regressions — added.** Full VAD-to-gate pipeline coverage now exercises `agent stop`, `hey agent stop`, and partial `hey agent stop` revised to final `hey agent stop monitoring xyz`. The first two interrupt early and complete as global stops; the revised final dispatches `stop monitoring xyz` while retaining the early interruption.
- **Suggestion — addressed.** Approved multiword prefixes now accept punctuation after `said` and `you`; `I said, stop` and `can you, stop` are covered.
- **Nit — addressed.** The two stale test comments now describe intentional final transcription and no longer refer to the removed suppression flag.
- **Documentation — addressed.** The public voice-gate docstring, `VadConfig`/processor guidance, canonical voice reference, sample configuration comment, and PR description now state the closed global-STOP grammar, excluded categories, wake-stripped classification, chime independence, and final-transcript authority.
- **Deferred:** none.

### Validation

- 224 focused voice/voice-gate tests passed locally
- 26 documentation/configuration contract tests passed locally
- repository-wide Ruff and `git diff --check` passed locally
- all GitHub checks pass on `3f4f0927`, including strict docs, Python 3.11/3.12, CodeQL, DCO, SPDX, dependency map, lock validation, and docs publish

wenxind-nvidia:
## Suggestions

1. [voice_gate.py:115](https://github.com/NVIDIA/xr-ai/blob/3f4f0927dd41abdb28f3699a7e171c5576e56dc4/agent-sdk/xr-ai-voice/xr_ai_voice/_processors/voice_gate.py#L115) calls the private [VoiceGate._matches_partial_stop()](https://github.com/NVIDIA/xr-ai/blob/3f4f0927dd41abdb28f3699a7e171c5576e56dc4/utils/xr-ai-voicegate/xr_ai_voicegate/gate.py#L183-L187) across package boundaries. Make this a supported public method, or keep the complete partial-classification operation behind a public VoiceGate API, rather than coupling xr-ai-voice to an underscore-private implementation detail.

2. [voice_gate.py:120-122](https://github.com/NVIDIA/xr-ai/blob/3f4f0927dd41abdb28f3699a7e171c5576e56dc4/agent-sdk/xr-ai-voice/xr_ai_voice/_processors/voice_gate.py#L120-L122) — Keep bounded probing after a wake acknowledgement so an appended STOP can still interrupt. With staged STT such as first probe “hey agent” and second probe “hey agent stop,” the first result matches the wake phrase and returns True, causing [_run_partial_probes()](https://github.com/NVIDIA/xr-ai/blob/3f4f0927dd41abdb28f3699a7e171c5576e56dc4/agent-sdk/xr-ai-voice/xr_ai_voice/_processors/vad_stt.py#L333-L343) to exit before seeing STOP. This also occurs when the listening chime is disabled or not initialized because chime emission can be a no-op while the handler still returns True. Add staged cases such as [“hey agent”, “hey agent stop”] and [“hey agent st”, “hey agent stop”], with the chime enabled and disabled.

wenxind-nvidia (approved):

blongs-nv:
Reviewed at `3f4f0927dd41abdb28f3699a7e171c5576e56dc4` by a bot.

## Suggestions

1. [`_phrases.py:20`](https://github.com/NVIDIA/xr-ai/blob/3f4f0927dd41abdb28f3699a7e171c5576e56dc4/utils/xr-ai-voicegate/xr_ai_voicegate/_phrases.py#L20) admits `hang on` but not `hold on`, which is the same interjection and does not match; `never mind` and `excuse me` are in the same family. Now that the lead-in set is a closed grammar, each of these needs its own entry and case rather than falling out of a general rule, so it is worth deciding the intended breadth once rather than per report.

2. Keeping the completed transcript authoritative means a stop-like utterance is transcribed twice, once by the probe at [`vad_stt.py:323`](https://github.com/NVIDIA/xr-ai/blob/3f4f0927dd41abdb28f3699a7e171c5576e56dc4/agent-sdk/xr-ai-voice/xr_ai_voice/_processors/vad_stt.py#L323) and again at VAD finalization, where the suppressed path transcribed once. That is the right trade for correctness, but it is a real per-utterance cost against a hosted STT endpoint and belongs in the `stop_probe_after_s` guidance next to the rest of the contract.

Treating a revisable partial as an interruption rather than a commitment is the right correction, and it removes more code than it adds. The closed grammar is also a genuine correctness gain over the open two-word prefix it replaces, beyond the wake-prefix and punctuation cases already raised: I ran both patterns over a negation corpus, and `don't stop`, `do not stop`, `never stop`, `please don't stop`, `should I stop`, `we should stop`, `you said stop`, `say stop`, `the word stop`, `don't shut up`, `do not be quiet`, and `not quiet` all matched under the merged pattern and correctly do not now, while every canonical interruption form still matches and `stop doing it`, `stop right now`, `stop for now`, `I said, stop`, and `can you, stop` are newly covered. No blockers.

blongs-nv (approved):

nvddr:
@wenxind-nvidia @blongs-nv Follow-up disposition for `8003bd96`:

- **Continued probing after wake acknowledgement — addressed.** Wake recognition may play the optional chime but now returns “continue”; only a gate-aware STOP terminates bounded probing. Staged full-pipeline cases cover `hey agent` → `hey agent stop` and `hey agent st` → `hey agent stop`, each with the chime enabled and disabled.
- **Interruption ownership — clarified.** `VadSttProcessor` is again the sole owner of early `InterruptionFrame` emission for both raw and wake-stripped STOP matches. Chime configuration, initialization, and playback no longer control probe termination.
- **Bounded STT cost — addressed.** `VadConfig` and the canonical voice reference now state that phrase-gated input can make up to three partial STT requests plus the authoritative final request; setting `stop_probe_after_s` to zero disables the additional requests and early path.
- **Public classification API — deferred.** This focused bug fix does not broaden the public API. The existing private cross-package classifier remains an internal integration detail; promoting or redesigning that surface should be handled only with an explicit API decision.
- **Additional lead-in vocabulary — not changed.** The documented closed grammar remains intentional; this PR does not add speculative synonyms without observed requirements.

Validation on the follow-up: 228 focused voice/voice-gate tests and 26 documentation/configuration tests passed locally; repository-wide Ruff and diff checks passed; all GitHub checks are green on `8003bd96`.

PR #456, opened 2026-09-01
🍒 fix(services): migrate Hugging Face transfers to Xet

yanziz-nvidia:
## Summary

- cherry-pick merged [`4e4ed94a`](https://github.com/NVIDIA/xr-ai/commit/4e4ed94abcb15c01523b0445c5087e1f89a3da9b) from #439 to `main`
- migrate Hugging Face downloads from the deprecated `HF_HUB_ENABLE_HF_TRANSFER` path to Xet high-performance mode across vLLM, STT, Magpie TTS, and Piper TTS
- forward effective Xet controls into vLLM containers and include them in the launch fingerprint
- replace the unconditional `hf_transfer` install with guarded in-container Xet validation and repair, skipping bootstrap when Xet is disabled
- keep Piper's Xet cache under `model_cache` and document cache ownership and the legacy `hf_transfer` workaround

## Why

The existing Docker startup path unconditionally installs the deprecated `hf_transfer` helper from PyPI before starting vLLM. This makes startup depend on live package access and can fail in restricted or offline deployments.

This change uses the supported Xet path instead. Docker mode validates the image's Xet stack, conditionally repairs it when necessary, and fails clearly if the integration remains incompatible. Operators can set `HF_HUB_DISABLE_XET=1` to disable Xet and skip bootstrap validation and installation entirely.

Because the launch contract changes, existing persistent vLLM containers are recreated once on their next launch.

## Testing

- 138 focused vLLM, Piper, service-layout, and dependency-map tests passed
- Ruff passed for every changed Python file
- strict Sphinx documentation build passed
- `git diff --check` passed
- stable patch ID matches source commit `4e4ed94a`

### DIFF
--- a/DEPENDENCIES.md
+++ b/DEPENDENCIES.md
@@ -257,7 +257,8 @@ artifacts; do not commit them.
 - Runtime dependencies:
   - `vllm>=0.14.0`
   - `pyyaml>=6.0`
-  - `hf-transfer>=0.1.4`
+  - `huggingface-hub>=0.32.0`
+  - `hf-xet>=1.1.2,<2.0.0`
   - `xr-ai-logging` → [`xr-ai-logging`](utils/xr-ai-logging/) (local, editable)
   - `xr-ai-vllm` → [`xr-ai-vllm`](utils/xr-ai-vllm/) (local, editable)
 - Optional dependency groups: none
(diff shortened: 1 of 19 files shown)


wenxind-nvidia (commented):
## Suggestions

- `utils/xr-ai-vllm/xr_ai_vllm/_docker.py:178-185` — **Honor the explicit Xet opt-out during bootstrap.** `build_run_argv()` forwards and fingerprints the effective `HF_HUB_DISABLE_XET=1`, but the generated container command still imports, installs, and revalidates `hf-xet` unconditionally. On an offline or custom image without a compatible Xet/Hub stack, startup therefore fails before `vllm serve` even though the operator explicitly disabled Xet. Gate the Xet validation and repair commands on the effective post-`extra_env` disable value, while retaining that value in the fingerprint, and add a test asserting that disabled Xet emits neither command.

Reviewed head: 05a0eeec88c9c07f17b34c79252bab95b1d7cc6a

yanziz-nvidia:
thanks @wenxind-nvidia fixed above comment

PR #458, opened 2026-09-01
Move XR render routing rules from supervisor prompt to tool descriptions

blongs-nv:
Large PR: yes

The render demo supervisor prompt had grown to ~150 lines, restating each subagent's responsibilities alongside their tool definitions. This moves routing rules and instruction-phrasing contracts into the subagent tool `DESCRIPTION` constants (the surface the routing model actually reads), leaving the supervisor prompt with cross-cutting turn discipline only. The vision agent selects between two descriptions, and the worker now wires recorded video only when `video_history_enabled` is set (default off, matching the shipped hub config), so a video-less deployment never advertises past-frame capability.

Live testing surfaced perception-routing misses the offline tiers couldn't reproduce, driven by stale context: the transcript store persists across sessions under a reused participant id, so `[Recent conversation]` could carry hours-old turns that the model imitated ("You are not holding anything." replayed for a new question). Fixed structurally: recall is session-scoped (cut at the participant's last departure, with a 10-minute window covering worker restarts; departure records expire with that window), the same cutoff governs completion of a clipped utterance so a pre-disconnect "Make a..." never splices onto a reconnecting user's first words, a clarifying ask for a clipped utterance expires after 8 seconds, and an unset request timestamp falls back to the wall clock so the window can't silently disable. Recolors misrouted to object_agent are refused and reported back for re-delegation, and `spatial_ops.shape()` rejects color words instead of fuzzy-matching them to a shape.

New eval coverage: a `live_perception` tier that builds real multi-turn history and judges replies from the transcript store (including a poisoned-history case that fails without session scoping), plus offline cases for describe-view routing, camera imperatives, compound partial failure, and object_agent recolor refusal; the train/test audit now also scans supervisor.py and the live tier's texts, and the harness gained a `reply_excludes` assertion.

## Why this lands as one PR

Every group here is a consequence of the same live investigation, and the groups only pass the eval battery together:

- **Tool-description migration** is the fix itself: routing moved from the prompt to the descriptions. Splitting it from the prompt slimming would ship a prompt that says less while the descriptions still say nothing, and the routing model would regress on the very cases this PR adds.
- **Session-scoped recall and the clipped-utterance lifecycle** share one code path (`_recent_conversation`) and one cutoff. The poisoned-history live case fails without both, and the reconnect regression test fails if either is applied alone.
- **Recorded-video capability selection** exists because the migrated vision description advertises past-frame lookups; once routing lives in the description, the description has to reflect what the deployment can actually do.
- **Recolor and shape routing** were the two remaining misroutes the same live sessions exposed; they are the refusal-and-redelegate counterpart of the description contract.
- **The live_perception driver** is the catch-first test for the session-scoping change. Landing the fix without its catcher would leave the regression undetectable, since the offline tiers cannot reproduce a long live history.

Prompt and description text is token-sensitive at temperature 0: several wordings that looked equivalent flipped offline cases. Splitting the groups would mean re-running the full battery per PR with a prompt that is mid-transition, and the intermediate states are not ones we would want deployed.

## Validation and risk

Validated per the sample's eval policy, all on the final head: live_manip 13/13, live_perception 4/4, offline scenarios/precision/utterances at main's baseline (54/69, 13/16, 37/40; the remaining failures pre-exist on clean main), supervisor delegation and subagent tiers unchanged from head-before-review (18/23, 49/50, identical failing cases), 1366 non-GPU tests, ruff clean. Risk is concentrated in the routing model's reading of the new descriptions; the live tiers exercise the real stack end to end, and the two deliberate prompt/description duplications are marked in comments as intentional backstops.

### DIFF
--- a/DEPENDENCIES.md
+++ b/DEPENDENCIES.md
@@ -583,6 +583,7 @@ artifacts; do not commit them.
   - `xr_render_demo_live_explore` → `xr_render_demo_eval.live_explore:run`
   - `xr_render_demo_live_garble` → `xr_render_demo_eval.live_garble:run`
   - `xr_render_demo_live_manip` → `xr_render_demo_eval.live_manip:run`
+  - `xr_render_demo_live_perception` → `xr_render_demo_eval.live_perception:run`
   - `xr_render_demo_live_pose_matrix` → `xr_render_demo_eval.live_pose_matrix:run`
   - `xr_render_demo_live_smoke` → `xr_render_demo_eval.live_smoke:run`
 
--- a/agent-samples/xr-render-demo/eval/README.md
+++ b/agent-samples/xr-render-demo/eval/README.md
@@ -23,6 +23,7 @@ uv run xr_render_demo_live_pose_matrix
 uv run xr_render_demo_live_manip
 uv run xr_render_demo_live_garble
 uv run xr_render_demo_live_explore
+uv run xr_render_demo_live_perception
 ```
 
 <!-- Compatibility anchors for headings consolidated into the documentation. -->
--- a/agent-samples/xr-render-demo/eval/pyproject.toml
+++ b/agent-samples/xr-render-demo/eval/pyproject.toml
@@ -32,6 +32,7 @@ xr_render_demo_live_pose_matrix = "xr_render_demo_eval.live_pose_matrix:run"
 xr_render_demo_live_manip = "xr_render_demo_eval.live_manip:run"
 xr_render_demo_live_garble = "xr_render_demo_eval.live_garble:run"
 xr_render_demo_live_explore = "xr_render_demo_eval.live_explore:run"
+xr_render_demo_live_perception = "xr_render_demo_eval.live_perception:run"
 
 [tool.hatch.build.targets.wheel]
 packages = ["xr_render_demo_eval"]
--- a/agent-samples/xr-render-demo/eval/xr_render_demo_eval/harness.py
+++ b/agent-samples/xr-render-demo/eval/xr_render_demo_eval/harness.py
@@ -800,6 +800,57 @@ UTTERANCES = (
         expected_colors=(("sphere-0", (0.0, 0.4, 1.0)),),
     ),
     Case(
+        name="basics_describe_view_is_camera",
+        # A describe-the-view request means the physical world; reciting
+        # SCENE OBJECTS is never the answer.
+        request="Describe what you can see.",
+        scene=(
+            {"id": "sphere-0", "type": "sphere",
+             "position": {"x": 0.0, "y": 1.6, "z": -1.5},
+             "color": {"r": 0, "g": 0.8, "b": 0}, "size": 0.1},
+            {"id": "box-0", "type": "box",
+             "position": {"x": 0.5, "y": 1.4, "z": -1.2},
+             "color": {"r": 1, "g": 0.5, "b": 0}, "size": 0.1},
+        ),
+        vision="A cluttered desk with a laptop and a coffee mug.",
+        required_tools=frozenset({"look_at_current_frame"}),
+        forbidden_tools=_FORBID_MUTATIONS,
+        reply_contains="desk",
+    ),
+    Case(
+        name="basics_use_the_camera_imperative",
+        # Telling the agent to consult the camera re-enters the pending
+        # question; the literal words are never the delegation.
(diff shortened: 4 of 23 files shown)


wenxind-nvidia:
## Suggestions

1. Apply the participant departure cutoff to pending clarification detection. `_recent_conversation()` derives `pending` from the unfiltered final two transcript entries before applying `_left_at_us`. If a user says “Make a…”, disconnects, then reconnects with the same participant ID within eight seconds and says “cube,” the old request becomes “Make a cube.” despite the session cutoff. Filter pending candidates to the current session as well, and add a reconnect-after-truncation regression test.

2. Select the vision description from configured recording capability, not merely whether a `VideoMemoryTools` object exists. `app.py` always creates and passes `VideoMemoryTools`, so `make_vision_agent()` always advertises recorded-video support in the shipped sample, while the hub configuration disables recording and the video-memory YAML has no `recordings_dir`. Introduce an explicit history-enabled configuration or omit the video dependency when recording is disabled.

3. Use the worker YAML loader in the live-perception driver. `_text_memory_dir()` manually splits the YAML line, so valid quoted values or inline comments produce a different path from the worker and can make the evaluator poll the wrong transcript store. Load the configuration through `load_config()` or `yaml.safe_load()`.

wenxind-nvidia (approved):

nvddr (commented):
## Blockers

- `PR description` — **Justify the accepted large scope explicitly.** This 20-file, 867-line change is being accepted as a large PR, but the description does not include the repository's `Large PR: yes` marker or explain why the tool-description migration, session-history behavior, clipped-utterance lifecycle, recorded-video capability selection, recolor/shape routing, and new live-evaluation driver need to land together. Add the marker and a concise explanation of why these groups cannot be split and how their combined risk was validated.

- `agent-samples/xr-render-demo/worker/xr_render_demo_worker/supervisor.py:242` — **Do not carry a clipped request across participant sessions.** The new departure cutoff filters `[Recent conversation]`, but `pending` is derived from the unfiltered final transcript pair. A participant who disconnects after “Make a…”, reconnects within eight seconds, and says “cube” can therefore execute the prior session's mutation. Filter entries to the current session before selecting the pending pair and add a rapid reconnect regression test.

- `agent-samples/xr-render-demo/worker/xr_render_demo_worker/agents/vision/agent.py:131` — **Select the description from actual recording capability.** The new live-only description is selected only when `video is None`, but `app.py` always constructs `VideoMemoryTools`, while the shipped hub configuration disables recording and the video-memory configuration has no `recordings_dir`. The default sample therefore still advertises recorded-video support and routes historical questions to a disabled capability, contrary to the PR description. Pass the configured recording capability explicitly, or omit the video tool when history is disabled, and test the shipped wiring.

## Nits

- `agent-samples/xr-render-demo/worker/xr_render_demo_worker/supervisor.py:231` — `_left_at_us` retains every departed participant indefinitely, although entries stop affecting recall after ten minutes. Prune expired departure records so session-unique participant IDs cannot accumulate for the worker's lifetime.

- `agent-samples/xr-render-demo/eval/xr_render_demo_eval/live_perception.py:178` — An unknown case filter runs zero cases and exits successfully with `0 passed, 0 failed`. Validate requested names against `CASES` and fail with the unknown names, matching the other evaluation entry points.

Reviewed head: e97a8d4e178820ac2a258bcd54056c7ffd172e04

nvddr (commented):
## Blockers

- `agent-samples/xr-render-demo/worker/xr_render_demo_worker/app.py:149` — **Guard cleanup when video history is disabled.** The shipped/default configuration now makes `video` `None`, but the `finally` block unconditionally executes `await video.close()`. Every normal default-config shutdown—and any failure after client construction—therefore raises `AttributeError`, potentially masking the original error and making the default stack exit uncleanly. Close it only when constructed and add cleanup coverage for the disabled-video path.

## Nits

- `agent-samples/xr-render-demo/worker/xr_render_demo_worker/config.py:42` — Avoid Python truthiness for this boolean. A quoted YAML value such as `"false"` is nonempty and becomes `True`, silently enabling the historical-video capability this change is intended to gate. Require an actual YAML boolean or parse accepted boolean strings explicitly, and cover the invalid/quoted case.

Reviewed head: 0c10e8c871593b95e254ebe9a70671d885eb59e0

nvddr (approved):

PR #459, opened 2026-09-01
Fix active tea guide instruction routing

nvddr:
## Summary

- restore one focused active foreground agent per workflow step, built from the step prompt and tool list in `workflow.yaml`
- pass each active turn as structured `{request, state}` input with the step's projected state, matching the original NAT execution shape
- keep guide questions read-only while binding explicit next/skip commands to the authorized transition
- preserve live step tools for mixed and visual requests, and keep timer requests from being captured as background monitoring
- validate workflow tool names at startup and retain the original sparse legacy `active_context()` shape
- let the VLM judge water presence through a two-value response contract, then require two consecutive positive frames
- confirm heating only from two complete, explicit-unit readings above 50 Celsius or 122 Fahrenheit
- expand the live routing evaluation and use an explicit 80 percent acceptance threshold for model variation

## Architecture

`workflow.yaml` is authoritative for active foreground behavior. The worker constructs a private focused agent definition for every workflow step at initialization. Each definition owns that step's system prompt and tool names; a turn selects the active definition and supplies only projected step state.

The implementation keeps the native runtime and shared LLM service. A private query guard exposes at most one state-changing workflow control for an explicit command, and the advance tool's request schema fixes `skip` to the authorized value. Explicit background requests receive only their matching background tools. No public API is added.

Water and heating observations use small, step-specific VLM response contracts. Their consecutive-evidence and temperature-threshold checks are deterministic, avoiding a redundant second model decision for these two classifications.

## Large PR rationale

This release hotfix is larger than a normal prompt correction because the regression came from replacing the original per-step NAT foreground shape with one generic active prompt. Restoring that boundary, preserving background applications, hardening the two visual transition checks used by the demo, and adding regression coverage are coupled changes: splitting them would leave intermediate release commits with either broken active-guide behavior or unprotected state transitions.

## Validation

- `uv run --project tests pytest -q tests/test_tea_making_sample.py` — 98 passed
- broad live-model evaluation — 64/74 passed (86.5 percent), above the checked-in 80 percent acceptance threshold; every miss remains visible
- `uvx --from ruff==0.15.16 ruff check ...` — passed for all changed Python files
- `git diff --check` — passed

The remaining live-eval misses are non-mutating model variations around unrelated requests and unavailable or out-of-step live tools. Deterministic tests cover workflow-control isolation and binding, every focused step's tool catalog, two-frame water-presence confirmation, and exact temperature parsing.


wenxind-nvidia (commented):
## Blockers

- `agent-samples/tea-making-sample/worker/tea_making_worker/foreground.py:109` — **Advance authorization does not constrain `skip`.** “Next” and “Skip” expose the same `workflow__advance(skip: bool)` tool. The model can use `skip=true` for “Next,” applying fallback state and skipping an incomplete step. The eval validates only the tool name and argument schema. Bind the authorized `skip` value and assert it in evals.

- `agent-samples/tea-making-sample/worker/tea_making_worker/foreground.py:722` — **Procedural keywords suppress valid live requests.** Any occurrence of `details` or `current step` removes every live step tool. For example, “Read the details on the tea label” loses `current_view`. Base retained the step-authorized tools. Restrict this classification to whole-query procedural requests or prioritize explicit live-fact cues.

- `agent-samples/tea-making-sample/worker/tea_making_worker/foreground.py:112` — **Common explicit commands are rejected.** “Could you please continue?”, “Proceed to the next tea step,” and “Please restart my tea guide” receive no workflow tool. Base exposed the controls for the model to interpret; this PR makes its incomplete regex an authorization prerequisite. Expand the grammar and add regression cases.

- `agent-samples/tea-making-sample/eval/eval.py:197` — **The new routing evaluation knowingly fails.** The evaluator exits nonzero for any mismatch, while the PR reports only 60/70 cases passing. Remaining failures include next-step and unavailable out-of-step routing, which are part of this PR’s surface. Make the declared cases pass or encode an explicit passing acceptance contract.

- `docs/source/reference/tea-making-sample.md:204` — **Canonical documentation now describes the old behavior.** It still says active turns receive workflow controls, use an exact refusal, and append configured foreground instructions to the active policy. The new implementation query-gates controls, removes the exact refusal, and excludes the configured prompt from ordinary active turns. Update this section and the obsolete exact-refusal eval description at line 440.

- `agent-samples/tea-making-sample/worker/tea_making_worker/workflow.py:153` — **The large scope is unmarked and includes unused expansion.** This is a `+911/-116` PR without a large-PR marker or explanation of why it cannot be split. It also expands `active_context()` even though production no longer calls it. Remove that unused expansion and either split the remaining work or add the required large-PR marker and rationale.

Reviewed head: `1aaf615e442bfcd2f3867509e6fc2eaab37edbe4`

blongs-nv:
Reviewed at `1aaf615e442bfcd2f3867509e6fc2eaab37edbe4` by a bot.

## Blockers

1. **Stripping `?` before matching means questions expose the state-changing control.**

   [`foreground.py:113`](https://github.com/NVIDIA/xr-ai/blob/1aaf615e442bfcd2f3867509e6fc2eaab37edbe4/agent-samples/tea-making-sample/worker/tea_making_worker/foreground.py#L113) normalizes with `.strip(" .!?")`, so `Next?`, `next step?`, `Skip?`, `Continue?`, `Can you skip?`, and `Restart?` all match the command grammar and expose `workflow__advance` or `workflow__restart`. [`_TEA_PROMPT`](https://github.com/NVIDIA/xr-ai/blob/1aaf615e442bfcd2f3867509e6fc2eaab37edbe4/agent-samples/tea-making-sample/worker/tea_making_worker/foreground.py#L61) states "Questions using these words are not commands", so the gate is widest in exactly the direction this change exists to close.

2. **The two query gates subtract in sequence and can leave the turn with no tools at all.**

   [`foreground.py:415-421`](https://github.com/NVIDIA/xr-ai/blob/1aaf615e442bfcd2f3867509e6fc2eaab37edbe4/agent-samples/tea-making-sample/worker/tea_making_worker/foreground.py#L415-L421) applies `_workflow_tools_for_query` first, which drops every control when the utterance is not an exact command, then `_guide_tools_for_query`, which keeps only controls. A procedural question that is not also an exact command therefore reaches the model with an empty tool set while its step prompt tells it to call one; "What is the next step, and is the water hot enough?" is such an utterance.

## Suggestions

1. [`_background_tools_for_query`](https://github.com/NVIDIA/xr-ai/blob/1aaf615e442bfcd2f3867509e6fc2eaab37edbe4/agent-samples/tea-making-sample/worker/tea_making_worker/foreground.py#L600) replaces the step tool set rather than merging into it, so at `start_steeping`, whose completion message says "Say next to monitor it", asking to monitor the timer swaps `clock__timer` out for the change-watch tools.

2. `_build_tea_agents` runs at startup with the workflow in hand but never checks each step's `voice.tools` names against the tool catalogue, so a typo in `workflow.yaml` surfaces as a failed turn when a participant reaches that step rather than at launch.

3. Several new response patterns cannot fail on content: the refusal cases for weather, arithmetic, and unrelated scene descriptions accept any reply containing the word `tea`, and the next-step cases at `fill_water` accept `water`, which names the current step as well as the next.

4. The control-leak assertion in `tests/test_tea_making_sample.py` intersects the question turn's tool names with `_WORKFLOW_CONTROLS` itself. Removing `workflow__status` from that set makes `_workflow_tools_for_query` stop stripping it, so it reaches every question turn, and the assertion still passes because the leaked name has left the set being intersected.

5. Only `fill_water` is ever driven through `_prepare_turn` offline, so hardcoding that step id, or building `tool_names` from the wrong step, breaks nothing in the suite. One parametrized walk over `workflow.steps` would also cover `_select_tools` and the misspelled-tool error path.

6. Several assertions compare prompt text against the constants that generate it, and one pins a prose fragment of `workflow.yaml` inside a foreground test, so they break on rewording and pass on any behavior.

Making a read-only question structurally unable to reach a state-changing tool is the right shape for this fix, and the `{request, state}` envelope is correctly escaped and bounded. The blockers above should be resolved before merge.

yanziz-nvidia:
reviewed by yanziz-review-bot

**Summary**

This PR introduces focused per-step agents, query-gated tools, and deterministic water/temperature evidence handling. Other than the existing findings in previous comments, I found two new merge-blocking false-positive paths in the latest evidence parsers. [GitHub checks are green](https://github.com/NVIDIA/xr-ai/pull/459/checks); locally, 75 focused tests passed in-sandbox, the socket-bind test passed outside it, and targeted reproductions confirmed both defects.

| Severity | Finding |
|---|---|
| 🚫 BLOCKER | [`workflow.py:93`](https://github.com/NVIDIA/xr-ai/blob/31389d0414e79c2da22a65d12c9501dd5d048ac9/agent-samples/tea-making-sample/worker/tea_making_worker/workflow.py#L93) — `_water_visible` accepts any `water visible:` response unless its suffix contains one of a limited set of negative words. Both `water visible: the kettle appears empty` and `water visible: the lid is open` return true, so two repeated observations commit `water_filled=true` and announce that the vessel has water without a concrete water cue. Require an approved positive cue and reject empty/dry/absent/obscured language before counting evidence. |
| 🚫 BLOCKER | [`workflow.py:96`](https://github.com/NVIDIA/xr-ai/blob/31389d0414e79c2da22a65d12c9501dd5d048ac9/agent-samples/tea-making-sample/worker/tea_making_worker/workflow.py#L96) — `_temperature_reading_c` searches for one unit-bearing number anywhere in the response and ignores contradictory context. `No current reading; the target is 93 Celsius.` parses as `(93.0, false)`, so two such observations commit `heating_started=true`. Accept only a full response matching the reading protocol—or use a structured result—and treat explanatory or negative responses as unavailable/ambiguous. |
| 💡 SUGGESTION | none |
| 🧹 NIT | none |

- `workflow.py:93` — Require an accepted water response to contain a concrete positive water cue, reject empty/dry/absent/obscured responses, and add two-observation regression tests proving those responses cannot set `water_filled`.
- `workflow.py:96` — Parse only a complete exact temperature response or a structured result, and add a regression proving that a missing-reading response mentioning a target temperature cannot set `heating_started`.

nvddr:
Addressed in `fe7cdce0bd79ba72f0bd039e1b42398e1a234e50`.

### Wenxin

- **Addressed:** `workflow__advance` is now wrapped in a private literal request schema, so next/continue can only send `skip=false` and skip can only send `skip=true`. The eval also validates the expected argument.
- **Addressed:** procedural filtering now matches whole-query guide questions; mixed and explicit live-fact requests retain their step tools.
- **Addressed:** the explicit-command grammar covers the reported polite continue, proceed, and restart forms, with regressions.
- **Addressed:** the live eval now has an explicit passing contract: every case reports `PASS`/`MISS`, and the run fails below 80 percent instead of requiring a stochastic model to pass every utterance. The latest run passed 64/74 (86.5 percent).
- **Addressed:** canonical docs now describe query-gated controls, semantic refusal wording, focused YAML-owned active prompts, and the eval threshold.
- **Addressed:** the unused expanded `active_context()` payload was removed; it retains the original sparse legacy shape. The PR description now includes the large-PR rationale.
- **Not split:** this is a release hotfix restoring the per-step NAT boundary. Splitting agent selection, transition safety, and its regression coverage would leave intermediate release commits with either broken active routing or unsafe transitions.

### Brent

- **Addressed:** bare questions such as `Next?`, `Next step?`, and `Restart?` cannot receive mutation tools. Syntactic indirect requests such as `Can you skip?` remain accepted as natural explicit commands; their transition argument is now structurally bound, so the model cannot reinterpret the operation.
- **Addressed:** the sequential filters no longer erase live tools from mixed procedural/live requests.
- **Addressed:** timer/steeping requests retain `clock__timer` unless the user explicitly names a background watcher. Explicit background requests still intentionally use the isolated background tool set rather than merging unrelated step controls.
- **Addressed:** all YAML step tool names are validated against the private catalog at startup, including a typo regression.
- **Addressed:** response patterns now require actual next-step/refusal content; the control-leak assertion uses a literal protected set; and an all-step test exercises focused selection and declared live tools.
- **Partial:** retained a few narrow prompt-contract checks because repository policy requires prompt rules to have corresponding eval coverage, but behavioral tests now carry the routing and mutation guarantees instead of assertions derived from `_WORKFLOW_CONTROLS`.

### Yanzi

These two cases were **not** fixed at the `31389d0` head you reviewed; they are fixed in this update.

- **Addressed:** water evidence now requires an approved concrete cue and rejects negative/empty/dry/absent/obscured language. Repeated `water visible: the kettle appears empty` and `water visible: the lid is open` cannot set `water_filled`.
- **Addressed:** temperature parsing now accepts only a complete exact reading response. `No current reading; the target is 93 Celsius.` is unavailable and cannot set `heating_started`; a target-only response is treated as ambiguous.

Validation: 96 focused tests passed, Ruff 0.15.16 passed, `git diff --check` passed, and the live-model eval passed at 86.5 percent.

nvddr:
Final water-detection disposition for `76c5422dcee2db140b35975a5b8f0b66ce217808`:

- The latest demo run showed the VLM repeatedly recognizing water but answering `water visible: yes` or `<yes>`. The worker's cue-word whitelist rejected those judgments and reset the evidence counter; isolated `waterline` answers passed but never occurred consecutively.
- The fix removes that brittle semantic re-parsing. The VLM now judges the image directly under a small prompt and returns `water present` or `water absent`; the worker validates only that response contract rather than requiring particular descriptive words.
- Two positive frames remain only as temporal debounce against a single bad visual inference. They no longer impose a second semantic test, and the tested run now advances correctly.
- We did not add a separate text-LLM pass: the VLM is already the language model with access to the pixels, while another model call would add latency and another tool/formatting failure point.
- Exact negative and malformed responses still cannot advance the step, including the two false-positive cases raised in review.

Validation: 98 focused tests passed, including two-frame completion and negative/malformed-response regressions; Ruff and `git diff --check` passed.

wenxind-nvidia:
Reviewed at `76c5422dcee2db140b35975a5b8f0b66ce217808`.

## Suggestions

1. [`foreground.py:685-723`](https://github.com/NVIDIA/xr-ai/blob/76c5422dcee2db140b35975a5b8f0b66ce217808/agent-samples/tea-making-sample/worker/tea_making_worker/foreground.py#L685-L723) claims to expose background tools only for explicit requests, but it authorizes an entire service catalog from noun matches alone. For example, both “Should I start recording the transcript?” and “If I say start recording the transcript, will you do it?” expose `transcript__start`, `transcript__stop`, and `transcript__status`. The model can therefore begin transcript persistence from a deliberative or hypothetical question, despite the background services being documented as opt-in. Parse the requested background operation and expose only that operation; mutating start/stop tools should be unavailable for informational or hypothetical turns. Add negative deterministic coverage for each background mutator.

2. [`workflow.yaml:30-32`](https://github.com/NVIDIA/xr-ai/blob/76c5422dcee2db140b35975a5b8f0b66ce217808/agent-samples/tea-making-sample/yaml/workflow.yaml#L30-L32) still describes `heating_started` as accepting a heater-on indicator or any unit-bearing temperature. The implementation now ignores heater indicators and requires two readings above 50 Celsius or 122 Fahrenheit. Update this generated configuration guidance to match the actual contract.

3. [`eval.py:210-212`](https://github.com/NVIDIA/xr-ai/blob/76c5422dcee2db140b35975a5b8f0b66ce217808/agent-samples/tea-making-sample/eval/eval.py#L210-L212) computes detailed validation errors but discards them when printing a `MISS`. Include `errors` in the output so schema, response-pattern, forbidden-content, and length failures can be diagnosed without manually comparing each case.

wenxind-nvidia (approved):

yanziz-nvidia (approved):

PR #460, opened 2026-09-01
Add repository-wide resolved dependency manifest

blongs-nv:
## Summary

- add `dependency-manifest/`, a virtual uv project depending on every Python project in the repo, with its `uv.lock` committed
- exclude it from the per-project `lock-check` loop; mark the lock as generated
- document regeneration as part of the dependency refresh in `DEPENDENCIES.md`

## Why

Per-project lockfiles stay gitignored, so nothing changes for contributors. Dependency analysis tooling wants a resolved lockfile for exact versions, and one aggregate lock at the `exclude-newer` cutoff provides the complete set in a single place without placing a lock where `uv run` or `uv sync` would consume it.

It is regenerated only when the cutoff advances, with `uv --config-file uv.toml lock --upgrade --project dependency-manifest`.

## Validation

- `uv lock --check --project dependency-manifest` passes (352 packages at cutoff `2026-08-27T17:15:48Z`)
- `generate_dependency_map.py --check` and `pre-commit run --all-files` pass


yanziz-nvidia:
i thought we don't want to commit in uv.lock file?

yanziz-nvidia:
Can we target at release branch?

yanziz-nvidia:
reviewed by yanziz-review-bot

**Summary**
This adds a virtual aggregate project and a 352-package lock currently covering all 37 in-repo Python projects and declared extras. The generated dependency map, aggregate `uv lock --check`, and all current [PR checks](https://github.com/NVIDIA/xr-ai/pull/460/checks) pass. Other than the existing lockfile-policy and target-branch questions, the aggregate lock lacks freshness enforcement.

| Severity | Finding |
|---|---|
| 🚫 BLOCKER | [`.github/workflows/lock-check.yml:93`](https://github.com/NVIDIA/xr-ai/blob/e80a07324085b7ba07c22b9465bf2eeaca9601f0/.github/workflows/lock-check.yml#L93-L99) - This exclusion removes `dependency-manifest` from the only lock-validation job, while the documentation defers its regeneration until a dependency refresh. After an ordinary dependency or extra change, CI verifies the changed project's ignored lock and `DEPENDENCIES.md`, but never compares the committed aggregate lock against the changed metadata; newly added projects and extras are also manually maintained. Dependency tooling can therefore silently receive stale or incomplete data. I reproduced the first path by adding one dependency to an existing project: `uv lock --check --project dependency-manifest` rejected the committed lock. Add a dedicated aggregate `--check` step and validate that the manifest contains every discovered project and extra. |
| 💡 SUGGESTION | none |
| 🧹 NIT | [`DEPENDENCIES.md:53`](https://github.com/NVIDIA/xr-ai/blob/e80a07324085b7ba07c22b9465bf2eeaca9601f0/DEPENDENCIES.md#L53-L58) - “Complete resolved dependency set” overstates the artifact's scope: all 37 aggregated projects declare `hatchling` as a build requirement, but the lock contains no `hatchling` package record. If runtime/install dependencies are the intended scope, say “complete resolved runtime dependency set” and note that build-system requirements are excluded; otherwise capture those requirements separately. |

- `.github/workflows/lock-check.yml:93` - Add a dedicated `uv --config-file uv.toml lock --check --project dependency-manifest` CI step, and extend dependency-map validation to require every discovered project and optional extra in the aggregate manifest.
- `DEPENDENCIES.md:53` - Scope the completeness claim to runtime dependencies or explicitly include build-system requirements.

blongs-nv:
**Release branch:** If we have this on main and cherry-pick to the release branch, we can rerun it whenever we do another cutoff in the future.

**Bot blocker (`--check` step and completeness validation):** Deliberately not gated. A CI check on this file would make every ordinary dependency bump or new project fail until someone regenerates a 6.5k-line shared lock, which is exactly the contributor friction the no-lockfile policy exists to avoid. The lock is a snapshot at the qualification cutoff, refreshed together with the cutoff; between refreshes it lags by design, the same way the cutoff itself lags newly published packages. That trade is documented in DEPENDENCIES.md.

**Bot nit ("complete resolved dependency set"):** Agreed, wording adjusted to runtime dependencies with a note that build-system requirements are not part of a uv lock.

nvddr:
Closing in favor of cherry-pick version.

PR #461, opened 2026-09-01
fix(services): clean up Piper process groups

wenxind-nvidia:
## Summary

- replace Piper's detached long-lived supervisor/child pair with one foreground server
- keep Piper in the launcher's dedicated process group, verify ownership before group-wide cleanup, and retain PID-only cleanup for unverified or non-Piper local servers
- keep reused Piper bootstraps monitorable in normal stacks while allowing them to exit from persistent-only `exit_after_ready` launchers
- preserve `startup_timeout_s` through Uvicorn readiness, report foreground startup failures concisely, propagate incomplete `model_servers --stop` cleanup, and cover normal and aborted lifecycles end to end

## Root cause

Piper implemented persistence inside the service wrapper even though `model-servers` already launches it with `launch_mode="persist"`. The wrapper started the real Uvicorn/ONNX server in a new session and then stayed alive. Every later stack start created another wrapper that idled while reusing the same healthy server. Those wrappers were not listeners, so port-based cleanup discovered only the detached server and could leave Piper-related processes behind.

The bootstrap now uses `exec` so Uvicorn remains the foreground service without an extra supervisor. The launcher places each `uv` wrapper and its children in a dedicated session and stamps the entry-point command in a private ownership marker. Piper records that existing process group without calling `setsid()`, which keeps it reachable by the launcher's abort-time `SIGTERM` and `SIGKILL`. A directly started Piper process records a group only when it already leads its own dedicated session; otherwise cleanup safely falls back to the listener PID.

Persistent cleanup verifies the Piper listener marker, recorded group, session, and—when the listener is not the group leader—the launcher's group-leader marker before signaling the complete group. A reuse invocation retains the existing healthy-listener behavior and signals readiness. Other local server cleanup remains PID-scoped.

For reuse, the launcher grants ready-process exit only when `exit_after_ready=True` and the process uses `launch_mode="persist"`, as in `model-servers`. Reuse invocations in ordinary monitored stacks remain alive as health proxies, so their launcher continues to detect service termination without accumulating wrappers across repeated persistent-only starts.

PID-scoped cleanup now performs a bounded poll after `SIGKILL`, matching the process-group cleanup path. This avoids reporting a successful asynchronous kill as a cleanup failure while still returning nonzero if the process remains alive.

## Validation

- Python 3.11 broader PR suite (174 passed)
- `uv run --project tests pytest tests/test_launcher_stack.py tests/test_vllm_stop.py tests/test_piper_tts.py -q` (59 passed)
- abort regression launches a Piper-shaped child that ignores `SIGTERM`, verifies it remains in the launcher's session, exercises `SIGKILL` escalation, and verifies no live group member survives
- reuse lifecycle tests cover monitored reuse, explicit persistent ready-exit, transient health-check failure recovery, and launch/reuse/group cleanup end to end
- PID cleanup regression reproduces delayed process disappearance after `SIGKILL`
- Ruff 0.15.16 and `git diff --check` pass


nvddr (dismissed):
Please pare this back to the reproduced Piper cleanup failure. Two parts currently broaden the change beyond what that fix needs:

1. The reuse-identity work adds `managed_server_on_port` to the public `xr_ai_vllm` API, adds an `xr-ai-vllm` dependency to the non-vLLM Piper service, introduces another environment marker, and adds migration/docs/test surface. Reuse of an already-healthy listener predates this bug; changing its identity policy is separate hardening. It is also not sufficient reason to broaden a public API. Please remove that scope from this PR (or handle it in a focused follow-up).

2. `stop_persistent_servers` now uses process-group signaling for every marker-owned local server, changing pip-vLLM lifecycle behavior even though the reproduced orphan is Piper. Please scope group termination to the Piper service unless there is a separate reproduced failure demonstrating that the other local model servers require this behavior.

The foreground `exec`, ready-file handoff, Piper process-group cleanup, startup-timeout preservation, and failure propagation from `model_servers --stop` are directly connected to this fix and should remain. Keeping the PR there would substantially reduce its API, dependency, compatibility, and regression surface.

wenxind-nvidia:
Addressed the scope review in `3c86829`:

- removed `managed_server_on_port`, Piper's `xr-ai-vllm` runtime dependency, the service-identity marker/policy, migration entry, and their associated test/documentation surface;
- limited process-group termination to Piper (`tts`) and restored the original PID-only cleanup path for every other local server, with a regression test that prevents group signaling for pip-vLLM;
- retained the foreground `exec`, ready-file handoff, end-to-end startup timeout, Piper group cleanup, and `model_servers --stop` failure propagation.

The affected suite passes (159 tests), the Piper/cleanup tests pass under Python 3.11 and 3.12 (28 each), and all PR workflows are green.

yanziz-nvidia (approved):

nvddr (commented):
## Blockers

- `utils/xr-ai-vllm/xr_ai_vllm/__init__.py:226` — **Do not signal an unverified process group.** Piper cleanup now treats the managed-port environment marker as proof that the listener's entire process group is repository-owned. However, the bootstrap sets that marker and calls `execvpe()` without creating or verifying a dedicated group. A direct or programmatic Piper launch can inherit a shell, CI harness, or supervisor group, so `model_servers --stop` can signal unrelated processes or its own caller. The base implementation created the managed child with `start_new_session=True`. Add a private trustworthy group-ownership marker, isolate non-launcher starts, or fall back to PID-only cleanup when group ownership is not established.

## Nits

- `services/piper-tts/piper_tts_server/__main__.py:428` — Startup timeouts and early Uvicorn failures now escape `asyncio.run()` as raw tracebacks. The base wrapper translated these into concise `[piper_tts_server] …` failures. Restore equivalent exception translation around the foreground server while preserving the nonzero exit.

Reviewed head: 3c86829485cb8cacd4c4f31509d817e316d89f67

nvddr (commented):
## Blockers

- `services/piper-tts/piper_tts_server/__main__.py:327` — **Preserve the launcher's abort-time SIGKILL reachability.** The launcher deliberately starts `uv` in a dedicated session and `_shutdown()` signals that process group, including a SIGKILL escalation after 20 seconds. Because the `uv` child has a different PID while initially sharing that group, `_ensure_owned_process_group()` calls `setsid()` and moves Piper into a second session. The initial SIGTERM may be forwarded, but if Piper has not exited when the launcher escalates, SIGKILL reaches only the original `uv` group and the detached Piper process survives. I reproduced that topology and survival through escalation with uv 0.12.3. Keep Piper reachable by abort cleanup or make the launcher aware of the new group, and add an abort regression test that verifies no process survives.

Reviewed head: 212590fe48f9e1758a9452f533fffa2ca875b4a6

wenxind-nvidia:
Addressed the abort-time cleanup blocker in 7659dfc. Piper no longer calls setsid() when it is a uv child: the launcher marks its dedicated session, Piper records and remains in that group, and persistent cleanup verifies the separate launcher group leader before signaling the group. Direct/unverified starts retain PID-only fallback. Added an abort regression that makes the Piper-shaped child ignore SIGTERM, exercises the launcher SIGKILL escalation, and verifies no live process remains. Focused tests: 56 passed; Ruff and DCO pass.

wenxind-nvidia:
Both comments were valid and are addressed in f31dc62. Reused Piper bootstraps now remain alive as health proxies for normal monitored run_stack calls. The launcher grants bootstrap exit only for the explicit exit_after_ready=True + launch_mode=persist combination used by model-servers, preventing both premature stack shutdown and accumulated reuse wrappers. PID-scoped cleanup now polls for up to 5 seconds after SIGKILL before deciding cleanup failed. Added regressions for monitored reuse, the constrained ready-exit protocol, transient health failures, and delayed post-SIGKILL disappearance. Validation: 174 broader Python 3.11 tests and 59 focused tests passed; Ruff and DCO pass.

nvddr (approved):

PR #462, opened 2026-09-02
fix(model-servers): stabilize DGX Spark vLLM cold starts

wenxind-nvidia:
## Summary

- add explicit `kv_cache_memory_bytes` support to the Cosmos and Nemotron Omni vLLM launchers
- pin the DGX Spark caches to 1.5 GiB for Cosmos and 2 GiB for Omni so cold checkpoint I/O cannot distort automatic UMA memory profiling
- deliberately retain the Spark `gpu_memory_utilization` values because the bundled vLLM versions still use them for the initial free-memory admission check
- keep the Blackwell and dual-Ada profiles on their existing fractional allocations
- document the two distinct Spark failure stages: the early `cudaMemGetInfo` failure and the later cold-start profiling error

## Why

DGX Spark shares system and GPU memory. During a cold model load, Linux page-cache activity can be counted by vLLM as model or non-Torch overhead, reducing the computed KV-cache budget or making it negative. Nemotron Omni can therefore fail before Cosmos starts, and a later cached attempt can pass with identical settings.

Fixed cache bytes bypass that variable sizing calculation. The 2 GiB Omni budget covers four full 32,768-token requests under the pinned vLLM 0.20 hybrid Mamba and attention cache layout while preserving unified-memory headroom for Cosmos and the rest of the stack. The 1.5 GiB Cosmos budget remains above its measured 1.12 GiB requirement.

The fractional values remain intentional: vLLM 0.20 calls its startup admission check before branching to the explicit-cache path. Passing both values avoids falling back to the higher vLLM default while still making the cache allocation deterministic.

Upstream context: https://github.com/vllm-project/vllm/issues/35920

## Testing

- `uv run --project tests pytest -q tests/test_model_servers.py tests/test_vlm_server_config.py` (60 passed)
- `uv run --project tests pytest -q tests/test_config_documentation.py tests/test_api_documentation.py` (16 passed)
- `uvx ruff check services/vlm-server/vlm_server/__main__.py services/nemotron-omni-llm/nemotron_omni_llm_server/__main__.py tests/test_model_servers.py tests/test_vlm_server_config.py`
- `python3 -m py_compile services/vlm-server/vlm_server/__main__.py services/nemotron-omni-llm/nemotron_omni_llm_server/__main__.py`

Signed-off-by: Wenxin Ding <wenxind@nvidia.com>


yanziz-nvidia:
why not just bump up utilization from 0.2 to 0.21? that should be simplest and quick win?

nvddr (commented):
## Nits

- `services/vlm-server/vlm_server/__main__.py:94` and `services/nemotron-omni-llm/nemotron_omni_llm_server/__main__.py:114` — The new field is documented as a positive integer, but `int()` silently truncates YAML numbers such as `1.5` to `1`, forwarding a one-byte KV-cache allocation instead of reporting a configuration error. Reject non-integral values in both launchers and add a fractional value to the invalid-configuration tests.

Reviewed head: 88d438b2de6b684222d0199763d61686aa0289aa

nvddr (approved):

wenxind-nvidia:
@yanziz-nvidia fractional partition is not robust in this case. Change it to absolute number of bytes should be more robust.

yanziz-nvidia (approved):

PR #465, opened 2026-09-03
Add repository-wide resolved dependency manifest

blongs-nv:
This is https://github.com/NVIDIA/xr-ai/pull/460 but for the release branch instead of main.


yanziz-nvidia (approved):

wenxind-nvidia (commented):
## Blockers

1. `.github/workflows/lock-check.yml:98` - **The committed aggregate lock is never validated.** The workflow excludes `dependency-manifest`, so dependency changes or manual lock corruption can leave `dependency-manifest/uv.lock` inaccurate while CI passes. I reproduced this by changing an existing project dependency: its normal lock check passed, but regenerating the aggregate changed the committed lock. A dedicated CI step should regenerate the aggregate and verify `git diff --exit-code`; with uv 0.11.7, `uv lock --check` alone did not detect changed editable-project metadata.

2. `dependency-manifest/pyproject.toml:13` - **Manifest completeness is manually maintained and unchecked.** The generator discovers every project, but the change at `.github/scripts/generate_dependency_map.py:45` only adds a documentation group. It does not verify that the aggregate declares every discovered project, source, and optional extra. A newly added project can therefore be omitted while all current checks pass. Generate this list mechanically or validate it against project discovery.

The author rejected the same concern on PR 460 because the lock is intentionally refreshed only at dependency cutoffs. I still consider it valid: package-index cutoff lag is intentional, but silently lagging repository dependency metadata makes the new artifact no longer describe the release branch.

No nits found. The current artifact itself is correct: all 37 projects and all declared extras are represented, the 351-package lock resolves, generated documentation is current, and CI is passing.

Reviewed head: b3767e4083099df9828c4da46f37e5d17426903e

blongs-nv:
Both points are addressed in b8546c76.

**Completeness.** `dependency-manifest/pyproject.toml` is now generated from the same project discovery the dependency map uses, so every project and every declared extra is included mechanically. The file is marked generated and is not edited by hand.

**Validation.** A new `dependency-manifest` workflow re-resolves the aggregate lock from scratch and compares it byte for byte with the committed file, which catches the editable-metadata case that `uv lock --check` misses. It pins the uv version it resolves with, since lock output varies across uv releases, and the same version is recorded in DEPENDENCIES.md for whoever regenerates. A pre-commit hook does the regeneration automatically.

**Scope of the check.** Both run when `uv.toml`, `dependency-manifest/`, or the generator changes, not on every PR. That is deliberate: the manifest is defined as the resolved dependency set as of the cutoff, and it is the cutoff change that produces a new snapshot for analysis. Between cutoffs it lags repository metadata by design, the same way the cutoff itself lags the package index; DEPENDENCIES.md now says so explicitly, along with the yank caveat. Ordinary dependency changes never touch the directory, which keeps the no-lockfile policy intact for contributors.

wenxind-nvidia (commented):
## Blockers

- `.github/scripts/generate_dependency_manifest.py:50` - **The generator rejects a repository-supported Python configuration.** It requires every `requires-python` value to be byte-identical, while `DEPENDENCIES.md:24-27` explicitly permits projects to declare different ranges. Even equivalent formatting fails, and a valid Python-3.12-only project would prevent the next dependency refresh. Normalize equivalent specifiers and represent narrower ranges with appropriate markers or resolution cohorts.

- `.pre-commit-config.yaml:27` - **Local regeneration does not enforce the uv version required for reproducibility.** CI byte-compares output from uv 0.10.7 because lock output can vary between releases, but the hook and documented command use whichever `uv` is on `PATH`. A dependency refresh can therefore produce a lock that deterministically fails CI. Make the generator invoke the pinned version or fail fast when another version is used.

## Nits

- `.github/scripts/generate_dependency_manifest.py:62` - Quote generated TOML source keys. I reproduced that a valid distribution name such as `foo.bar` becomes the nested structure `{"foo": {"bar": ...}}` instead of the required `"foo.bar"` source entry.

- `DEPENDENCIES.md:49` - “All generated `uv.lock` files remain gitignored” contradicts the committed exception introduced three lines later. Change this to “All generated per-project lockfiles.”

## Previous Findings

Both original findings are resolved under the clarified contract that this artifact is a snapshot of the repository at the last qualification cutoff, rather than a continuously current representation of `HEAD`:

- Manifest membership and extras are now generated mechanically.
- The new workflow re-resolves and byte-compares the lock when cutoff or manifest-related files change.

The 14 focused generator tests pass, the current lock reproduces locally, and all current CI checks pass.

Reviewed head: b8546c76ef571eeb0ea45d53e9d31cb5405bd3c8

wenxind-nvidia (approved):

PR #466, opened 2026-09-03
fix(demo): let observation agents judge tea evidence

nvddr:
## Summary

- route every visual workflow caption through the step observation LLM instead of using water/temperature parsers or caption regexes as the primary judge
- keep deterministic safeguards around typed state, stale revisions, explicit advancement, and consecutive completion proposals
- use one semantic confirmation for water and two for heating/steeping; retain deterministic timer handling
- clarify VLM and observation prompts, including the previously contradictory completion instruction
- update focused tests, eval setup, and architecture documentation

## Why

The release branch had accumulated step-specific Python classification for water and heating, while other steps still used caption regexes as hard semantic gates. That made natural VLM descriptions brittle and diverged from the intended VLM-caption -> observation-agent -> guarded tool-call flow.

Regex matching now sees only an internal accepted/rejected guard token. The observation LLM is the semantic judge; Python validates and counts its completion proposals before allowing a commit.

## Validation

- `uv run --project . pytest -q test_tea_making_sample.py` from `tests/`: 80 passed
- Ruff 0.15.16 on changed Python files: passed
- `git diff --check`: passed
- live local Omni observation evaluation: tea identification, water synonyms, temperature target/unit negatives, two hot readings, tea-beside-water negative, and two immersion confirmations all behaved as expected
- live routing eval: 64/74 passed (86.5%, above the configured 80% threshold); remaining misses are existing foreground-routing cases outside this change
- docs build not run: the available Sphinx environment lacks `myst_parser`


yanziz-nvidia:
reviewed by yanziz-review-bot

**Summary**
This PR moves semantic observation classification into the LLM and adds deferred guidance speech. CI is green; focused changed-path tests passed locally, while custom reproductions exposed one unenforced workflow guard and two voice-ordering failures.

| Severity | Finding |
|---|---|
| 🚫 BLOCKER | [`workflow.py:531`](https://github.com/NVIDIA/xr-ai/blob/e652a8c830413dafea4d1fc9d31413b36edf8dc7/agent-samples/tea-making-sample/worker/tea_making_worker/workflow.py#L531) — Heating completion is accepted solely from a completion-shaped patch; the wrapper never verifies that `temperature__threshold` ran or returned `above=true`. Two direct `workflow__commit({"heating_started": true})` calls therefore complete the step without any temperature reading. Additionally, [`threshold_c` is model-controlled](https://github.com/NVIDIA/xr-ai/blob/e652a8c830413dafea4d1fc9d31413b36edf8dc7/agent-samples/tea-making-sample/worker/tea_making_worker/workflow_tools.py#L103), so 45°C can pass if the model supplies 40°C. Bind the 50°C threshold in code and require a successful same-observation threshold result before accepting `heating_started`. |
| 🚫 BLOCKER | [`guidance_voice.py:118`](https://github.com/NVIDIA/xr-ai/blob/e652a8c830413dafea4d1fc9d31413b36edf8dc7/agent-samples/tea-making-sample/worker/tea_making_worker/guidance_voice.py#L118) — Every queued notice is flushed after the foreground turn, even when that turn advanced or reset the workflow. For example, a queued “The vessel has water. Say next when ready.” notice is spoken after the user says “next” and the foreground response has moved to heating, giving a stale and contradictory instruction. Carry the originating step/revision with each notice and discard mismatches, or explicitly invalidate pending notices after state-changing workflow tools. |
| 💡 SUGGESTION | [`guidance_voice.py:177`](https://github.com/NVIDIA/xr-ai/blob/e652a8c830413dafea4d1fc9d31413b36edf8dc7/agent-samples/tea-making-sample/worker/tea_making_worker/guidance_voice.py#L177) — The one-second release delay does not cover the actual utterance lifecycle: [`UserStoppedSpeakingFrame` is emitted before STT begins](https://github.com/NVIDIA/xr-ai/blob/e652a8c830413dafea4d1fc9d31413b36edf8dc7/agent-sdk/xr-ai-voice/xr_ai_voice/_processors/vad_stt.py#L209). When STT and gate processing exceed one second, the notice is released before `USER_QUERY_TOPIC` marks the foreground busy and is then interrupted by the eventual response. Replace the timing heuristic with an explicit post-STT/gate outcome event. |
| 🧹 NIT | none |

- `workflow.py:531` — Require a same-observation, `above=true` result from a code-bound 50°C threshold check before accepting `heating_started`; add regressions for direct commit, altered threshold, and missing or ambiguous readings.
- `guidance_voice.py:118` — Invalidate deferred notices when the foreground turn changes workflow state, preferably by carrying and checking the notice’s step and revision.
- `guidance_voice.py:177` — Replace `release_delay_s` with an explicit utterance-resolution signal emitted after STT and voice-gate processing; test with STT delayed beyond one second.

blongs-nv:
Reviewed at `e652a8c830413dafea4d1fc9d31413b36edf8dc7` by a bot.

## Blockers

1. **A malformed commit is counted as contradicting evidence and wipes confirmed observations.**

   [`workflow.py:530`](https://github.com/NVIDIA/xr-ai/blob/e652a8c830413dafea4d1fc9d31413b36edf8dc7/agent-samples/tea-making-sample/worker/tea_making_worker/workflow.py#L530) treats any non-completing patch as a rejection, and `CommitRequest.evidence` defaults to `"rejected"`, so an invalid patch resets `evidence_hits` to zero before `commit` rejects it for being invalid. On `heat_water` at two confirmations: a hot reading takes hits to 1, the model then commits a non-writable field, hits go to 0, and the next hot reading only returns to 1. A schema slip is a failed judgment, not a negative one, and should take the `None` path that `ToolLoopError` already uses.

2. **`fill_water` drops to one confirmation in the same change that makes the judgment semantic.**

   [`workflow.yaml:111`](https://github.com/NVIDIA/xr-ai/blob/e652a8c830413dafea4d1fc9d31413b36edf8dc7/agent-samples/tea-making-sample/yaml/workflow.yaml#L111) goes from `consecutive: 2` to `consecutive: 1`, which makes the gate a no-op: the first completion proposal both increments the counter and satisfies it. This is the step whose caption handling was rewritten three times in the preceding change because the model kept deviating, and it is now the one step where a single judgment completes it outright.

## Suggestions

1. The new `_VoiceIOProcessor` test asserts only that the callbacks fire, so removing `await self.push_frame(frame, direction)` at [`io.py:133`](https://github.com/NVIDIA/xr-ai/blob/e652a8c830413dafea4d1fc9d31413b36edf8dc7/agent-sdk/xr-ai-voice/xr_ai_voice/_processors/io.py#L133) passes it while `UserStartedSpeakingFrame` and `UserStoppedSpeakingFrame` stop reaching everything downstream of the processor, for every voice sample rather than just this one. One assertion on the sink's frame kinds closes it.

2. [`guidance_voice.py:64`](https://github.com/NVIDIA/xr-ai/blob/e652a8c830413dafea4d1fc9d31413b36edf8dc7/agent-samples/tea-making-sample/worker/tea_making_worker/guidance_voice.py#L64) publishes a notice immediately whenever `_busy()` is false, but `_pending` can still hold older notices during the release window and mid-`_flush`. A participant's notices can be spoken out of order among themselves; checking `_pending` in `notify`, or a per-participant flushing flag, keeps them sequential.

3. `_foreground` is cleared only by a foreground record or an interruption, and the record is published after the contribution publish succeeds. A subscriber raising there leaves that participant permanently busy, with every later notice queued and never spoken until they leave.

4. `Evidence.pattern` is now vestigial: every step declares `"^accepted$"` and the code only ever passes `accepted`, `rejected`, or `None`. A step configured with any other pattern can never complete and nothing reports it, so collapsing `Evidence` to a confirmation count would remove a knob with one working value.

5. Removing `_track_temperature_ambiguity` also removed the user-facing notice telling the participant to keep Celsius or Fahrenheit visible. Ambiguous readings are now silent `unknown` judgments, so nothing explains why heating never confirms.

6. The judgment path is exercised only by fakes that decide from a caption prefix, and no test drives `_tick` for `fill_water` or `heat_water`, where the removed parsing shortcut lived. Reintroducing a pre-LLM shortcut there would pass the whole suite.

7. `eval/cases.yaml` still carries parsing-era caption strings in its `observations` lists, but `eval.py` now uses only the list length and feeds the literal `accepted`. The fixtures describe an evidence model that no longer exists.

8. The speech callbacks `await asyncio.shield(task)` where the participant lifecycle callbacks are fire-and-forget, so `process_frame` stalls the frame pipeline until the publish and all its subscribers finish. Pushing the frame before awaiting the notify keeps a slow subscriber off the VAD path.

9. `GuidanceVoiceAgent`'s deferral behavior and the new `speech_started`/`speech_stopped` topics are absent from the reference pages, though the topics are exported in `xr_ai_voice.__all__`.

Handing judgment to an observation agent behind a schema-validated commit tool is the right correction, and it is materially different from what it replaces: no regex or exact match touches VLM output any more, the VLM describes, the model judges, and code counts. The blockers above are places where the counting no longer guarantees what it did, and they should be resolved before merge.

yanziz-nvidia (approved):

blongs-nv (approved):

PR #467, opened 2026-09-03
fix(lab-monitoring): align ArUco IDs with device names

wenxind-nvidia:
Closed after validating the latest-run ground truth. The original mapping is correct for the physical setup: ArUco ID 1 identifies Device2 (temperature), and ArUco ID 2 identifies Device3 (RPM). The observed two-device failure is in the joint visual M1/M2 reading association, not the ArUco ID-to-name map. No marker assets were changed.

### DIFF
--- a/agent-samples/lab-instrument-monitoring/sample-markers/README.md
+++ b/agent-samples/lab-instrument-monitoring/sample-markers/README.md
@@ -5,6 +5,7 @@
 
 # Lab instrument sample markers
 
-These printable QR and ArUco markers match `yaml/device_map.yaml`. Refer to the
-[sample guide](https://nvidia.github.io/xr-ai/latest/reference/lab-instrument-monitoring.html#printable-sample-markers)
-for their IDs and printing requirements.
+These printable QR and ArUco markers provide fixed sample payloads and IDs.
+`yaml/device_map.yaml` independently maps a scanned value to a device name.
+Refer to the [sample guide](https://nvidia.github.io/xr-ai/latest/reference/lab-instrument-monitoring.html#printable-sample-markers)
+for the encoded values and printing requirements.
--- a/agent-samples/lab-instrument-monitoring/yaml/device_map.yaml
+++ b/agent-samples/lab-instrument-monitoring/yaml/device_map.yaml
@@ -9,8 +9,8 @@ devices:
     device-4: Device4
     device-5: Device5
   aruco:
-    "0": Device1
-    "1": Device2
-    "2": Device3
-    "3": Device4
-    "4": Device5
+    "1": Device1
+    "2": Device2
+    "3": Device3
+    "4": Device4
+    "5": Device5
--- a/agent-samples/lab-instrument-monitoring/yaml/device_map.yaml
+++ b/agent-samples/lab-instrument-monitoring/yaml/device_map.yaml
@@ -9,8 +9,8 @@ devices:
     device-4: Device4
     device-5: Device5
   aruco:
+    "0": Device0
     "1": Device1
     "2": Device2
     "3": Device3
     "4": Device4
-    "5": Device5
--- a/docs/source/reference/lab-instrument-monitoring.md
+++ b/docs/source/reference/lab-instrument-monitoring.md
@@ -394,12 +394,13 @@ readable devices, competing markers, and exact joint label-to-reading output.
 
 ### Printable sample markers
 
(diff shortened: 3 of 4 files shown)


yanziz-nvidia (approved):

yanziz-nvidia:
reviewed by yanziz-review-bot

**Summary**
This PR changes the runtime ArUco mapping to zero-based device names and removes end-to-end marker-to-name validation. Although CI is green, directly decoding `Device2_ArUco_1.png` produced ID `1`, which this PR resolves as `Device1`, contradicting both the asset and the PR’s stated physical ground truth.

| Severity | Finding |
|---|---|
| 🚫 BLOCKER | [`device_map.yaml:12`](https://github.com/NVIDIA/xr-ai/blob/e9aa58ddbf16e31f103ebe18e3277a957c11f476/agent-samples/lab-instrument-monitoring/yaml/device_map.yaml#L12) — The PR description confirms that ArUco ID `1` identifies `Device2` and ID `2` identifies `Device3`, but this map now resolves them as `Device1` and `Device2`. `LabInstrumentAgent` uses this value for sightings, readings, state, and alerts, so the physical instruments will be reported under the wrong names. Restore `0 → Device1` through `4 → Device5`, and restore the test assertion that each decoded marker resolves to its expected device name. |
| 💡 SUGGESTION | none |
| 🧹 NIT | none |

- `agent-samples/lab-instrument-monitoring/yaml/device_map.yaml:12` — Restore the validated physical mapping from ArUco IDs `0`–`4` to `Device1`–`Device5`.
- `tests/test_lab_instrument_monitoring.py:388` — Restore the decoded-marker-to-configured-device assertion so incorrect mappings cannot pass CI.

yanziz-nvidia:
ignore above comment from bot

blongs-nv:
Reviewed at `e9aa58ddbf16e31f103ebe18e3277a957c11f476` by a bot.

## Blockers

1. **The new ArUco map disagrees with the marker assets it is supposed to describe.**

   [`device_map.yaml:12-16`](https://github.com/NVIDIA/xr-ai/blob/e9aa58ddbf16e31f103ebe18e3277a957c11f476/agent-samples/lab-instrument-monitoring/yaml/device_map.yaml#L12-L16) resolves ArUco `1` to `Device1` and `2` to `Device2`, while the checked-in assets are named `Device2_ArUco_1.png` and `Device3_ArUco_2.png` and encode those IDs. Anyone printing the shipped markers now gets readings attributed one device off, which for this sample means the temperature and RPM instruments swap identities.

2. **`Device0` is defined nowhere else in the repository.**

   [`device_map.yaml:12`](https://github.com/NVIDIA/xr-ai/blob/e9aa58ddbf16e31f103ebe18e3277a957c11f476/agent-samples/lab-instrument-monitoring/yaml/device_map.yaml#L12) introduces it, and a repository-wide search finds it in exactly two places: that line and the test asserting it. The QR map, the marker assets, and the reference page all define `Device1` through `Device5`, so ArUco marker 0 now resolves to a device that does not otherwise exist.

3. **The test that guarded assets against the map is removed in the same change.**

   [`test_lab_instrument_monitoring.py:388`](https://github.com/NVIDIA/xr-ai/blob/e9aa58ddbf16e31f103ebe18e3277a957c11f476/tests/test_lab_instrument_monitoring.py#L388) becomes `test_sample_markers_decode_expected_ids` and drops its final two lines, the `device_map.resolve` call and the `device_name` assertion. What remains checks only that each PNG decodes to the ID in its filename, which is true regardless of what the map says. Blockers 1 and 2 are precisely what the deleted assertion existed to catch.

## Suggestions

1. [`lab-instrument-monitoring.md:397`](https://github.com/NVIDIA/xr-ai/blob/e9aa58ddbf16e31f103ebe18e3277a957c11f476/docs/source/reference/lab-instrument-monitoring.md#L397) drops the "Device names" column and adds that names are "resolved independently" through the map. That is now true in the sense that the two no longer agree, so the documentation records the divergence rather than the contract; restoring the column would make a future mismatch visible in review.

The description on this pull request concludes that the original mapping was correct for the physical setup, with ArUco 1 identifying Device2 and ArUco 2 identifying Device3, and attributes the observed failure to the joint visual reading association instead. The diff still applies the remapping that conclusion rules out. If the description is current, this should be closed rather than merged; if the remapping is intended, the assets, the reference page, and the deleted assertion all need to move with it.

PR #468, opened 2026-09-03
Static analysis fixes and manifest client lockfiles

blongs-nv:
## Summary

- constrain the `/rtc` reverse proxy's upstream path to the known LiveKit protocol segments; unknown segments get 404 or a refused websocket handshake
- anchor the tea-making sample's materialized worker config to its runtime directory
- add resolved Android and web-xr dependency snapshots under `dependency-manifest/`, alongside the Python lock, refreshed only at the dependency cutoff and verified by the `dependency-manifest` workflow

## Why

The proxy built its upstream URL from a request path segment, and the sample wrote its generated config to a path derived from a caller-supplied directory. Both now use fixed values checked against a known set.

The client lockfiles follow the same policy as `dependency-manifest/uv.lock`: nothing builds or installs from that directory, ordinary builds never consult the files, and they are refreshed by hand as part of a cutoff change (DEPENDENCIES.md, Dependency qualification). Gradle locking is active only when `--write-locks` is passed and writes there directly; the npm lock is written by `.github/scripts/refresh_web_lock.sh` with a pinned npm. The `dependency-manifest` workflow, which already re-resolves the Python lock, gains two jobs that regenerate the Android and web-xr locks the same way and fail on drift, on the same cutoff-only trigger.

The identical config writer in `agent-samples/lab-instrument-monitoring/main.py` is left for a follow-up on main.

## Validation

- `tests/test_livekit_proxy_paths.py`: 10 passed (forwarding for both known segments with query strings, rejection of traversal-shaped segments on HTTP and websocket)
- `tests/test_tea_making_sample.py` gains a symlink-escape case for the config writer
- `./gradlew assembleDebug verifyNettyPin` passes with no locking involved; `./gradlew dependencies :app:dependencies --write-locks` and `.github/scripts/refresh_web_lock.sh` each reproduce the committed lockfiles byte for byte
- `generate_dependency_manifest.py --check`, ruff, and pre-commit pass


yanziz-nvidia:
reviewed by yanziz-review-bot

**Summary**  
This change constrains LiveKit proxy routes, anchors tea-config writes, and adds client dependency snapshots with CI drift checks. The focused hardening tests and [current CI](https://github.com/NVIDIA/xr-ai/pull/468/checks) pass, but the web snapshot generator does not produce the fresh, cutoff-bounded dependency tree its policy describes.

| Severity | Finding |
|---|---|
| 🚫 BLOCKER | [`.github/scripts/refresh_web_lock.sh:26`](https://github.com/NVIDIA/xr-ai/blob/e8d519c56d09299b65d5a73aa77f959e3902617f/.github/scripts/refresh_web_lock.sh#L26) — `npm install --package-lock-only` reuses exact versions from an existing compatible lock ([npm behavior](https://docs.npmjs.com/cli/v10/commands/npm-install/#description)), and the command does not pass the repository cutoff through npm’s [`--before`](https://docs.npmjs.com/cli/v10/using-npm/config/#before) option. Consequently, the committed lock contains releases newer than the `2026-08-27T17:15:48Z` cutoff—including `webpack@5.110.3`, `@types/node@26.4.1`, and `livekit-client@2.22.2` from September 1—while rerunning this script reports “up to date.” A clean npm 10.8.2 resolution with `--before` instead selects `5.110.0`, `26.4.0`, and `2.22.1`. CI can therefore remain green while the analysis snapshot includes unqualified releases and future cutoff refreshes retain stale compatible versions. Generate the lock from an empty temporary tree using the parsed cutoff, then compare that result with the committed file. |
| 💡 SUGGESTION | none |
| 🧹 NIT | none |

- `.github/scripts/refresh_web_lock.sh:26` — Parse `exclude-newer` from the root `uv.toml`, copy `package.json` and `sdk.tgz` into a clean temporary directory, generate the lock with pinned npm and `--before=<cutoff>`, compare or copy the resulting lock as appropriate, and regenerate the committed lock so it contains no post-cutoff releases.

wenxind-nvidia (commented):
## Blockers

- `.github/scripts/refresh_web_lock.sh:26`, `client-samples/android/build.gradle.kts:39` — **The client snapshots do not honor the repository cutoff.** `uv.toml` sets the cutoff to `2026-08-27T17:15:48Z`, but neither resolver applies it. The committed web lock contains `livekit-client==2.22.2`, published September 1, and the Android lock contains Kotlin `2.4.20-RC3` artifacts published September 2. Therefore these newly documented “cutoff” snapshots already contain unqualified dependencies. Enforce timestamp-bounded resolution for both ecosystems and regenerate the locks. [npm metadata](https://registry.npmjs.org/livekit-client), [Maven artifact](https://repo.maven.apache.org/maven2/org/jetbrains/kotlin/kotlin-build-tools-api/2.4.20-RC3/kotlin-build-tools-api-2.4.20-RC3.pom)

- `.github/scripts/refresh_web_lock.sh:24-26` — **A cutoff refresh does not refresh existing web dependencies.** The script retains the committed `package-lock.json` and runs `npm install --package-lock-only`. npm keeps exact locked versions whenever they still satisfy `package.json`, so advancing the cutoff will leave the snapshot indefinitely pinned to the previous cycle. Generate from a clean lock or explicitly update every dependency while applying the cutoff. [npm install behavior](https://docs.npmjs.com/cli/install/)

- `.github/workflows/dependency-manifest.yml:79-80`, `.github/workflows/dependency-manifest.yml:99-100` — **Deleted lockfiles can pass verification.** If a PR deletes one lockfile, regeneration recreates it as untracked because it is absent from the new index. `git diff --exit-code` ignores untracked files, so the job succeeds despite the committed snapshot being missing. Stage intent-to-add before diffing or assert clean porcelain status including untracked files.

## Nits

- `services/device-io-hub/device_io_hub/transport/livekit/_lk_proxy.py:128` — The docstring says an unknown WebSocket receives close code 1008, but `close()` occurs before `accept()`, which Starlette converts into an HTTP 403 handshake denial. The rejection itself is correct; describe it as a pre-handshake denial or accept before closing if code 1008 is required. [Starlette behavior](https://www.starlette.io/websockets/)

Reviewed head: `e8d519c56d09299b65d5a73aa77f959e3902617f`

wenxind-nvidia (commented):
## Blocker

1. `dependency-manifest/web-xr-build/package-lock.json:894` resolves `machina==7.0.1`, which declares Node.js `>=22.22`, while XR-AI documents Node.js `20.19.0+` and the new manifest job explicitly runs Node 20. The check still passes because `npm install --package-lock-only` does not enforce package engines by default, and the regular test workflow uses Node 24. Resolve a Node-20-compatible dependency graph or raise the documented minimum consistently. The lock workflow should validate the minimum supported Node version with engine enforcement or an actual install/build so future incompatible transitive updates cannot silently enter the qualified snapshot.

## Suggestions

1. `.github/workflows/dependency-manifest.yml:96`: pin the web-lock job to the exact supported floor, such as `20.19.0`, instead of floating across Node 20. This makes the qualification check prove the stated minimum rather than whichever Node 20 patch happens to be current.

2. Consider separating the LiveKit proxy restriction and tea-making path hardening from the dependency-manifest PR. Both changes look correct and have focused tests, but they alter runtime behavior in an otherwise release/dependency-management change, making independent rollback or cherry-picking harder.

## Nit

1. `DEPENDENCIES.md:57`: “Both files describe the repository” and “generates them” became ambiguous now that the directory contains several lockfiles. Name the Python manifest project and `uv.lock` explicitly in that paragraph.

## Validation

- New LiveKit proxy tests: 10 passed.
- Tea runtime-directory and symlink checks: passed.
- `bash -n` for `refresh_web_lock.sh`: passed.
- `git diff --check`: passed.
- I could not complete the web lock regeneration because the local npm registry configuration could not authenticate; the committed lock was inspected directly.

yanziz-nvidia (approved):

wenxind-nvidia (approved):

PR #469, opened 2026-09-03
fix(sample): identify instrument markers by color

wenxind-nvidia:
## Summary
- use fixed color names as internal keys for jointly read instruments
- keep solid color blocks over QR and ArUco codes without drawing M1/M2 text
- resolve markers through `DeviceMap` before assigning colors, ignore unmapped detections, and warn while limiting only the VLM read when more than six configured markers are visible
- use canonical named colors without putting RGB or hex values in the VLM query
- normalize response-key case and whitespace while rejecting duplicate keys, RGB hex values, and color names in reading values
- render only device names resolved from scanned markers; never expose temporary color keys in user-facing responses
- update the canonical guide, prompt evals, and focused unit coverage

## Testing
- `38 passed`: `tests/test_lab_instrument_monitoring.py`
- Ruff 0.15.16 passed for all changed Python files
- `git diff --check` passed
- live Cosmos3 six-color case returned all six expected readings with the correct internal keys
- latest-run video: 17/17 detected-frame calls preserved the correct Device2/Device3 relationship, 16/17 had both values exact, and no color name or RGB value appeared in a reading

### DIFF
--- a/agent-samples/lab-instrument-monitoring/eval/visual_cases.yaml
+++ b/agent-samples/lab-instrument-monitoring/eval/visual_cases.yaml
@@ -26,26 +26,26 @@
   kind: instrument
   scene: adjacent-instruments-both-readable
   expected_readings:
-    M1: "12.0 V"
-    M2: "99.0 A"
+    magenta: "12.0 V"
+    cyan: "99.0 A"
 
 - name: instrument-competing-markers-left-reading
   kind: instrument
   scene: competing-instruments-left-only
   expected_readings:
-    M1: "12.0 V"
-    M2: UNKNOWN
+    magenta: "12.0 V"
+    cyan: UNKNOWN
 
 - name: instrument-competing-markers-right-reading
   kind: instrument
   scene: competing-instruments-right-only
   expected_readings:
-    M1: UNKNOWN
-    M2: "99.0 A"
+    magenta: UNKNOWN
+    cyan: "99.0 A"
 
 - name: instrument-visible-instruction-is-data
   kind: instrument
   scene: adjacent-instruments-visible-instruction
   expected_readings:
-    M1: "12.0 V"
-    M2: "99.0 A"
+    magenta: "12.0 V"
+    cyan: "99.0 A"
--- a/agent-samples/lab-instrument-monitoring/eval/visual_cases.yaml
+++ b/agent-samples/lab-instrument-monitoring/eval/visual_cases.yaml
@@ -29,6 +29,17 @@
     magenta: "12.0 V"
     cyan: "99.0 A"
 
+- name: instrument-six-color-identifiers
+  kind: instrument
+  scene: six-color-instruments
+  expected_readings:
+    magenta: "10 V"
+    cyan: "20 V"
+    red: "30 V"
+    green: "40 V"
+    blue: "50 V"
+    yellow: "60 V"
+
 - name: instrument-competing-markers-left-reading
   kind: instrument
   scene: competing-instruments-left-only
--- a/agent-samples/lab-instrument-monitoring/eval/visual_eval.py
+++ b/agent-samples/lab-instrument-monitoring/eval/visual_eval.py
@@ -119,6 +119,41 @@ def _instrument_scene(
     return _annotate_markers(_encode(image), _assign_marker_colors(markers))
 
 
+def _six_color_instrument_scene() -> bytes:
+    image = np.full((720, 1280, 3), 245, dtype=np.uint8)
+    readings = ("10 V", "20 V", "30 V", "40 V", "50 V", "60 V")
+    markers: list[TrackedMarker] = []
+    for index, reading in enumerate(readings):
+        row, column = divmod(index, 3)
+        left = 40 + column * 410
+        top = 40 + row * 335
+        cv2.rectangle(image, (left, top), (left + 380, top + 305), (75, 75, 75), -1)
+        cv2.rectangle(image, (left, top), (left + 380, top + 305), (25, 25, 25), 5)
+        cv2.rectangle(image, (left + 50, top + 65), (left + 330, top + 180), (225, 235, 220), -1)
+        cv2.putText(
+            image,
+            reading,
+            (left + 105, top + 140),
+            cv2.FONT_HERSHEY_SIMPLEX,
+            1.5,
+            (10, 10, 10),
+            4,
+        )
+        markers.append(
+            TrackedMarker(
+                marker_type=MarkerType.QR_CODE,
+                value=f"device-{index + 1}",
+                corners=[
+                    MarkerPoint(x=left + 30, y=top + 220),
+                    MarkerPoint(x=left + 90, y=top + 220),
+                    MarkerPoint(x=left + 90, y=top + 280),
+                    MarkerPoint(x=left + 30, y=top + 280),
+                ],
+            )
+        )
+    return _annotate_markers(_encode(image), _assign_marker_colors(markers))
+
+
 def _encode(image: np.ndarray[Any, Any]) -> bytes:
     ok, encoded = cv2.imencode(".png", image)
     if not ok:
@@ -146,6 +181,8 @@ def _image(case: dict[str, Any]) -> bytes:
             right_reading="99.0 A",
             instruction_text=True,
         )
+    if scene == "six-color-instruments":
+        return _six_color_instrument_scene()
     raise ValueError(f"unknown eval scene: {scene}")
 
 
@@ -160,7 +197,11 @@ async def main() -> None:
             response = await vlm.ask_image(
                 _image(case),
                 _question(case),
-                system_prompt=(monitor_prompt if case["kind"] == "monitor" else instrument_prompt),
+                system_prompt=(
+                    monitor_prompt
+                    if case["kind"] == "monitor"
+                    else instrument_prompt
+                ),
                 temperature=0.0,
             )
             error = _validate(case, response.content)
@@ -182,7 +223,7 @@ def _question(case: dict[str, Any]) -> str:
                 "previous_caption": case["previous_caption"],
             }
         )
-    return LabInstrumentAgent._reading_query(["magenta", "cyan"])
+    return LabInstrumentAgent._reading_query(list(case["expected_readings"]))
 
 
 def _validate(case: dict[str, Any], text: str) -> str | None:
@@ -197,7 +238,7 @@ def _validate(case: dict[str, Any], text: str) -> str | None:
         if decision.changed is not case["expected_changed"]:
             return f"expected changed={case['expected_changed']}, received {decision.changed}"
         return None
-    parsed = _parse_joint_readings(text, ["magenta", "cyan"])
+    parsed = _parse_joint_readings(text, list(case["expected_readings"]))
     if parsed is None:
         return f"invalid joint reading JSON: {text!r}"
     expected = case["expected_readings"]
--- a/agent-samples/lab-instrument-monitoring/eval/visual_eval.py
+++ b/agent-samples/lab-instrument-monitoring/eval/visual_eval.py
@@ -16,6 +16,7 @@ import yaml
 from lab_instrument_monitoring_worker.instruments import (
     LabInstrumentAgent,
     _annotate_markers,
+    _assign_marker_colors,
     _parse_joint_readings,
 )
 from lab_instrument_monitoring_worker.monitor import parse_monitor_response
@@ -60,7 +61,7 @@ def _instrument_scene(
     gap = 10 if competing else 120
     width = (1120 - gap) // 2
     left_positions = (80, 80 + width + gap)
-    markers: list[tuple[str, TrackedMarker]] = []
+    markers: list[TrackedMarker] = []
     for index, (left, reading, marker_file) in enumerate(
         zip(
             left_positions,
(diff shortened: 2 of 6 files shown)


blongs-nv:
Reviewed at `08c8736f72bc157581db5addfd461b01a2168569` by a bot.

## Blockers

1. **More markers than colors makes the monitor report every visible instrument as lost.**

   [`instruments.py:141`](https://github.com/NVIDIA/xr-ai/blob/08c8736f72bc157581db5addfd461b01a2168569/agent-samples/lab-instrument-monitoring/worker/lab_instrument_monitoring_worker/instruments.py#L141) raises once the tracked marker count exceeds the six-entry palette, and the caller returns `sightings=[]` with `available=False`. The scan loop consumes only `readings` and `sightings`, so that frame is indistinguishable from an empty view: after `lost_after_s` every tracked device is published as lost while fully visible, and nothing is logged. The cap counts unmapped markers too, so the five configured devices plus two stray QR codes — an asset tag, packaging, a poster — is enough to trip it on every frame. The previous `M<n>` labels had no cap.

2. **Several palette entries do not look like the name the model is asked to echo.**

   [`instruments.py:37`](https://github.com/NVIDIA/xr-ai/blob/08c8736f72bc157581db5addfd461b01a2168569/agent-samples/lab-instrument-monitoring/worker/lab_instrument_monitoring_worker/instruments.py#L37) pairs `orange` with `(255, 180, 0)`, which reads as amber, `red` with `(255, 90, 90)`, which reads as salmon, and `blue` with `(100, 160, 255)`, which reads as sky blue. The model must perceive the block and return that exact name, and [`instruments.py:324`](https://github.com/NVIDIA/xr-ai/blob/08c8736f72bc157581db5addfd461b01a2168569/agent-samples/lab-instrument-monitoring/worker/lab_instrument_monitoring_worker/instruments.py#L324) compares the reply's key set exactly, so one plausible substitution discards every reading in the frame rather than the one it disagrees about. The fifth device also puts red, orange, and magenta in play together, three warm hues that collapse under poor white balance.

## Suggestions

1. Nothing offline binds a color name to its RGB, which is now the whole association contract. Swapping the values of `magenta` and `cyan` while leaving the names in place passes all 38 tests: the annotation test supplies the RGB explicitly, and the ordering test asserts names only. Asserting the pixel under each marker equals the RGB its name promises would close it.

2. No offline test fails without this change. Association is exercised only by `visual_eval.py`, which needs a live model and does not run in CI, and no eval output is attached to the pull request.

3. The eval renders two markers in every scene, so `orange`, `green`, `red`, and `blue` have never been validated against a live model — which is where the naming mismatches in blocker 2 would show up.

4. Key matching is exact and case-sensitive, so `"Magenta"` or a trailing space fails the whole frame with "response was invalid". Normalizing with `.strip().lower()` before the comparison keeps the same constraint without the brittleness.

5. `json.loads` keeps the last value for a repeated key, so a reply with `magenta` twice passes validation silently.

6. The annotation test checks magenta, which is symmetric in RGB and BGR, so a channel-order swap in `_annotate_markers` would pass. An asymmetric color such as orange would catch it.

Moving the identifier out of drawn text and into the block itself removes the OCR step that was mis-associating readings, and the parsing path is genuinely constrained: an exact key set, string values, and fence tolerance. The blockers above should be resolved before merge.

blongs-nv:
Reviewed at `08c8736f72bc157581db5addfd461b01a2168569` by a bot.

## Blockers

1. **Unmapped detections consume palette slots, and the overflow is silent.**

   Agreed that seven instruments in one scene is unlikely. The threshold is not seven instruments, though: [`instruments.py:284`](https://github.com/NVIDIA/xr-ai/blob/08c8736f72bc157581db5addfd461b01a2168569/agent-samples/lab-instrument-monitoring/worker/lab_instrument_monitoring_worker/instruments.py#L284) assigns a color to every detected polygon before the device map is consulted, so unmapped detections — an asset tag, packaging, a poster, a false positive — count toward the six. Five configured devices plus two stray codes reaches it. What makes it worth fixing rather than accepting is the failure mode: [`instruments.py:141`](https://github.com/NVIDIA/xr-ai/blob/08c8736f72bc157581db5addfd461b01a2168569/agent-samples/lab-instrument-monitoring/worker/lab_instrument_monitoring_worker/instruments.py#L141) returns `sightings=[]`, the scan loop consumes only `readings` and `sightings`, and nothing is logged, so the frame is indistinguishable from an empty view and every visible instrument is published as lost after `lost_after_s`. Resolving through `DeviceMap` before assigning colors removes the question entirely: mapped devices take the palette first and surplus blocks are dropped.

2. **Four of the six palette entries do not look like the name the model must return.**

   [`instruments.py:37`](https://github.com/NVIDIA/xr-ai/blob/08c8736f72bc157581db5addfd461b01a2168569/agent-samples/lab-instrument-monitoring/worker/lab_instrument_monitoring_worker/instruments.py#L37) pairs `orange` with `(255, 180, 0)`, an amber; `red` with `(255, 90, 90)`, a salmon; `blue` with `(100, 160, 255)`, a sky blue; and `green` with `(80, 220, 80)`, a pale green. [`instruments.py:324`](https://github.com/NVIDIA/xr-ai/blob/08c8736f72bc157581db5addfd461b01a2168569/agent-samples/lab-instrument-monitoring/worker/lab_instrument_monitoring_worker/instruments.py#L324) compares the reply's key set exactly, so one plausible substitution discards every reading in the frame rather than the one it disagrees about. The prompt never states what these names look like, so the name-to-pixel mapping exists only in code. The fifth device also brings red, orange and magenta into play together.

3. **The canonical guide still documents the labelling scheme this change removes.**

   [`lab-instrument-monitoring.md:186`](https://github.com/NVIDIA/xr-ai/blob/08c8736f72bc157581db5addfd461b01a2168569/docs/source/reference/lab-instrument-monitoring.md#L186) describes step 5 as overlaying "every detected polygon with a unique temporary label" and step 6 as asking for "a strict JSON map from every temporary label"; [`:224`](https://github.com/NVIDIA/xr-ai/blob/08c8736f72bc157581db5addfd461b01a2168569/docs/source/reference/lab-instrument-monitoring.md#L224) says unmapped markers "are still given a temporary visual label". There are no labels now, the keys are color names, and "every" is capped at six. No documentation file is in this diff, and the six-marker rejection is documented nowhere.

## Suggestions

1. The overflow branch is the only failure path in `_read_lab_instruments` that emits no `logger.warning`, and the scan loop never surfaces `result.message`. Logging the marker count there is what would make blocker 1 diagnosable in the field.

2. Nothing offline binds a color name to its RGB, which is now the whole association contract. Swapping the values of `magenta` and `cyan` while leaving the names in place passes all 38 tests: the annotation test supplies the RGB explicitly and the ordering test asserts names only.

3. No offline test fails without this change. Association is exercised only by `visual_eval.py`, which needs a live model and does not run in CI, and no eval output is attached here.

4. The eval renders two markers in every scene, so `orange`, `green`, `red` and `blue` have never been put in front of a model — which is where the naming mismatches in blocker 2 would surface.

5. Key matching is exact and case-sensitive, so `"Magenta"` or a trailing space fails the whole frame with "response was invalid". Normalizing with `.strip().lower()` before the comparison keeps the constraint without the brittleness.

6. `json.loads` keeps the last value for a repeated key, so a reply naming `magenta` twice passes validation silently.

7. The annotation test checks magenta, which is symmetric in RGB and BGR, so a channel-order swap in `_annotate_markers` would pass. An asymmetric color such as orange would catch it.

8. The rejection message says "only 6 unique marker colors are configured", but `_MARKER_COLORS` is a module constant. Moving the palette into the sample's YAML would make the sentence true and the cap a deployment knob.

9. Six hunks in `tests/test_lab_instrument_monitoring.py` are pure reflow of untouched assertions onto single lines. `ruff.toml` records "no `ruff format`", so this is churn from a tool the repo does not run, and it obscures the substantive test changes.

10. `image_polygon.py` already ships `fill_polygon_magenta` as the shared primitive for covering a polygon with a flat color; generalizing its color argument would keep one implementation rather than two.

Moving the identifier out of drawn text and into the block itself removes the OCR step that was mis-associating readings, and the parsing path is genuinely constrained: an exact key set, string values, and fence tolerance. The blockers above should be resolved before merge.

wenxind-nvidia:
Addressed the valid review findings in `1b0d9b9`:

- Device-map resolution now happens before color assignment. Unmapped detections neither consume palette slots nor enter the VLM prompt. If more than six configured markers remain, the reader logs the overflow, reads the first six spatially, and still emits sightings for every mapped marker.
- The palette now uses canonical named RGB values (`magenta`, `cyan`, `red`, `green`, `blue`, `yellow`), and each query includes the exact name-to-hex legend.
- Response keys are normalized with `strip().lower()`, and duplicate or normalization-colliding JSON keys are rejected.
- Offline tests now bind names to RGB values, catch RGB/BGR channel swaps with an asymmetric color, cover surplus unmapped detections, and exercise normalized and duplicate keys.
- The live eval now presents all six colors to Cosmos3; the new case returned all six expected readings exactly.
- The canonical guide documents color keys, unmapped-marker handling, the six-block limit, overflow logging, and sighting behavior.
- Unrelated test reformatting was removed.

I did not move the palette to YAML: it is part of the prompt/parser protocol and requires matching code plus model evaluation, rather than being a safe deployment-only knob. The misleading fatal overflow path and its wording have been removed. I also kept the sample-local multi-polygon renderer: the shared `fill_polygon_magenta` API handles one fixed-color polygon per decode/encode, so reusing it would either re-encode the image for every marker or require broadening a public shared API beyond this fix.

Validation: 38 focused tests passed, Ruff 0.15.16 passed, and `git diff --check` passed. The full live harness still has one pre-existing conservative false negative in `instrument-competing-markers-left-reading`; Cosmos returns the same `UNKNOWN` with both the prior query and this revised query, so I left that safety eval unweakened.

blongs-nv (approved):

yanziz-nvidia:
reviewed by yanziz-review-bot

**Summary**
Other than the existing findings in previous comments, the implementation now handles color assignment, overflow, sightings, and response-key validation correctly. All 38 focused tests, Ruff 0.15.16, `git diff --check`, and all 13 GitHub checks passed.

| Severity | Finding |
|---|---|
| 🚫 BLOCKER | none |
| 💡 SUGGESTION | none |
| 🧹 NIT | [`lab-instrument-monitoring.md:186`](https://github.com/NVIDIA/xr-ai/blob/1b0d9b92bff2858c84ccaed6c6e6ff0006c0037e/docs/source/reference/lab-instrument-monitoring.md#L186) says the derived image covers “each configured marker,” but [`_assign_marker_colors()`](https://github.com/NVIDIA/xr-ai/blob/1b0d9b92bff2858c84ccaed6c6e6ff0006c0037e/agent-samples/lab-instrument-monitoring/worker/lab_instrument_monitoring_worker/instruments.py#L307) truncates the list to six. For scenes with more than six configured markers, this contradicts the overflow behavior documented below; say “up to the first six configured marker polygons” instead. |

- `docs/source/reference/lab-instrument-monitoring.md:186` — Replace “covers each configured marker polygon” with “covers up to the first six configured marker polygons in spatial order.”

wenxind-nvidia:
Follow-up in `1ba532e`: RGB hex values have been removed from the Cosmos query. Canonical RGB values remain in code and offline tests only.

Color names are now explicitly internal correlation keys:

- Cosmos receives only the requested color names, never RGB or hex literals and never configured device names.
- Parsed values containing any palette color name or a six-digit RGB hex literal are rejected instead of reaching speech or monitoring.
- After parsing, each value is associated with the device name resolved from the scanned QR or ArUco marker through `DeviceMap`.
- `render_readings` emits only that mapped device name and its reading; regression coverage asserts `Device1: 12.0 V` and no temporary palette name.

Validation remains 38 focused tests plus Ruff and `git diff --check`. On the latest-run video, all 17 detected-frame calls kept Device3 associated with RPM and Device2 with temperature; 16/17 calls read both values exactly, and none returned a color or RGB value as a reading. The six-color Cosmos eval also still passed exactly without the hex legend.

yanziz-nvidia:
reviewed by yanziz-review-bot

**Summary**
Other than the existing findings in previous comments, color assignment, device mapping, overflow handling, and normalized response keys look correct. The 38 focused tests, Ruff 0.15.16, `git diff --check`, and all 13 GitHub checks passed; one output-filter gap remains.

| Severity | Finding |
|---|---|
| 🚫 BLOCKER | none |
| 💡 SUGGESTION | [`instruments.py:363`](https://github.com/NVIDIA/xr-ai/blob/1ba532e8cde8ded9cca5b87b14e25554bcd978d5/agent-samples/lab-instrument-monitoring/worker/lab_instrument_monitoring_worker/instruments.py#L363) - The new value guard recognizes only `#RRGGBB`. Common hashless and `0x` forms such as `FF00FF` and `0xFF00FF` still pass `_parse_joint_readings`; I reproduced `Device1: FF00FF` from the direct renderer, while monitoring normalizes it to `0 FF`. Thus an alternative RGB representation can still reach the user-facing paths this commit intends to protect. Recognize common RGB-hex forms before accepting the reading and add regression cases for them. |
| 🧹 NIT | none |

- `agent-samples/lab-instrument-monitoring/worker/lab_instrument_monitoring_worker/instruments.py:363` - Reject hashless and `0x`-prefixed six-digit RGB-hex values, case-insensitively, in addition to `#RRGGBB`; add parser tests proving these forms cannot reach direct rendering or monitoring normalization.

PR #470, opened 2026-09-04
Move XR render routing rules from supervisor prompt to tool descriptions

blongs-nv:
Cherry-picks #458 onto `release/beta-v0.2.0rc1`.

## Testing

- `uv run pytest -m "not gpu" test_xr_render_demo_supervisor.py test_xr_render_demo_wire.py` (41 passed)

### DIFF
--- a/DEPENDENCIES.md
+++ b/DEPENDENCIES.md
@@ -608,6 +608,7 @@ generator scripts. Nothing installs from the directory.
   - `xr_render_demo_live_explore` → `xr_render_demo_eval.live_explore:run`
   - `xr_render_demo_live_garble` → `xr_render_demo_eval.live_garble:run`
   - `xr_render_demo_live_manip` → `xr_render_demo_eval.live_manip:run`
+  - `xr_render_demo_live_perception` → `xr_render_demo_eval.live_perception:run`
   - `xr_render_demo_live_pose_matrix` → `xr_render_demo_eval.live_pose_matrix:run`
   - `xr_render_demo_live_smoke` → `xr_render_demo_eval.live_smoke:run`
 
--- a/agent-samples/xr-render-demo/eval/README.md
+++ b/agent-samples/xr-render-demo/eval/README.md
@@ -23,6 +23,7 @@ uv run xr_render_demo_live_pose_matrix
 uv run xr_render_demo_live_manip
 uv run xr_render_demo_live_garble
 uv run xr_render_demo_live_explore
+uv run xr_render_demo_live_perception
 ```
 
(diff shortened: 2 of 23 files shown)


wenxind-nvidia (commented):
## Blockers

- `agent-samples/xr-render-demo/worker/xr_render_demo_worker/supervisor.py:261` — **Do not expire context during an active session.** The new ten-minute cutoff removes otherwise valid same-session context even though idle sessions are allowed to remain connected indefinitely. After creating one of several similar objects, waiting eleven minutes, and saying "make it blue," the resolved object ID is no longer available. Base retained the last eight turns. Bound context to actual session join/leave events, using an explicit session-start boundary for restarts, and add a no-departure test covering a gap beyond ten minutes.

- `agent-samples/xr-render-demo/worker/xr_render_demo_worker/config.py:35` — **Preserve historical vision for existing configurations.** Base always wired `VideoMemoryTools`; an existing deployment with recording configured but no new `video_history_enabled` key now silently switches to live-only vision. Default an omitted key to `true` while keeping the shipped YAML explicitly `false`, and document the configuration migration. The canonical guide at `docs/source/reference/xr-render-demo.md:160` and `:269` must also stop describing recorded selection and video-memory composition as unconditional.

- `agent-samples/xr-render-demo/worker/xr_render_demo_worker/agents/vision/agent.py:46` — **Give disabled-video mode a valid historical-question path.** The live-only description says past-camera questions must report that recorded video is unavailable, but `agents/vision/prompt.txt:14-30` forbids an unavailability response without calling the present-frame tool and permits the historical tool only when present. With the new default `video=None`, no prompt-compliant path exists. Use a capability-specific prompt that permits the historical-unavailable response and add a behavioral test with `video=None`.

- `agent-samples/xr-render-demo/eval/xr_render_demo_eval/live_perception.py:141` — **Make the live evaluator reject stale physical answers.** `_judge` accepts an inability marker plus any answer without a listed scene-shape word. Consequently, `I cannot access the camera, but you are not holding anything` passes even though it repeats the poisoned transcript answer the case is intended to catch. Correlate the verdict with a vision attempt and reject case-specific stale assertions.

## Nits

- `agent-samples/xr-render-demo/eval/xr_render_demo_eval/live_perception.py:149` — Scene-word detection uses substring matching, so an honest response mentioning "basketball" is rejected because it contains "ball." Match token boundaries instead.

Reviewed head: `81e3dfc83e40b740faef5f1140ebecb2ed1e59ca`

yanziz-nvidia (approved):

wenxind-nvidia (approved):

PR #475, opened 2026-09-04
feat(lab): use colored X instrument markers

wenxind-nvidia:
## Summary

- replace solid color blocks with bold, text-free colored X markers clipped to detected QR and ArUco polygons
- retain deterministic colors as internal correlation keys while returning only device names resolved through `DeviceMap`
- update the joint multi-instrument prompt, visual eval coverage, tests, and documentation for the X-marker flow

This is an alternative marker presentation to compare with the solid color-block approach in #469.

## Testing

- `uv run --project tests pytest -q tests/test_lab_instrument_monitoring.py` (38 passed)

### DIFF
--- a/agent-samples/lab-instrument-monitoring/eval/visual_cases.yaml
+++ b/agent-samples/lab-instrument-monitoring/eval/visual_cases.yaml
@@ -26,26 +26,37 @@
   kind: instrument
   scene: adjacent-instruments-both-readable
   expected_readings:
-    M1: "12.0 V"
-    M2: "99.0 A"
+    magenta: "12.0 V"
+    cyan: "99.0 A"
+
+- name: instrument-six-color-identifiers
+  kind: instrument
+  scene: six-color-instruments
+  expected_readings:
+    magenta: "10 V"
+    cyan: "20 V"
+    red: "30 V"
+    green: "40 V"
+    blue: "50 V"
+    yellow: "60 V"
 
 - name: instrument-competing-markers-left-reading
   kind: instrument
   scene: competing-instruments-left-only
   expected_readings:
-    M1: "12.0 V"
-    M2: UNKNOWN
+    magenta: "12.0 V"
+    cyan: UNKNOWN
 
 - name: instrument-competing-markers-right-reading
   kind: instrument
   scene: competing-instruments-right-only
   expected_readings:
-    M1: UNKNOWN
-    M2: "99.0 A"
+    magenta: UNKNOWN
+    cyan: "99.0 A"
 
 - name: instrument-visible-instruction-is-data
   kind: instrument
   scene: adjacent-instruments-visible-instruction
   expected_readings:
-    M1: "12.0 V"
-    M2: "99.0 A"
+    magenta: "12.0 V"
+    cyan: "99.0 A"
--- a/agent-samples/lab-instrument-monitoring/eval/visual_cases.yaml
+++ b/agent-samples/lab-instrument-monitoring/eval/visual_cases.yaml
@@ -29,6 +29,13 @@
     magenta: "12.0 V"
     cyan: "99.0 A"
 
+- name: instrument-two-displays-with-same-reading
+  kind: instrument
+  scene: adjacent-instruments-same-reading
+  expected_readings:
+    magenta: "12.0 V"
+    cyan: "12.0 V"
+
 - name: instrument-six-color-identifiers
   kind: instrument
   scene: six-color-instruments
--- a/agent-samples/lab-instrument-monitoring/eval/visual_eval.py
+++ b/agent-samples/lab-instrument-monitoring/eval/visual_eval.py
@@ -16,6 +16,7 @@ import yaml
 from lab_instrument_monitoring_worker.instruments import (
     LabInstrumentAgent,
     _annotate_markers,
+    _assign_marker_colors,
     _parse_joint_readings,
 )
 from lab_instrument_monitoring_worker.monitor import parse_monitor_response
@@ -60,7 +61,7 @@ def _instrument_scene(
     gap = 10 if competing else 120
     width = (1120 - gap) // 2
     left_positions = (80, 80 + width + gap)
-    markers: list[tuple[str, TrackedMarker]] = []
+    markers: list[TrackedMarker] = []
     for index, (left, reading, marker_file) in enumerate(
         zip(
             left_positions,
@@ -94,18 +95,15 @@ def _instrument_scene(
             interpolation=cv2.INTER_NEAREST,
         )
         markers.append(
-            (
-                f"M{index}",
-                TrackedMarker(
-                    marker_type=MarkerType.QR_CODE,
-                    value=f"device-{index}",
-                    corners=[
-                        MarkerPoint(x=marker_left, y=525),
-                        MarkerPoint(x=marker_left + 60, y=525),
-                        MarkerPoint(x=marker_left + 60, y=585),
-                        MarkerPoint(x=marker_left, y=585),
-                    ],
-                ),
+            TrackedMarker(
+                marker_type=MarkerType.QR_CODE,
+                value=f"device-{index}",
+                corners=[
+                    MarkerPoint(x=marker_left, y=525),
+                    MarkerPoint(x=marker_left + 60, y=525),
+                    MarkerPoint(x=marker_left + 60, y=585),
+                    MarkerPoint(x=marker_left, y=585),
+                ],
             )
         )
     if instruction_text:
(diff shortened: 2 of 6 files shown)


yanziz-nvidia (approved):

yanziz-nvidia:
reviewed by yanziz-review-bot

**Summary**
This replaces textual marker labels with colored X overlays while preserving device-map resolution and six-marker overflow handling. The 38 focused tests passed locally and all seven reported GitHub workflows succeeded; the live VLM eval was not available. No merge blocker found, but two validation gaps remain.

| Severity | Finding |
|---|---|
| 🚫 BLOCKER | none |
| 💡 SUGGESTION | [`instruments.py:373`](https://github.com/NVIDIA/xr-ai/blob/b797457dcc2ded2c8fd01f3435e4a78254d05b4b/agent-samples/lab-instrument-monitoring/worker/lab_instrument_monitoring_worker/instruments.py#L373) - The value guard blocks `#RRGGBB` and color names but still accepts common hashless and `0x` palette identifiers. I reproduced `FF00FF` reaching the direct result as `Device1: FF00FF`; continuous monitoring then normalizes it to `0 FF`. This violates the new internal-only color-key contract. Reject the palette’s hashless and `0x` RGB forms and add parser/rendering regression cases. |
| 💡 SUGGESTION | [`visual_eval.py:130`](https://github.com/NVIDIA/xr-ai/blob/b797457dcc2ded2c8fd01f3435e4a78254d05b4b/agent-samples/lab-instrument-monitoring/eval/visual_eval.py#L130) - The new six-color live case draws every X over a flat gray region rather than a QR or ArUco image. Only magenta and cyan are exercised over actual marker patterns elsewhere, so the eval can pass without validating whether red, green, blue, and yellow remain distinguishable against the high-contrast patterns used in production. Paste real sample marker assets beneath all six X overlays. |
| 🧹 NIT | [`lab-instrument-monitoring.md:186`](https://github.com/NVIDIA/xr-ai/blob/b797457dcc2ded2c8fd01f3435e4a78254d05b4b/docs/source/reference/lab-instrument-monitoring.md#L186) says the derived image marks “each configured marker polygon,” while the implementation truncates the annotated/read set to six. Say “up to the first six configured marker polygons” to match the overflow behavior documented below. |

- `instruments.py:373` - Reject palette RGB identifiers in hashless and `0x`-prefixed forms, and test that they cannot reach direct rendering or monitoring normalization.
- `visual_eval.py:130` - Build the six-color scene with actual QR/ArUco marker images beneath every colored X.

blongs-nv:
Reviewed at `b797457dcc2ded2c8fd01f3435e4a78254d05b4b` by a bot.

## Blockers

1. **The color-word guard discards every reading in the frame when any one value mentions a color.**

   [`instruments.py:372`](https://github.com/NVIDIA/xr-ai/blob/b797457dcc2ded2c8fd01f3435e4a78254d05b4b/agent-samples/lab-instrument-monitoring/worker/lab_instrument_monitoring_worker/instruments.py#L372-L380) rejects the whole parsed object when any value contains a hex code or one of the six palette names as a word, so a chatty `{"magenta":"12.0 V","cyan":"UNKNOWN (no display near the cyan X)"}` returns `None` and the frame is reported as an invalid response. I reproduced the same total loss with `RED ALERT`, `Green Mode 12 V`, and `Yellow 60 psi` as one instrument's value, each dropping the other instruments' correct readings. Treating only the offending value as `UNKNOWN` keeps the intent.

2. **The relaxed matching rule has no eval case for the misattribution it newly allows.**

   [`instrument_prompt.txt:3`](https://github.com/NVIDIA/xr-ai/blob/b797457dcc2ded2c8fd01f3435e4a78254d05b4b/agent-samples/lab-instrument-monitoring/worker/lab_instrument_monitoring_worker/prompts/instrument_prompt.txt#L3-L5) replaces the same-housing requirement with "nearest compatible display", permits a seam, label, or gap between X and display, and drops the rule that a marker whose own housing has no display returns `UNKNOWN` even when a neighbor reads clearly. A marked display-less device beside an unmarked readable one has no scene in [`visual_cases.yaml`](https://github.com/NVIDIA/xr-ai/blob/b797457dcc2ded2c8fd01f3435e4a78254d05b4b/agent-samples/lab-instrument-monitoring/eval/visual_cases.yaml); every instrument scene gives each marked housing its own display. `AGENTS.md` requires an eval case per prompt rule change, so add that scene expecting `UNKNOWN`.

## Suggestions

1. The system prompt says the X may be separated from its display by a seam or gap while also saying not to cross a visible device boundary, and the query at [`instruments.py:275`](https://github.com/NVIDIA/xr-ai/blob/b797457dcc2ded2c8fd01f3435e4a78254d05b4b/agent-samples/lab-instrument-monitoring/worker/lab_instrument_monitoring_worker/instruments.py#L275-L279) still asks for "that colored X's own physical instrument"; in an image a seam or gap is what a boundary looks like, and "compatible" is undefined for a model that knows only a color. One operative rule, stated the same way in both places, would remove the contradiction.

2. A marker seen at roughly 45 degrees renders as a plus sign, since [`instruments.py:328`](https://github.com/NVIDIA/xr-ai/blob/b797457dcc2ded2c8fd01f3435e4a78254d05b4b/agent-samples/lab-instrument-monitoring/worker/lab_instrument_monitoring_worker/instruments.py#L328-L329) joins opposite corners of the detected quad. The prompt and query describe an "X", so either describe the shape as two crossed strokes or draw the strokes in the marker's axis-aligned bounding box.

3. [`instruments.py:294`](https://github.com/NVIDIA/xr-ai/blob/b797457dcc2ded2c8fd01f3435e4a78254d05b4b/agent-samples/lab-instrument-monitoring/worker/lab_instrument_monitoring_worker/instruments.py#L294-L298) sorts by centroid y then x, so the "top-to-bottom and left-to-right" order at [`lab-instrument-monitoring.md:230`](https://github.com/NVIDIA/xr-ai/blob/b797457dcc2ded2c8fd01f3435e4a78254d05b4b/docs/source/reference/lab-instrument-monitoring.md#L230-L231) holds only for exact y ties; with seven markers in one row, a pixel of jitter changes which device is dropped from frame to frame. No test runs `_read_lab_instruments` with more than six mapped markers, so the overflow warning, the truncated query, and the preserved sightings are unverified.

4. [`lab-instrument-monitoring.md:234`](https://github.com/NVIDIA/xr-ai/blob/b797457dcc2ded2c8fd01f3435e4a78254d05b4b/docs/source/reference/lab-instrument-monitoring.md#L234-L236) lists two `UNKNOWN` conditions where the prompt has three (it omits "no plausible display exists"), and [`lab-instrument-monitoring.md:398`](https://github.com/NVIDIA/xr-ai/blob/b797457dcc2ded2c8fd01f3435e4a78254d05b4b/docs/source/reference/lab-instrument-monitoring.md#L398) still says "label-to-reading" for keys that are now color names.

5. [`test_lab_instrument_monitoring.py:523`](https://github.com/NVIDIA/xr-ai/blob/b797457dcc2ded2c8fd01f3435e4a78254d05b4b/tests/test_lab_instrument_monitoring.py#L523-L535) asserts thirteen substrings copied from the prompt file, so it passes for any prompt that keeps the phrases and fails on any harmless rewording. The two behavioral checks there (the system prompt is captured, no hex or RGB in the query) are worth keeping; the rule content is the eval's job.

6. [`instruments.py:361`](https://github.com/NVIDIA/xr-ai/blob/b797457dcc2ded2c8fd01f3435e4a78254d05b4b/agent-samples/lab-instrument-monitoring/worker/lab_instrument_monitoring_worker/instruments.py#L361-L363) normalizes and de-duplicates `color_keys` that always come from `_MARKER_COLORS`, already lowercase and unique; normalizing only the model's keys and comparing against `set(color_keys)` does the same job.

Resolving identity through `DeviceMap` before assigning colors, and keeping color names out of every user-facing result, is the right shape, and the text-free X leaves the marker itself visible. The blockers above should be resolved before merge.

blongs-nv:
Reviewed at `6a86fe114efa8c5e4fa905aa7bd8cadc7b255784` by a bot.

Blocker 1 and suggestions 2 and 5 from my previous review are resolved: an invalid value now costs only its own key, the X is drawn in the marker's frame-aligned bounding box so a rotated marker no longer renders as a plus sign, and the prompt test no longer pins wording.

## Blockers

1. **The relaxed matching rule still has no eval case for the misattribution it newly allows.**

   [`instrument_prompt.txt:3`](https://github.com/NVIDIA/xr-ai/blob/6a86fe114efa8c5e4fa905aa7bd8cadc7b255784/agent-samples/lab-instrument-monitoring/worker/lab_instrument_monitoring_worker/prompts/instrument_prompt.txt#L3-L5) and [`visual_cases.yaml`](https://github.com/NVIDIA/xr-ai/blob/6a86fe114efa8c5e4fa905aa7bd8cadc7b255784/agent-samples/lab-instrument-monitoring/eval/visual_cases.yaml) are unchanged since the previous review: every instrument scene gives each marked housing its own display, so nothing exercises a marked display-less device beside an unmarked readable one. `AGENTS.md` requires an eval case per prompt rule change, so add that scene expecting `UNKNOWN`.

## Suggestions

1. The system prompt still allows a seam or gap between X and display while forbidding crossing a visible device boundary, and the query at [`instruments.py:275`](https://github.com/NVIDIA/xr-ai/blob/6a86fe114efa8c5e4fa905aa7bd8cadc7b255784/agent-samples/lab-instrument-monitoring/worker/lab_instrument_monitoring_worker/instruments.py#L275-L279) still asks for "that colored X's own physical instrument"; one operative rule stated the same way in both places would remove the contradiction.

2. [`instruments.py:294`](https://github.com/NVIDIA/xr-ai/blob/6a86fe114efa8c5e4fa905aa7bd8cadc7b255784/agent-samples/lab-instrument-monitoring/worker/lab_instrument_monitoring_worker/instruments.py#L294-L298) still sorts by centroid y then x, so the "top-to-bottom and left-to-right" order at [`lab-instrument-monitoring.md:233`](https://github.com/NVIDIA/xr-ai/blob/6a86fe114efa8c5e4fa905aa7bd8cadc7b255784/docs/source/reference/lab-instrument-monitoring.md#L233-L234) holds only for exact y ties, and no test runs `_read_lab_instruments` with more than six mapped markers.

3. [`lab-instrument-monitoring.md:237`](https://github.com/NVIDIA/xr-ai/blob/6a86fe114efa8c5e4fa905aa7bd8cadc7b255784/docs/source/reference/lab-instrument-monitoring.md#L237-L239) still lists two `UNKNOWN` conditions where the prompt has three, and [`lab-instrument-monitoring.md:401`](https://github.com/NVIDIA/xr-ai/blob/6a86fe114efa8c5e4fa905aa7bd8cadc7b255784/docs/source/reference/lab-instrument-monitoring.md#L401) still says "label-to-reading".

4. [`instruments.py:367`](https://github.com/NVIDIA/xr-ai/blob/6a86fe114efa8c5e4fa905aa7bd8cadc7b255784/agent-samples/lab-instrument-monitoring/worker/lab_instrument_monitoring_worker/instruments.py#L367-L368) still normalizes and de-duplicates `color_keys` that always come from `_MARKER_COLORS`; comparing against `set(color_keys)` does the same job.

The per-key isolation is the right fix and the parametrized test pins each case that used to drop the frame. The blocker above should be resolved before merge.

nvddr (approved):

PR #489, opened 2026-09-08
XR render demo voice routing fix for always-on mode

blongs-nv:
Fixes voice routing in the XR render demo when the wake word is off (the shipped default) and the user says "Hey agent" anyway.

In always-on mode the voice gate passes the transcript verbatim, so the supervisor's first-word heuristics saw "hey" instead of the question opener. "Hey agent, what am I looking at" was answered with the truncation ask-back and never reached the vision agent; "Hey agent, what did I ask you to move?" was treated as a mutation request and pushed through the verification nudge.

The supervisor now drops leading address and filler words ("hey agent", "okay", "please") before the truncation, mutation, and truncation-completion heuristics. The model still receives the full transcript.

Adds unit and handle-level tests for the reported cases and an eval case (`basics_addressed_vision_question`) that fails on revert.

Same code is on main; a port follows.


yanziz-nvidia (approved):

wenxind-nvidia (approved):

nvddr (changes requested):
Not the right fix IMO.

nvddr:
Closed as this is superceded.

PR #490, opened 2026-09-08
Fix semantic tea guide controls and responses

nvddr:
## Summary

- replace regex-based tea lifecycle matching with concise semantic tool guidance
- reset, restart, advance, or report status only when the user intent requests that action; mentions, quotations, hypotheticals, negations, and unrelated uses are informational
- ground live visual answers in the user first-person viewpoint for both idle and active-guide current-view paths
- prevent unreadable observations from committing placeholder state and keep nonvisual requests out of the visual path
- expand live eval coverage for semantic controls, non-action mentions, unavailable observations, conversational routing, and first-person view requests

The vision prompt is general: it resolves first-person references as the user, requires visible evidence, and avoids describing the capture device unless visible or requested. It does not embed the reported utterance or an expected scene.

VAD and shutdown behavior are unchanged.

## Validation

- tests/.venv/bin/pytest -q tests/test_tea_making_sample.py: 66 passed
- worker/.venv/bin/python eval/eval.py: 87/95 passed (91.6%; fuzzy gate 80%)
- Python compilation and git diff checks passed


yanziz-nvidia:
fix DCO

yanziz-nvidia (approved):

PR #501, opened 2026-09-08
Add on-demand client image capture

nvddr:
## Summary

- make the existing `get_current_frame` tool transparently request a client still when no fresh streamed frame is available
- route correlated, participant-targeted image requests through DeviceIOHub and return encoded images with LiveKit byte streams
- add capture support to the shared web/WebXR, Android, Apple, and native StreamKit clients
- use simple-vlm-example as the reference sample while preserving the same image reference for VLM and agentic workloads
- document a future analogous short-video fallback without adding a provisional video protocol

## Capture contract

`CurrentFrameTool` is intentionally a live-frame-or-still selector. Every invocation, including a periodic monitoring call, uses a fresh frame already observed by DeviceIOHub when possible and otherwise requests one client still. Installing `onImageCaptureRequested` is the client-side opt-in boundary. Applications that must consume streamed frames only can use `LiveFrameSource` directly without installing a capture handler.

The web, WebXR, Android, and Apple samples expose one persisted **Camera Mode** selector. **Off** is the first-launch default and enables neither path. **On-demand images** installs the still handler without publishing video. **Live video** publishes the selected source and leaves the still handler unset. The selection survives disconnect, reconnect, and app relaunch; connection lifecycle changes stop active resources without silently changing the user's authorization choice.

The existing Android and Apple frame-injection APIs remain publishing APIs. A glasses or other external-camera adapter forwards SDK frames to `injectVideoFrame` only in **Live video**, retains or requests a frame for `onImageCaptureRequested` in **On-demand images**, and stops or discards the feed in **Off**. This routing is entirely client-side and adds no StreamKit public API.

A missing or failing capture handler returns an immediate correlated rejection on `camera.capture.response`; the tool does not wait for the full capture timeout. Capture protocol types and endpoint methods remain private implementation details, and the public `xr_ai_hub` surface is unchanged.

## Validation

- `140 passed` across the focused capture, hub, image-tool, sample-worker, web-camera, and documentation tests
- `81 passed` across capture, IPC codec, dependency-manifest, and third-party-notice regressions
- Ruff passed repository-wide
- web JavaScript capture and camera-mode persistence tests passed
- Android `assembleDebug` and Netty-pin verification passed in CI
- all six native StreamKit stub-mode tests passed
- the Python and web dependency snapshots re-resolved unchanged at the repository cutoff

## Notes

- Android LiveKit is upgraded to 2.28.2, with the committed Android lock snapshot refreshed.
- Apple LiveKit is upgraded to 2.16.0, with resolved notice versions synchronized.
- Native LiveKit C++ requires v1.10.2 or newer and the official `include/` layout.
- Exact client SDK pins required by a feature may advance without moving the repository-wide resolver cutoff; the affected lock and notices must be refreshed.
- Client video-segment fallback is intentionally unimplemented.


yanziz-nvidia:
reviewed by yanziz-review-bot

**Summary**

This adds on-demand still capture across the hub and clients. Other than the existing findings in previous comments, I found one new blocker. Validation: 72 Python tests, seven JavaScript tests, and the native stub session test passed; device capture was not exercised.

| Severity | Finding |
|---|---|
| 🚫 BLOCKER | [Android `LiveKitBackend.kt:248`](https://github.com/NVIDIA/xr-ai/blob/85324c281ada6d71df906793798a7ee811135476/client-samples/android/app/src/main/java/com/nvidia/xrai/streamkitsample/streamkit/backends/livekit/LiveKitBackend.kt#L248) — **Still capture fails after turning video off.** `stopCamera()` calls `setCameraEnabled(false)`, which retains the publication but stops camera capture in [LiveKit 2.28.2](https://github.com/livekit/client-sdk-android/blob/v2.28.2/livekit-android-sdk/src/main/java/io/livekit/android/room/participant/LocalParticipant.kt#L440). Consequently, `localCameraTrack` remains non-null and this branch waits for a frame from the stopped track. Once the hub’s cached frame expires, visual queries time out. Reuse only an actively capturing track; otherwise use the temporary, unpublished capture path. |
| 💡 SUGGESTION | none |
| 🧹 NIT | none |

- Fix Android `captureImage()` to distinguish an active camera from a retained muted publication. Add a regression covering camera on → camera off → still request, verifying a JPEG returns while video stays unpublished or muted.

nvddr:
Disposition: fixed in c9828e5a after rebasing PR 501 onto latest main (4c221c34). LiveKit retains the Android camera publication after setCameraEnabled(false), so track presence alone did not mean capture was active. The localCameraTrack accessor now returns the retained track only when its publication is unmuted. After video is stopped, captureImage therefore uses the existing temporary, unpublished capture path and leaves the video publication muted. Scope is limited to LiveKitBackend.kt. Rebase verification: 72 focused Python tests passed, the web camera capture suite passed, all 6 native stub-mode tests passed, and git diff --check passed. The Android build could not run in this environment because no Android SDK or ANDROID_HOME is installed.

wenxind-nvidia (commented):
## Blockers

1. **Native clients accept capture commands from any room participant.**
   [StreamSession.cpp:128](https://github.com/NVIDIA/xr-ai/blob/c9828e5adae2e0118ea0a4f581df4eae3050a5fc/client-samples/native/StreamKit/src/StreamSession.cpp#L128) automatically invokes the capture callback, but the LiveKit delegate discards the sender identity without checking `hub_identity`. Previously, these messages were merely forwarded to application code; now a peer can trigger an opted-in client's camera. Responses remain hub-targeted—this is unauthorized capture, not demonstrated image disclosure. Validate capture-control senders against the configured hub before invoking the handler.

2. **Rejected image uploads leak LiveKit readers and buffered data.**
   [_room_client.py:319](https://github.com/NVIDIA/xr-ai/blob/c9828e5adae2e0118ea0a4f581df4eae3050a5fc/services/device-io-hub/device_io_hub/transport/livekit/_room_client.py#L319) returns on oversized or invalid uploads without closing the reader. LiveKit retains its event subscription and queues subsequent chunks, undermining the intended 8 MiB limit. I reproduced this with the qualified LiveKit version, including retention after the sender's end-of-stream event. Close the reader in `finally` and test rejection/cancellation paths. [LiveKit's reader contract](https://docs.livekit.io/reference/python/livekit/rtc/data_stream.html#ByteStreamReader)

3. **Android's first on-demand capture never requests camera permission.**
   [AppViewModel.kt:200](https://github.com/NVIDIA/xr-ai/blob/c9828e5adae2e0118ea0a4f581df4eae3050a5fc/client-samples/android/app/src/main/java/com/nvidia/xrai/streamkitsample/AppViewModel.kt#L200) calls `captureImage()` directly. On a fresh installation, connecting and enabling the microphone does not grant camera permission. LiveKit throws `SecurityException`; the session logs it, and the agent eventually times out without showing a permission prompt. The existing manual camera button handles permission correctly. Route still capture through an Activity-owned permission request without enabling video publication.

4. **Overlapping Android captures can invalidate the camera's callback iteration.**
   [CameraImageCapture.kt:111](https://github.com/NVIDIA/xr-ai/blob/c9828e5adae2e0118ea0a4f581df4eae3050a5fc/client-samples/android/app/src/main/java/com/nvidia/xrai/streamkitsample/streamkit/CameraImageCapture.kt#L111) removes its renderer synchronously inside the frame callback. LiveKit 2.28.2 iterates the underlying mutable sink set directly. With two pending captures on the same active track, the first removal causes `ConcurrentModificationException` before the second sink receives its frame. Defer removal until outside that iteration and add overlapping-capture coverage.

5. **The required native SDK cannot be discovered in its official layout.**
   [CMakeLists.txt:26](https://github.com/NVIDIA/xr-ai/blob/c9828e5adae2e0118ea0a4f581df4eae3050a5fc/client-samples/native/StreamKit/CMakeLists.txt#L26) upgrades the requirement to client-sdk-cpp v1.10.2, but discovery and include paths still require `inc/livekit/`. The official SDK uses `include/livekit/`. Explicit configuration therefore fails; automatic discovery instead falls back to the stub backend. Update discovery/include paths together with the upgrade. I reproduced the configuration failure using a layout-only fixture. [Upstream layout](https://github.com/livekit/client-sdk-cpp/blob/v1.10.2/docs/building.md#build-output)

Reviewed head: c9828e5adae2e0118ea0a4f581df4eae3050a5fc

nvddr:
Disposition for Wenxin blockers — fixed in 7f4553df:

1. Native capture authorization: LiveKitBackend now carries the UserDataPacketEvent sender identity into its private delivery boundary and drops non-status data from participants other than configured hub_identity. The public StreamingBackend API is unchanged. A native regression verifies peer capture control is rejected and hub capture control is delivered.
2. Reader retention: image byte-stream handling now closes ByteStreamReader in finally for validation rejection, oversize aborts, normal completion, exceptions, and cancellation. Rejection and cancellation regressions were added.
3. Android permission: physical still capture now awaits a shared Activity-owned camera permission launcher. Granting permission continues directly to captureImage and does not call startCamera or publish video.
4. Android overlap: renderer removal and continuation completion are posted outside LiveKit capture callback iteration, with removal occurring before a temporary track can be disposed. This lets overlapping sinks receive the same frame without mutating LiveKit mutable sink set during iteration.
5. Native SDK layout: discovery, validation, diagnostics, and include paths now consistently use the official include/livekit layout.

Validation: 74 focused Python tests passed; 7 web capture tests passed; all 6 native stub-mode tests passed; Ruff and git diff --check passed; the native backend compiled against the pinned v1.10.2 upstream headers; and an official-layout CMake fixture configured successfully. Android compilation/device capture could not be exercised because this environment has no Android SDK.

blongs-nv:
Reviewed at `7f4553df709d09097e74d16929b8476881df1473` by a bot.

## Blockers

1. **Every existing `CurrentFrameTool` user now activates the participant's camera, including background pollers.**

   [`agent-sdk/xr-ai-tools/xr_ai_tools/current_frame.py:82`](https://github.com/NVIDIA/xr-ai/blob/7f4553df709d09097e74d16929b8476881df1473/agent-sdk/xr-ai-tools/xr_ai_tools/current_frame.py#L82) The capture source is constructed unconditionally and the fallback at [line 120](https://github.com/NVIDIA/xr-ai/blob/7f4553df709d09097e74d16929b8476881df1473/agent-sdk/xr-ai-tools/xr_ai_tools/current_frame.py#L120) has no opt-out. Tea-making's change watch polls every 2 s and lab-instrument's monitor every 5 s; with video off each tick sends a capture request that opens and closes the web camera. Make capture opt-in per tool instance and say in the model-facing description at [line 89](https://github.com/NVIDIA/xr-ai/blob/7f4553df709d09097e74d16929b8476881df1473/agent-sdk/xr-ai-tools/xr_ai_tools/current_frame.py#L89) that the tool may activate the camera.

2. **The protocol has no negative response.**

   [`client-samples/web/StreamKit/StreamSession.js:286`](https://github.com/NVIDIA/xr-ai/blob/7f4553df709d09097e74d16929b8476881df1473/client-samples/web/StreamKit/StreamSession.js#L286) A client with no handler returns silently, and denied permissions and handler errors look the same, so every turn on such a client stalls the full `frame_timeout_s` before reporting a timeout. Add a reject payload on the response topic and fail the pending future immediately.

3. **Hostile image bytes abort the agent turn.**

   [`agent-sdk/xr-ai-tools/xr_ai_tools/current_frame.py:126`](https://github.com/NVIDIA/xr-ai/blob/7f4553df709d09097e74d16929b8476881df1473/agent-sdk/xr-ai-tools/xr_ai_tools/current_frame.py#L126) The handler catches `OSError` and `ValueError`, but Pillow raises `DecompressionBombError` for a 67-byte PNG header claiming 14000x14000 and `SyntaxError` from `verify()` on a bad PNG CRC; neither becomes `FrameUnavailable`. Catch `Exception` around the PIL call.

4. **Two client SDK bumps postdate the dependency cutoff.**

   [`dependency-manifest/android/gradle.lockfile:148`](https://github.com/NVIDIA/xr-ai/blob/7f4553df709d09097e74d16929b8476881df1473/dependency-manifest/android/gradle.lockfile#L148) client-sdk-android 2.28.2 (published 2026-09-07) and client-sdk-cpp 1.10.2 (2026-09-08) are newer than the 2026-08-27 cutoff in `uv.toml`, and the lockfile still records 2.7.0. [`THIRD_PARTY_NOTICES.md:143`](https://github.com/NVIDIA/xr-ai/blob/7f4553df709d09097e74d16929b8476881df1473/THIRD_PARTY_NOTICES.md#L143) lists the Swift SDK at 2.13.0 while [`Package.swift:20`](https://github.com/NVIDIA/xr-ai/blob/7f4553df709d09097e74d16929b8476881df1473/client-samples/ios-visionos/StreamKit/Package.swift#L20) pins 2.16.0. Either move the cutoff and refresh locks and notices in this PR, or hold the bumps.

5. **Public API growth with no recorded approval.**

   [`agent-sdk/xr-ai-hub/xr_ai_hub/__init__.py:51`](https://github.com/NVIDIA/xr-ai/blob/7f4553df709d09097e74d16929b8476881df1473/agent-sdk/xr-ai-hub/xr_ai_hub/__init__.py#L51) Five new exports, new endpoint methods, and three wire types join the public surface; only `ClientImageCaptureSource` is needed outside the hub, and its error is already re-raised as `FrameUnavailable`.

## Suggestions

1. [`_room_client.py:330`](https://github.com/NVIDIA/xr-ai/blob/7f4553df709d09097e74d16929b8476881df1473/services/device-io-hub/device_io_hub/transport/livekit/_room_client.py#L330) A header with no trailer parks the receive task, and up to 8 MiB, for the room's lifetime, and nothing limits how many such streams one participant can open. Wrap the receive in a deadline and cap concurrent streams per participant.

2. `frame_timeout_s` now also bounds cold camera start, exposure settle, encode and upload, and the web handler ignores `timeoutMs` ([`core.js:451`](https://github.com/NVIDIA/xr-ai/blob/7f4553df709d09097e74d16929b8476881df1473/client-samples/web/App/core.js#L451)). Add a separate capture timeout enforced locally in `StreamSession.js`.

3. Delivery needs a `Subscribe.DATA` subscription, but [`_capture.py:44`](https://github.com/NVIDIA/xr-ai/blob/7f4553df709d09097e74d16929b8476881df1473/agent-sdk/xr-ai-hub/xr_ai_hub/_capture.py#L44) only checks the roster, so a video-only endpoint times out silently on every call. Any agent can also forge the `camera.capture.request` topic through plain return data at [`_hub.py:520`](https://github.com/NVIDIA/xr-ai/blob/7f4553df709d09097e74d16929b8476881df1473/services/device-io-hub/device_io_hub/ipc/_hub.py#L520); reserve the namespace.

4. The failure text at [`_capture.py:66`](https://github.com/NVIDIA/xr-ai/blob/7f4553df709d09097e74d16929b8476881df1473/agent-sdk/xr-ai-hub/xr_ai_hub/_capture.py#L66) is spoken to users through TTS in the tea-making, lab-instrument, and simple-vlm samples. Rephrase it for end users and set the processing status before the capture starts.

5. Docs to bring along: a `STREAMKIT_PARITY.md` row, the new types and topics in `server-runtime.md`, a migrations entry for the client SDK floors, and [`clients.md:46`](https://github.com/NVIDIA/xr-ai/blob/7f4553df709d09097e74d16929b8476881df1473/docs/source/getting_started/clients.md#L46), which says streams are not broadcast to peers when that holds only with `hubIdentity` set. The `clientControl` swallow at [`core.js:496`](https://github.com/NVIDIA/xr-ai/blob/7f4553df709d09097e74d16929b8476881df1473/client-samples/web/App/core.js#L496) is now dead code.

6. Untested: the `_receive_image_capture` size and mime limits on the accept path, the hub cancel and drop branches, `StreamSession` request validation, and the sample capture handler.

7. [`current_frame.py:142`](https://github.com/NVIDIA/xr-ai/blob/7f4553df709d09097e74d16929b8476881df1473/agent-sdk/xr-ai-tools/xr_ai_tools/current_frame.py#L142) `image.verify()` is a no-op for JPEG and WebP, so the PR body overstates the validation, and the `ImageFrame` docstrings for the sequence and timestamp fields do not describe stills.

The per-participant request correlation, the same-identity rejoin handling, and the four client implementations are careful work, and the reader now closes on every exit path. The blockers above should be resolved before merge.

nvddr:
Disposition for Brent's blockers — addressed in 537ee28a:

1. Current-frame capture semantics: the transparent live-frame-or-still fallback is the feature contract, so I did not add a second per-tool opt-in that would make existing `CurrentFrameTool` call sites behave differently. The client capture handler remains the explicit opt-in boundary. The model-facing tool description, SDK reference, client guide, parity table, and PR description now state that every invocation—including background polling—may briefly activate an opted-in client's camera when video is off, and streamed-only applications are directed to `LiveFrameSource`.
2. Negative response: all four StreamKit implementations now return a correlated rejection on `camera.capture.response` when no handler is installed, the handler fails, or it returns invalid image data. DeviceIOHub routes that response to the pending request, which raises immediately instead of waiting for `frame_timeout_s`. Rejection routing is covered in Python, web, and native regressions.
3. Hostile image bytes: Pillow validation is now an exception boundary. `DecompressionBombError`, `SyntaxError`, and other decoder/plugin exceptions become `FrameUnavailable`; a decompression-bomb regression covers the path.
4. Dependency records: regenerated the Android lock snapshot at 2.28.2, synchronized Swift direct/transitive notice versions with `Package.resolved`, documented all three client SDK floors, and clarified the repository rule for targeted exact client SDK pins. The repository-wide cutoff was intentionally not advanced because that would refresh unrelated Python and web dependencies; both of those snapshots re-resolved unchanged.
5. Public API surface: removed all capture exports from `xr_ai_hub` and `device_io_hub.ipc`, made capture source/errors and processor/connector operations private, and kept the wire records internal. No new Python public API remains.

Validation: 108 focused Python tests passed; 81 capture/codec/dependency tests passed; 42 documentation-contract tests passed; the web capture suite passed; all six native StreamKit tests passed; Ruff and `git diff --check` passed; Android lock regeneration passed. Local Android compilation remains unavailable because this environment has no Android SDK.

blongs-nv:
Reviewed at `537ee28abac58969050cd47acd052e12c9392a4d` by a bot.

Blockers 2, 3, 4, and 5 from my previous review are resolved: every client now returns an immediate rejection, the tool converts any Pillow error to `FrameUnavailable`, the Android lockfile and Swift notices are refreshed under the amended pin policy, and the capture surface is private.

## Blockers

1. **Background pollers still cycle the camera on every shipped sample client.**

   [`agent-sdk/xr-ai-tools/xr_ai_tools/current_frame.py:83`](https://github.com/NVIDIA/xr-ai/blob/537ee28abac58969050cd47acd052e12c9392a4d/agent-sdk/xr-ai-tools/xr_ai_tools/current_frame.py#L83) Keeping the transparent fallback and putting the opt-in on handler installation is a fair design call, but the web ([`core.js:451`](https://github.com/NVIDIA/xr-ai/blob/537ee28abac58969050cd47acd052e12c9392a4d/client-samples/web/App/core.js#L451)), Android ([`AppViewModel.kt:195`](https://github.com/NVIDIA/xr-ai/blob/537ee28abac58969050cd47acd052e12c9392a4d/client-samples/android/app/src/main/java/com/nvidia/xrai/streamkitsample/AppViewModel.kt#L195)), and Apple ([`AppModel.swift:280`](https://github.com/NVIDIA/xr-ai/blob/537ee28abac58969050cd47acd052e12c9392a4d/client-samples/ios-visionos/App/AppModel.swift#L280)) samples all install it unconditionally, so tea-making's change watch ([`change_watch.py:291`](https://github.com/NVIDIA/xr-ai/blob/537ee28abac58969050cd47acd052e12c9392a4d/agent-samples/tea-making-sample/worker/tea_making_worker/change_watch.py#L291)) and lab-instrument's monitor still open and close the camera every 2 to 5 s whenever video is off. Either move the pollers to `LiveFrameSource`, as the new docs advise for callers that must not initiate capture, or gate the sample handlers on a user setting.

## Suggestions

1. [`_room_client.py:334`](https://github.com/NVIDIA/xr-ai/blob/537ee28abac58969050cd47acd052e12c9392a4d/services/device-io-hub/device_io_hub/transport/livekit/_room_client.py#L334) A header with no trailer still parks the receive task, and up to 8 MiB, for the room's lifetime, with no cap on concurrent streams per participant. Add a deadline and a per-participant cap.

2. `frame_timeout_s` still bounds cold camera start, exposure settle, encode and upload together, and the web handler ignores `timeoutMs`. Add a separate capture timeout enforced locally in `StreamSession.js`.

3. [`_capture.py:47`](https://github.com/NVIDIA/xr-ai/blob/537ee28abac58969050cd47acd052e12c9392a4d/agent-sdk/xr-ai-hub/xr_ai_hub/_capture.py#L47) still checks only the roster, so a video-only endpoint times out silently on every call, and the `camera.capture.request` topic can still be forged through plain return data. Check the data subscription and reserve the namespace.

4. Both failure strings at [`_capture.py:69`](https://github.com/NVIDIA/xr-ai/blob/537ee28abac58969050cd47acd052e12c9392a4d/agent-sdk/xr-ai-hub/xr_ai_hub/_capture.py#L69) and [`line 99`](https://github.com/NVIDIA/xr-ai/blob/537ee28abac58969050cd47acd052e12c9392a4d/agent-sdk/xr-ai-hub/xr_ai_hub/_capture.py#L99) are spoken through TTS in three samples. Rephrase for end users and set the processing status before the capture starts.

5. Still open from the docs list: the new types and topics in `server-runtime.md`, [`clients.md:48`](https://github.com/NVIDIA/xr-ai/blob/537ee28abac58969050cd47acd052e12c9392a4d/docs/source/getting_started/clients.md#L48) claiming streams are not broadcast to peers when that holds only with `hubIdentity` set, and the dead `clientControl` branch at [`core.js:496`](https://github.com/NVIDIA/xr-ai/blob/537ee28abac58969050cd47acd052e12c9392a4d/client-samples/web/App/core.js#L496).

6. [`current_frame.py:148`](https://github.com/NVIDIA/xr-ai/blob/537ee28abac58969050cd47acd052e12c9392a4d/agent-sdk/xr-ai-tools/xr_ai_tools/current_frame.py#L148) `image.verify()` is a no-op for JPEG and WebP, and the `ImageFrame` docstrings for the sequence and timestamp fields do not describe stills.

The rejection protocol is a clean addition, mirrored on all four platforms with tests on each, and the hub accepts it through the same size-checked path. The blocker above should be resolved before merge.

yanziz-nvidia (approved):

nvddr:
Disposition for Brent's repeated camera-activation blocker — fixed in `1ea08100` with the Android visibility follow-up in `6d50df2e`.

The blocker was protecting the client-side authorization boundary: a background `CurrentFrameTool` caller must not be able to activate a sample client's camera merely because the client connected. The graphical samples previously installed `onImageCaptureRequested` unconditionally, so they had opted every connected user into that behavior.

The web, WebXR, Android, and Apple samples now use one mutually exclusive, persisted **Camera Mode** selector:

- **Off** (first-launch default): no still handler and no live publication.
- **On-demand images**: installs the still handler without publishing video.
- **Live video**: publishes the selected source and leaves the still handler unset.

Disconnects and reconnects stop active resources but do not reset the selected authorization mode; the choice also survives app/page relaunch. Mode transitions re-check the current choice after asynchronous camera start so a late permission or startup completion cannot leave video active after the user selects another mode.

Frame injection follows the same boundary without changing StreamKit's public API. `injectVideoFrame` remains a publishing operation and is used only in **Live video**. A glasses/external-camera adapter keeps or requests its latest SDK frame and returns encoded bytes from `onImageCaptureRequested` in **On-demand images**, and stops or discards frames in **Off**. The Android virtual camera exercises both the publishing and unpublished-still paths; the Apple and Android injection docs now state the routing contract explicitly.

Validation: 140 focused Python/JavaScript/documentation tests passed, including a regression for mode exclusivity and persistence; JavaScript syntax and `git diff --check` passed; and the Android `assembleDebug` CI job passed at the final head.

blongs-nv (approved):

wenxind-nvidia (commented):
## Blockers

1. **[P1] Selecting Camera Mode “Off” does not cancel an in-flight still capture.**
   [`core.js:768`](https://github.com/NVIDIA/xr-ai/blob/6d50df2efe271eb07d8dd231e81145f198bfc54f/client-samples/web/App/core.js#L768) only removes the handler for future requests. An existing request retains its handler and uncancelled operation.

   Reproduced using the actual web core and StreamSession: start capture → select **Off** while camera acquisition is pending → acquisition completes → **a JPEG is captured and uploaded while the mode remains Off**. WebXR shares this path.

   The same missing cancellation exists in [Android](https://github.com/NVIDIA/xr-ai/blob/6d50df2efe271eb07d8dd231e81145f198bfc54f/client-samples/android/app/src/main/java/com/nvidia/xrai/streamkitsample/AppViewModel.kt#L330) and [Apple](https://github.com/NVIDIA/xr-ai/blob/6d50df2efe271eb07d8dd231e81145f198bfc54f/client-samples/ios-visionos/App/AppModel.swift#L418). Cancel pending captures when leaving on-demand mode, recheck authorization after asynchronous permission/acquisition, and release acquired resources. Add an Off-during-capture regression.

Reviewed head: 6d50df2efe271eb07d8dd231e81145f198bfc54f

nvddr:
Disposition for the PR527 restack:

- Rebased PR501 onto the latest PR527 head, 02799e36. PR501 is now at 5a6efdf8.
- Reused the private byte-stream transport added by PR527 on Android, iOS/visionOS, native, and web. The capture response keeps its topic, request correlation, MIME type, name, size, and hub-only destination, while the actual send is delegated to the shared LiveKit byte writer.
- Removed the sendImage / SendImage methods from the public StreamingBackend surfaces and removed the public native StreamSession image-send method. No public byte-stream API was added. Web uses a symbol-keyed internal channel; the other clients use LiveKit-only internal/private methods.
- Reused the PR527 bounded byte-stream reader in DeviceIOHub with the existing 8 MiB capture limit plus idle and total timeouts.
- Resolved the newer-main filename overlap without combining unrelated concepts: the existing private media-capture topic module remains in _capture.py, and the on-demand image coordinator remains private in _image_capture.py.
- Regenerated the one dependency-map line required by the final PR527 manifest-sync commit.

Validation:
- 163 focused Python tests passed.
- 13 focused web and byte-stream tests passed.
- Native stub build and all 7 tests passed.
- Native production code built against LiveKit C++ 1.10.2; the byte-stream test passed with the real SDK.
- Ruff and the generated dependency-map check passed locally.
- Current CI already has DCO, Ruff, SPDX, generated dependency map, Android lockfiles, and web lockfile checks green; remaining jobs are still running.

Android and Swift compilation could not be run locally because this environment has no Android SDK or Swift toolchain, so those are left to the active PR checks.

nvddr:
CI update: all current PR501 checks are now green, including Android assembleDebug, Python 3.11 and 3.12 tests, docs build/publish, CodeQL, dependency and lockfile checks, Ruff, SPDX, and DCO.

blongs-nv:
Reviewed at `5a6efdf812ec16b7448057fe8b3e05dc9f079436` by a bot.

The blocker from my previous review is resolved. The web, WebXR, Android, and Apple samples now default to an Off camera mode ([`core.js:154`](https://github.com/NVIDIA/xr-ai/blob/5a6efdf812ec16b7448057fe8b3e05dc9f079436/client-samples/web/App/core.js#L154), [`AppViewModel.kt:110`](https://github.com/NVIDIA/xr-ai/blob/5a6efdf812ec16b7448057fe8b3e05dc9f079436/client-samples/android/app/src/main/java/com/nvidia/xrai/streamkitsample/AppViewModel.kt#L110), [`AppModel.swift:142`](https://github.com/NVIDIA/xr-ai/blob/5a6efdf812ec16b7448057fe8b3e05dc9f079436/client-samples/ios-visionos/App/AppModel.swift#L142)), install the capture handler only in On-demand mode, and re-check the mode after every asynchronous camera start, so a background poller can no longer wake a camera the user has not authorized. The shared byte-stream reader also gives the hub idle and total deadlines on capture uploads.

## Suggestions

1. The Live video status reads "Starting…" ([`core.js:294`](https://github.com/NVIDIA/xr-ai/blob/5a6efdf812ec16b7448057fe8b3e05dc9f079436/client-samples/web/App/core.js#L294), [`MainActivity.kt:623`](https://github.com/NVIDIA/xr-ai/blob/5a6efdf812ec16b7448057fe8b3e05dc9f079436/client-samples/android/app/src/main/java/com/nvidia/xrai/streamkitsample/MainActivity.kt#L623), [`ContentView.swift:268`](https://github.com/NVIDIA/xr-ai/blob/5a6efdf812ec16b7448057fe8b3e05dc9f079436/client-samples/ios-visionos/App/ContentView.swift#L268)) and stays there forever after a denied permission or a failed start. Show a failed state, or fall back to Off, when `startCamera` does not end with the camera active.

2. [`StreamSessionTests.cpp:192`](https://github.com/NVIDIA/xr-ai/blob/5a6efdf812ec16b7448057fe8b3e05dc9f079436/client-samples/native/StreamKit/Tests/StreamKitTests/StreamSessionTests.cpp#L192) The native rejection test no longer asserts that a rejection is sent, because the mock backend cannot observe the LiveKit-only send; it now passes whether or not the payload goes out. Add a test seam so the rejection topic, mime type, and body are asserted again.

The mode selector is a clean resolution, and routing the capture response through the shared byte-stream writer removes a public backend method on every platform while keeping the topic, correlation, and hub-only destination intact. Nothing above blocks; the other suggestions from my previous review still stand as written there.

nvddr:
Disposition for Brent final suggestions — addressed in e5214ca8:

1. Failed live-camera starts no longer leave the sample UI at Starting indefinitely. Web/WebXR, Android, and Apple now fall back to persisted Off only when startup fails while the client is still connected and still in Live mode. Disconnect and reconnect continue to preserve the user authorization choice. The web regression covers failure, inactive state, and persisted Off.
2. Native rejection coverage is restored through a private StreamSession capture-response sender seam. The regression now asserts the camera.capture.response topic, request correlation, rejection MIME type, and exact JSON body. StreamingBackend and the public byte-stream API remain unchanged.

Validation: the focused web camera and byte-stream suites pass; all seven native stub tests pass; the native session and byte-stream tests compile and pass against LiveKit C++ 1.10.2; git diff --check passes. Android and Apple compilation are left to CI.

nvddr:
Updated disposition after restacking PR 501 onto the merged PR 527 byte-stream transport (ad6a2ea5):

- Camera startup failures now leave the UI in a truthful persisted **Off** state on web, Android, and Apple, but only when the client is still connected and still in Live mode. Disconnect/reconnect does not reset the user authorization choice.
- Restored native capture-response coverage through a private StreamSession test seam. The tests now assert the response topic, correlation/request ID, MIME type, and rejection body without adding anything to the public StreamingBackend API.
- Reused PR 527 private byte-stream transport; no cross-API-boundary or unrelated changes.

Validation:
- web byte-stream + camera-capture tests: pass
- native stub suite: 7/7 pass
- native real-LiveKit session + byte-stream tests: 2/2 pass
- generated dependency map and git diff check: clean

New head: f3ce829b

nvddr:
Disposition for Wenxin’s later P1 blocker (Off during an in-flight still capture) — fixed in 248a0487.

- Clearing the existing onImageCaptureRequested handler now cancels every in-flight capture on web, Android, and Apple. Because camera-mode transitions already clear that handler when leaving On-demand images, selecting Off takes effect immediately without adding or broadening a public API.
- A locally revoked request returns the existing correlated capture-rejection response, so the hub does not wait for the capture timeout.
- The sample handlers re-check On-demand authorization after asynchronous permission/acquisition. Existing cancellation/finally paths release temporary web tracks, Android tracks/renderers, and Apple tracks.
- Added an Off-during-capture regression using the real web StreamSession and camera-mode transition. It verifies abort, cleanup, idle UI state, correlated rejection, and no image response.

Validation: web byte-stream and camera-capture suites pass; git diff check is clean. Local Android compilation cannot run without an installed Android SDK, so the pushed CI run is the platform compile check.

wenxind-nvidia (commented):
## Blockers

1. **Selecting Off can still transmit a captured image.**
   [StreamSession.js:315](https://github.com/NVIDIA/xr-ai/blob/248a048779b9447ca71278624de160ecc28887dd/client-samples/web/StreamKit/StreamSession.js#L315) checks cancellation before sending, but does not propagate it through the asynchronous byte-stream upload. I reproduced: image ready → stream opening pending → select Off → rejection sent → stream opens and JPEG bytes are written anyway. Propagate cancellation through the private send path and check it before subsequent writes. Add a regression test that revokes capture while stream opening is pending.

2. **Off does not promptly release an acquired web camera.**
   [core.js:510](https://github.com/NVIDIA/xr-ai/blob/248a048779b9447ca71278624de160ecc28887dd/client-samples/web/App/core.js#L510) releases the temporary camera only after capture unwinds. However, exposure configuration and [video playback startup](https://github.com/NVIDIA/xr-ai/blob/248a048779b9447ca71278624de160ecc28887dd/client-samples/web/App/core.js#L636) can remain pending without observing cancellation. I reproduced Off sending rejection while the acquired track remained running until `video.play()` resolved. Register cancellation cleanup immediately after acquiring the owned stream and make playback startup abortable. The latest fix covers cancellation during camera acquisition, but not these later stages.

3. **A capture rejection can hide an available live frame.**
   [current_frame.py:128](https://github.com/NVIDIA/xr-ai/blob/248a048779b9447ca71278624de160ecc28887dd/agent-sdk/xr-ai-tools/xr_ai_tools/current_frame.py#L128) immediately raises on capture failure without reconsidering streaming frames. With an initially stale frame, a fresh frame arriving after 10 ms, and a no-handler rejection arriving after 30 ms, the base returns the frame; this head reports that the client could not provide a picture. This matters because Live mode intentionally disables the capture handler. Recheck for an already-fresh streaming frame before propagating the rejection; no additional waiting is necessary.

4. **Rapid Live → Off → Live transitions can leave mobile video stopped.**
   Both [Android's mode switch](https://github.com/NVIDIA/xr-ai/blob/248a048779b9447ca71278624de160ecc28887dd/client-samples/android/app/src/main/java/com/nvidia/xrai/streamkitsample/AppViewModel.kt#L346) and [Apple's equivalent](https://github.com/NVIDIA/xr-ai/blob/248a048779b9447ca71278624de160ecc28887dd/client-samples/ios-visionos/App/AppModel.swift#L427) allow overlapping transitions. Off starts asynchronous shutdown; selecting Live before shutdown finishes calls `startCamera()`, which returns because `isCameraActive` is still true. Shutdown then clears it, leaving Live selected without restarting video. The new mode selector makes this ordering reachable. Serialize transitions or reconcile the latest desired mode after shutdown, with a delayed-stop regression test.

Reviewed head: 248a048779b9447ca71278624de160ecc28887dd

nvddr:
Disposition for Wenxin’s review at 248a0487 — revised in c7674455 after a holistic lifecycle review.

The earlier disposition treated selecting Off as retroactive cancellation. That was stronger than the intended product contract and created an unreliable coupling between a UI preference and camera/transport operations already in progress. The partial cancellation behavior from 248a0487 has therefore been removed.

The explicit contract is now:
- On-demand images authorizes acceptance of new capture requests.
- A request snapshots that authorization when StreamSession accepts it. If Off is selected afterward, that accepted request finishes and releases its resources.
- Off rejects subsequent requests. Disconnect and camera.capture.cancel retain their existing pending-work cancellation behavior.

This ordering is documented in clients.md and STREAMKIT_PARITY.md and covered by a regression that proves an accepted request completes after Off while the next request is rejected. Therefore findings 1 and 2 are not implemented as Off-time revocation semantics.

The two independent correctness findings were addressed:
- CurrentFrameTool now performs one non-waiting fresh-live-frame recheck after a capture rejection, so a frame that arrived while the rejection was in flight wins.
- Camera mode transitions now reconcile the latest desired mode after delayed shutdown. Web serializes mode start/stop transitions privately; Android and Apple restart Live after shutdown when Live became the latest selection. A delayed-stop regression covers Live → Off → Live.

No public API or protocol changes were added.

Validation: 39 focused capture/tool tests, 50 documentation tests, both web capture/byte-stream suites, Ruff, generated dependency map, and git diff check pass. Android CI is pending on the pushed head.

nvddr:
Rebased PR 501 onto the latest main (5a091644) and force-pushed with an exact lease.

- New head: 0219cb98
- Rebase completed without conflicts.
- Range-diff confirms all 11 PR commits are patch-identical; only ancestry changed.

Post-rebase validation:
- 39 focused capture/tool tests pass
- 50 documentation tests pass
- web byte-stream and camera-capture suites pass
- native stub suite: 7/7 pass
- native real-LiveKit session and byte-stream tests: 2/2 pass
- generated dependency map and git diff check are clean

nvddr:
Rebased again onto current main (6de5bf20) and force-pushed with an exact lease.

- New head: e9474894
- No conflicts; all 11 PR commits are patch-identical by range-diff.
- Focused capture tests: 39 passed; web capture/byte-stream suites pass; generated dependency map and diff check are clean.

PR #506, opened 2026-09-09
docs: support withdrawn release tags

yanziz-nvidia:
## Summary

- exclude withdrawn documentation tags supplied through `XR_AI_DOCS_WITHDRAWN_TAGS`
- keep latest-release selection and sphinx-multiversion navigation consistent
- add focused regression coverage for selection, regex filtering, and workflow wiring

Cherry-pick of release PR #505: https://github.com/NVIDIA/xr-ai/pull/505

Release-branch commit cherry-picked:
- https://github.com/NVIDIA/xr-ai/commit/d487f8564c5d216d7ef8a2e90d1b7e0979847ddd

## Test plan

- `uvx --from pytest pytest --noconftest -q tests/test_docs_versions.py` (18 passed)
- `uvx ruff==0.15.16 check .github/scripts/docs_release_policy.py .github/scripts/select_latest_docs_release.py docs/source/conf.py tests/test_docs_versions.py`
- `git diff --check origin/main...HEAD`

### DIFF
--- a/.github/scripts/docs_release_policy.py
+++ b/.github/scripts/docs_release_policy.py
@@ -1,0 +1,23 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Shared policy for release tags published in the documentation site."""
+from __future__ import annotations
+
+import os
+import re
+
+_WITHDRAWN_TAGS_ENV = "XR_AI_DOCS_WITHDRAWN_TAGS"
+
+
+def withdrawn_tags() -> frozenset[str]:
+    """Return release tags excluded from documentation publication."""
+    value = os.environ.get(_WITHDRAWN_TAGS_ENV, "")
+    return frozenset(filter(None, re.split(r"[\s,]+", value)))
+
+
+def tag_whitelist(semver_pattern: str) -> str:
+    """Return an anchored semantic-version pattern excluding withdrawn tags."""
+    withdrawn = "|".join(re.escape(tag) for tag in sorted(withdrawn_tags()))
+    exclusion = rf"(?!(?:{withdrawn})$)" if withdrawn else ""
+    return rf"^{exclusion}{semver_pattern}$"
--- a/.github/scripts/select_latest_docs_release.py
+++ b/.github/scripts/select_latest_docs_release.py
@@ -7,6 +7,11 @@ from __future__ import annotations
 import re
 import sys
 from dataclasses import dataclass
+from pathlib import Path
+from runpy import run_path
+
+_POLICY = run_path(str(Path(__file__).with_name("docs_release_policy.py")))
+_withdrawn_tags = _POLICY["withdrawn_tags"]
 
 _SEMVER = re.compile(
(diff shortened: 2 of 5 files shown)


blongs-nv:
Reviewed at `951c3127f0c2ca75d1eed302e713cabbd07bd98a` by a bot.

## Suggestions

1. [`docs.yaml:28`](https://github.com/NVIDIA/xr-ai/blob/951c3127f0c2ca75d1eed302e713cabbd07bd98a/.github/workflows/docs.yaml#L28-L30) lists the repository settings the workflow depends on, and the new `XR_AI_DOCS_WITHDRAWN_TAGS` variable is not among them; nothing else in the repo names it. One sentence there giving the format, that an unset variable withdraws nothing, and that clearing it republishes the tag on the next docs push would make the dependency findable.

2. The variable exists only inside Actions, so a local `make versions` renders and lists the withdrawn tag, and [`SKILL.md:43`](https://github.com/NVIDIA/xr-ai/blob/951c3127f0c2ca75d1eed302e713cabbd07bd98a/skills/getting-started/SKILL.md#L43-L46) tells the agent to run the selector in a fresh clone "matching what the docs site serves", which now holds only while the withdrawn tag is not the highest one. A note in the skill and the Makefile header, or a checked-in default the selector falls back to, keeps the local answer honest.

3. Changing the variable rebuilds nothing, since [`docs.yaml:36`](https://github.com/NVIDIA/xr-ai/blob/951c3127f0c2ca75d1eed302e713cabbd07bd98a/.github/workflows/docs.yaml#L36-L39) has only path-filtered push and pull_request triggers, so withdrawing a tag is set-the-variable-then-find-something-to-push until a `workflow_dispatch` trigger is added.

4. [`test_docs_versions.py:100`](https://github.com/NVIDIA/xr-ai/blob/951c3127f0c2ca75d1eed302e713cabbd07bd98a/tests/test_docs_versions.py#L100-L110) replaced the old whitelist test and always sets two withdrawn tags, so the empty-variable path every ordinary deploy takes has no whitelist coverage; add an assertion that an unset variable yields the original pattern, and a selector case where every tag is withdrawn returns nothing (the Makefile's fall back to `main` depends on it).

5. [`test_docs_versions.py:113`](https://github.com/NVIDIA/xr-ai/blob/951c3127f0c2ca75d1eed302e713cabbd07bd98a/tests/test_docs_versions.py#L113-L120) matches the workflow as text, so it still passes if the `env` block moves under the publish job where the build never sees it; `pyyaml` is already a test dependency, so parsing the file and asserting the key under the top-level `env` and the path in both `paths` lists pins what matters.

6. A repository variable needs the same port to every buildable branch that a checked-in constant would (a tag pushed from a branch without this change rebuilds the site with the withdrawn tag back in), while being invisible to git history and local builds. Since #505 already chose this shape, this is a direction to converge on rather than a change for this PR.

The port is byte-identical to what merged on the release branch, the lookahead excludes exactly the withdrawn tag while admitting its dot-suffixed successor, and the env reaches the selector and every sphinx-multiversion build. No blockers, ready to merge.

nvddr (approved):

blongs-nv (approved):

wenxind-nvidia (approved):

PR #507, opened 2026-09-09
Replace Piper with Pocket TTS

nvddr:
## Summary

- remove the Piper service, preset, configuration, and compatibility surface
- add a Pocket TTS 3.0.2 service that preserves the existing OpenAI-compatible speech endpoint and model-stack integration
- switch checked-in profiles to the CC0 `bill_boerst` voice
- record the Pocket package, CC-BY-4.0 model, voice provenance, reviewed model revisions, and exact acceptable-use terms in third-party notices
- update samples, documentation, generated dependency metadata, and tests

## Compatibility

This intentionally provides no Piper compatibility alias. Existing configurations must migrate from `piper_tts`, `piper_tts_server`, and `services/piper-tts/` to their Pocket TTS equivalents.

## Validation

- `185 passed` in focused service/model/launcher tests
- `1537 passed, 10 deselected` in the full non-GPU test suite
- generated dependency map and dependency manifest are current
- changed-file SPDX and diff checks pass


wenxind-nvidia (commented):
## Nits

- `services/pocket-tts/pyproject.toml:13`: **Avoid unused CUDA dependencies for the CPU-only service.** Pocket introduces a transitive PyTorch dependency without a CPU-wheel source. On Linux this resolves CUDA-enabled PyPI Torch and its CUDA/cuDNN/NCCL/Triton dependencies, although this wrapper loads the model on CPU. Piper did not need that dependency branch. Select a service-scoped, platform-appropriate CPU Torch source to avoid unnecessary downloads and disk usage.

- `tests/test_pocket_tts.py:237`: **Preserve the backend-independent lifecycle regression tests.** Deleting `test_piper_tts.py` also removes the real launcher-abort/process-group test and coverage for unverified group ownership, reuse-monitor recovery, and HTTP-startup timeout cancellation. The replacement exec test mocks `_ensure_owned_process_group` and `execvpe`, so it cannot catch those regressions. Port the applicable existing tests to the Pocket names instead of removing them; the launcher-abort test does not need model weights.

- `dependency-manifest/pyproject.toml:40`: **Keep the qualification snapshot out of this replacement.** This file and `dependency-manifest/uv.lock` are regenerated even though `uv.toml` is unchanged. The base contract defines them as the last qualification snapshot, not the live project inventory, with updates deferred to the next cutoff refresh. Drop these two file changes; keep the regenerated live inventory in `DEPENDENCIES.md`.

Reviewed head: b5d9cb04ac1a40abc5d81bc07f85a6452448c870

blongs-nv:
Reviewed at `b5d9cb04ac1a40abc5d81bc07f85a6452448c870` by a bot.

## Blockers

1. **Abandoned synthesis requests keep running under the one generation lock, so barge-in and long replies regress.**

   [`__main__.py:127`](https://github.com/NVIDIA/xr-ai/blob/b5d9cb04ac1a40abc5d81bc07f85a6452448c870/services/pocket-tts/pocket_tts_server/__main__.py#L127-L128) serializes every request behind a thread lock on a model that [pins torch to one thread](https://github.com/kyutai-labs/pocket-tts/blob/main/pocket_tts/models/tts_model.py) and runs non-streaming `generate_audio`, while [`streaming_tts.py:280`](https://github.com/NVIDIA/xr-ai/blob/b5d9cb04ac1a40abc5d81bc07f85a6452448c870/agent-sdk/xr-ai-voice/xr_ai_voice/_processors/streaming_tts.py#L280) posts every sentence of a reply concurrently and [`streaming_tts.py:364`](https://github.com/NVIDIA/xr-ai/blob/b5d9cb04ac1a40abc5d81bc07f85a6452448c870/agent-sdk/xr-ai-voice/xr_ai_voice/_processors/streaming_tts.py#L364) cancels those tasks on interruption. Cancelling the client closes the socket but never reaches the executor thread, so each abandoned sentence still takes the lock and generates to completion, and the stop acknowledgement and the next reply's first sentence wait behind speech nobody will hear. The 30 s preset timeout inherited from Piper counts that queue time, so trailing sentences of a long reply time out client-side and are dropped while the server keeps synthesizing them. Check `request.is_disconnected()` before taking the lock, or bound the client to one in-flight request per participant, and record the measured real-time factor on the reference hosts.

2. **The launcher contract kept verbatim from Piper lost its tests, and the one synthesis test cannot fail.**

   [`test_pocket_tts.py:1`](https://github.com/NVIDIA/xr-ai/blob/b5d9cb04ac1a40abc5d81bc07f85a6452448c870/tests/test_pocket_tts.py) replaces 22 tests with 10; gone are all four `_ensure_owned_process_group` cases, the reuse-monitor failure limit, the wildcard and non-loopback probe hosts, the serve-error translation, the never-starts timeout, and both process-group integration tests, while the code they covered is unchanged and [`ai-services.md:359`](https://github.com/NVIDIA/xr-ai/blob/b5d9cb04ac1a40abc5d81bc07f85a6452448c870/docs/source/components/ai-services.md#L359-L370) still promises the group-cleanup behavior. [`test_pocket_tts.py:66`](https://github.com/NVIDIA/xr-ai/blob/b5d9cb04ac1a40abc5d81bc07f85a6452448c870/tests/test_pocket_tts.py#L44-L67) returns a constant from a fake tensor whose `clamp`, `__mul__`, and `short` are identities, so the int16 conversion at [`__main__.py:129`](https://github.com/NVIDIA/xr-ai/blob/b5d9cb04ac1a40abc5d81bc07f85a6452448c870/services/pocket-tts/pocket_tts_server/__main__.py#L129-L136) is never exercised, and nothing drives the FastAPI app, so the 503-before-load, 400 mapping, and media types are untested. Port the dropped cases, back the fake with numpy, and add one `TestClient` pass.

3. **The provenance record names model revisions the code never downloads.**

   [`THIRD_PARTY_NOTICES.md:45`](https://github.com/NVIDIA/xr-ai/blob/b5d9cb04ac1a40abc5d81bc07f85a6452448c870/THIRD_PARTY_NOTICES.md#L45-L47) records `4925226` as the gated revision reviewed and `e81d79e` as the "ungated no-voice-cloning fallback revision", but the pinned package's `config/english.yaml` fetches gated weights at `39592ff`, fallback weights and the tokenizer at `d29db79` from the separate repo `kyutai/pocket-tts-without-voice-cloning`, and only the `bill_boerst` embedding at `e81d79e`. The section also says the service downloads `kyutai/pocket-tts` and voices from `kyutai/tts-voices`, but every default run downloads the tokenizer and voice embedding from the ungated repo and touches `tts-voices` only when a voice is given as an `hf://` path. Record the repos and revisions the wheel actually pins, with the second repo's license.

## Suggestions

1. [`__main__.py:100`](https://github.com/NVIDIA/xr-ai/blob/b5d9cb04ac1a40abc5d81bc07f85a6452448c870/services/pocket-tts/pocket_tts_server/__main__.py#L100) inherits pocket-tts's silent fallback: any failure to fetch the gated weights loads the ungated repo with no log line, and the predefined voice works either way, so [`credentials.md:40`](https://github.com/NVIDIA/xr-ai/blob/b5d9cb04ac1a40abc5d81bc07f85a6452448c870/docs/source/getting_started/credentials.md#L40-L44) overstates that the default stack requires accepting the terms. Log which weights loaded next to "Pocket TTS ready", and consider loading the ungated config explicitly since the wrapper never uses voice cloning.

2. [`pyproject.toml:13`](https://github.com/NVIDIA/xr-ai/blob/b5d9cb04ac1a40abc5d81bc07f85a6452448c870/services/pocket-tts/pyproject.toml#L13) pulls `torch>=2.5.0` from PyPI, which the manifest resolves to 2.11.0 with CUDA 13 runtime packages on both x86_64 and aarch64, for a CPU-only model in its own venv next to vLLM; a per-package PyTorch CPU index, as stt-server already does for `lightning`, avoids a second multi-gigabyte CUDA install.

3. [`__main__.py:102`](https://github.com/NVIDIA/xr-ai/blob/b5d9cb04ac1a40abc5d81bc07f85a6452448c870/services/pocket-tts/pocket_tts_server/__main__.py#L102) passes `voice:` straight to `get_state_for_audio_prompt`, which accepts any local path or URL and clones it when the gated weights loaded, and the built-in name map also exposes CC-BY-NC voices (`cosette`, `jean`) by bare name; [`pocket_tts_server.yaml:7`](https://github.com/NVIDIA/xr-ai/blob/b5d9cb04ac1a40abc5d81bc07f85a6452448c870/services/pocket-tts/pocket_tts_server.yaml#L7-L9) frames the key as a catalog pick. Say that arbitrary recordings are cloned and fall under the model's consent clause, name the non-commercial voices, or restrict the wrapper to the predefined names.

4. A mistyped name (`bill_boerts`) is treated as a file path, so with the ungated fallback the process exits with the upstream "accept the terms" message, sending the user to fix credentials for a typo; a membership check against the predefined-voice map at startup restores what Piper's voice validation did.

5. [`presets/pocket_tts.py:9`](https://github.com/NVIDIA/xr-ai/blob/b5d9cb04ac1a40abc5d81bc07f85a6452448c870/agent-sdk/xr-ai-models/xr_ai_models/presets/pocket_tts.py#L9) keeps the 30 s timeout sized for a 100 ms ONNX call, and [`ai-services.md:542`](https://github.com/NVIDIA/xr-ai/blob/b5d9cb04ac1a40abc5d81bc07f85a6452448c870/docs/source/components/ai-services.md#L542-L546) dropped Piper's latency figure without a replacement; one measured first-audio number per reference profile belongs in both places.

6. [`__main__.py:137`](https://github.com/NVIDIA/xr-ai/blob/b5d9cb04ac1a40abc5d81bc07f85a6452448c870/services/pocket-tts/pocket_tts_server/__main__.py#L137-L138) returns `pcm` with no way to convey the rate, which changed from 22050 to 24000; an `X-Sample-Rate` header keeps that path usable. [`__main__.py:163`](https://github.com/NVIDIA/xr-ai/blob/b5d9cb04ac1a40abc5d81bc07f85a6452448c870/services/pocket-tts/pocket_tts_server/__main__.py#L163-L164) accepts `voice` and `speed` and ignores both.

7. Docs: [`xr-render-demo.md:198`](https://github.com/NVIDIA/xr-ai/blob/b5d9cb04ac1a40abc5d81bc07f85a6452448c870/docs/source/reference/xr-render-demo.md#L198) reworded a heading without the compatibility anchor the style guide requires (`tts-piper`); [`migrations.md:39`](https://github.com/NVIDIA/xr-ai/blob/b5d9cb04ac1a40abc5d81bc07f85a6452448c870/docs/source/reference/migrations.md#L39-L44) files a service replacement under "Removed SDK compatibility surfaces" above that section's own intro; [`ai-services.md:344`](https://github.com/NVIDIA/xr-ai/blob/b5d9cb04ac1a40abc5d81bc07f85a6452448c870/docs/source/components/ai-services.md#L344) still says "ONNX process".

The licensing work is careful (MIT package, CC-BY-4.0 model with the acceptable-use text quoted and dated, CC0 default voice), the rename is thorough, and the launcher plumbing lines up on both sides. The blockers above should be resolved before merge.

blongs-nv:
On the manifest nit: the manifest lists every project as a path member, so removing Piper has to remove its entry or the lock stops resolving. Deferring to the next cutoff only applies to version bumps.

nvddr:
Disposition for the review feedback (final squashed head: `6a7dac0c`):

### Blockers

1. **Abandoned synthesis queued behind Pocket's generation lock — fixed.**
   - Requests now wait on a cancellable asyncio lock before entering the executor.
   - The connection is checked after acquiring that lock, so disconnected queued sentences do not start generation.
   - A cancellation flag also closes the executor-submission race.
   - An already-running native generation remains non-cancellable; Piper had the same limitation for an active executor call. This change targets the new Pocket-specific serialized backlog.

2. **Lost lifecycle and ineffective synthesis coverage — fixed.**
   - Ported the applicable reuse monitoring, host probing, startup timeout, error translation, process-group ownership, launcher-abort, and managed-reuse tests.
   - The fake audio tensor is numpy-backed, so clamp/scaling/int16 conversion is exercised.
   - Added health/error/media-type route coverage. This invokes the FastAPI route directly with a controlled executor because the current test environment's TestClient/httpx combination hangs during executor shutdown; the subprocess integration tests exercise the real Uvicorn lifecycle.

3. **Incorrect artifact provenance — fixed.**
   - Recorded the actual Pocket TTS 3.0.2 pins: gated weights at `39592ff...`, fallback weights/tokenizer at `d29db79...`, and the `bill_boerst` embedding at `e81d79e...`.
   - Distinguished those artifact revisions from the model-record revisions at which the acceptable-use text was reviewed.
   - Recorded both model repositories and the source recording provenance.

### Suggestions

1. **Silent gated fallback / credentials wording — fixed.**
   - Startup logs now identify gated voice-cloning weights versus the ungated fallback.
   - Credentials docs explain that gated access is attempted but is not required for the checked-in voice.
   - We did not force the ungated configuration: retaining upstream selection preserves gated/fallback operation, while the wrapper exposes no cloning input.

2. **Unnecessary CUDA Torch installation — fixed.**
   - Added a service-scoped CPU PyTorch index and direct Torch declaration, following the repository's scoped-source pattern.
   - The Pocket service lock now resolves `torch+cpu` and removes its CUDA/cuDNN/NCCL/Triton branch.
   - Pocket remains CPU-only; no GPU configuration was added.

3. **Unreviewed/non-commercial voices and arbitrary cloning inputs — fixed by restriction.**
   - This release accepts only `bill_boerst`, whose source Voice-Zero recording is CC0.
   - Other predefined voices, URLs, local recordings, and arbitrary voice-cloning inputs are rejected.

4. **Mistyped voice producing a misleading credential failure — fixed.**
   - Unsupported names fail before importing/loading the model and identify the sole supported voice.

5. **30-second client timeout and latency number — intentionally unchanged.**
   - The 30-second value is inherited from Piper, so changing it was not established as a new regression.
   - No reference-host measurement was available; we did not invent a timeout or publish an unmeasured latency claim.
   - The concrete new queueing regression is addressed as described above.

6. **PCM sample-rate header and ignored request `voice`/`speed` fields — intentionally unchanged.**
   - These are inherited OpenAI-compatible service semantics rather than new Pocket regressions: the shared request schema already accepts those fields, the selected server model/voice remains configuration-owned, and raw PCM rate is known by the configured backend/client path.
   - Changing them would broaden this licensing/replacement PR into an API behavior change.

7. **Documentation defects — fixed.**
   - Restored the `tts-piper` compatibility anchor, moved the migration into a dedicated local-speech section, and replaced stale ONNX wording.

### Dependency manifest

The Piper path member was replaced by Pocket in both the generated manifest and lock. That change cannot be deferred because leaving the removed Piper path makes the aggregate lock unresolvable. The CPU Torch source also required a regenerated lock; the root qualification cutoff was not changed. Both the aggregate manifest check and every-project lock check now pass.

### Deliberately outside scope

- No Piper compatibility alias or fallback.
- No additional Pocket voice catalog.
- No arbitrary voice cloning.
- No GPU support.
- No within-sentence Pocket streaming redesign; the existing voice pipeline continues to submit and play sentence batches as it did with Piper.

wenxind-nvidia (approved):

blongs-nv (approved):

PR #508, opened 2026-09-09
Complete reciprocal dependency license notices

nvddr:
## Summary

- add complete license and attribution texts for audited LGPL and MPL dependencies and bundled native libraries
- document corresponding-source and redistribution obligations for `num2words`, `pycountry`, `tqdm`, `soxr`/libsoxr, SoundFile/libsndfile and its MP3 codecs, OpenCV/FFmpeg, `certifi`, and PyZMQ/libzmq/libsodium
- explicitly elect `Artistic-1.0-Perl` for dual-licensed `text-unidecode`; XR AI does not rely on its alternative GPL grant
- preserve the reviewed Pocket TTS model terms and accepted-use text in the compliance record
- add a guard test binding audited versions across the qualified lock, notice table, and bundled license-file records

This PR changes no dependencies or runtime APIs. It is based directly on `release/v0.2.0-beta` after #507 merged.

## Validation

- `10 passed` in `tests/test_third_party_notices.py`
- repository-wide SPDX header check passes
- `git diff --check` passes

PR #510, opened 2026-09-09
Refresh vulnerable vLLM dependencies and runtime image

nvddr:
## Summary

- require vLLM 0.27 or newer and refresh the aggregate scanner lock to vLLM 0.28.0
- update every shipped vLLM runtime configuration to the NVIDIA 26.08 image containing vLLM 0.27.1
- auto-detect DGX Spark for the standalone Nano and Omni launchers and default NVFP4 workloads to four sequences there; other configured profiles retain their existing sequence limits
- retain the existing optional Omni `moe_backend` forwarding for explicitly configured custom profiles while all shipped profiles leave backend selection to vLLM
- use vLLM 26.08 native Mamba and causal-convolution implementations instead of compiling `mamba-ssm` and `causal-conv1d` into containers
- preserve both isolated-service PyNvVideoCodec/protobuf inventories in the scanner manifest through mutually exclusive environments
- document the stack-upgrade cleanup path for stale managed containers and old vLLM images

## Security disposition

The vLLM floor is `>=0.27.0`, which fixes CVE-2026-73558 and the older reported findings fixed at or below 0.26.0. The scanner lock resolves vLLM 0.28.0 at the repository qualification cutoff, while shipped runtime YAML uses the qualified NVIDIA 26.08 image with vLLM 0.27.1; both satisfy the security floor.

The currently unfixed vLLM advisories CVE-2026-53923, CVE-2026-9540, and CVE-2026-12491 are explicitly accepted pending upstream fixed releases. Hydra, Lightning, NLTK, and setuptools findings remain excluded because their patched versions cannot be selected through a compatible refresh with the current NeMo/vLLM constraints. The two manifest environments retain protobuf 6.33.5 and 6.33.6; CVE-2026-0994 is fixed starting at 6.33.5.

## License review

The refreshed lock new packages were audited. `nccl4py` and `nvtx` are Apache-2.0, `nvidia-ml-py` and `torchcodec` are BSD-family, and the CUTLASS/NVDisasm wheels use NVIDIA proprietary/EULA terms. After incorporating current main, the CUDA 13 packages newly inventoried through Pocket TTS were also checked against their exact-release PyPI metadata; they use NVIDIA proprietary/Other terms, with `nvidia-nvtx` declaring Apache-2.0. No reciprocal license was introduced. All nine versions audited by `THIRD_PARTY_NOTICES.md` are unchanged, so no notices update is required.

## Validation

- 235 focused CPU tests passed on the rebased tree in independent review
- dependency map and manifest reproducibility checks passed
- repository-wide Ruff, SPDX-on-changed-files, and diff checks passed
- strict Sphinx documentation and generated public-API reference checks passed
- an earlier reviewed head passed 1,579 non-GPU tests with 11 GPU tests deselected
- NVIDIA 26.08 Nano FP8 startup and text inference passed on RTX 6000 Ada
- NVIDIA 26.08 Omni FP8 startup and multimodal inference passed on RTX 6000 Ada
- direct image inspection confirmed the external `mamba-ssm` and `causal-conv1d` distributions are absent and vLLM supplies the native model/ops implementations

GPU inference was not run locally for the 96G Blackwell or DGX Spark profiles. Argument-generation and GPU-identity tests cover Spark auto-detection and its four-sequence default.


wenxind-nvidia (commented):
## Blockers

1. **The standalone Nano service misses the new image’s required hardware settings.**  
   [nemotron3_nano_llm_server.yaml:71](https://github.com/NVIDIA/xr-ai/blob/17c82b94841a52f39cf1bf5e526770c567b8cd50/services/nemotron3-nano-llm/nemotron3_nano_llm_server.yaml#L71) switches to 26.08, but retains `max_num_seqs: 8`. Its launcher also cannot forward `moe_backend`. NVIDIA requires four sequences for Nano V3 NVFP4 on Spark, and `flashinfer_cutlass` on B200/B300. The PR applies workarounds to Omni profiles but misses this documented standalone entry point. [NVIDIA release notes](https://docs.nvidia.com/deeplearning/frameworks/vllm-release-notes/rel-26-08.html).

   I reproduced the generated command with mocked hardware: eight sequences and no MoE-backend flag. Apply the corresponding settings to the standalone Nano path and add argument-generation tests.

2. **The scanner override hides a dependency version installed by vLLM.**  
   [generate_dependency_manifest.py:30](https://github.com/NVIDIA/xr-ai/blob/17c82b94841a52f39cf1bf5e526770c567b8cd50/.github/scripts/generate_dependency_manifest.py#L30) overrides `pynvvideocodec>=2.2`, leaving only **2.2.2** in the scanner lock. However, the newly resolved vLLM **0.28.0** requires **PyNvVideoCodec==2.0.4**, and runtime service projects do not apply this override. [Official package metadata](https://pypi.org/pypi/vllm/0.28.0/json).

   Consequently, version-specific scanning misses a real runtime dependency. Preserve scanner-visible inventories for both conflicting environments instead of substituting one version globally. This is an inventory-coverage issue—not a claim that 2.0.4 is vulnerable.

## Suggestions

1. **Validate the checked-in GPU configurations with the new image.**  
   [test_gpu_llm_servers.py:332](https://github.com/NVIDIA/xr-ai/blob/17c82b94841a52f39cf1bf5e526770c567b8cd50/tests/test_gpu_llm_servers.py#L332) removes Omni’s expected-failure marker, but the reported validation excludes GPU tests. The fixtures also use one sequence rather than the shipped configurations. Record startup and inference results for the affected hardware profiles, including an Omni multimodal request.

## Nits

None.

Validation: **180 focused CPU tests passed**; GPU inference was not run locally. Dependency verification included an independent review.

Reviewed head: 17c82b94841a52f39cf1bf5e526770c567b8cd50

nvddr:
Addressed both blockers and the GPU-validation suggestion in rebased commit `231abe45`.

1. **Standalone Nano settings:** reduced the shipped configuration to four sequences and added `moe_backend` forwarding for B200/B300 (SM100-series), with argument-generation coverage for Ada, SM100, and DGX Spark. A direct RTX 6000 Ada check showed that forcing `flashinfer_cutlass` there fails, so the launcher deliberately forwards it only on SM100-series GPUs; Ada and Spark retain vLLM's automatic backend selection.
2. **Scanner inventory:** removed the global dependency overrides. The manifest now resolves mutually exclusive `repository` and `vllm` environments in one universal lock, preserving both PyNvVideoCodec 2.2.2 and the vLLM runtime's 2.0.4 (as well as both protobuf lines). Tests assert this separation and inventory coverage.
3. **GPU validation:** with `nvcr.io/nvidia/vllm:26.08-py3` (vLLM 0.27.1), Nano FP8 started and completed text inference on an RTX 6000 Ada; Omni FP8 started and completed a multimodal image request on the same GPU. B200/B300 and DGX Spark hardware were unavailable, so I did not claim direct validation for those profiles.

Validation after rebasing onto `main`: 1,579 non-GPU tests passed (11 GPU tests deselected), focused tests passed, the dependency manifest reproduced exactly, and Ruff/SPDX/diff checks passed. The PR remains one squashed, DCO-signed commit.

blongs-nv:
Reviewed at `231abe45b57d8f67b48892a436bfcbbd229258ae` by a bot.

## Blockers

1. **The standalone Omni config moves to 26.08 without either of the 26.08 NVFP4 settings.**

   [`nemotron_omni_llm_server.yaml:33`](https://github.com/NVIDIA/xr-ai/blob/231abe45b57d8f67b48892a436bfcbbd229258ae/services/nemotron-omni-llm/nemotron_omni_llm_server.yaml#L33) keeps `max_num_seqs: 384` and sets no `moe_backend`, while [line 58](https://github.com/NVIDIA/xr-ai/blob/231abe45b57d8f67b48892a436bfcbbd229258ae/services/nemotron-omni-llm/nemotron_omni_llm_server.yaml#L58) selects the 26.08 image. The wrapper picks the NVFP4 checkpoint on every SM100+ device ([`__main__.py:99`](https://github.com/NVIDIA/xr-ai/blob/231abe45b57d8f67b48892a436bfcbbd229258ae/services/nemotron-omni-llm/nemotron_omni_llm_server/__main__.py#L99-L100)), so B200 users get NVFP4 MoE without `flashinfer_cutlass` and Spark users get 384 sequences where the release notes require 4. Only the two profile YAMLs under `agent-samples/model-servers/yaml/` carry the settings, and [`test_spark_omni_uses_explicit_kv_cache`](https://github.com/NVIDIA/xr-ai/blob/231abe45b57d8f67b48892a436bfcbbd229258ae/tests/test_model_servers.py#L599-L626) is the only test that runs a shipped Omni YAML through the wrapper. This is the same gap the Nano fix in this push closed, on the other NVFP4 entry point. The Nano fix also shows the cleaner shape: the wrapper knows the arch and the model it chose, so apply the per-model rules there with YAML able to override.

2. **The two wrappers now disagree about where CUTLASS MoE is allowed, and the 96 GB profile sits on the wrong side.**

   The Nano wrapper forwards `--moe-backend` only when `major == 10` ([`__main__.py:141`](https://github.com/NVIDIA/xr-ai/blob/231abe45b57d8f67b48892a436bfcbbd229258ae/services/nemotron3-nano-llm/nemotron3_nano_llm_server/__main__.py#L139-L142)), and the new test pins SM120 to no backend ([`test_model_servers.py:560`](https://github.com/NVIDIA/xr-ai/blob/231abe45b57d8f67b48892a436bfcbbd229258ae/tests/test_model_servers.py#L560-L563)). The Omni wrapper forwards it unconditionally ([`__main__.py:168`](https://github.com/NVIDIA/xr-ai/blob/231abe45b57d8f67b48892a436bfcbbd229258ae/services/nemotron-omni-llm/nemotron_omni_llm_server/__main__.py#L168-L169)), and the [`96G_blackwell` profile](https://github.com/NVIDIA/xr-ai/blob/231abe45b57d8f67b48892a436bfcbbd229258ae/agent-samples/model-servers/yaml/96G_blackwell/nemotron_omni_llm_server.yaml#L4-L7) sets `flashinfer_cutlass` for a "single Blackwell 96 GB", which is RTX PRO 6000, SM120. The release notes name B200/B300 only, and the PR reply says forcing CUTLASS failed on Ada. Either SM120 needs the backend, in which case the Nano gate and its test are wrong, or it does not, in which case the 96 GB Omni profile forces a backend the same PR refuses to send to the same GPU for Nano. Pick one and make both wrappers apply it.

## Suggestions

1. [`nemotron3_nano_llm_server.yaml:44`](https://github.com/NVIDIA/xr-ai/blob/231abe45b57d8f67b48892a436bfcbbd229258ae/services/nemotron3-nano-llm/nemotron3_nano_llm_server.yaml#L44) drops the standalone Nano to 4 sequences on every GPU. The 26.08 requirement is Spark-only; on the Ada and B200 configurations the comments above it describe, this halves concurrency for no stated reason, and the test at [`test_model_servers.py:592`](https://github.com/NVIDIA/xr-ai/blob/231abe45b57d8f67b48892a436bfcbbd229258ae/tests/test_model_servers.py#L592) now pins 4 for compute major 8 too. Gate the cap on Spark the way the backend is gated on SM100, or say in the comment why 4 is right everywhere.

2. The manifest now records both environments honestly (protobuf 6.33.5 with grpcio-tools 1.81.1 in the repository fork, 6.33.6 in the vLLM fork; pynvvideocodec 2.0.4 and 2.2.2), which resolves the earlier override concern. The written rules still say otherwise: [`AGENTS.md:125`](https://github.com/NVIDIA/xr-ai/blob/231abe45b57d8f67b48892a436bfcbbd229258ae/AGENTS.md#L125-L127) says only a `uv.toml` change regenerates the manifest, the generated header at [`pyproject.toml:6`](https://github.com/NVIDIA/xr-ai/blob/231abe45b57d8f67b48892a436bfcbbd229258ae/dependency-manifest/pyproject.toml#L6) repeats it, and [`DEPENDENCIES.md:57`](https://github.com/NVIDIA/xr-ai/blob/231abe45b57d8f67b48892a436bfcbbd229258ae/DEPENDENCIES.md#L57-L61) describes one dependency set with no mention of the two extras or the conflict. Per the same-commit rule, name the security refresh as a second legitimate trigger and describe the two forks in DEPENDENCIES.md.

3. The three profile YAMLs pass `extra_pip: []` on the claim that 26.08 needs no mamba-ssm, but the Omni wrapper default at [`__main__.py:144`](https://github.com/NVIDIA/xr-ai/blob/231abe45b57d8f67b48892a436bfcbbd229258ae/services/nemotron-omni-llm/nemotron_omni_llm_server/__main__.py#L144) is still the two-package list and the standalone YAML does not set the key, so the standalone path still compiles mamba-ssm in-container on every start. Flip the default and drop the per-profile overrides.

4. [`DEFAULT_IMAGE`](https://github.com/NVIDIA/xr-ai/blob/231abe45b57d8f67b48892a436bfcbbd229258ae/utils/xr-ai-vllm/xr_ai_vllm/__init__.py#L59), `_DEFAULT_VLLM_IMAGE` in vlm_server, five docstrings, twelve YAML `vllm_image` values and the docs row all say 26.08 with nothing tying them together. One test that globs the YAMLs and asserts the set of images equals the default would make the next bump a one-line change.

5. The security disposition for vLLM is coherent: OSV shows one advisory fixed in 0.27.0 (CVE-2026-73558) and everything else at or below 0.26.0, so `>=0.27` is the minimal correct floor, and the excluded setuptools, hydra, lightning and nltk findings check out against the pins. Two gaps: vllm still carries unfixed OSV entries (CVE-2026-53923, CVE-2026-9540, CVE-2026-12491) that the disposition should name as accepted, and the lock records vllm 0.28.0 while every shipped config runs 0.27.1 from the image. The "refresh protobuf to 6.33.6" line should go: 6.33.6 fixes no advisory (CVE-2026-0994 is fixed in 6.33.5), and the repository fork still records 6.33.5.

6. Licenses: the new lock packages (nccl4py Apache-2.0, nvtx Apache-2.0, nvidia-ml-py BSD, torchcodec BSD-3, nvidia-cutlass-dsl-libs-* NVIDIA proprietary, nvidia-cuda-nvdisasm NVIDIA EULA) add no reciprocal license, and all nine audited versions in `THIRD_PARTY_NOTICES.md` are unchanged so the #508 guard holds. [`DEPENDENCIES.md:46`](https://github.com/NVIDIA/xr-ai/blob/231abe45b57d8f67b48892a436bfcbbd229258ae/DEPENDENCIES.md#L46-L49) says a lock refresh audits license metadata; the PR body should state that this was done for the proprietary NVIDIA wheels.

## Nits

- [`pyproject.toml:15`](https://github.com/NVIDIA/xr-ai/blob/231abe45b57d8f67b48892a436bfcbbd229258ae/services/nemotron3-nano-llm/pyproject.toml#L15-L17) still explains the floor as "0.12.0 is the minimum the model card calls for" directly above `vllm>=0.27.0`; the floor is now a CVE floor.
- [`test_dependency_manifest.py:145`](https://github.com/NVIDIA/xr-ai/blob/231abe45b57d8f67b48892a436bfcbbd229258ae/tests/test_dependency_manifest.py#L140-L145) checks that both pynvvideocodec versions are in the lock but not that the lock's extras and conflict match the generator; a sibling assertion on `lock["manifest"]` would catch a regenerate-without-relock.

yanziz-nvidia (approved):

yanziz-nvidia:
needs to test a bit before merging

nvddr:
Addressed the remaining review at `cc836039`.

- Standalone Nano and Omni now auto-detect DGX Spark and default to four sequences there; other GPUs retain Nano 8 / Omni 384 unless YAML overrides them.
- Both wrappers now apply `flashinfer_cutlass` only to the NVFP4 model on SM100 (B200/B300). The RTX PRO 6000 SM120 profile no longer forces it.
- Omni defaults `extra_pip` to an empty list and the redundant profile overrides are gone. Inspection of `nvcr.io/nvidia/vllm:26.08-py3` confirmed that the external `mamba-ssm` and `causal-conv1d` distributions are absent; 26.08 instead carries native/adapted vLLM Mamba and causal-convolution ops plus the native Nano Omni model implementation.
- All shipped YAML image pins are checked against the shared dispatcher default, and vlm-server no longer carries a second image constant.
- The manifest generation contract now allows targeted security refreshes, documents the mutually exclusive repository/vLLM environments, and tests the lock extras/conflict as well as both dependency inventories.
- The PR body now names the three accepted unfixed vLLM advisories, explains lock vLLM 0.28.0 versus runtime vLLM 0.27.1, removes the inaccurate protobuf-refresh claim, and records the proprietary-wheel license audit.
- The upgrade docs tell users to stop persistent services, pull 26.08, let fingerprinting recreate stale containers, and remove old images only after successful startup while retaining the model cache.
- Updated the stale Nano dependency comment to describe the CVE floor.

Current validation: **236 relevant non-GPU tests passed** (4 GPU tests deselected), manifest regeneration/check passed, repository-wide Ruff passed, strict Sphinx docs/public-API checks passed, SPDX passed for every changed file, and `git diff --check` passed. I also attempted the entire non-GPU suite, but `test_agent_runtime.py::test_agent_exposes_existing_unary_and_streaming_tools` did not complete within a 60-second isolated run in this environment; the prior reviewed head's recorded 1,579-test run remains in the PR body. B200/B300 and Spark were not available for direct GPU execution.

nvddr:
Follow-up: the first CI run exposed that the PR was two commits behind `main`, whose Pocket TTS update changes the aggregate Torch/CUDA inventory. I merged current `main`, pinned the manifest generator to exact Python 3.11.14 for reproducible clean-run resolution, regenerated the 358-package lock, added a resolver-command regression test, and extended the license audit to the newly inventoried CUDA 13 wheels.

Final head: `fa162a88`. All eight workflows are green: **tests, Docs, dependency-manifest, lock-check, lint, SPDX, DCO, and CodeQL**. Current-base focused validation is **237 passed, 4 GPU tests deselected**.

nvddr:
Scope correction for the B200/B300 feedback:

- Removed the automatic `flashinfer_cutlass` / `--moe-backend` handling for SM100 from both Nemotron wrappers.
- Removed Nano’s newly added `moe_backend` configuration surface and the B200/B300-specific comments and test cases.
- Restored Omni’s pre-existing behavior: `moe_backend` is forwarded only when a user explicitly configures it; there is no hardware-specific default.
- Restored the 96G profile guidance to leave backend selection to vLLM.
- Kept the pre-existing Blackwell/NVFP4 checkpoint selection unchanged; this correction is limited to the B200/B300-specific wrapper override introduced in this PR.

Validation:
- `uv run pytest -q tests/test_model_servers.py` — 59 passed locally.
- All PR checks pass on Python 3.11 and 3.12, including strict docs, lockfiles, Ruff, DCO, and CodeQL.

Fixed in 4f35ef94.

wenxind-nvidia (commented):
## Blockers

None newly identified.

## Suggestions

1. **Update the PR description to match the final scope.** The summary still claims automatic `flashinfer_cutlass` selection for B200/B300, and the validation section claims coverage of those hardware-specific settings. The [latest scope correction](https://github.com/NVIDIA/xr-ai/pull/510#issuecomment-5671302102) explicitly removes that behavior and its tests. Describe Spark auto-detection and Omni's explicitly configured backend forwarding instead.

## Nits

1. **Update the remaining stale docstrings.**

   - [`serve()` documentation](https://github.com/NVIDIA/xr-ai/blob/12b286339263d969b21cfaeddfe057e449c19850/utils/xr-ai-vllm/xr_ai_vllm/__init__.py#L109) still presents `mamba-ssm` and `causal-conv1d` as packages needed for Omni, contradicting the new native implementation.
   - [Nano](https://github.com/NVIDIA/xr-ai/blob/12b286339263d969b21cfaeddfe057e449c19850/services/nemotron3-nano-llm/nemotron3_nano_llm_server/__main__.py#L23) and [Omni](https://github.com/NVIDIA/xr-ai/blob/12b286339263d969b21cfaeddfe057e449c19850/services/nemotron-omni-llm/nemotron_omni_llm_server/__main__.py#L23) still document unconditional sequence defaults of 8 and 384; mention the new Spark default of 4.

The scanner-inventory issue is addressed: both PyNvVideoCodec versions are retained without global overrides. Independent review found no additional manifest issue.

Validation: **235 focused CPU tests passed**, dependency-map check passed, and all GitHub checks are green. GPU inference was not rerun locally.

Reviewed head: 12b286339263d969b21cfaeddfe057e449c19850

wenxind-nvidia (approved):

blongs-nv (approved):

nvddr:
Addressed Wenxin’s follow-up review at `358d04fa`:

- Updated the PR summary to remove the deleted B200/B300 behavior and its claimed test coverage.
- Described the actual final behavior: Spark auto-detection with the four-sequence NVFP4 default, plus Omni forwarding `moe_backend` only when explicitly configured.
- Corrected the `serve()` docstring so it no longer says Omni needs external `mamba-ssm` / `causal-conv1d` wheels.
- Corrected the Nano and Omni `max_num_seqs` docstrings to include their Spark defaults.
- Validation now states precisely which supported GPU profiles were and were not exercised directly.

PR #514, opened 2026-09-11
Add GPU streaming for Pocket TTS

nvddr:
## Summary

- stream native Pocket TTS PCM chunks through the typed model client and voice pipeline
- support `cpu`, `cuda`, and `auto` placement, with CUDA warmup before readiness
- use CUDA in the shipped GPU profiles while retaining non-streaming WAV and CPU fallback behavior
- document the adapter and deployment changes

## Measured latency

End-to-end local HTTP streaming on an RTX 6000 Ada after warmup:

- 5 words: first audio in 76.1 ms; complete in 317.7 ms
- 18 words: first audio in 44.5 ms; complete in 687.6 ms

## Testing

- `112 passed` — model configuration, protocols, and OpenAI-compatible clients
- `12 passed` — focused Pocket TTS configuration, backend, and HTTP cases
- Ruff and `git diff --check`
- manual CUDA server/client streaming validation

The full repository suite was not run.


yanziz-nvidia:
reviewed by yanziz-review-bot

**Summary**

The PR adds native PCM streaming and CUDA placement. Other than the existing findings in previous comments, I found two cancellation issues below. Focused validation passed 142 model/server tests and 18 voice TTS tests; broader testing encountered local sandbox/dependency failures, and CUDA execution was not validated.

| Severity | Finding |
|---|---|
| 🚫 BLOCKER | [`__main__.py:174–177`](https://github.com/NVIDIA/xr-ai/blob/e9de8f926a77ce632860910fc665ac990b57eb5f/services/pocket-tts/pocket_tts_server/__main__.py#L174) — **Cancellation releases the model lock before generation finishes.** Returning here abandons the upstream iterator. Pocket TTS 3.0.2’s background generation/decoder threads keep running because their cleanup is not protected by `finally` ([upstream implementation](https://github.com/kyutai-labs/pocket-tts/blob/v3.0.2/pocket_tts/models/tts_model.py#L674)). A reproduction confirmed a live worker with the generation lock available. Subsequent requests can therefore overlap on a model explicitly documented as not thread-safe, while abandoned work continues consuming resources. Drain and discard remaining output off the event loop while retaining ownership, or implement cancellation that stops and joins upstream workers before releasing the lock. |
| 💡 SUGGESTION | [`__main__.py:277–289`](https://github.com/NVIDIA/xr-ai/blob/e9de8f926a77ce632860910fc665ac990b57eb5f/services/pocket-tts/pocket_tts_server/__main__.py#L277) — **Disconnect cleanup can close an executing generator.** Under Uvicorn’s ASGI 2.3 path, Starlette’s cancelled AnyIO scope immediately cancels the unshielded `await next_task` again. The executor thread continues, so `stream.close()` raises `ValueError: generator already executing`; reproduced using the actual response implementation. Shield cleanup from the enclosing cancellation scope and finish the outstanding worker operation before closing the iterator. |
| 🧹 NIT | none |

- Keep generation ownership until upstream workers finish after cancellation; add a regression covering interruption followed immediately by another request.
- Protect disconnect cleanup from repeated AnyIO cancellation; test disconnect while `next()` is blocked.

wenxind-nvidia (commented):
The two cancellation issues identified as blockers in my review are already covered in [the other review](https://github.com/NVIDIA/xr-ai/pull/514#issuecomment-5641848497), so I am not duplicating their details here.

## Suggestion

1. **Exercise the new streaming branch in voice-pipeline tests.**
   [streaming_tts.py:291](https://github.com/NVIDIA/xr-ai/blob/e9de8f926a77ce632860910fc665ac990b57eb5f/agent-sdk/xr-ai-voice/xr_ai_voice/_processors/streaming_tts.py#L291)

   Existing voice tests use non-streaming TTS doubles. Add a streaming double covering early audio delivery, sentence ordering, participant-specific interruption, iterator cleanup, and `TTSStoppedFrame` ordering.

Reviewed head: e9de8f926a77ce632860910fc665ac990b57eb5f

nvddr:
@yanziz-nvidia and @wenxind-nvidia, addressed in the latest two commits:

- `47443b5d`: cancellation now drains the upstream Pocket TTS iterator while retaining the model lock; disconnect cleanup is shielded, waits for an outstanding `next()`, and only then closes the iterator. Added regressions for a canceled request followed immediately by another request and for disconnect while generation is blocked.
- `97a9e76a`: added a streaming TTS double and voice-pipeline coverage for first-chunk delivery, sentence ordering, participant-scoped interruption, async-iterator cleanup, and `TTSStoppedFrame` ordering.

Validation: `155 passed` across `tests/test_voice_pipeline.py` and `tests/test_pocket_tts.py`; Ruff 0.15.16 passes.

nvddr:
## Review disposition

- **Yanzi — cancellation/model-lock blocker:** Accepted and fixed in `47443b5d`; canceled streams are drained while retaining model ownership.
- **Yanzi — disconnect cleanup suggestion:** Accepted and fixed in `47443b5d`; cleanup is shielded, waits for an outstanding `next()`, then closes the iterator.
- **Wenxin — streaming voice-pipeline coverage:** Accepted and fixed in `97a9e76a`; tests cover early delivery, sentence order, participant-scoped interruption, iterator cleanup, and `TTSStoppedFrame` order.

No review findings were deferred or rejected. Validation: 155 affected tests pass and Ruff is clean.

yanziz-nvidia (approved):

yanziz-nvidia:
reviewed by yanziz-review-bot

**Summary**

The earlier model-lock and generator-close issues are fixed, and streaming voice-pipeline coverage is added. Re-review found one new cleanup performance issue. Locally, 145 model/server tests and 21 voice TTS tests passed; CUDA execution was not validated.

| Severity | Finding |
|---|---|
| 🚫 BLOCKER | none |
| 💡 SUGGESTION | [`__main__.py:64–68`](https://github.com/NVIDIA/xr-ai/blob/97a9e76a368538ce1e02befc8f137a5e432ff0a2/services/pocket-tts/pocket_tts_server/__main__.py#L64) — **Disconnect cleanup busy-loops under AnyIO cancellation.** Starlette’s cancelled scope repeatedly cancels each new `await asyncio.shield(task)`, so this loop continually retries until generation finishes. An actual `StreamingResponse` disconnect with a 500 ms blocked worker produced **32,643 shield calls**, consuming unnecessary CPU throughout cleanup. Wrapping cleanup in `anyio.CancelScope(shield=True)` reduced this to three calls while preserving orderly closure. Shield against the enclosing scope and add an ASGI disconnect regression; the current test’s two explicit `Task.cancel()` calls do not exercise this behavior. |
| 🧹 NIT | none |

- Shield cleanup from AnyIO scope cancellation while preserving protection against explicit task cancellation. Test a real `http.disconnect` during blocked generation and verify cleanup waits without repeatedly retrying.

wenxind-nvidia (commented):
## Blockers

1. **Repeated cancellations can deadlock speech generation.** In [`__main__.py:291–309`](https://github.com/NVIDIA/xr-ai/blob/97a9e76a368538ce1e02befc8f137a5e432ff0a2/services/pocket-tts/pocket_tts_server/__main__.py#L291), the async generation lock is released before cleanup finishes. If cancellation occurs during PCM conversion, the generator can remain suspended holding the backend lock. Subsequent cancelled requests occupy executor threads waiting for that lock, eventually preventing the queued `stream.close()` from running.

   Reproduced against the actual endpoint with deliberately delayed PCM conversion and the default 14-thread executor: **15 unfinished requests, with cleanup queued and the model lock held**. Keep the async generation lock through outstanding-worker completion and iterator closure; add a cancellation-under-load regression.

## Suggestions

The cleanup busy-loop is already covered by [the latest review](https://github.com/NVIDIA/xr-ai/pull/514#issuecomment-5641964186). Independently confirmed approximately 34,000 shield retries during a 500 ms blocked generation. This confirms the existing suggestion rather than adding a duplicate finding.

## Nits

None.

The earlier executing-generator closure issue is fixed, and the requested voice-streaming coverage is present. Validation: **145 model/server tests and 20 voice-streaming tests passed**; two launcher integration tests excluded. CUDA inference was not tested locally.

Reviewed head: 97a9e76a368538ce1e02befc8f137a5e432ff0a2

blongs-nv:
Reviewed at `97a9e76a368538ce1e02befc8f137a5e432ff0a2` by a bot.

## Blockers

1. **The new client is silent against any pre-PR server, which is the normal upgrade path.** [`_openai_compat.py:851`](https://github.com/NVIDIA/xr-ai/blob/97a9e76a368538ce1e02befc8f137a5e432ff0a2/agent-sdk/xr-ai-models/xr_ai_models/_openai_compat.py#L851) requires `x-audio-sample-rate`; the old server ignores the unknown `stream` field and returns a full PCM body with no header, so every sentence raises KeyError, the sender loop logs and drops it, and the agent says nothing. The [preset switch](https://github.com/NVIDIA/xr-ai/blob/97a9e76a368538ce1e02befc8f137a5e432ff0a2/agent-sdk/xr-ai-models/xr_ai_models/presets/pocket_tts.py#L8) sends every sample through this path, and [`run()`](https://github.com/NVIDIA/xr-ai/blob/97a9e76a368538ce1e02befc8f137a5e432ff0a2/services/pocket-tts/pocket_tts_server/__main__.py#L536-L545) reuses any healthy listener on 8105, so a stale server survives the upgrade. Fall back to `synthesize()` when the header is absent, or fail naming the version mismatch.

2. **Four new public names with no recorded approval.** [`__init__.py:86`](https://github.com/NVIDIA/xr-ai/blob/97a9e76a368538ce1e02befc8f137a5e432ff0a2/agent-sdk/xr-ai-models/xr_ai_models/__init__.py#L86-L93) exports `TTSChunk`, `StreamingTTSService`, `PocketTTS` and `KIND_POCKET_TTS`; AGENTS.md makes `__all__` the public API and this rule was enforced on #445. Record the approval, or keep the last two private: a `capabilities: {streaming: true}` flag on `openai_compat` with `stream()` on `OpenAICompatTTS` would need neither a new kind nor a new class.

## Suggestions

1. The cancel-path fix drains under the model lock, which is right, but the asyncio lock at [`__main__.py:290`](https://github.com/NVIDIA/xr-ai/blob/97a9e76a368538ce1e02befc8f137a5e432ff0a2/services/pocket-tts/pocket_tts_server/__main__.py#L290) is still held for the whole response, so a second participant waits on the first client's network pacing rather than on generation. Generating from one worker into a queue and releasing when the library generator is exhausted would drop that wait and the per-chunk task creation.

2. [`__main__.py:466`](https://github.com/NVIDIA/xr-ai/blob/97a9e76a368538ce1e02befc8f137a5e432ff0a2/services/pocket-tts/pocket_tts_server/__main__.py#L466) runs warmup on the default executor inside the startup timeout; a hung CUDA init leaves a non-daemon thread that `asyncio.run` joins without limit, so the process cannot exit. Run it in the same daemon thread as `_load`.

3. The stream path has no error conversion: a backend failure after headers are sent gives the client a truncated body and an "incomplete PCM sample" error. Pull the first chunk before building the `StreamingResponse` so pre-generation failures still map to 400/499 like the non-stream path.

4. Placement and docs: [`dual_48G_ada/stt_server.yaml:14`](https://github.com/NVIDIA/xr-ai/blob/97a9e76a368538ce1e02befc8f137a5e432ff0a2/agent-samples/model-servers/yaml/dual_48G_ada/stt_server.yaml#L14) still says GPU 0 is fully allocated to vlm; [`cuda_visible_devices`](https://github.com/NVIDIA/xr-ai/blob/97a9e76a368538ce1e02befc8f137a5e432ff0a2/services/pocket-tts/pocket_tts_server/__main__.py#L20-L29) is missing from the config-key docstring; the `stream` field and response headers are documented nowhere; and dropping the CPU torch index at [`pyproject.toml:14`](https://github.com/NVIDIA/xr-ai/blob/97a9e76a368538ce1e02befc8f137a5e432ff0a2/services/pocket-tts/pyproject.toml#L14-L16) makes every Linux install pull CUDA torch while the code default stays `cpu`. One [`migrations.md`](https://github.com/NVIDIA/xr-ai/blob/97a9e76a368538ce1e02befc8f137a5e432ff0a2/docs/source/reference/migrations.md#L42-L48) entry should cover GPU placement, the footprint change, and warmup inside `startup_timeout_s`. The manifest lock still records the CPU index; the next cutoff refresh will flip torch to CUDA for every member.

5. [`streaming_tts.py:365`](https://github.com/NVIDIA/xr-ai/blob/97a9e76a368538ce1e02befc8f137a5e432ff0a2/agent-sdk/xr-ai-voice/xr_ai_voice/_processors/streaming_tts.py#L365) calls the voice gate unguarded while the WAV path wraps it; a raising gate drops the sentence.

6. Still untested after the second push: device selection (`auto`, cuda unavailable, invalid value), the 400 for `stream` without `pcm`, warmup, and the client's missing-header and partial-sample paths. The `sys.modules` monkeypatch pattern in `test_pocket_tts.py` makes each one short.

## Nits

- [`__main__.py:260`](https://github.com/NVIDIA/xr-ai/blob/97a9e76a368538ce1e02befc8f137a5e432ff0a2/services/pocket-tts/pocket_tts_server/__main__.py#L260) `getattr(req, "stream", False)` exists for a test double; use `req.stream`.
- The PR body's "retaining CPU fallback" holds only for the standalone yaml; the profiles set `cuda` and raise without it.

nvddr:
## Additional review disposition

- **Wenxin blocker:** Accepted and fixed in `88c21fb2`. Generation now runs in one producer worker and retains ownership through iterator closure; cancellation under load is covered.
- **Yanzi cleanup suggestion:** Accepted and fixed. Cleanup uses an AnyIO shielded cancel scope, with a real ASGI disconnect regression.
- **Blongs blockers:** Accepted and fixed. Clients fall back against pre-streaming servers, and the new Pocket/streaming implementation remains private with no expansion of the public `xr_ai_models` API.
- **Blongs suggestions and nit:** Accepted and fixed. The patch decouples generation from network pacing, moves warmup into the daemon loader, validates the first chunk before headers, completes GPU/API/dependency documentation, preserves audio when gate observation fails, adds the requested tests, and uses `req.stream` directly.
- **Code-quality comments:** Accepted and fixed; both no-effect statements were removed.

Nothing was deferred or rejected. Validation: 276 affected tests pass; Ruff 0.15.16, dependency-map generation, Pocket TTS lock resolution, and `git diff --check` pass.

wenxind-nvidia (approved):

PR #515, opened 2026-09-12
Add StreamKit client-to-agent file transfer

blongs-nv:
Adds transport-neutral `sendBytes` and `sendFile` APIs across Web, iOS/visionOS, Android, and native StreamKit. DeviceIOHub validates and reassembles LiveKit byte streams, then delivers complete `FileMessage` objects to opted-in agents.

File traffic uses a separate bounded IPC lane with size, concurrency, timeout, participant-session, and routing controls, allowing applications to send images and other files without one-off data-channel framing or interference with real-time media.

Validation includes the 1,629-test CPU suite, Android Kotlin compilation, Swift package compilation, native stub CTest, and an end-to-end browser file-transfer run.


yanziz-nvidia:
reviewed by yanziz-review-bot

**Summary**

This adds file transfer across four clients and a separate hub IPC lane. The 36 focused Python tests pass, but a dependency compatibility issue blocks merge. Native compilation was not verified because CMake is unavailable.

| Severity | Finding |
|---|---|
| 🚫 BLOCKER | [`_room_client.py:358`](https://github.com/NVIDIA/xr-ai/blob/1911d92bc809fa6e0c0a242d1e036a2329ec964c/services/device-io-hub/device_io_hub/transport/livekit/_room_client.py#L358) — **Raise the LiveKit dependency floor.** `livekit>=1.0` permits versions without `DataStreamOptions`, now used unconditionally during connection. Reproduced an `AttributeError` with 1.1.8, preventing hub startup even when files are unused. This API first appears in 1.1.15; require at least that version and validate startup at the minimum. |
| 💡 SUGGESTION | [`_processor.py:769`](https://github.com/NVIDIA/xr-ai/blob/1911d92bc809fa6e0c0a242d1e036a2329ec964c/agent-sdk/xr-ai-hub/xr_ai_hub/_processor.py#L769) — **Recheck file subscriptions before dispatching queued callbacks.** With one callback blocked and another file queued, calling `unsubscribe("alice")` still allows the queued callback to execute afterward. The guard checks participant membership and session but ignores the current subscription. Also require `Subscribe.FILE` before dispatch, covering both unsubscribe and filter changes. |
| 🧹 NIT | none |

**Actionables**

- Require `livekit>=1.1.15` in the hub manifest, regenerate the dependency map, and test connection initialization at that minimum.
- Check the current participant file subscription before dequeued-file dispatch; add coverage for unsubscribe and removing `Subscribe.FILE` while a callback is blocked.

nvddr (commented):
## Summary review

This adds a coherent cross-platform file-transfer path: completed transfers are isolated from real-time IPC, bounded by size and queue limits, and protected against stale participant sessions. The implementation and documentation are generally clear, and the focused Python tests and current CI checks pass.

### Blocker

- **Raise the Python LiveKit dependency floor.** `services/device-io-hub/pyproject.toml` still permits `livekit>=1.0`, while `RoomClient.connect()` now unconditionally constructs `rtc.DataStreamOptions`. I reproduced that an allowed version, LiveKit 1.1.8, does not expose that symbol, so DeviceIOHub fails at startup even when file transfer is unused. `DataStreamOptions` is available starting in 1.1.15; require at least that version, regenerate the dependency map/lock, and validate startup at the minimum supported version.

### Suggestions

- **Honor file unsubscription for messages waiting in the callback queue.** `_run_file_callbacks()` rechecks participant/session liveness but not the current subscription. Consequently, after `unsubscribe("alice")` or changing the filter to remove `Subscribe.FILE`, an already queued file still invokes the callback. Recheck that the participant remains subscribed with `Subscribe.FILE` immediately before dispatch, and add coverage for unsubscribe/filter changes while another callback is blocked.

- **Make return-audio cleanup session-aware during same-identity reconnects.** `_handle_left()` now awaits file-task cancellation and participant notification before popping `_return_audio` by participant identity. During either await, the same identity can reconnect. The connector correctly ignores the stale departure using its session ID, but the old leave handler still closes the identity's return-audio entry. That can discard the new session's first response or briefly expose buffered audio from the prior session. Tag return-audio entries with their participant session, or detach the departing session's entry before the first suspension.

### Description and documentation

The description accurately reflects the architecture and local-completion semantics, and the documentation explains opt-in delivery and the operational limits clearly. One qualification would improve the validation claim: the visible PR checks cover Python, Android, docs, and linting, but do not include Swift compilation, native LiveKit compilation, or the browser end-to-end run. If those were run locally, label them as local validation.

Focused verification: 23 passed, 3 skipped. The current PR checks are green.

nvddr:
Before this lands, please extract the raw byte-stream mechanics into a private/internal transport seam so PR #501 can reuse them after rebasing without going through the public file-transfer API.

The narrow shape should be:

- A private client-side byte-stream writer that owns LiveKit stream creation, chunk writing, close-on-success/failure, and connection-generation checks. Public `sendBytes` / `sendFile` should call it, but its inputs should be raw internal wire options rather than public `FileSendOptions`.
- A private hub-side bounded byte-stream reader that owns declared/observed size checks, idle/total timeouts, chunk accumulation, and reader closure. The public file receiver should call it while retaining its existing file admission limits and `FileMessage` dispatch.

Please do not expose this helper from the StreamKit public API or make it part of the public/custom `StreamingBackend` contract. It should be package/module internal on Swift, Kotlin, and Web, and a non-installed detail interface/header in C++.

PR #501 can then use the same private transport primitives for its correlated capture response, while keeping all capture semantics separate: its reserved internal wire topic, request/session correlation, rejection payload, pending-request resolution, and capture-specific limits. It must not call public `sendBytes` / `sendFile`, use `_streamkit.file`, produce `FileMessage`, or require `Subscribe.FILE`.

This keeps the reusable API deliberately narrow: shared byte transport only; public file delivery and private capture delivery remain separate consumers.

nvddr (commented):
## Re-review summary

The earlier dependency-floor, queued-unsubscribe, same-identity return-audio, and private byte-stream seam concerns are addressed at `a312be5d`. The focused Python, JavaScript, and native stub suites pass, and the current PR checks are green. Two connection-ordering cases still risk reporting or causing silent file loss.

### Blockers

- **Preserve participant lifecycle ordering across the separate file lane.** `ConnectorEndpoint.notify_participant_joined()` sends the join on the main PUSH socket, while `push_file()` sends the completed file on a different PUSH socket. `HubEndpoint.run()` and `_run_files()` consume those sockets in independent tasks, and `_run_files()` drops the file when the main task has not installed the session yet ([`_hub.py:354-407`](https://github.com/NVIDIA/xr-ai/blob/a312be5d4f9722f1802e5cee146db7d4889f7cda/services/device-io-hub/device_io_hub/ipc/_hub.py#L354-L407)). ZeroMQ ordering does not extend across sockets, so a small transfer sent immediately after join/reconnect can reach the file task first and be discarded. The same race exists downstream: participant events arrive on the main SUB socket, files arrive on the file SUB socket, and the processor drops a file until it has consumed the join and installed the per-participant file subscription ([`_processor.py:719-768`](https://github.com/NVIDIA/xr-ai/blob/a312be5d4f9722f1802e5cee146db7d4889f7cda/agent-sdk/xr-ai-hub/xr_ai_hub/_processor.py#L719-L768), [`_processor.py:874-890`](https://github.com/NVIDIA/xr-ai/blob/a312be5d4f9722f1802e5cee146db7d4889f7cda/agent-sdk/xr-ai-hub/xr_ai_hub/_processor.py#L874-L890)). Put the lifecycle/session marker on the ordered file path, add an explicit barrier/ack, or temporarily buffer new-session files until that lane has observed the join. Add tests that deliberately deliver a file before the corresponding main-lane join at both boundaries.

- **Invalidate native transfers across LiveKit reconnects.** Native file sends capture `connect_generation_` and treat the stream as active when the room, connected flag, and generation still match ([`LiveKitBackend.cpp:639-645`](https://github.com/NVIDIA/xr-ai/blob/a312be5d4f9722f1802e5cee146db7d4889f7cda/client-samples/native/StreamKit/src/Backends/LiveKit/LiveKitBackend.cpp#L639-L645)). Teardown and a new explicit connection advance that generation, but `ApplyConnectionState()` only clears `is_connected_` when LiveKit enters reconnecting/disconnected state and does not advance the generation ([`LiveKitBackend.cpp:728-742`](https://github.com/NVIDIA/xr-ai/blob/a312be5d4f9722f1802e5cee146db7d4889f7cda/client-samples/native/StreamKit/src/Backends/LiveKit/LiveKitBackend.cpp#L728-L742)). If reconnecting and reconnected both occur between chunk/close checks, the predicate becomes true again for the old writer and the send may return local success on a different underlying connection. Web, Android, and Swift now make this transition irreversible for the captured epoch. Advance a dedicated byte-stream epoch whenever the room leaves `connected`, and cover an active → inactive → active transition.

### Suggestions

- **Narrow the PR description's isolation claim.** The separate bounded IPC lane prevents file callbacks and IPC queues from blocking real-time dispatch, but file bytes still share the LiveKit connection, network bandwidth, and the connector event loop. “Without interference with real-time media” is stronger than the implementation guarantees; “without blocking the real-time IPC path” would be accurate.

- **Qualify non-CI validation as local.** The visible checks cover Python, Android, docs, lint, and static analysis, but not Swift compilation, a LiveKit-enabled native build, or the browser end-to-end run. If those were performed manually, label them as local validation in the description.

### Verification

- Focused Python: 36 passed.
- JavaScript byte-stream transport: 1 passed.
- Native stub build/CTest: 7 passed.
- `git diff --check`: clean.
- Current GitHub checks: green.

No additional comment/readability nits; the new internal transport seams and session-aware cleanup are generally clear.

yanziz-nvidia:
reviewed by yanziz-review-bot

**Summary**

Other than the existing findings in previous comments, I found a defect in the new session-ordering buffer and a reproducible test failure. Local validation: 144 Python tests passed, one failed twice; all three JavaScript transport tests passed. CI is green; native and mobile builds were not rerun.

| Severity | Finding |
|---|---|
| 🚫 BLOCKER | [`_file_ordering.py:107`](https://github.com/NVIDIA/xr-ai/blob/bc58ace701537d1038cdb021a857e4dc2aad4b91/agent-sdk/xr-ai-hub/xr_ai_hub/_file_ordering.py#L107) — **Retain pending files from sessions not yet observed.** During a rapid reconnect, session B’s file can overtake the lifecycle sequence `join A → leave A → join B`. Processing `join A` discards B’s buffered file solely because its session differs. Reproduced at both hub and processor: B’s subsequent join delivers nothing. Retain unknown sessions until their matching join, confirmed departure, or expiry. |
| 💡 SUGGESTION | [`test_file_transfer.py:1091`](https://github.com/NVIDIA/xr-ai/blob/bc58ace701537d1038cdb021a857e4dc2aad4b91/tests/test_file_transfer.py#L1091) — **Allow upstream queue capacity to recover in the subscriber-isolation test.** On macOS, this test repeatedly fails sending `stream-1`: the connector’s HWM=1 queue still reports full after the healthy subscriber received `stream-0`. Use bounded retries when `push_file()` returns false so this test measures subscriber isolation without assuming immediate upstream capacity recovery. |
| 🧹 NIT | none |

**Actionables**

- Retain unmatched, non-departed sessions in `FileSessionOrderer.participant_joined()`. Add hub and processor coverage for `file B → join A → leave A → join B`.
- Add bounded admission retries to the slow-subscriber test while preserving its delivery assertions.

yanziz-nvidia:
LGTM modulo blocker comment (but I think blocker comment is not a must) feel free to create an issue to defer

yanziz-nvidia (approved):

nvddr (commented):
## Re-review summary

The two blockers from my previous review are substantially addressed at `dd6acdef`: the split IPC lanes now share a bounded session orderer, and native byte streams capture an epoch invalidated by SDK reconnects. The follow-up future-session fix in `ca96bf2e` also retains session B files while session A lifecycle events drain. One intermediate-state ordering hole remains.

### Blocker

- **Do not publish an old-session file while later lifecycle events are already queued.** `_drain_file_session_events()` says it applies every pending lifecycle event, but each intermediate join calls `_apply_file_session_event()`, which immediately publishes that session's buffered files ([`_hub.py:460-479`](https://github.com/NVIDIA/xr-ai/blob/dd6acdefcab6662a48aaaecf0872aa5907c2c6fb/services/device-io-hub/device_io_hub/ipc/_hub.py#L460-L479), [`_hub.py:510-522`](https://github.com/NVIDIA/xr-ai/blob/dd6acdefcab6662a48aaaecf0872aa5907c2c6fb/services/device-io-hub/device_io_hub/ipc/_hub.py#L510-L522)). I reproduced this at the current head with `file A → join A → leave A → join B` queued before the drain: the hub's authoritative main-lane state was already `session-b`, yet it published the buffered `session-a` file while applying the intermediate join. Because the participant and file publications use different sockets, a processor can observe that file after join A but before the already-sent leave/join B and invoke the callback with stale data. Fold all currently queued lifecycle events to their final state before publishing ready files, or at minimum require `_participant_sessions[pid] == msg.participant_session_id` immediately before hub publication. Add a regression where an old-session file is buffered and all of join A, leave A, and join B are pending before one drain; assert A is never published. The existing drain test routes the stale file only after the lifecycle queue is empty, so it misses this ordering.

### Suggestions

- **Preserve valid Web attribute keys such as `__proto__`.** `fileStreamOptions()` copies application attributes into a normal `{}` with property assignment ([`LiveKitBackend.js:107-131`](https://github.com/NVIDIA/xr-ai/blob/dd6acdefcab6662a48aaaecf0872aa5907c2c6fb/client-samples/web/StreamKit/Backends/LiveKit/LiveKitBackend.js#L107-L131)). An own `__proto__` key—for example from `JSON.parse()`—passes every documented validation rule, but assignment invokes the inherited setter and silently omits the attribute. Use a null-prototype record or `Object.fromEntries()` and cover this valid edge key.

- **The previous description qualifications remain.** “Without interference with real-time media” is stronger than the shared LiveKit/network path guarantees; “without blocking the real-time IPC path” is accurate. Swift compilation, a LiveKit-enabled native build, and browser E2E are not visible CI checks, so label them as local validation if that is how they were run.

### Verification

- Focused Python lifecycle/file/transport tests: 145 passed.
- JavaScript byte-stream transport: 3 passed.
- Native stub build/CTest: 7 passed (with four existing missing-field-initializer warnings in the expanded test).
- `git diff --check`: clean.
- Current GitHub checks at `dd6acdef`: green; the branch is behind current `main`.

No additional documentation or comment-readability blockers found.

nvddr (changes requested):
Re-review summary for `d2a423e3`: **1 blocker, 2 suggestions**.

## Blocker

1. **Revalidate the participant session after asynchronous encoding, before publishing the file.**

   `HubEndpoint._publish_file()` checks `_file_session_is_active(msg)` before `await asyncio.to_thread(encode, ...)`, but publishes immediately when that await returns. The authoritative session can change during the encode: session A can pass the check, leave/rejoin can update `_participant_sessions` to session B, and the completed session-A file is then sent to subscribers.

   I reproduced this deterministically on the current head by pausing `asyncio.to_thread`, changing `hub._participant_sessions["alice"]` from `session-a` to `session-b`, and resuming it; `send_multipart` was still awaited once. This means the new guard fixes the previously reported ordering case at function entry, but does not yet guarantee that stale bytes cannot cross a reconnect boundary.

   Please move the authoritative check to after encoding, or repeat it immediately after encoding and before `send_multipart`, and add a regression test that changes the active session while encoding is suspended.

## Suggestions

1. The PR description says the separate IPC lane avoids “interference with real-time media.” It isolates and bounds the hub IPC path, but file traffic still shares the LiveKit connection and underlying network resources with media. Consider narrowing this to the concrete guarantee, such as preventing file IPC/backpressure from blocking the real-time IPC lane.

2. Consider labeling the Android, Swift, native stub, and browser E2E items in the Validation paragraph as local/manual validation where applicable. The visible CI checks cover Python, Android assembly, docs, lint, dependency locks, SPDX, DCO, and CodeQL, but do not make all of those claimed runs independently visible to a reviewer.

## Confirmed resolved / verification

- The earlier authoritative-session-at-entry issue is fixed for files that are already stale when `_publish_file()` begins.
- Web attributes now preserve `__proto__` as an ordinary key.
- The shared byte-stream option setup in the newest commit is consistent across the four clients; I did not find a new client-side correctness issue in that refactor.
- Focused Python suite: **205 passed**.
- Browser byte-stream and capture suites: **17 passed**.
- Native stub build and CTest: **7/7 passed**.
- Current visible CI checks are green.

nvddr (approved):
Re-review summary for `c2d44b72`: **0 blockers, 0 suggestions**.

The previous blocker is resolved. `_publish_file()` now revalidates the authoritative participant session after the asynchronous encoding step and before publication, and the new deterministic regression test covers a session change while encoding is suspended.

The PR description also now accurately limits the isolation claim to the real-time IPC path and distinguishes CI coverage from local validation. The new comment explains the non-obvious race without narrating implementation history.

Verification on the current head:

- Focused Python IPC/image-capture suite: **206 passed**.
- Browser byte-stream and capture suites: **17 passed**.
- Native stub build and CTest: **7/7 passed**.
- `git diff --check`: clean.
- Current visible CI checks: green.

PR #518, opened 2026-09-14
Fix web agent audio playback after autoplay block

nvddr:
## Summary

- prime LiveKit remote-audio playback from the existing Connect button gesture
- retry playback from later Start Microphone and typed-message gestures
- keep playback recovery best-effort so it cannot block connection, microphone publication, or data sends
- document browser autoplay as a connected-but-silent cause

## Why

Remote agent audio can arrive before or without microphone capture. The remote-track handler ignored a rejected play() promise, so receive-only sessions had no path that explicitly retried LiveKit’s playback gate. Browser autoplay policies may require a click or tap; Room.startAudio() is LiveKit’s explicit recovery mechanism.

Tracks subscribed after Connect still perform their own play attempt. Start Microphone and typed-message gestures provide later retries if that attempt remains blocked.

## Validation

- CI pytest passes on Python 3.11 and 3.12
- all current PR checks pass
- no tests run locally, per request (preserved existing processes)


yanziz-nvidia (approved):

blongs-nv:
Reviewed at `60a7d52f8445ef5ebf02d933db920fc8dd9bfb08` by a bot.

No blockers. The call is the right API at the right moment: it is the first await on the click path, so `play()` runs inside the gesture, and it is idempotent on repeated presses. The premise checks out against livekit-client 2.22.2: the SDK's own retry hook fires only from `localParticipant.createTracks()`, and this backend publishes through `createLocalAudioTrack`, so nothing ever re-ran playback before this change.

## Suggestions

1. [`LiveKitBackend.js:299`](https://github.com/NVIDIA/xr-ai/blob/60a7d52f8445ef5ebf02d933db920fc8dd9bfb08/client-samples/web/StreamKit/Backends/LiveKit/LiveKitBackend.js#L299) puts a rethrowing call on the microphone's critical path. `Room.startAudio()` rejects if any attached element's `play()` fails (a stale element, an AbortError from a concurrent detach, Safari's deferred play), and `core.js` then shows the raw error and never publishes the microphone. Playback recovery is best-effort, so catch and warn: `await this.#room.startAudio().catch((e) => console.warn('remote audio playback still blocked', e));`

2. The fix only reaches users who press Start Microphone. The quickstart documents a type-only path with spoken replies ([`quickstart.md:175`](https://github.com/NVIDIA/xr-ai/blob/60a7d52f8445ef5ebf02d933db920fc8dd9bfb08/docs/source/getting_started/quickstart.md#L175-L179)); Chrome and Firefox are fine because Connect grants sticky activation, but a Safari user who only types stays silent with no hint. Calling `startAudio()` at the top of `send()` too, or reflecting `room.canPlaybackAudio` in the existing audio-status text, closes that without a new control.

3. [`LiveKitBackend.js:255`](https://github.com/NVIDIA/xr-ai/blob/60a7d52f8445ef5ebf02d933db920fc8dd9bfb08/client-samples/web/StreamKit/Backends/LiveKit/LiveKitBackend.js#L255) still says playback "will resume on next user gesture"; it now resumes only on the Start Microphone gesture, so point the comment at `startAudio()`. [`troubleshooting.md:347`](https://github.com/NVIDIA/xr-ai/blob/60a7d52f8445ef5ebf02d933db920fc8dd9bfb08/docs/source/guides/troubleshooting.md#L347) lists only the UDP firewall cause for "connects but no audio"; add the autoplay cause and its remedy.

## Nits

- With `hubIdentity` isolation the hub track can subscribe after the click; the call then only resumes the AudioContext, and the later `el.play()` at line 254 succeeds on Chromium via sticky activation but can still fail on WebKit. A clause in the new comment would stop the next reader assuming the call covers not-yet-subscribed tracks.
- The PR body's "LiveKit requires it to run from a click or tap handler" is the browser's rule, not LiveKit's; the SDK calls `startAudio` outside gestures itself.

blongs-nv (approved):

nvddr:
Disposition for the review at `60a7d52f` (addressed in `1eb189cd`):

1. **Microphone critical path — addressed.** Playback recovery now goes through a best-effort helper that catches and warns, so a playback rejection cannot prevent microphone publication.
2. **Typed/receive-only path — addressed.** Connect primes playback before token/network waits, and `send()` retries it from typed-message gestures. The microphone gesture remains an additional retry.
3. **Stale comment and troubleshooting — addressed.** The track-subscription comment now names the actual retry paths, and the troubleshooting guide covers browser autoplay and the available remedies.
4. **Later-subscribed tracks — addressed.** The Connect comment explicitly notes that those tracks still make their own `play()` attempt.
5. **Policy wording — addressed.** The PR description now attributes the gesture constraint to browser autoplay policy and describes `Room.startAudio()` as LiveKit’s recovery mechanism.
6. **CI fixture — addressed.** The mocked LiveKit room now implements and records `startAudio()`, with assertions for the Connect and typed-send attempts.

CI pytest passes on Python 3.11 and 3.12, and all current PR checks pass. No tests were run locally, per request; existing processes were preserved.

blongs-nv:
Reviewed at `1eb189cdf6c55138588c5c8cf173b8af0e18df45` by a bot.

Everything from my earlier comment is addressed at this head: the recovery call is best-effort on all three gestures, so a playback failure cannot block the microphone; typed-only sessions are covered via Connect and send; the handler comment and troubleshooting entry are updated. Nothing further from me.

PR #520, opened 2026-09-15
Add XR AI application-building guidance

blongs-nv:
Adds a complete entry point for developers and coding agents to build XR AI applications. The `build-your-app` skill handles new or existing checkouts, then follows a canonical guide for choosing SDK layers, adapting the reference application, running it, and verifying it without hardware.

Introduces an ignored top-level `apps/` workspace for application code that is not intended to become a repository sample. Repository dependency inventory, SPDX, lint, documentation catalogs, and related checks exclude that boundary while continuing to cover repository-owned code.

Also makes reused voice response identifiers diagnosable. The runtime remembers recently completed response IDs, ignores later output under those IDs with one warning, and documents when to use a unique ID or omit it.

Testing:

- `cd tests && UV_PROJECT_ENVIRONMENT=../.venv uv run pytest -v --tb=short --color=yes -m "not gpu"` — 1,669 passed, 11 deselected
- Focused application-boundary, dependency, documentation, model-server, and voice-runtime suite — 145 passed
- `uv run --with-requirements docs/requirements.txt make -C docs strict`


wenxind-nvidia (commented):
## Blockers

None identified.

## Suggestions

1. **Preserve the repository's dependency qualification cutoff.**
   [`building-your-app.md:132`](https://github.com/NVIDIA/xr-ai/blob/0615f33e6f19aa5f1f8242273c0948dfa5800191/docs/source/guides/building-your-app.md#L132) uses plain `uv sync`. Both copied projects contain `[tool.uv.sources]`, which stops discovery of the root `uv.toml`; fresh environments can therefore resolve packages newer than the qualified cutoff.

   From the documented application directory, use:

   ```bash
   uv --config-file ../../uv.toml sync
   uv --config-file ../../uv.toml sync --project worker
   ```

   Apply the same explicit configuration to subsequent `uv run` commands that may resolve dependencies.

## Nits

1. **Qualify the closed-response guarantee.**
   The new [voice documentation](https://github.com/NVIDIA/xr-ai/blob/0615f33e6f19aa5f1f8242273c0948dfa5800191/docs/source/reference/agent-sdk-voice.md#L72) says closed-stream output is ignored and warned once. However, `_closed_streams` retains only 1,024 identifiers; older identifiers can be accepted again after eviction. Describe this as protection for **recently closed streams**, rather than permanent deduplication. No behavioral change is needed.

Validation: **44 focused tests passed**, including the new warning regression; two environment-dependent tests excluded. All GitHub checks are green. Independent review found no additional application-workflow issue.

Reviewed head: 0615f33e6f19aa5f1f8242273c0948dfa5800191

nvddr (changes requested):
## Summary

This is a useful direction: the skill is deliberately small, the main guide is centralized, the explicit root `uv.toml` usage preserves qualified dependency resolution, and the hardware-free configuration check is practical. I followed the copy/rename workflow in a clean worktree; both documented `uv sync` commands and the configuration-load command succeeded. The focused runtime/docs suite also passes (46 tests), and CI is green.

I am requesting changes because the new canonical guide currently conflicts with another pending change and because its central “supported non-sample application” promise does not yet match the repository behavior it documents.

## Blockers

1. **Align this guide with #522 before either change merges.** `building-your-app.md` lines 107, 113, and 159–167 tell readers to preserve reuse-only `Process` declarations, model `deployment` separation, and implicit `VoiceAgent` STT/TTS health checks. #522 removes those exact contracts and updates the reference application. Merging both PRs as currently written leaves this new canonical guide stale. Please rebase/update the guide for the intended post-#522 behavior, or make the dependency and merge order explicit and ensure the corresponding guide update lands with it.

2. **Make the “not an upstream sample” workflow internally consistent.** The guide says this is the supported path for an application that will not be a repository sample, but requires putting it under `agent-samples/`; lines 201–214 then explain that the standard hooks, dependency inventory, docs, and tests *do* treat it as a sample and recommend leaving pre-commit uninstalled. That is a local prototype carve-out, not yet a clean supported application boundary. Please either provide a repository-recognized local-application location/exclusion path, or narrow both the guide and PR description so they accurately describe the limitations. In either case, avoid making “do not install the repository hooks” part of the normal workflow.

3. **Close the copy recipe’s documentation trap.** Applying the documented rename table to the copied README changes the existing sample links into nonexistent `reference/my-app.html` links while retaining “Simple VLM example” / “This sample” prose. The reuse table’s generic “describe the application” direction is easy to miss. Please explicitly tell readers to remove the copied sample-reference links and replace the README’s sample identity with application-owned documentation (or link back to this general guide).

## Suggestions

- **Keep the runtime change separate or describe its narrower contract exactly.** The closed-response warning and its 42-line regression test are unrelated to application-building guidance and make this documentation PR harder to reason about. If retained, the PR body should say “recently closed incremental stream,” not any “completed voice response”: a one-message `VoiceOutput(text=..., response_id=..., final=True)` is not recorded in `_closed_streams`, so the same ID is accepted again without a warning. Either make that behavior intentional in the wording/tests or remember that completion too.
- The `VoiceOutput.response_id` field docstring should also say **recently** closed; the 1,024-entry cache does not provide a permanent guarantee.
- `quick_validate.py skills/build-your-app` is not reproducible from this repository because no such script is present. Give the full owning tool/path (and version if external), or omit that test entry from the PR description.

## Nits

- Consider splitting the long configuration `python -c` command across lines or providing a small checked-in validation entry point; the current command is correct, but unnecessarily hard to scan and edit.

## Validation

- Exact documented copy, rename, root/worker sync, and configuration-load workflow: passed.
- `tests/test_voice_runtime.py` and `tests/test_docs_versions.py`: 46 passed.
- Current GitHub checks: all green.
- The prior review’s explicit root-config and “recently closed” narrative-doc requests are addressed at head `58419f1`.

nvddr (commented):
## Review summary

Reviewed current head `5cc92ec`. The #522 behavior alignment and copied-README cleanup are resolved. Three items remain:

- [`building-your-app.md:184`](https://github.com/NVIDIA/xr-ai/blob/5cc92ece531ed28f8134215ad5dc0bc9a2e2c0a6/docs/source/guides/building-your-app.md#L184) still launches `../model-servers`. Current `main` moved that project in #523, so from `agent-samples/my-app` the path must be `../../model-server-samples/model-servers`.
- The non-sample workflow still depends on leaving repository hooks uninstalled, even though projects under `agent-samples/` are discovered by dependency generation, documentation catalogs, and repository tests, and the repository contract requires dependency-map regeneration after `pyproject.toml` changes. This needs either a recognized local-application boundary or a normal fork/sample workflow that runs and updates those checks.
- [`_runtime.py:392`](https://github.com/NVIDIA/xr-ai/blob/5cc92ece531ed28f8134215ad5dc0bc9a2e2c0a6/agent-sdk/xr-ai-voice/xr_ai_voice/_runtime.py#L392) now treats a one-message finite `VoiceOutput` with a `response_id` as closed and drops later reuse. Because that changes public field behavior, add migration guidance telling applications to use unique identifiers or omit `response_id` for finite responses.

Validation: 44 focused runtime/docs tests passed with two environment-dependent tests excluded; `git diff --check` passed; current GitHub checks are green. The PR is behind current `main`, which exposed the #523 path mismatch in the merged-tree review.

nvddr:
Why is the runtime change in this PR?

blongs-nv:
> Why is the runtime change in this PR?

It addressed something that confused the test run agent when it was trying to follow the skill, and was small. I can move it to a separate PR if preferred though.

nvddr (approved):

yanziz-nvidia (approved):

wenxind-nvidia (approved):

PR #522, opened 2026-09-16
Simplify consumer model profiles and startup readiness

wenxind-nvidia:
Consumer workers may now report ready without confirming STT, TTS, LLM, or VLM availability. Start the shared model-server stack and wait for its launcher to return before starting consumers. Model-server startup and reuse checks remain in place.

Consumer samples previously repeated model ownership declarations and polled model health after the shared stack had already checked readiness. This change removes 17 redundant deployment blocks, explicit health settings, and skipped model-process declarations from four consumer samples. Workers keep their configured adapters and endpoints. The SDK's existing support for omitted deployment metadata is sufficient; the launcher profile loader retains its original contract.

`VoiceAgent` waits only for explicitly supplied probes. The remaining startup gates are:

- Simple VLM's streaming image warmup, without a preceding health request.
- XR Render's tool-shaped LLM warmup, including for hosted LLMs.
- Tea making's RAG capability probe. RAG still checks embedding health; hosted embedding entries require `readiness: none` when no health route exists, or the provider's supported `health_path`.

Custom `VoiceAgent` users can retain the previous speech-health gate with `probes={"stt": stt.health, "tts": tts.health}`. The public API contract and migration guide document this behavior, with sample-specific warmup details in the sample guides. Repeated guidance links to these references, and NIM YAML comments link to the canonical customization procedure.

HTTP inference retry behavior is unchanged. Model implementations, managed profiles, adapters, endpoint addresses, and ports are unchanged.

Validation passed for the targeted test selection below, repository Ruff, strict Sphinx documentation and generated API references, changed-file SPDX headers, and `git diff --check`. Reproduce the test selection from the repository root:

```bash
uv --config-file uv.toml run --project tests pytest -q -c tests/pyproject.toml \
  tests/test_models_openai_compat.py tests/test_models_config.py \
  tests/test_models_protocols.py tests/test_models_riva.py \
  tests/test_launcher_config.py tests/test_launcher_stack.py \
  tests/test_launcher_credentials.py tests/test_voice_session.py \
  tests/test_consumer_model_startup.py tests/test_simple_vlm_example_worker.py \
  tests/test_rag_service.py
```

The selection verifies loader validation, consumer profiles without deployment metadata, no implicit model-health calls, retained warmups and RAG probes, model clients, and voice lifecycle. Live GPU qualification was not run; no live model stack was restarted.


nvddr (changes requested):
## High-level summary

The core cleanup is a good direction: consumer samples should declare only the processes they own, model profiles should describe the adapters/endpoints they consume, and startup gates should be explicit warmups or capability probes. The sample wiring changes are straightforward, the migration note calls out the `VoiceAgent` behavior change, and the focused tests plus green CI give good coverage.

I do not think the PR is ready as one change, though. It combines that cleanup with a new retry policy for every OpenAI-compatible inference call and broadens an exported launcher profile contract. Both materially increase the review surface and public behavior of a PR whose stated goal is simplification.

## Blockers

1. **Separate the HTTP retry policy, or define its at-least-once semantics explicitly.**

   `_inference_response()` retries POSTs after receiving 502/503/504 response headers. In particular, a 504 does not prove the upstream inference was not processed, so a retry can duplicate a costly/billable request. There is no idempotency key or caller opt-in. This is independent of removing health polling and changes LLM, VLM, STT, TTS, embeddings, and streaming behavior at once; it accounts for most of the new implementation/test complexity.

   The description and tests also say “accepted responses are never replayed,” while the implementation and `test_inference_retries_before_response...` deliberately retry 502/503/504 responses after their headers have been accepted. The actual boundary is narrower: once a response other than those retry statuses is handed to the consumer, body/stream failures are not retried.

   My preference is to move retries to a follow-up PR with a clear policy decision about 502/503/504, request cost, idempotency, and configurability. If retries must remain here, please at least document the at-least-once behavior accurately and explain why it is coupled to this readiness cleanup.

2. **Do not broaden the exported launcher profile contract without a concrete use case/approval.**

   `load_deployment_profile()` now accepts entries with no `deployment` object. That is a public behavior expansion, but consumer launchers no longer inspect model profiles at all. The only in-tree production caller is `model-servers`, whose shipped profiles already contain deployment metadata. The PR's own docs say consumer launchers do not inspect these profiles.

   Please revert the launcher-loader behavior, its compatibility tests, and the claims that `deployment` is optional in both loaders. The worker-side SDK already defaults omitted deployment metadata to `external`, which is sufficient for the consumer-profile cleanup. If an out-of-tree launcher use case requires the broader contract, call that out and handle it as an explicitly approved API change.

## Suggestions

- **Consolidate the documentation.** The same “no implicit health polling / start model-servers first / Simple VLM and XR Render retain warmups” explanation is repeated across architecture, AI services, adding a sample, customizing model servers, both SDK references, quickstart, migrations, and class docstrings. Keep the public `VoiceAgent` contract and migration guidance authoritative, keep sample-specific warmups in each sample guide, and link from the remaining pages. The retry paragraph is likewise repeated at module level and on four client classes.
- **Tighten the PR description.** Lead with the breaking behavior: consumer workers may report ready without STT/TTS/LLM/VLM availability, so operators must wait for `model_servers`; then list the explicit warmups that remain and the migration for custom `VoiceAgent` users. If retries remain, give them a separate section and replace “accepted responses are never replayed” with the exact retry boundary. Prefer reproducible test commands or the CI result over the unauditable “735 focused tests” count.
- **Remove duplicated workflow comments from the six GPU-specific NIM YAMLs.** The canonical customization guide already owns the copy/remove-deployment procedure. A short pointer there is easier to keep correct than six nearly identical instructions.

## Nits

- `EndpointSpec.health_check` still says “Whether readiness requires...” and `_http_health()` says “readiness is assumed.” Under the new contract these should describe only whether an explicit `health()` call performs a request.
- `docs/source/reference/xr-render-demo.md` still says every `models.json` role maps to “an adapter, endpoint, and deployment,” although this PR removes deployment from every entry.
- `docs/source/components/ai-services.md` similarly says every role selects through `adapter`/`endpoint`/`deployment`; make `deployment` explicitly model-server-only/optional there.

## Validation

- Focused local review suite: **250 passed**.
- Current GitHub CI is green; Python 3.11 reports **1,718 passed, 1 skipped, 9 deselected**.
- I did not run live GPU qualification.

nvddr (approved):

yanziz-nvidia (approved):

PR #527, opened 2026-09-18
Add StreamKit byte-stream transport

blongs-nv:
Adds a private byte-stream transport layer that StreamKit features can share without expanding the public SDK API.

- Add client-side writers for Android, Swift, web, and C++ that own LiveKit stream creation, writes, closure, error cleanup, and connection checks.
- Add a bounded hub-side reader with size validation, idle and total timeouts, accumulation, and guaranteed closure.
- Update the minimum LiveKit dependencies required for byte streams.

Validated with the Android build, Python and web tests, Swift package compilation, native stub compilation and tests, and a native build against LiveKit C++ v1.10.2.


wenxind-nvidia (commented):
## Blockers

- `client-samples/native/StreamKit/src/Backends/LiveKit/ByteStreamTransport.cpp:41`: **The native SDK upgrade leaves the existing backend incompatible.** The new writer uses `localParticipant().lock()`, matching the newly required SDK's weak-pointer return type. However, `LiveKitBackend.cpp:166` still assigns that result to `const auto*`, and other calls use `localParticipant()->...`. These cannot compile against the same SDK. Update the existing backend alongside the dependency upgrade, update the native setup documentation, and validate a real-SDK build. Stub tests exclude these paths.

Validation: 22 Python tests passed, 3 skipped; 3 web tests passed. GitHub checks are green. Real native and Swift builds remain unverified.

Reviewed head: 58a6dc7c3e9f1610d9d4c7beec8f791d5a8e600d

nvddr (changes requested):
## Summary

The split is architecturally sound: the new transport remains internal on every platform, the public `StreamingBackend` API is unchanged, and the native compatibility issue reported on the prior commit is fixed. I built and linked the native implementation against the official LiveKit C++ v1.10.2 SDK.

### 🚫 Blocker

- `client-samples/native/StreamKit/Tests/StreamKitTests/ByteStreamTransportTests.cpp:19`: **Make the new test valid when the real LiveKit SDK is enabled.** The test passes a null room and expects the stub-only `stub-*` result. With `LIVEKIT_SDK_ROOT` pointing to the official v1.10.2 SDK, the production code builds and links, but this test aborts with `ByteStreamConnectionChanged`. Keep a stub-only test conditional on stub mode, or inject/mock the writer/participant so the same target also passes in an SDK-enabled build. Please keep this focused on the new byte-stream test; the separate existing backend test also assumes stub behavior but predates this PR.

### 💡 Suggestion

- `client-samples/web/StreamKit/Backends/LiveKit/ByteStreamTransport.js:55`: **Clear the losing timeout after failure cleanup.** When `writer.close()` resolves first, the two-second timer created for `Promise.race` remains scheduled. The three focused Node tests consequently keep the process alive for about two seconds. Retain the bounded close behavior but clear the timer when the race completes.

### API boundary

The current visibility is appropriate: Kotlin and Swift are internal, the C++ header stays under `src`, Python uses a private module, and Web does not re-export the helper from the StreamKit entry point. When PR #501 consumes this transport, it should use a package-internal capability/adapter and remove `sendImage` from the public backend contract; the raw byte-stream operation should not be added to public `StreamingBackend`.

Validation at `5104bf0337e0963828c39dce92cd261065ffc784`:

- Python focused tests: 25 passed
- Web focused tests: 3 passed
- Native stub build/tests: 7 passed
- Native official LiveKit C++ v1.10.2: production code compiled and linked; the new stub-assuming test failed as described
- Swift compilation was not available locally

nvddr (approved):

wenxind-nvidia (approved):

PR #529, opened 2026-09-18
Disconnect LiveKit room before stopping its Docker server

wenxind-nvidia:
DeviceIOHub currently starts Docker shutdown concurrently with media cleanup and `Room.disconnect()`. The LiveKit server can disappear before room teardown finishes, creating a shutdown-order race associated with the reported FFI handle assertions when stopping Simple VLM on DGX Spark.

Register Docker shutdown first in `AsyncExitStack` so it executes last, after worker loops, media streams, room disconnection, and the web/token services finish cleanup. Preserve the remaining cleanup order and propagation of cleanup failures.

This PR contains only the connector change and shutdown regression tests. It is independent of the Magpie wrapper (#528) and the NIM sample (#512), from which it was extracted.

Validation: all 38 `tests/test_device_io_hub_startup.py` tests pass, including a blocked room disconnect proving Docker cannot start early, disconnect failure, and later cleanup execution. Pinned Ruff and whitespace checks pass. These tests verify shutdown ordering; they do not reproduce native FFI destruction on Spark.


nvddr (commented):

Reviewed head: 0c420c9bbe1112a70cf01e12a25d0c5b96b1050e

blongs-nv:
Reviewed at `0c420c9bbe1112a70cf01e12a25d0c5b96b1050e` by a bot.

## Blockers

1. **A hung room disconnect now leaks the LiveKit container.**

   [`connector.py:170`](https://github.com/NVIDIA/xr-ai/blob/0c420c9bbe1112a70cf01e12a25d0c5b96b1050e/services/device-io-hub/device_io_hub/transport/livekit/connector.py#L170) Docker stop is now strictly behind `RoomClient.disconnect()`, whose `await self._room.disconnect()` ([`_room_client.py:368`](https://github.com/NVIDIA/xr-ai/blob/0c420c9bbe1112a70cf01e12a25d0c5b96b1050e/services/device-io-hub/device_io_hub/transport/livekit/_room_client.py#L368)) has no timeout and waits on FFI callbacks that the very failure mode this PR targets may never deliver. The launcher then SIGKILLs the hub after 20 s ([`_stack.py:54`](https://github.com/NVIDIA/xr-ai/blob/0c420c9bbe1112a70cf01e12a25d0c5b96b1050e/utils/xr-ai-launcher/xr_ai_launcher/_stack.py#L54)), which skips the `atexit` stop and leaves `xr-ai-livekit-server` holding 7880 until the next launch's `docker rm -f`; on `main` the eager task issued `docker stop` regardless. Keep the new order but wrap the disconnect step in `asyncio.wait_for` with a timeout that still fits `docker stop --time 10` plus its 5 s wait inside the 20 s window.

## Suggestions

1. Add the hang case to the new test: a `disconnect` that never releases, asserting `docker-stop` still occurs and `stop()` returns within the bound. [`test_device_io_hub_startup.py:87`](https://github.com/NVIDIA/xr-ai/blob/0c420c9bbe1112a70cf01e12a25d0c5b96b1050e/tests/test_device_io_hub_startup.py#L87-L129) always sets `release_disconnect`, so only the happy ordering is pinned.

2. Record the shutdown budget in the comment. [`connector.py:168`](https://github.com/NVIDIA/xr-ai/blob/0c420c9bbe1112a70cf01e12a25d0c5b96b1050e/services/device-io-hub/device_io_hub/transport/livekit/connector.py#L168-L169) explains the order but not that disconnect, web and token drains, and a 15 s Docker stop are now additive inside the launcher's 20 s SIGKILL window, which is the constraint the next person would need before re-parallelizing it.

3. Pin that a disconnect failure survives a later cleanup failure. One more parametrization with `docker.stop` also raising, asserting the disconnect error is the `__context__` of what `stop()` raises, would cover the "propagation of cleanup failures" claim in the PR body.

The reordering is the minimal idiomatic way to express that the server outlives the client, and the new test genuinely fails on the old eager task. The blocker above should be resolved before merge.

wenxind-nvidia:
Addressed the shutdown review in 7be4739c.

- Preserve disconnect-before-Docker ordering, but bound room, web, and token cleanup with a shared four-second deadline. This reserves the nominal 15 seconds for Docker stop/reaping plus one second of margin within the launcher's 20-second shutdown window.
- Add never-returning room, web, and token cleanup regressions: each times out, subsequent cleanup still runs, and Docker stop is awaited.
- Use an outer exit stack for Docker cleanup so a later Docker error retains the earlier disconnect error as its exception context. Add the requested two-failure regression.
- Document the additive shutdown budget beside the deadline.

Validation: all 42 DeviceIOHub startup/shutdown tests pass; pinned Ruff and whitespace checks pass. All findings in this review are addressed. The changes remain limited to connector shutdown and its tests; this does not claim to fix separate LiveKit native-handle ownership bugs or constitute a DGX Spark reproduction.

yanziz-nvidia (approved):

nvddr (approved):

blongs-nv (approved):

